PR #132, opened 2017-07-13
Initial pass at the GeneratorReference schema.

ssteinbach:
Adds support for a `otio.schema.GeneratorReference` class, and sketches out a SMPTEBars version of that class.  Also adds support for this in the RV adapter.

Questions:

- Should we follow the pattern in the Transition schema (with an Enum defining specific generators, and then a "generic" fallback)?

### DIFF
--- a/opentimelineio/schema/generator_reference.py
+++ b/opentimelineio/schema/generator_reference.py
@@ -1,0 +1,77 @@
+"""
+Generators are media references that _produce_ media rather than refer to it.
+"""
+
+from .. import (
+    core,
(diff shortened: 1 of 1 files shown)


reinecke:
This seems like a really good place to work from.

+1 for using an enum for generator kind, this will encourage us to standardize a lot of the more common types and also hopefully drive the community to help expose what sorts of generators are in use based on their submissions for enum additions.

ssteinbach:
This was still from the old private repo.  Switching to the public fork now.

Issue #142, opened 2017-08-01
Color Decision List (CDL) support


jminor:
Some people would like OTIO to handle Color Decision List metadata in a structured way.

For example, an EDL can contain ASC_SOP and ASC_SAT like this:
*ASC_SOP (1.0000 1.0000 1.0000)(0.0000 0.0122 0.0305)(1.0000 1.0000 1.0000)
*ASC_SAT 1.0000

For reference:
https://en.wikipedia.org/wiki/ASC_CDL

jminor:
@adaro can you take a look at this to see if it parses your CDL data correctly?

jminor:
CDL metadata is now supported in the EDL adapter (both read and write).

CDL metadata on a Clip in OTIO looks like this:
```
                        "metadata": {
                            "cdl": {
                                "asc_sat": 0.9, 
                                "asc_sop": {
                                    "offset": [
                                        1.0, 
                                        0.0122, 
                                        0.0305
                                    ], 
                                    "power": [
                                        1.0, 
                                        0.0, 
                                        1.0
                                    ], 
                                    "slope": [
                                        0.1, 
                                        0.2, 
                                        0.3
                                    ]
                                }
                            }
                         }
```

We still need to add support for CDL to the FCP XML adapter.

jminor:
Here is a reference for how CDL data is stored in AAF (as a Video Effect).
https://www.amwa.tv/downloads/specifications/AS-05_AAF_Effects_protocol_v1.pdf

jminor:
Here is a reference for how CDL data is stored in FCP X XML (as an XML comment inside a color correction video filter).
https://developer.apple.com/library/content/documentation/FinalCutProX/Reference/FinalCutProXXMLFormat/Adjustments/Adjustments.html#//apple_ref/doc/uid/TP40011227-CH14-SW8

jminor:
I think CDL support in OTIO would be a great candidate for expanding our support for effects. It really is a color correction effect with well defined parameters, so we can model it as a schema.Effect object on a Clip. Each adapter can then transcribe that to/from the various other formats as appropriate.

ssteinbach:
Addressed by #143

PR #143, opened 2017-08-01
Added CDL support to the EDL adapter.

jminor:
Needs testing with real CDLs.
Needs write support.
Needs vetting in terms of where this should be stored (on a clip, media reference, as an effect, etc.) 
Addresses part of #142


jminor:
@adaro can you take a look at this to see if it parses your CDL data correctly?

jminor:
This is what the CDL info looks like in OTIO. Is this too verbose?

```
"metadata": {
    "cdl": {
        "asc_sat": 0.9, 
        "asc_sop": {
            "offset": [
                1.0, 
                0.0122, 
                0.0305
            ], 
            "power": [
                1.0, 
                0.0, 
                1.0
            ], 
            "slope": [
                0.1, 
                0.2, 
                0.3
            ]
        }
    }, 
}
```

jminor:
FCP X XML supports CDL information as a Color Filter (like a video effect)
https://developer.apple.com/library/content/documentation/FinalCutProX/Reference/FinalCutProXXMLFormat/Adjustments/Adjustments.html#//apple_ref/doc/uid/TP40011227-CH14-SW18

We could model it this way also, which would be a step towards fleshing out OTIO's Effects support.

Research is needed into how AAF models CDL information.

jminor:
Also, there is some nice open source python code for dealing with various flavors of CDL here: https://github.com/shidarin/cdl_convert

jminor:
and more here: https://github.com/andrewbunday/ale2ccc

jminor:
FYI, I'm going to add write support to this also before accepting the PR.

jminor:
@ssteinbach and @adaro can you take one more look? I added support for writing CDL info back out to an EDL.

opentimelineio/adapters/cmx_3600.py
```diff
@@ -36,9 +36,15 @@
 import os
 import re
 import math
+import collections
 
 import opentimelineio as otio
 
+
+class EDLParseError(otio.exceptions.OTIOError):
```
ssteinbach: Should we make a more specific subclass of `OTIOError` along the lines of "AdapterParseError"?

opentimelineio/adapters/cmx_3600.py
```diff
@@ -591,6 +641,28 @@ def write_to_string(input_otio):
         if url:
             lines.append("* FROM CLIP: {}".format(url))
 
+        cdl = clip.metadata.get('cdl')
+        if cdl:
+            asc_sop = cdl.get('asc_sop')
+            asc_sat = cdl.get('asc_sat')
+            if asc_sop:
```
ssteinbach: Is it valid to have a CDL with SAT but not SOP data? Or the other way?  That would be an argument for pushing for a schema.  This code looks good to me for now though.

opentimelineio/adapters/cmx_3600.py
jminor: Sure. I've been adding adapter-specific exception classes like this, so having a superclass to those makes sense.

opentimelineio/adapters/cmx_3600.py
jminor: I made them separate like that in case you had one and not the other. Some of the reference I read implied that could happen, but I haven't seen it in practice. There was also some documentation that implied that you could stack multiple SOPs or SATs on a single clip (in an FCP XML not an EDL). I think we should defer support for that until we do the FCP XML side of this, which is also when we will likely want to treat this as a video effect instead of metadata.

PR #144, opened 2017-08-09
Fix spelling of 'serializable'

alexwidener:
Before development and integration of OTIO gets too far along, I thought we might want to fix the spelling of this. Feel free to reject and keep it as it currently is, it was only a quick refactor. I ran all tests and everything is passing.

### DIFF
--- a/bin/otioview.py
+++ b/bin/otioview.py
@@ -138,7 +138,7 @@ class Main(QtGui.QMainWindow):
             self.sequences_widget.setVisible(False)
         elif isinstance(
             file_contents,
-            otio.schema.SerializeableCollection
+            otio.schema.SerializableCollection
         ):
             for s in file_contents:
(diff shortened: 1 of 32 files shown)


jminor:
Oops, thanks for catching that.

Since `SerializableCollection.1` also changed, we'll need to add code to handle old OTIO files that have `SerializeableCollection.1` objects in them. We previously renamed `Filler` to `Gap` and you can see how we handled that here: #122 The essential piece was `core.register_type(Gap, "Filler")`

Also, we could consider renaming SerializableCollection to just Collection...

alexwidener:
The Travis repository is still private, so I'm _guessing_ that what failed was the linter.

ssteinbach:
Whoops.  We'll flip the switch on making that public.  Sorry about that!  See #146

PR #148, opened 2017-08-10
GeneratorReference implementation.

ssteinbach:
Adds support for a otio.schema.GeneratorReference class, and sketches out a SMPTEBars version of that class. Also adds support for this in the RV adapter.

Questions:

- Should we follow the pattern in the Transition schema (with an Enum defining specific generators, and then a "generic" fallback)?

Reposted with the public fork (was #132)


reinecke:
This looks totally reasonable to me. I like the idea of using an enum for generator kind, It will encourage some schema unification around the generator types commonly used (flat color, smpte bars, timecode overlay, etc.)

I'd be curious to know what kinds of generators people are encountering so we can try this approach against them, but it seems pretty solid to me.

bashesenaxis:
I guess the only concern I would have with this, is that using a generic "parameters" field leaves it quite open to what you put in there. 
So I definitely agree with the idea of using an enum for commonly used kinds. And perhaps also a parameter spec for each common kind? Or do we want to rely on convention to make sure that all generator implementations of the adapters line up?

quelsolaar:
I'm generally in favor of using Enums as much as possible in the API, strings are error prone. Especialy since the underlying implementation will use a readable text string so that its easy to add new enums.  If we use strings they should be backed up by defines for common usages.

ssteinbach:
Rebased onto latest master.

contrib/adapters/extern_rv.py
```diff
@@ -204,6 +204,29 @@ def _write_timeline(tl, to_session):
     return result
 
 
+def _create_media_reference(mr, to_session):
+    if hasattr(mr, "media_reference") and mr.media_reference:
+        if isinstance(mr.media_reference, otio.media_reference.External):
```
jminor: Didn't this move to core?

ssteinbach: yes!  good catch.

Issue #73, opened 2017-03-03
PySide vs PyQt


jminor:
otioview uses PySide, but some folks might need to use PyQt. If we need to support both, then this shim might help: https://github.com/mottosso/Qt.py

jminor:
We have decided to stick with PySide.

PR #160, opened 2017-09-09
PySide2 support

alexwidener:
This adds support for PySide2. It doesn't add switching support, which is what Qt.py provides. 

I saw the comment that was left on #73 and know that you're choosing to stick with PySide, so here is the update using PySide2.


ssteinbach:
This looks great.  Unfortunately, internally at least we still need pyside 1 support.  How tricky would it be to shim this or make a fallback so that those of us with pyside 1 and not 2 can still keep going?

Thanks!

alexwidener:
It's not difficult, it's just ... messy. 

You'd have to do something like

```
try:
    from PySide.QtGui import X, Y
except ImportError:
    from PySide.QtWidgets import X, Y
```

and continue on for each of the cases. That's the benefit of Qt.py, you just have 

```
from Qt import QtGui
```

even though Qt.py has it's own drawbacks like no completion stubs. 

I didn't submit this thinking it would get merged in, I just submitted it in case people _wanted_ to use it, in case they were in an environment running PySide 2 and wanted to merge the branch in themselves. We do stuff like that all the time at my current studio, picking and choosing feature we want added or fixed. 

My thought was that this would just stay open until it's needed, at which point it can be updated.

If you don't mind me asking, what are the reasons for choosing to not drop in the Qt.py file to the OTIO repository? It's MIT licensed.

melevine:
From #73, I believe the concern is related to the licensing model in PyQt*.  If that's the case, you could fork Qt.py and remove all PyQt support.  You'll just want to make sure you namespace the module to avoid collisions with something like `from opentimelineview._Qt import QtWidgets`.  This could also give you the flexibility to reinstate Qt.py if the licensing concerns ever change.

alexwidener:
If that were true, then wouldn't Qt.py be breaking the ToS/License for PyQt? 

This package is not shipping PySide1/2 or PyQt4/5 with the code. The package that's used at runtime would be determined by the end user, who is in charge of the library that they're using.

I'm not a fan of stripping the support for PyQt out of Qt.py - it will probably have unintended consequences, and honestly seems unnecessary until we receive further information about the reason for excluding Qt.py.

jminor:
The short answer is that our legal council asked us to avoid using Qt.py in OTIO due to potential licensing issues. I know that's annoying from a technical point of view, but it's what we're trying to adhere to.

I think it is fine to keep this PR open for a while in the hopes that a more elegant solution presents itself.

jminor:
FYI, we're going to revisit this soon. I have a rebase of this PR working for me locally.

alexwidener:
Let me know if I can be of help. Can't guarantee the time, but I can try.

jminor:
I filed a PR onto your PR. If that looks good to you, you can accept my PR and it will update your PR to the latest.

alexwidener:
Never received a notification for it, apologies - I'll check it out when I get home this evening!

jminor:
@avrata can you test this on Linux before we accept the PR?

avrata:
From a quick first check, I am getting some drawing issues with the borders of the clips in otioview:

otioview_pyside2

I did not have a chance to dig into why yet though.

-Robyn

alexwidener:
I can build and run on win/mac/linux tomorrow evening and see if I can reproduce this if nobody else gets to it first.

jminor:
@alexwidener that would be very helpful. Thanks!

alexwidener:
Might not be able to get to it tonight, the Qt server is giving me the appropriate wheel at a whopping 27 kB/s right now. 

```
  Downloading http://download.qt.io/snapshots/ci/pyside/5.9/latest/pyside2/PySide2-5.9.0a1.dev1526973458-5.9.6-cp36-cp36m-win_amd64.whl (99.8MB)
    57% |██████████████████▌             | 57.6MB 27kB/s eta 0:25:14
```
Looks like I'll have to get to it tomorrow night, apologies.

alexwidener:
I see the same issue on Windows with Python3.6.5, PySide2-PySide2-5.9.0a1.dev1526973458

I've tried tracking it down, but nothing has changed in how the QRectF for the items in the `Track._populate()` method is calculated, and nothing has changed in the `_BaseItem` that looks suspicious. 

My best guess is that a default behavior changed from Qt4 to Qt5 in brushes or pens that isn't immediately obvious in the Qt/PySide2 docs. Unfortunately, I'm not great in the whole painting area with Qt, so I can't be of much help.

jminor:
Hey @bashesenaxis could you take a look at this drawing glitch? I think that's the only thing holding back this PR.

bashesenaxis:
I am unsure how to commit my changes to this pull request, but they are fairly minimal. So I`ll just outline here what I did...
In the \_\_init__ of _BaseItem and Track
```python
pen = QtGui.QPen()
pen.setWidth(0)
self.setPen(pen)
```
then itemChange in _BaseItem changes like this to make sure we keep using the same pen
```
def itemChange(self, change, value):
    if change == QtWidgets.QGraphicsItem.ItemSelectedHasChanged:
        pen = self.pen()
        pen.setColor(
            QtGui.QColor(0, 255, 0, 255) if self.isSelected()
            else QtGui.QColor(0, 0, 0, 255)
        )
        self.setPen(pen)
        self.setZValue(
            self.zValue() + 1 if self.isSelected() else self.zValue() - 1
        )

    return super(_BaseItem, self).itemChange(change, value)
```
Your hunch was correct in that a default had changed. The default pen-with seems to have gone from 0 to 1:
http://doc.qt.io/qt-5/sourcebreaks.html#changes-to-qt-gui

jminor:
Oh good. @alexwidener do you have time to try @bashesenaxis's fix manually?

If not, @bashesenaxis I can show you how to make a PR onto this PR.

alexwidener:
I can test this tonight.

alexwidener:
screen shot 2018-06-13 at 8 24 01 pm

Looks like it's back to normal. 

Is the black box in the left corner of the view normal? Clicking it does nothing.

alexwidener:
Did a merge conflict resolution inside of Github just now, not around an environment to be able to test right currently. If you can, could you do a fetch and open for otioview to make sure the file path is still working after that conflict resolution?

PR #161, opened 2017-09-10
Make tests cross platform

alexwidener:
The tests were running on `*nix` only, so I moved all of the test manifest creation to happen in a temp directory and clean up the file when finished.


jminor:
Thanks @alexwidener !

PR #163, opened 2017-09-14
Add nested sequence support to OTIOView

ssteinbach:
- Add support for nested sequences to `opentimelineview` library (prior version supported nested Stacks but not nested Sequences).
- this is done by adding support for `otio.schema.Sequence` to the `StackView`
- Fix small bugs associated with using `source_range` directly instead of going through the `trimmed_range` wrapper
- Renames `Track` widget class to `TrackWidget` to disambiguate it

### DIFF
--- a/opentimelineview/timeline_widget.py
+++ b/opentimelineview/timeline_widget.py
@@ -256,8 +256,9 @@ class TrackWidget(QtGui.QGraphicsRectItem):
         self._populate()
 
     def _populate(self):
-        for i, item in enumerate(self.track):
-            timeline_range = self.track.range_of_child_at_index(i)
+        for item in self.track:
+            timeline_range = item.trimmed_range_in_parent()
+
             rect = QtCore.QRectF(
                 0,
                 0,
@@ -280,8 +281,10 @@ class TrackWidget(QtGui.QGraphicsRectItem):
                 continue
 
             new_item.setParentItem(self)
-            new_item.setX(otio.opentime.to_seconds(timeline_range.start_time) *
-                          TIME_MULTIPLIER)
+            new_item.setX(
+                otio.opentime.to_seconds(timeline_range.start_time) 
+                * TIME_MULTIPLIER
+            )
             new_item.counteract_zoom()
 
 
--- a/opentimelineview/timeline_widget.py
+++ b/opentimelineview/timeline_widget.py
@@ -61,12 +61,11 @@ class _BaseItem(QtGui.QGraphicsRectItem):
     def itemChange(self, change, value):
         if change == QtGui.QGraphicsItem.ItemSelectedHasChanged:
             self.setPen(
-                QtGui.QColor(0, 255, 0, 255) if self.isSelected()
+                QtGui.QColor(0, 255, 0, 255) if self.isSelected() 
                 else QtGui.QColor(0, 0, 0, 255)
             )
             self.setZValue(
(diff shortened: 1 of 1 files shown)


opentimelineview/timeline_widget.py
```diff
@@ -338,34 +338,32 @@ def __init__(self, stack, *args, **kwargs):
...
-            t for t in self.stack if t.kind == otio.schema.SequenceKind.Audio
-        )
+        if isinstance(self.stack, otio.schema.Stack):
+            has_video_tracks = any(
+                t.kind == otio.schema.SequenceKind.Video for t in self.stack
+            )
+            has_audio_tracks = any(
+                t.kind == otio.schema.SequenceKind.Audio for t in self.stack
+            )
+        else:
```
jminor: This would be better as an elif checking for schema.Sequence, with another else to complain about unknown composition types (which may exist in the future).

Also, self.stack could be self.composition to make it clear that it isn't always a Stack.

opentimelineview/timeline_widget.py
```diff
@@ -338,34 +338,32 @@ def __init__(self, stack, *args, **kwargs):
...
-        has_audio_tracks = any(
-            t for t in self.stack if t.kind == otio.schema.SequenceKind.Audio
-        )
+        if isinstance(self.stack, otio.schema.Stack):
+            has_video_tracks = any(
+                t.kind == otio.schema.SequenceKind.Video for t in self.stack
+            )
+            has_audio_tracks = any(
+                t.kind == otio.schema.SequenceKind.Audio for t in self.stack
+            )
```
jminor: Would be great to show tracks with unknown types also. Silently ignoring them could be super confusing to people with odd track types.

Checking the track types is also a great thing for the validator to report.

opentimelineview/timeline_widget.py
```diff
@@ -338,35 +338,45 @@ def __init__(self, stack, *args, **kwargs):
...
-        has_video_tracks = any(
-            t for t in self.stack if t.kind == otio.schema.SequenceKind.Video
-        )
-        has_audio_tracks = any(
-            t for t in self.stack if t.kind == otio.schema.SequenceKind.Audio
-        )
+        if isinstance(self.composition, otio.schema.Stack):
+            # non audio tracks are sorted into one area
+            has_video_tracks = any(
+                t.kind != otio.schema.SequenceKind.Audio
```
subinplakkannickal, line 349: Is it required otio.schema.SequenceKind.Video?

opentimelineview/timeline_widget.py
```diff
@@ -338,35 +338,45 @@ def __init__(self, stack, *args, **kwargs):
...
+                for t in self.composition
+            )
+            has_audio_tracks = any(
+                t.kind == otio.schema.SequenceKind.Audio
+                for t in self.composition
+            )
+        elif isinstance(self.composition, otio.schema.SequenceKind):
+            has_video_tracks = (
+                self.composition.kind != otio.schema.SequenceKind.Audio
```
subinplakkannickal, line 358: Is it required otio.schema.SequenceKind.Video?

ssteinbach: ah, actually look a bit closer - one is == and the other one is !=.  So this is saying anything that isn't an audio track should be assumed to be a video track, and _only_ things marked as audio tracks should be assumed to be audio tracks.  Have you seen this cause a problem with something?

PR #164, opened 2017-09-28
Otiostat program: validate and print information about your .otio file.

ssteinbach:
This PR adds the `bin/otiostat.py` program, which prints various information about `.otio` files out and can be used to validate `.otio` files.


ssteinbach:
Removing the "[WIP]" tag for now.  We're going to check this in and gather features as time goes on (if people find it useful).

PR #167, opened 2017-10-05
Burnin adapter (second attempt)

repsac:
Removed enums.
Constants are now global (as opposed to a class namespace).
Updating the MIT license with preamble.

### DIFF
--- a/contrib/adapters/ffmpeg_burnins.py
+++ b/contrib/adapters/ffmpeg_burnins.py
@@ -49,14 +49,19 @@ def _system_font():
         fonts = ('arial.ttf', 'times.ttf', 'couri.ttf')
 
     system_font = None
+    backup = None
     for font in fonts:
         font = os.path.join(font_path, font)
         if os.path.exists(font):
             system_font = font
             break
     else:
-        raise OSError("Could not determine system font.")
-    return system_font
+        if os.path.exists(font_path):
+            for each in os.listdir(font_path):
+                ext = os.path.splitext(each)[-1]
+                if ext[1:].startswith('tt'):
+                    system_font = os.path.join(font_path, each)
+    return system_font or backup
 
 
 # Default valuues
@@ -254,8 +259,6 @@ class Burnins(object):
         :param enum align: alignment, must use provided enum flags
         :param dict options:
         """
-        ifont = ImageFont.truetype(options['font'],
-                                   options['font_size'])
         resolution = self.resolution
         data = {
             'text': options.get('expression') or text,
@@ -264,10 +267,7 @@ class Burnins(object):
             'font': options['font'],
             'opacity': options['opacity']
         }
-        data.update(_drawtext(align, resolution,
-                              ifont.getsize(text),
-                              [options['x_offset'],
-                               options['y_offset']]))
+        data.update(_drawtext(align, resolution, text, options))
         self.filters['drawtext'].append(DRAWTEXT % data)
 
         if options.get('bg_color') is not None:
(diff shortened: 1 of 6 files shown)


repsac:
Test server doesn't have PIL. That is going to make this difficult.

`ImportError: No module named PIL`

contrib/adapters/ffmpeg_burnins.py
```diff
@@ -0,0 +1,391 @@
+# OpenTimelineIO bundles ffmpeg_burnins, which is available under the MIT
+# License.  For details contrib/adapters.
```
jminor: Can you move the text "OpenTimelineIO bundles ffmpeg_burnins, which is available under the MIT License.  For details contrib/adapters." to the bottom of the top-level LICENSE.txt file here:
https://github.com/PixarAnimationStudios/OpenTimelineIO/blob/master/LICENSE.txt

jminor: and mention the exact filename of the ffmpeg_burnins.py file, to clarify which file(s) the MIT license applies to.

jminor: The PIL issue can be resolved by editing the `.travis.yml` file. Look for `pip install` and add PIL to the list.

repsac: I may have found a smarter way to do this. All examples of `ffmpeg` I looked at before were using `drawbox`, but just now found one that doesn't use it. Doing some tests now (if this works it will make the code cleaner too)

repsac: You can make fun of me later for missing this (so much easier)

```
ffmpeg -loglevel panic -i BUSH0005.MOV -vf "drawtext=text='Bottom Right+50':x=1437.5:y=h-text_h-5:fontcolor=white@1.0:fontsize=48:fontfile=/System/Library/Fonts/Menlo.ttc:box=1:boxborderw=5:boxcolor=black" test.mov
```

jminor: Nice!

repsac: From a syntax perspective this is cleaner, however PIL getting the dimensions of the text is really useful and kind of needed. If you're using either of the right alignment parameters it helps to know how to offset the text from the right side (offsets appear to be calculated from the left side of the text). 

I think PIL will be needed, sorry for the complication.

repsac: The alignment parameters are a convenience so that people don't have to bang their head against a keyboard getting the right x value (which can change on resolution). If you you the right alignments and give it an offset the tool can figure out what the actual X position of the text is.

repsac:
The system font thing is going to be a pain. On my mac it uses Menlo, and even if I could figure out the system font to use on travis it is going to throw the box dimensions (for the x position) off when calculating the right alignments.

Yeah at this point I am not sure how to get this through.

jminor:
You can either just drop the test (honestly that would be fine),
or maybe bundle a specific font so it is consistent?
Something like this: https://fonts.google.com/specimen/Open+Sans maybe?

repsac:
I have an update that only parses the system font when using 'right' alignments. As long as we don't use those in the tests this may work. Still poking at it.

jminor:
Ed, this looks great, the tests all pass now, and we have your CLA on file. Is there anything else you want to change before we accept it?

repsac:
Nope, that is all I can think of for now.

jminor:
Accepted. Thanks Ed! This will be super handy.

repsac:
Awesome. Will be interesting to see how it appeals to users.

PR #172, opened 2017-10-18
Use an empty string for a locator comment if the marker name is None

michaeljefferies:


jminor:
Thanks @michaeljefferies! This change looks great.

It may seem like overkill for a small change like this, but we'll need a signed CLA from you or your company in order to accept your change. See here for details: https://github.com/PixarAnimationStudios/OpenTimelineIO/wiki/Contributing

michaeljefferies:
Sorry, finally got around to filling out the CLA and sent it in.

jminor:
Great! Thanks.

PR #175, opened 2017-10-25
Fixed "E722 do not use bare except" lint issue.

jminor:


contrib/adapters/ale.py
```diff
@@ -52,7 +52,7 @@ def _parse_data_line(line, columns, fps):
         if "Start" in metadata:
             try:
                 start = otio.opentime.from_timecode(metadata["Start"], fps)
-            except:
+            except Exception:
```
ssteinbach: Do you know which exception this would raise?  I'm assuming its a `TypeError` from `from_timecode` (since the `if "Start" in metadata` avoids the `KeyError` from the dictionary).  I would prefer that to catching `Exception`.

ssteinbach: Also, can do `metadata.pop("Start")` here and skip the `del` below.

contrib/adapters/ale.py
```diff
@@ -65,15 +65,15 @@ def _parse_data_line(line, columns, fps):
                         metadata["Duration"], fps
                     )
                     del metadata["Duration"]
-                except:
+                except Exception:
```
ssteinbach: Same thing, `TypeError` here I think.

contrib/adapters/ale.py
```diff
@@ -65,15 +65,15 @@ def _parse_data_line(line, columns, fps):
...
                     raise ALEParseError("Invalid Duration timecode: {}".format(
                         metadata["Duration"]
                     ))
             if metadata.get("End", "") != "":
                 try:
                     end = otio.opentime.from_timecode(metadata["End"], fps)
                     del metadata["End"]
-                except:
+                except Exception:
```
ssteinbach: ... and here ( + the `.pop()` if you want to compress this).

examples/shot_detect.py
```diff
@@ -215,7 +215,7 @@ def _verify_ffprobe():
         )
     try:
         out, err = proc.communicate()
-    except:
+    except Exception:
```
ssteinbach: This one I think we can leave as is, since its less clear what the child process will raise.  Googling suggests that typical types would be `OSError` and `ValueError`.

jminor:
Notes addressed.

examples/shot_detect.py
avrata: And super minor, but I wonder if this exception block works even as is due to `out` and `err` being used before they are defined (but I might be missing something).

jminor: Oops. Thanks for catching that. Fixed. Also, I've now actually tested that code path.

PR #176, opened 2017-10-31
Adding Timecode support

repsac:
Now users can add timecode burnins to their media.


contrib/adapters/ffmpeg_burnins.py
```diff
@@ -356,3 +401,11 @@ def _drawtext(align, resolution, text, options):
...
         y_pos = 'h-text_h-%d' % (options['y_offset'])
     return {'x': x_pos, 'y': y_pos}
+
+
+def _frames_to_timecode(frames, framerate):
+    return '{0:02d}:{1:02d}:{2:02d}:{3:02d}'.format(
+        int(frames / (3600*framerate)),
+        int(frames / (60*framerate) % 60),
+        int(frames / framerate % 60),
+        int(frames % framerate))
```
jminor: FYI, there's code for this in opentime.to_timecode() if you want to use it. That code will (eventually) handle drop frame as well.

repsac: Good suggestion. Tho ffmpeg_burnins.py is still a stand-alone module I use separate from `otio` and I try to keep my modules as light as possible concerning import dependencies. I'll kick around the idea later on.

Issue #174, opened 2017-10-23
FCP adaptor using file timebase instead of clipitem 


al3ph:
Noticed this when using audio clips which have a clipitem timebase of 24, and a file timebase of 30, but when viewing in otioview.py the clip is too short, it looks like its using end/30fps instead of 24fps.

					<masterclipid>masterclip-10</masterclipid>
						<name>tumbl-105.0-171013.DX_MIX.aif</name>
						<enabled>TRUE</enabled>
						<duration>5912</duration>
						<rate>
							<timebase>24</timebase>
							<ntsc>FALSE</ntsc>
						</rate>
						<start>0</start>
						<end>755</end>
						<in>0</in>
						<out>755</out>

                        "source_range": {
                            "OTIO_SCHEMA": "TimeRange.1", 
                            "duration": {
                                "OTIO_SCHEMA": "RationalTime.1", 
                                "rate": 29.97, 
                                "value": 755
                            }, 
                            "start_time": {
                                "OTIO_SCHEMA": "RationalTime.1", 
                                "rate": 29.97, 
                                "value": 0
                            }
                        }

I can supply an example if needed..

ta

al3ph:
Here we go, it's using the media framerate instead of the item framerate when setting the source_range.

def _parse_clip_item(clip_item, transition_offsets, element_map):
    markers = clip_item.findall('./marker')

    media_reference = _parse_media_reference(
        clip_item.find('./file'),
        element_map
    )

   src_rate = _parse_rate(clip_item.find('./file'), element_map)

jminor:
@al3ph can you confirm that @bashesenaxis's PR #177 fixes the issue you found?

alatdneg:
Worked it out, yup that did the trick cheers!

alatdneg:
Not sure if writing out the xml works though(I suspect that'll have the same kind of problem), as I only tested reading xml..

PR #177, opened 2017-10-31
FCP adapter uses clip item rate instead of source rate

bashesenaxis:
addresses #174 
- uses clip item rate instead of source rate
- edited clipitem in premiere_example.xml that has different rate compared to its linked file
- fixed bug in timeline_widget, transitions don`t have trimmed_range_in_parent method
- fixed bug in timeline_widget, left over rename from stack to composition

### DIFF
--- a/opentimelineio/adapters/fcp_xml.py
+++ b/opentimelineio/adapters/fcp_xml.py
@@ -274,7 +274,6 @@ def _parse_clip_item(clip_item, transition_offsets, element_map):
         clip_item.find('./file'),
         element_map
     )
-    src_rate = _parse_rate(clip_item.find('./file'), element_map)
     item_rate = _parse_rate(clip_item, element_map)
 
     # transition offsets are provided in timeline rate. If they deviate they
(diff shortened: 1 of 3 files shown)


opentimelineview/timeline_widget.py
```diff
@@ -255,8 +255,8 @@ def __init__(self, track, *args, **kwargs):
         self._populate()
 
     def _populate(self):
-        for item in self.track:
-            timeline_range = item.trimmed_range_in_parent()
+        for n, item in enumerate(self.track):
+            timeline_range = self.track.trimmed_range_of_child_at_index(n)
```
ssteinbach: @jminor this is another case where we might want to consider folding transitions back into the regular `item` pattern.

jminor: I agree.

jminor: That shouldn't block this change, we can do that later.

opentimelineio/adapters/fcp_xml.py
```diff
@@ -399,42 +418,48 @@ def _parse_track(track_e, kind, rate, element_map):
...
         transition_offsets = [
-            otio.opentime.RationalTime(),
-            otio.opentime.RationalTime()
+            otio.opentime.RationalTime(rate=rate),
+            otio.opentime.RationalTime(rate=rate)
         ]
 
         if track_item.tag == 'clipitem':
-            if start == -1:
+            if start.value == -1:
```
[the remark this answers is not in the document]
bashesenaxis: -1 is actually the values that you`ll encounter in an XML file. 
If start or end is -1, then it means that item is preceded or succeeded by a transition. As such, you`ll need to query neighboring items for the correct values.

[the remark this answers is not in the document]
jminor: Ok, got it.

opentimelineview/timeline_widget.py
ssteinbach: Agreed.

bashesenaxis:
I have rebased my changes and addressed the issue that was causing the unit test to fail. Item rate is now maintained through the read / write process.

bashesenaxis:
Cool :)
I`ve got no more changes for this.

PR #178, opened 2017-11-06
Addressing a bug with escaping slashes on Windows

repsac:
The work was primarily in the burnin module and not the adapter. Unit tests updated accordingly.


contrib/adapters/ffmpeg_burnins.py
```diff
@@ -313,6 +323,9 @@ def _add_burnin(self, text, align, options, draw):
         }
         data.update(options)
         data.update(_drawtext(align, resolution, text, options))
+        if 'font' in data and _is_windows():
+            data['font'] = data['font'].replace(os.sep, os.sep*2)
```
jminor: `os.sep*2` is misleading, it should be `r'\'+os.sep` to make it clear that you're escaping the separator (which happens to be a backslash).
Also, is there any harm in escaping these on all platforms?

jminor:
At some point, you should rebase onto the master, so your PRs don't show so many previously accepted commits.

contrib/adapters/ffmpeg_burnins.py
repsac: This is for `fontfile=`. When I moved my proxy generation tools back to Windows the single `\` was resulting in `fontfile=C:WindowsFontsarial.ttf` being passed to the subprocess, which is why I needed escapes. I am not sure how the escaping would be useful on *nix systems.

PR #179, opened 2017-11-11
Add check to see if PIL is importable before running tests.

ssteinbach:
Noticed that tests fail instead of skip if PIL isn't installed, so added a check for it.  Now unit tests skip instead of fail if PIL isn't installed for the contrib/burnin adapter.


repsac:
The one issue I have is the URL points to http://www.pythonware.com/products/pil/ and this is all older Windows versions.

`pil` actually uses Pillow (as I understand the original PIL project was abandoned).
https://python-pillow.org/

I think that should be the link.

Note: The import is the same between PIL and Pillow

PR #180, opened 2017-11-14
Time conversion

tuan-huy-truong:
Added 2 functions to convert into time string with microseconds as expected in ffmpeg

### DIFF
--- a/opentimelineio/opentime.py
+++ b/opentimelineio/opentime.py
@@ -568,6 +568,69 @@ def to_timecode(time_obj, rate):
     )
 
 
+def from_time(time_str, rate):
+    """Convert a time with microseconds string into a RationalTime.
+
+    :param time_str: (:class:`str`) A HH:MM:ss.ms time.
+    :param rate: (:class:`float`) The frame-rate to calculate timecode in
+        terms of.
+
+    :return: (:class:`RationalTime`) Instance for the timecode provided.
+    """
+
+    if ';' in time_str:
+        raise ValueError('Drop-Frame timecodes not supported.')
+
+    hours, minutes, seconds = time_str.split(":")
+
+    # Timecode is declared in terms of nominal fps
+    nominal_fps = math.ceil(rate)
+    value = (
+        (
+            # convert to frames
+            ((int(hours) * 60 + int(minutes)) * 60) + float(seconds)
+        ) * nominal_fps
+    )
+
+    return RationalTime(value, nominal_fps)
+
+
+def to_time(time_obj, rate):
+    """
+    Convert this timecode to time with microsecond, as formated in FFMPEG
+
+    :return: Number formated string of time
+    """
+    if time_obj is None:
+        return None
+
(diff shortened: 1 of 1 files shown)


opentimelineio/opentime.py
```diff
@@ -568,6 +568,69 @@ def to_timecode(time_obj, rate):
     )
 
 
+def from_time(time_str, rate):
```
jminor: Could you rename this to `from_time_string` or `from_hhmmss` or something to make it clear what the input is expected to be?

opentimelineio/opentime.py
```diff
@@ -568,6 +568,69 @@ def to_timecode(time_obj, rate):
...
+    :return: (:class:`RationalTime`) Instance for the timecode provided.
+    """
+
+    if ';' in time_str:
+        raise ValueError('Drop-Frame timecodes not supported.')
+
+    hours, minutes, seconds = time_str.split(":")
+
+    # Timecode is declared in terms of nominal fps
+    nominal_fps = math.ceil(rate)
```
jminor: This seems misleading. If I call `from_time("01:00:00.00", 24)` vs `from_time("01:00:00.00", 23.98)` I would expect to get different results. Maybe the conversion to nominal fps should be the responsibility of the caller?

tuan-huy-truong: Yep, you're right. At first I was coding from the timecode functions, but the ceiling doesn't make sense for absolute time value. I'll figure this out and create a new pull request. Also submitting my contributor licence in the day.
Thanks for the feedbacks

tuan-huy-truong:
```
time_string = "00:00:00.041666667"
t = otio.opentime.RationalTime(value=1, rate=24)
self.assertEqual(t, otio.opentime.from_time_string(time_string, 24))
AssertionError: otio.opentime.RationalTime(value=1, rate=24) != otio.opentime.RationalTime(value=1.000000008, rate=24)
```
        

I'm having some trouble with this kind of situation. 1st solution was to round my result. But I have the feeling that it won't be right with bigger values and floating fps.

Do you have any opinion in the way to figure this out?

here is the function i tried to clean, but I hesitate on what to leave
```
def from_time_string(time_str, rate):
    """Convert a time with microseconds string into a RationalTime.

    :param time_str: (:class:`str`) A HH:MM:ss.ms time.
    :param rate: (:class:`float`) The frame-rate to calculate timecode in
        terms of.

    :return: (:class:`RationalTime`) Instance for the timecode provided.
    """

    if ';' in time_str:
        raise ValueError('Drop-Frame timecodes not supported.')

    hours, minutes, seconds = time_str.split(":")

    # Timecode is declared in terms of nominal fps
    nominal_fps = math.ceil(rate)
    # here the round function to handle micro seconds losts, but
    # don't know if better put digits or round to int values
    value = round(
        (
            # convert to frames
            ((int(hours) * 60 + int(minutes)) * 60) + float(seconds)
        ) * rate)

    return RationalTime(value, rate)
```

PR #181, opened 2017-11-14
add fps arguments in functions to handle other frame rate than 24

tuan-huy-truong:
A rough way to handle custom frame rate for cmx parser

### DIFF
--- a/opentimelineio/adapters/__init__.py
+++ b/opentimelineio/adapters/__init__.py
@@ -120,8 +120,9 @@ def read_from_file(
         filepath,
         adapter_name=None,
         media_linker_name=media_linker.MediaLinkingPolicy.ForceDefaultLinker,
-        media_linker_argument_map=None
-        ):
+        media_linker_argument_map=None,
+        rate=24
+):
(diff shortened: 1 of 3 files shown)


tuan-huy-truong:
Hi, thanks for the feedbacks, I also opened Issue #182 (sorry for replicates by the way) because coding this started to taste only as a workaround as I wanted to be able to detect it instead of giving it as an argument. Thank you for the keyword arguments feature, it'll be simpler for me now to recode it.

jminor:
If you can find a reliable way to auto-detect what rate an EDL is, that would be great, but I've always seen other tools require you to specify it when you import, because EDL doesn't seem to have a place to specify it in the file itself. You could guess based on the last `:FF` segment of the timecode to see if there are any > 23, but you might get unlucky based on the specific data in the file and guess wrong which would be a huge mess.

tuan-huy-truong:
Indeed that was our conclusion as well. I haven't seen a tool able to resolve this either. Let's go for the specification then. Thanks for the answer.

PR #186, opened 2017-11-17
Time string functions

tuan-huy-truong:
New pull request for timestring formatting for FFmpeg.

Functions done by calling opentime.to_seconds to increase consistency.

It kinda works, even if some unit test doesn't works due to precision, that may be handle by the todo put in value_rescaled_to() function.

### DIFF
--- a/opentimelineio/opentime.py
+++ b/opentimelineio/opentime.py
@@ -568,7 +568,7 @@ def to_timecode(time_obj, rate):
     )
 
 
-def from_time(time_str, rate):
+def from_time_string(time_str, rate):
     """Convert a time with microseconds string into a RationalTime.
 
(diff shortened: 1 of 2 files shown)


tuan-huy-truong:
I chose to put the precision issue to the __eq__ function to avoid losts in time accumulation.

tuan-huy-truong:
Hi joshua, any feedback to improve if you don't agree to the provided solution?

jminor:
Instead of rounding inside `__eq__`, could you introduce an `approximately_equal` function with an optional precision argument, so that it is clear when rounding is important or not? That way we're not hiding the rounding inside an `==` which could be a gotcha for people. Thanks :)

tuan-huy-truong:
I'm good like so far. Thank you

PR #187, opened 2017-11-17
Custom fps cmx3600

tuan-huy-truong:
Clean branch for custom rate implementation based on adapter keyword arguments
Added 25fps edl in samples
Added test in test_cmx_3660.py


opentimelineio/adapters/cmx_3600.py
```diff
@@ -544,8 +544,9 @@ def expand_transitions(timeline):
     return timeline
 
 
-def read_from_string(input_str):
-    parser = EDLParser(input_str)
+def read_from_string(input_str, **edl_args):
+    rate = edl_args.get('rate', 24)
```
jminor: I think you could use `rate=24` instead of `**edl_args` here. I don't anticipate many more arguments, so it should be fine.

tests/test_cmx_3600_adapter.py
```diff
@@ -293,6 +294,17 @@ def test_fade_to_black_ends_with_gap(self):
...
+    def test_edl_25fps(self):
+        # EXERCISE
+        edl_path = EXEMPLE_25_FPS_PATH
+        fps = 25
+        timeline = otio.adapters.read_from_file(edl_path, rate=fps)
+        track = timeline.tracks[0]
+        self.assertEqual(track[0].source_range.duration.value, 161)
+        self.assertEqual(track[1].source_range.duration.value, 200)
+        self.assertEqual(track[2].source_range.duration.value, 86)
+        self.assertEqual(track[3].source_range.duration.value, 49)
```
jminor: Thanks for writing a test for this :)

opentimelineio/adapters/cmx_3600.py
tuan-huy-truong: I'll change it. Thanks.

tuan-huy-truong:
Sorry for several PRs. It's just that i messed up my branches here so I deleted everything and start from scratch. Will do next time. Thanks

opentimelineio/adapters/cmx_3600.py
```diff
@@ -544,8 +544,8 @@ def expand_transitions(timeline):
     return timeline
 
 
-def read_from_string(input_str, **edl_args):
-    rate = edl_args.get('rate', 24)
+def read_from_string(input_str, rate=24):
+    # rate = edl_args.get('rate', 24)
```
jminor: You could remove this comment.

opentimelineio/adapters/cmx_3600.py
```diff
@@ -590,21 +589,21 @@ def write_to_string(input_otio):
     for i, clip in enumerate(track):
         source_tc_in = otio.opentime.to_timecode(
             clip.source_range.start_time,
-            edl_rate
+            clip.source_range.start_time.rate
```
jminor: If the rate of clips in your timeline varies, then you'll get an EDL with mixed rates in it, which is unlikely to work properly. It would be better to use a single edl_rate to make sure they are all consistent.

I would suggest adding an `edl_rate=None` parameter to `write_to_string` and then add some code to detect the rate of the timeline when edl_rate is None (maybe `timeline.duration().rate`). That way the caller could override it, but it would auto-detect if you don't specify a rate.

tuan-huy-truong:
I'm ready. Thanks

jminor:
Thanks for the contribution!

tuan-huy-truong:
My pleasure.

PR #191, opened 2017-12-07
Add support for Nucoda EDLs

lsavidge:
Nucoda EDLs use a `* FROM FILE:` comment string to reference media in place of `* FROM CLIP:`.  This change introduces Nucoda EDL writing using an EDL `style` argument, and also supports Nucoda EDL reads.

### DIFF
--- a/opentimelineio/adapters/cmx_3600.py
+++ b/opentimelineio/adapters/cmx_3600.py
@@ -33,7 +33,6 @@
 # TODO: currently tracks with linked audio/video will lose their linkage when
 #       read into OTIO.
 
-import enum
 import os
 import re
 import math
@@ -73,12 +72,12 @@ channel_map = {
 
 
 # Currently, the 'style' argument determines
-# the comment string for the media reference.
+# the comment string for the media reference:
 #   'avid': '* FROM CLIP:' (default)
 #   'nucoda': '* FROM FILE:'
-class Style(enum.Enum):
-    avid = 0
-    nucoda = 1
+# When adding a new style, please be sure to add sufficient tests
+# to verify both the new and existing styles.
+VALID_EDL_STYLES = ['avid', 'nucoda']
 
 
 class EDLParser(object):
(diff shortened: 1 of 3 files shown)


opentimelineio/adapters/cmx_3600.py
```diff
@@ -33,6 +33,7 @@
 # TODO: currently tracks with linked audio/video will lose their linkage when
 #       read into OTIO.
 
+import enum
```
jminor: Python 2.7 doesn't have the `enum` module. (OTIO needs to support 2.7 and 3.5+) This came up last time we tried to use the enum module, but now I'm confused about why the Travis tests passed since it runs against both...

opentimelineio/adapters/cmx_3600.py
```diff
@@ -71,6 +72,15 @@ class EDLParseError(otio.exceptions.OTIOError):
 }
 
 
+# Currently, the 'style' argument determines
+# the comment string for the media reference.
+#   'avid': '* FROM CLIP:' (default)
+#   'nucoda': '* FROM FILE:'
```
jminor: Can you add some advice to future developers about adding new styles? Something along the lines of "When adding a new style, please make sure to add a test to verify that you are getting the results you want without disrupting the existing styles."

opentimelineio/adapters/cmx_3600.py
lsavidge: Ah, yes. Thank you. The enum34 package is required by flake8, which is why `enum` is available when the tests run and there's no error. 

Are you opposed to adding enum34 as a dependency? It seems like it could be useful elsewhere in the code as well.  I'm also fine rewriting it to not use `enum`.

jminor: Lets avoid the dependency for now. Also, maybe this explains why I'm not able to run flake8 on my machine :)

opentimelineio/adapters/cmx_3600.py
lsavidge: Done!

opentimelineio/adapters/cmx_3600.py
lsavidge: Just put up a revision that removes the `enum` dependency

lsavidge:
@jminor Nope!

PR #198, opened 2017-12-22
AAF adapter

jminor:
Here's a starting point for reading AAF compositions into OTIO.
As it stands, it can read a multitrack AAF composition, including Filler, SourceClips, Transitions and custom bin column metadata.
It does not yet support markers, nested compositions, any sort of effects, audio tracks, or a million other things.


jminor:
This is a work-in-progress towards addressing #1

markreidvfx:
For people who would like to test this, there are prebuilt pyaaf1 wheels [here](https://github.com/markreidvfx/pyaaf/releases) of the latest version for mac and windows users.

jminor:
Thanks @markreidvfx, and congrats on the 1.0.0 of pyaaf!
For OTIO, would you like us to pin to specific releases of pyaaf, or stay on the head of a specific branch?

markreidvfx:
@jminor staying at the head is preferred. I try and do a release every time I merge the dev branch into master.  Even though I've been more focused on pyaaf2, I will try and keep pyaaf1 maintained for people using it.

jminor:
Thanks @markreidvfx. We'll keep our master branch on the head of pyaaf1. However, when we do a release of OTIO, it probably makes sense to pin that to the current release of pyaaf1 at the time of the release.

jminor:
Landing this onto the AAF_adapter branch, so we can continue development there before landing to master.

PR #200, opened 2018-01-05
Support EDL dissolve writes

mikemahony:
Should address https://github.com/PixarAnimationStudios/OpenTimelineIO/issues/103.

### DIFF
--- a/opentimelineio/adapters/cmx_3600.py
+++ b/opentimelineio/adapters/cmx_3600.py
@@ -710,7 +710,7 @@ def write_to_string(input_otio, rate=None, style='avid'):
 
     edit_number = 0
     clips_and_events = []
-    for prev_child, child, next_child in lookahead_and_behind_enumerate(track):
+    for prv_child, child, nxt_child in _lookahead_and_behind_enumerate(track):
         if isinstance(child, otio.schema.Transition):
             continue
 
@@ -718,7 +718,7 @@ def write_to_string(input_otio, rate=None, style='avid'):
 
         event = EventLine(
             edit_number=edit_number,
-            reel=reel_from_clip(child),
+            reel=_reel_from_clip(child),
             kind=kind,
             rate=edl_rate
         )
@@ -733,18 +733,18 @@ def write_to_string(input_otio, rate=None, style='avid'):
         event.record_in = range_in_timeline.start_time
         event.record_out = range_in_timeline.end_time_exclusive()
 
-        if isinstance(next_child, otio.schema.Transition):
-            trans = next_child
+        if isinstance(nxt_child, otio.schema.Transition):
+            trans = nxt_child
             a_side_event = event
 
             # Shorten this, the A-side
             a_side_event.source_out -= trans.in_offset
             a_side_event.record_out -= trans.in_offset
 
-        if isinstance(prev_child, otio.schema.Transition):
+        if isinstance(prv_child, otio.schema.Transition):
             a_side, a_side_event = clips_and_events[-1]
             b_side_event = event
-            trans = prev_child
+            trans = prv_child
 
             # Add A-side cut
             cut_line = EventLine(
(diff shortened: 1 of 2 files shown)


opentimelineio/adapters/cmx_3600.py
```diff
@@ -687,123 +687,236 @@ def write_to_string(input_otio, rate=None, style='avid'):
...
+    for nxt in iterator:
+        yield (prv, cur, nxt)
+        prv, cur = cur, nxt
+    yield (prv, cur, None)
+
+
+def reel_from_clip(clip):
+    if (
+        clip.media_reference
+        and isinstance(clip.media_reference, otio.schema.Gap)
```
ssteinbach: A media_reference will never be a Gap. The `clip` argument might be a Gap.

tests/test_cmx_3600_adapter.py
```diff
@@ -415,17 +415,153 @@ def test_nucoda_edl_write(self):
...
+        expected = \
+            'TITLE: test_nucoda_timeline\n\n' \
+            '001  AX       V     C        ' \
+            '00:00:00:00 00:00:00:05 00:00:00:00 00:00:00:05\n' \
+            '* FROM CLIP NAME:  test clip1\n' \
+            '* FROM FILE: S:\\var\\tmp\\test.exr\n' \
+            '002  AX       V     C        ' \
+            '00:00:00:00 00:00:00:05 00:00:00:05 00:00:00:10\n' \
+            '* FROM CLIP NAME:  test clip2\n' \
+            '* FROM FILE: S:\\var\\tmp\\test.exr\n'
```
ssteinbach: Instead of backslashes you can wrap the whole thing in (), or you could use triple quotes.

tests/test_cmx_3600_adapter.py
```diff
@@ -415,17 +415,153 @@ def test_nucoda_edl_write(self):
...
+
+        self.assertEqual(result, expected)
+
+    def test_nucoda_edl_write_with_transition(self):
+        track = otio.schema.Track()
+        tl = otio.schema.Timeline(
+            "CrossDissolve_Day-Night_Long_1 from CZuber",
```
ssteinbach: Can you remove the person's name from this?
(same thing elsewhere in this file)

opentimelineio/adapters/cmx_3600.py
```diff
@@ -687,123 +687,236 @@ def write_to_string(input_otio, rate=None, style='avid'):
...
-                color = meta.get("color").upper()
-            comment = (marker.name or '').upper()
-            lines.append("* LOC: {} {:7} {}".format(timecode, color, comment))
 
-        # If we are carrying any unhandled CMX 3600 comments on this clip
-        # then output them blindly.
-        extra_comments = clip.metadata.get('cmx_3600', {}).get('comments', [])
-        for comment in extra_comments:
-            lines.append("* {}".format(comment))
+def generate_comment_lines(clip, style, edl_rate, from_or_to='FROM'):
```
ssteinbach: Top level functions like this should be prefixed with _ to make them hidden, so people outside this module won't accidentally find them (in help, etc.)

mikemahony: Sounds great.  Updated.  There are existing top-level functions that don't follow convention and will need similar treatment in the future.

opentimelineio/adapters/cmx_3600.py
mikemahony: Great catch!  This was existing logic but I've updated it here.

tests/test_cmx_3600_adapter.py
mikemahony: Much easier to read now. 👍

tests/test_cmx_3600_adapter.py
mikemahony: Done.  Thanks for catching!

opentimelineio/adapters/cmx_3600.py
jminor: If you're willing to _ those too, that'd be great, or I'm happy to do it after this lands.

mikemahony: Looks like there was only one more 🎉 .  Updated.

Issue #201, opened 2018-01-05
Algorithm for flattening a Stack of Tracks into one Track.


ssteinbach:
This is a common request. We should have a simple `flatten` function that does this.

jminor:
Addressed by #205

PR #205, opened 2018-01-10
Flatten Algorithm

jminor:
We now have a function that takes a Stack with many Tracks and returns a single Track with all of the items composed together.
Note: that flattening does not yet handle Transitions or nested Tracks, only Stacks, Clips and Gaps.

This also includes:
- Moved responsibility for updating parenting from _set_parent() to Composition.
- Added tests for equality and truthiness.
- Created stack_algo module.


jminor:
Addresses #201

opentimelineio/algorithms/stack_algo.py
```diff
@@ -0,0 +1,80 @@
...
+
+        track = in_stack[track_index]
+        if trim_range is not None:
+            track = track_algo.track_trimmed_to_range(track, trim_range)
+        for item in track:
+            if item.visible() or track_index == 0:
+                yield item
+            else:
+                trim = item.range_in_parent()
+                if trim_range is not None:
```
ssteinbach: when does `range_in_parent` return `None`?

opentimelineio/core/composition.py
```diff
@@ -430,9 +430,29 @@ def __getitem__(self, item):
         return self._children[item]
 
     def __setitem__(self, key, value):
-        value._set_parent(self)
+        old = self._children[key]
+        if old == value:
```
ssteinbach: should this be `is` or ==?

tests/test_composition.py
```diff
@@ -1180,12 +1234,38 @@ def test_deeply_nested(self):
         self.assertEqual(stack.transformed_time(fifty, clip), middle)
         self.assertEqual(stack.transformed_time(ninetynine, clip), last)
 
+        def _nest(self, item):
```
ssteinbach: Should this just be a member method of the test class?  that way you wouldn't have to explicitly pass in `self` below.  Could stay `_` prefixed.

Issue #207, opened 2018-01-10
Should be easy to get visible range of clip including adjacent transitions


jminor:
Currently, you have to do some work to find out the full range of media used by a clip when there is a transition into or out of that clip. OTIO should have a simple and clear way to ask for this without having to think about it much.

Maybe: `clip.visible_range()` or something like that.

jminor:
Fixed in #212

Issue #206, opened 2018-01-10
Should be easy to get just video tracks.


jminor:
There should be a convenience for getting the common types of tracks.

This pattern is too verbose:
```
        video_tracks = [
            t for t in timeline.tracks
            if t.kind == otio.schema.TrackKind.Video
        ]
        audio_tracks = [
            t for t in timeline.tracks
            if t.kind == otio.schema.TrackKind.Audio
        ]
```

jminor:
Fixed in #212

Issue #211, opened 2018-01-24
schema.track.neighbors_of is ambiguous as to order of neighbors


ssteinbach:
If you don't have neighbors of expand transitions, then its return value is ambiguous as to whether the return value is before or after.  Instead, return None if the neighbor doesn't exist.

Examples (assume all are clips):
```
[A, B, C] :: neighbors_of(B) -> (A, C)
[A, B, C] :: neighbors_of(A) -> (None, B)
[A, B, C] :: neighbors_of(C) -> (B, None)
[A] :: neighbors_of(A) -> (None, None)
```
In other words, it always should always return `(previous or None, next or None)`.  In the case of expanded transitions it would return `(previous or Gap, next or Gap)`.

Right now the last one for example returns `(A)` which is not particularly helpful.

jminor:
Fixed in #212

PR #212, opened 2018-01-24
Convenience methods for tracks, visible_range, etc.

jminor:
Added `video_tracks()` and `audio_tracks()` accessors to Timeline.
Added `visible_range()` to Item, to make it easy to get the range of media used including any adjacent Transitions in a Track.
Fixed `trimmed_range()` and `duration()` so they return copies of the TimeRange or RationalTime object, preventing the caller from inadvertently modifying stuff.

Addresses #206 
Addresses #207


jminor:
Also made it easier to use flatten like this:
```
flattened_track = otio.algorithms.flatten_stack(timeline.video_tracks())
```

opentimelineio/core/item.py
```diff
@@ -101,6 +101,13 @@ def trimmed_range(self):
 
         return self.available_range()
 
+    def visible_range(self):
```
avrata: Great! 

However would it make sense to also have clear `visible_range()` and `visible_range_in_parent()`? This seems to return one or the other depending on if there is a parent or not. 

Lets say we have a track with "[Head_Leader, Clip_A, 24frame_Transition, Clip_B]" (A and B being 100 frame clips). I could see wanting to ask Clip_B for its `visible_range()` (Lets say 101-200), `trimmed_range()` (113-200),  `visible_range_in_parent()` (268-367), and `trimmed_range_in_parent()` (280-367).. (I could totally be off by one or so in those numbers somewhere). 

As to what to return when there is no parent? `trimmed_range()` seems fine.

opentimelineio/schema/track.py
```diff
@@ -96,6 +96,25 @@ def trimmed_range_of_child_at_index(self, index, reference_space=None):
...
 
+    def visible_range_of_child(self, child):
+        """The range of this child's media visible to its parent Track.
+        Includes handles revealed by adjacent transitions (if any)."""
+
+        result = child.trimmed_range()
+
+        index = self.index(child)
+        if index > 0:
+            before = self[index-1]
```
ssteinbach: I'm (hoping) you can use the new `neighbors_of` implementation as a cleaner form of this.

opentimelineio/core/item.py
jminor: Yeah, I see what you're getting at. I'm planning to add an "Anatomy of a Clip" section to the wiki diagrams that we just updated which should address this directly. The differences between all these *_range() methods are confusing without pictures, which is kind of a bug in the API.

opentimelineio/schema/track.py
jminor: Done.

ssteinbach:
Also addresses #211

jminor:
@ssteinbach can you take another look? I replaced visible_range_of_child() with handles_of_child() which gets more directly at what visible_range() needs from its parent.

PR #213, opened 2018-01-24
Usability improvements to the schema.track.neighbors_of method.

ssteinbach:
Switch track.neighbors_of to return a collections.namedtuple: `(previous, next)`

- now returns a `collections.namedtuple`, so `result.previous` is always the previous item, and `result.next` is always the next one
- if set to not expand around transitions or if the item is not a transition, will return None as the next/previous item.
- Better documentation on how the method behaves.


ssteinbach:
Addresses #211

ssteinbach:
Closed as its a part of #212

PR #220, opened 2018-01-30
Add functions for filtering OTIO hierarchies.

ssteinbach:
Adds functions for filtering OTIO files:

- `otio.algorithms.filtered_items(<input otio object>, <unary function>)` - if the unary function returns an object, that object is placed into the hierarchy.  If it returns none, it is pruned.  Note that if you delete children, it will still visit them.
- `otio.algorithms.filtered_items_sequence_context(<input otio object>, <ternary function>)` ternary function that takes (prev, child, next) for items in tracks.  Otherwise will get passed (None, child, None).  behaves like above.
- both functions have arguments `types_to_prune` and `types_to_filter`.  
- If `types_to_prune` is not `None`, each item that is derived from a type in the `types_to_prune` list will be pruned from the result (the function will not be called on it)
- if `types_to_filter` is not `None`, each item that is not derived from a type in type to filter will be copied automatically to the result (the filter function will not be called on it either).
- if neither is specified or neither condition is met, then the object will be filtered by the passed in function.

Idea is to make common transformation patterns easier to write on OTIO objects - "turn all the transitions into shots" or "all gaps into a specific clip with media reference"

### DIFF
--- a/opentimelineio/algorithms/filter.py
+++ b/opentimelineio/algorithms/filter.py
@@ -41,7 +41,7 @@ def filtered_items(
     input_object,
     unary_filter_fn,
     types_to_prune=None,
-    types_to_pass_through=None,
+    types_to_filter=None,
 ):
     """Filter input_object with unary_filter_fn to make a new copy of object.
 
@@ -116,10 +116,7 @@ def filtered_items(
         if (types_to_prune and _isinstance_in(child, types_to_prune)):
             result = None
         # then try to pass through
-        elif (
-            types_to_pass_through
-            and _isinstance_in(child, types_to_pass_through)
-        ):
+        elif types_to_filter and not _isinstance_in(child, types_to_filter):
             result = child
         # finally call the user function
         else:
@@ -151,7 +148,7 @@ def filtered_with_sequence_context(
     input_object,
     reduce_fn,
     types_to_prune=None,
-    types_to_pass_through=None,
+    types_to_filter=None,
 ):
     """Filter input_object with reduce_fn to make a new copy of object.
 
@@ -211,14 +208,15 @@ def filtered_with_sequence_context(
     So the result is:
         [A, D, E, C]
 
-    Additionally, types_to_prune and types_to_pass_through let you skip or 
+    Additionally, types_to_prune and types_to_filter let you skip or 
     prune classes without having to build logic into your function.
 
     The order of operation is:
         for each thing to iterate on: 
             1 if it isinstance of anything in the prune list, prune it.
-            2 if it isinstance of anything in the pass_through_list, pass it 
-                through, without giving it to your function
+            2 if it is not isinstance of anything in the types_to_filter, 
+                or the types_to_filter is None, pass it through, without 
+                giving it to your function
             3 otherwise, call the reduce_fn and process according to the result.
     """
 
@@ -265,10 +263,7 @@ def filtered_with_sequence_context(
         if (types_to_prune and _isinstance_in(child, types_to_prune)):
             result = None
         # then try to pass through
-        elif (
-            types_to_pass_through
-            and _isinstance_in(child, types_to_pass_through)
-        ):
+        elif types_to_filter and not _isinstance_in(child, types_to_filter):
             result = child
         # finally call the user function
         else:
--- a/opentimelineio/algorithms/filter.py
+++ b/opentimelineio/algorithms/filter.py
@@ -29,7 +29,6 @@ import copy
 import opentimelineio as otio
 
 
-
 def _filtered_iterable(iterable, unary_filter_fn):
     # filter the root object
     new_parent = unary_filter_fn(copy.deepcopy(iterable))
@@ -49,10 +48,11 @@ def _filtered_iterable(iterable, unary_filter_fn):
 
     return new_parent
 
+
 def filtered_items(input_object, unary_filter_fn):
-    """Create a new copy of input_object whose children have been processed by 
-    unary_filter_fn.  If unary_filter_fn returns an object, that object will 
-    get placed there, otherwise if unary_filter_fn returns none, it will be 
+    """Create a new copy of input_object whose children have been processed by
+    unary_filter_fn.  If unary_filter_fn returns an object, that object will
+    get placed there, otherwise if unary_filter_fn returns none, it will be
     skipped.
     """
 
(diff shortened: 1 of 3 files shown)


opentimelineio/algorithms/filter.py
```diff
@@ -0,0 +1,285 @@
...
+def _is_in(thing, container):
+    return any(thing is item for item in container)
+
+
+def _isinstance_in(child, typelist):
+    return any(isinstance(child, t) for t in typelist)
+
+
+def filtered_items(
+    input_object,
```
reinecke: `input_object` feels a little abstract/ambiguous. It looks like it's made to take either `Composition` or `Timeline`. I don't know if we have a general term for those, but maybe the documentation could call out that it expects one of those.

opentimelineio/algorithms/filter.py
```diff
@@ -0,0 +1,285 @@
...
+    unary_filter_fn,
+    types_to_prune=None,
+    types_to_filter=None,
+):
+    """Filter input_object with unary_filter_fn to make a new copy of object.
+
+    unary_filter_fn will be called with the objects from input_object passed in
```
reinecke: again, `objects` is a bit ambiguous a term. Maybe use `item` instead since this is expecting things of the OTIO `Item` schema.

opentimelineio/algorithms/filter.py
```diff
@@ -0,0 +1,285 @@
...
+
+def _isinstance_in(child, typelist):
+    return any(isinstance(child, t) for t in typelist)
+
+
+def filtered_items(
+    input_object,
+    unary_filter_fn,
+    types_to_prune=None,
+    types_to_filter=None,
```
reinecke: Could use some documentation for `types_to_prune` and `types_to_filter`, just to make clear that these should be python classes. I'm assuming a typical thing might be to put something like `Transition` in to create a new Track with Transitions scrubbed?

opentimelineio/algorithms/filter.py
```diff
@@ -0,0 +1,285 @@
...
+):
+    """Filter input_object with unary_filter_fn to make a new copy of object.
+
+    unary_filter_fn will be called with the objects from input_object passed in
+    as arguments one a a time, including the input_object itself, in a depth
+    first traversal order.
+
+    If unary function returns an object, that object will be inserted into the
+    hierarchy, replacing whichever object was passed in to it.  If it returns
```
reinecke: Maybe for clarity this could say `replacing whichever object was passed in to it in the output.` This way it's clear that this isn't touching the original.

opentimelineio/algorithms/filter.py
```diff
@@ -0,0 +1,285 @@
...
+    If your unary function is:
+        def fn(thing):
+            if thing.name == B:
+                return tuple(B_1,B_2,B_3)
+            else:
+                return thing
+
+    filtered_items(track, fn) => [A,B_1,B_2,B_3,C]
+    """
+
```
reinecke: If I'm reading the diff right, could this implementation be:
```python
def filtered_items(
    input_object,
    unary_filter_fn,
    types_to_prune=None,
    types_to_filter=None,
):
    def reduce_fn(_, thing, _):
        return unary_filter_fn(thing)

    return filtered_with_sequence_context(
        input_object,
        reduce_fn,
        types_to_prune,
        types_to_filter,
    )
```

If that's the case, do we need both in the library?

opentimelineio/algorithms/filter.py
```diff
@@ -0,0 +1,285 @@
...
+    return None
+
+
+def filtered_with_sequence_context(
+    input_object,
+    reduce_fn,
+    types_to_prune=None,
+    types_to_filter=None,
+):
+    """Filter input_object with reduce_fn to make a new copy of object.
```
reinecke: This one-line description is the same as the one for `filtered_items`. Maybe it could be differentiated by being something like: `Filter input_object with reduce_fn to make a new copy of object given the context of surrounding items.`

opentimelineio/algorithms/filter.py
```diff
@@ -0,0 +1,285 @@
...
+        def fn(prev_item, thing, next_item):
+            if prev_item.name == A:
+                return None # prune the clip
+            else:
+                return thing
+
+    if filtered_with_sequence_context(track, fn) is called, the order will be:
+        fn(None, A, B) => A
+        fn(A, B, C) => None
+        fn(B, C, D) => C # !! note that it was passed B instead of D.
```
lsavidge: I believe this should read `note that it was passed B instead of None`

opentimelineio/algorithms/filter.py
```diff
@@ -0,0 +1,285 @@
...
+        def fn(prev_item, thing, next_item):
+            if prev_item.name == A:
+                return (D, E) # tuple of new clips
+            else:
+                return thing
+
+    if filtered_with_sequence_context(track, fn) is called, the order will be:
+        fn(None, A, B) => A
+        fn(A, B, C) => (D, E)
+        fn(B, C, D) => C # !! note that it was passed B instead of D.
```
lsavidge: Same here, something like `note that it was passed B instead of E` I think makes a little more sense

opentimelineio/algorithms/filter.py
```diff
@@ -0,0 +1,285 @@
...
+
+    Additionally, types_to_prune and types_to_filter let you skip or
+    prune classes without having to build logic into your function.
+
+    The order of operation is:
+        for each thing to iterate on:
+            1 if it isinstance of anything in the prune list, prune it.
+            2 if it is not isinstance of anything in the types_to_filter,
+                or the types_to_filter is None, pass it through, without
+                giving it to your function
```
lsavidge: My read of this is that if `types_to_filter` is `None` (which is the default), then we'll pass it through without giving it to the function - which I don't think is the case...?  Instead what about replacing `or the types_to_filter is None` with `and the types_to_filter is not None`?

tests/test_filter_algorithms.py
```diff
@@ -0,0 +1,386 @@
...
+
+class OTIOAssertions(object):
+    def assertJsonEqual(self, known, test_result):
+        """Convert to json and compare that (more readable)."""
+        self.maxDiff = None
+
+        known_str = otio.adapters.write_to_string(known, 'otio_json')
+        test_str = otio.adapters.write_to_string(test_result, 'otio_json')
+
+        self.assertMultiLineEqual(known_str, test_str)
```
lsavidge: This class seems like it could be useful in a bunch of places.  It might more logically live in tests/utils.py, or maybe even its own file.

lsavidge:
Thanks @ssteinbach, this is looking great!

tests/test_filter_algorithms.py
ssteinbach: Yes, good call.  I'll make an issue to bump this up, that might fit better in the next release than this one.

reinecke:
Looks like a really good and useful add, I like the simplifications!

Issue #222, opened 2018-02-01
Proper iteration when there are duplicate clips


michaeljefferies:
What is the proper way to iterate over clips in a track when there is the possibility that there are duplicate clips, but you want to get the range of the clip?

```
for clip in track.each_clip():
   r = track.range_of_child(clip)   # <- will always be the first clip when there are dupes
```

It seems like we need something like this:

```
for i, clip in track.enumerate_clip():
   r = track.range_of_child_at_index(i)
```

jminor:
OTIO doesn't support instancing, so if you have the same clip multiple times, they need to be copies, not the same Python object. Sorry the documentation isn't clear on that yet. Would it help for us to add some runtime checking to make this clear when developing?

You can (at the moment) have multiple clips that share the same media_reference, but even that will get turned into copies if you serialize and deserialize your timeline to/from disk.

The idea is that these objects are so light weight that there is no harm in copying them.

Do you have a use case that requires instancing? If so, we'd like to know more about it.

michaeljefferies:
Unless I'm mistaken, it's not an issue of instancing.  OTIO is using the python list index() call to determine the index of the child of the parent.  python defines index() as "Return the index in the list of the first item whose value is x."  The key is value, so if two clips are exact copies of each other, they have the same value.

The file I have has the same clip repeated several times in an fcp xml file.  When I read in the file, it will only return the first clip it finds.  My solution for now is to do this:

```
for track in timeline.tracks:
   for child in track.each_child():
      if hasattr(child, 'metadata'):
         child.metadata['uuid'] = uuid.uuid4().hex
```

That way, each clip will have a unique value.

jminor:
Ah, I think I see what you're saying. Since index() uses `==` instead of `is` then two clip objects that happen to be identical will get confused with each other. We found some similar issues like this recently in other areas of the API. Let me write a test to repro this and we'll try to fix it.

jminor:
Does #223 look like a reasonable fix for this?

I see one more call to index() inside the fcp_xml adapter which may need to be fixed as well. If you can test #223 on your sample, or send me your xml file for verification, then we can make sure that it works for you.

michaeljefferies:
Yes, that fixed things.  The index() call inside fcp_xml must be okay, since the xml file has fields like start and end that are different for the duplicate clips.

jminor:
Great! I'll fix up that one test and get this landed into master early next week. Thanks for reporting this!

[...]

PR #223, opened 2018-02-01
Fix iteration of dupes

jminor:
If you have duplicate Clips in a Track (identical, but not instances of the same object) then range_of_child() and other methods don't work as expected. This is because we were using index() to find objects in a Composition, but index() uses the `==` operator instead of `is`.

This PR introduces index_of_child() and fixes a range_of_child() and other related methods to work properly when duplicate clips are present.

Addresses #222


ssteinbach:
I think this is related to #228.  I suspect that making these types reference types instead of value types will map better to the way people expect them to behave and clarify a lot of these kinds of things.

ssteinbach:
We will want to revisit this, but we're hitting a case where we need to integrate this now to cover some bugs.

PR #227, opened 2018-02-08
Correctly represent Gaps

mikemahony:
`Gap`s were incorrectly being written as EDL events.  This will produce an actual gap in the record timecode.

### DIFF
--- a/opentimelineio/adapters/cmx_3600.py
+++ b/opentimelineio/adapters/cmx_3600.py
@@ -770,6 +770,7 @@ class EDLWriter(object):
         events = []
         for idx, child in enumerate(track):
             if isinstance(child, otio.schema.Transition):
+                # Transition will be captured in subsequent iteration.
                 continue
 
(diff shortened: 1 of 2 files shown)


opentimelineio/adapters/cmx_3600.py
```diff
@@ -775,26 +775,28 @@ def get_content_for_track_at_index(self, idx, title):
...
+                    )
                 )
-            else:
-                event = Event(
-                    child,
-                    self._tracks,
-                    track.kind,
-                    self._rate,
-                    self._style
+            elif not isinstance(child, otio.schema.Gap):
```
ssteinbach: I follow what the code is doing, but can you write a comment explaining this if statement?  Is it just that in EDL gaps are expressed as missing record timeline?

jminor:
In an earlier version of the EDL adapter, we treated BL events as Gaps. If that's no longer the case, we should just document the difference between a black segment and a gap.

PR #230, opened 2018-02-09
AAF adapter

jminor:
Added an adapter for reading Advanced Authoring Format (AAF) compositions into OTIO.
This contrib adapter requires PyAAF 1.0.0 to work: https://github.com/markreidvfx/pyaaf/releases


jminor:
Addresses #1

PR #232, opened 2018-02-14
[WIP] functions for trimming and translating time across hierarchy

ssteinbach:
Adds the `range_of` function, which:
- Lets you ask for the range of an OTIO object in the space of a parent or child
- with the trims applied from a print or child
- Examples: 
  - what range of the top level timeline does this clip occupy?
  - Which frame of the media is playing at frame 50 in the top timeline?

Additionally:
- `Item` has a `local_to_parent_transform` function and `Composable` has a `local_to_child_transform` method, which take into consideration effects which impact timing
- `TimeTransform`:
  - `TimeTransform` now has `scale` and `offset`, but no more rate argument and the math is clarified for how they compose with each other and with `RationalTime`/`TimeRange`.
  - `TimeTransform` has an `inverted` method which returns an inverted transform
  - `TimeTransform` behaves like a traditional transform matrix
- Fixed a bug in the `clamped` method on `TimeRange`

TODO:
- Add support for expanding transitions
- handle invert for scale of zero (freeze-frames create a 0 scale transform)
- smaller stuff:
  - refactor path stuff out
  - support up-and-down searches using common parent and getting relative times in sibling scopes (not just up or down)

### DIFF
--- a/opentimelineio/algorithms/time_transforms.py
+++ b/opentimelineio/algorithms/time_transforms.py
@@ -35,66 +35,65 @@ def _transform_range(range_to_transform, from_space, to_space, trim_to=None):
     found_trim      = trim_to is None
     found_transform = to_space is None 
 
-    # child_space = from_space
-    # parent_space = to_space
-    # if not parent_space.is_parent_of(child_space):
+    child_space = from_space
+    parent_space = to_space 
+    if (
+        (parent_space and not parent_space.is_parent_of(child_space))
+        or (not parent_space and child_space.is_parent_of(trim_to))
+    ):
+        child_space = to_space if to_space is not None else trim_to
+        parent_space = from_space
 
-    child_space = to_space if to_space is not None else trim_to
-    parent_space = from_space
+        while not found_transform or not found_trim:
+            parent_range = child_space.range_in_parent()
+            if not found_trim:
+                trim_parent_range = child_space.trimmed_range_in_parent()
+                if trim_parent_range.start_time != parent_range.start_time:
+                    start_max = max(trim_parent_range.start_time, parent_range.start_time)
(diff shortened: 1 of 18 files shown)


opentimelineio/algorithms/time_transforms.py
```diff
@@ -0,0 +1,107 @@
...
+    item      :: otio.core.Item
+    in_scope  :: otio.core.Item (if None, will default to item space)
+    trimmed_to:: otio.core.Item (if None, will default to item space)
+
+    For example:
+        # time ranges are written as (start frame, duration) in 24hz.
+        A1:: Clip       trimmed_range = (10, 20)
+        T1:: Track      [A1], trimmed_range = (2, 5)
+
+        range_of(A, in_scope=A1, trimmed_to=A1)  => (10, 20)
```
avrata: Should these use "A1" for the item (instead of "A") in this example?

opentimelineio/algorithms/time_transforms.py
```diff
@@ -0,0 +1,107 @@
...
+        from_space = from_space.parent()
+
+    return range_to_transform
+
+
+# @TODO: CURRENTLY ONLY WORKS IF item is a child of IN_SCOPE AND TRIMMED_TO
+def range_of(
+    item,
+    # @TODO: scope?  Or space? reference frame?
+    in_scope=None,    # must be a parent or child of item (Default is: item)
```
jminor: Maybe `relative_to` ?

ssteinbach: good note!

ssteinbach:
This work is proceeding along a different direction, expect a PR soon.

PR #233, opened 2018-02-15
Added minimal support for EDL speed effects.

jminor:
This only understands EDL FREEZE FRAME and M2 motion effects. It doesn't turn them into OTIO Effects, it just fudges the Clip source_range to match the EDL's record timecode, so the overall timing of the timeline will be correct, but it doesn't properly represent the correct source media that should be displayed.


jminor:
Also, the test EDL for this is fairly long, which uncovered a performance issue related to copy/deepcopy that I've attempted to address.

opentimelineio/adapters/cmx_3600.py
```diff
@@ -126,17 +126,26 @@ def add_clip(self, line, comments, rate=24):
...
+                motion = comment_handler.handled.get('motion_effect')
+                freeze = comment_handler.handled.get('freeze_frame')
+                if motion is not None or freeze is not None:
+                    # Adjust the clip to match the record duration
+                    # TODO: Once OTIO has speed effects, use them here.
+                    clip_handler.clip.source_range.duration = rec_duration
+
+                elif self.ignore_timecode_mismatch:
+                    # Pretend there was no problem by adjusting the record tc
+                    record_out = record_in + src_duration
```
avrata: Still a little weirded out that `record_out` is never used again, but I kinda understand why.

opentimelineio/opentime.py
```diff
@@ -41,6 +41,18 @@ def __init__(self, value=0, rate=1):
         self.value = value
         self.rate = rate
 
+    def __copy__(self):
+        return self.__deepcopy__()
+
+    def __deepcopy__(self, memodict={}):
```
avrata: Not at all related to my request, but was curious about memodict as I have never used that. However this did trigger my "no-multiple objects as default arguments" concern: 

https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html

jminor: Yeah, that thought occurred to me too. I want to ask @ssteinbach about it, as that default argument was present in other code already.

jminor:
FYI, I A/B tested a few variations of deepcopy implementations. The best I could do was to get the core-tests to run in 3.5s (as opposed to 15s without this optimization).

It looks like deepcopy is known to be slow, presumably because it is trying to deal with cycles: https://stackoverflow.com/questions/24756712/deepcopy-is-extremely-slow

In our case, we know that TimeRange and RationalTime are meant to be value types, so we can take a shortcut that saves a lot of overhead.

PR #237, opened 2018-02-27
Setuptools plugins

reinecke:
Submit for discussion, an approach to allow for plugin modules to be registered using setuptools entry points.

If the approach looks reasonable, I can do a pass to further document and finalize.

Here is an example of how to try this out:
```[ereinecke@airwolf:~/github/OpenTimelineIO]$ virtualenv plugintest
[ereinecke@airwolf:~/github/OpenTimelineIO]$ virtualenv plugintest
[ereinecke@airwolf:~/github/OpenTimelineIO]$ source plugintest/bin/activate
(plugintest) [ereinecke@airwolf:~/github/OpenTimelineIO]$ pip install .
(plugintest) [ereinecke@airwolf:~/github/OpenTimelineIO]$ cd examples/sample_plugin/
(plugintest) [ereinecke@airwolf:~/github/OpenTimelineIO]$ pip install .
(plugintest) [ereinecke@airwolf:~/github/OpenTimelineIO]$ cd ../..
(plugintest) [ereinecke@airwolf:~/github/OpenTimelineIO]$ python
>>> otio.adapters.read_from_string('6', 'otio_counter')
```

That call will return you a timeline with 6 tracks, per the adapter implementation.

### DIFF
--- a/examples/sample_plugin/otio_counter/__init__.py
+++ b/examples/sample_plugin/otio_counter/__init__.py
@@ -1,4 +1,5 @@
-import os
+import pkg_resources
+
 from opentimelineio.plugins import manifest
 
 """
@@ -42,6 +43,6 @@ as below.
 
 
 def plugin_manifest():
-    return manifest.manifest_from_file(
-        os.path.join(os.path.dirname(__file__), 'plugin_manifest.json')
+    return manifest.manifest_from_string(
+        pkg_resources.resource_string(__name__, 'plugin_manifest.json')
     )
--- a/examples/sample_plugin/otio_counter/__init__.py
+++ b/examples/sample_plugin/otio_counter/__init__.py
@@ -1,6 +1,45 @@
 import os
 from opentimelineio.plugins import manifest
 
+"""
+In the ``setup.py`` the top-level module was provided as the entry point. When
+OTIO loads a plugin, it then looks for a manifest object provided by the
+``plugin_manifest`` function. In this case we deserialize it from a json file
+we include in the module.
+
+Below is an example of what the manifest json might look like (also included
+as a file in this package.)
+
+.. codeblock:: JSON
+
+    {
+        "OTIO_SCHEMA" : "PluginManifest.1",
+        "adapters": [
+            {
+                "OTIO_SCHEMA": "Adapter.1",
+                "name": "track_counter",
+                "execution_scope": "in process",
+                "filepath": "adapter.py",
+                "suffixes": ["count"]
+            }
+        ]
+    }
+
+
+While this example shows an adapter, you may also define Media Linker plugins
+as below.
+
+.. codeblock:: JSON
+
+    {
+        "OTIO_SCHEMA" : "MediaLinker.1",
+        "name" : "linker_example",
+        "execution_scope" : "in process",
+        "filepath" : "linker.py"
+    }
+
+"""
+
 
 def plugin_manifest():
     return manifest.manifest_from_file(
--- a/examples/sample_plugin/otio_counter/__init__.py
+++ b/examples/sample_plugin/otio_counter/__init__.py
@@ -4,12 +4,18 @@ from opentimelineio.plugins import manifest
 
 """
 In the ``setup.py`` the top-level module was provided as the entry point. When
-OTIO loads a plugin, it then looks for a manifest object provided by the
-``plugin_manifest`` function. In this case we deserialize it from a json file
-we include in the module.
+OTIO loads a plugin, it will look for ``plugin_manifest.json`` in the package
+and register the plugins declared.
+
+If you wish to programmatically generate the manifest, you may return a
+:class:`Manifest` instance from your entry point's ``plugin_manifest``
+function.
+
+For example purposes, this simply deserializes it from the json file included
+in the package.
 
 Below is an example of what the manifest json might look like (also included
-as a file in this package.)
+as a file in this package).
 
 .. codeblock:: JSON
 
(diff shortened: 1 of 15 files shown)


opentimelineio/plugins/manifest.py
```diff
@@ -143,6 +145,19 @@ def load_manifest():
...
+        plugin_name = plugin.name
+        try:
+            plugin_entry_point = plugin.load()
+            plugin_manifest = plugin_entry_point.plugin_manifest()
+        except Exception:
+            logging.exception("could not load plugin: {}".format(plugin_name))
+            continue
+
+        result.adapters.extend(plugin_manifest.adapters)
+        result.adapters.extend(plugin_manifest.media_linkers)
```
mikemahony: Looks like this should be `result.media_linkers.extend`, yah?

mikemahony:
I really like this, @reinecke!  I think for our internal use we might like to add all our adapters and media linkers to a single plugin - any concern there?

image

Just that one `media_linker` comment - otherwise seems straightforward!

opentimelineio/plugins/manifest.py
reinecke: Good catch, fixed!

reinecke:
@mikemahony I think bundling adapters and linkers is a great idea! There are multiple cases where several adapters are related in a way where it would make sense to bundle them (e.x. ALE and EDL).

I'm going to try and do a pass at making the contrib adapters gracefully work as one of these modules and then I'll remove the `WIP` status.

jminor:
I like the direction this is going.

opentimelineio/plugins/manifest.py
```diff
@@ -143,6 +145,19 @@ def load_manifest():
     except ImportError:
         pass
 
+    # Discover setuptools-based plugins
+    for plugin in pkg_resources.iter_entry_points("opentimelineio.plugins"):
+        plugin_name = plugin.name
+        try:
+            plugin_entry_point = plugin.load()
+            plugin_manifest = plugin_entry_point.plugin_manifest()
+        except Exception:
```
ssteinbach: Is this a blank exception because we're not sure what kinds of exceptions could get thrown out of the other plugin's plugin loading?

reinecke: That was the thought. I didn't put too much thought into this behavior, I just implemented it this way to match the contrib adapter handling above.

That said, since people have a choice about the setuptools plugins (unlike the contrib ones) maybe it makes sense to have them hard  fail if there is an issue.

reinecke:
Things for me have gotten a bit crazy, but I'll try to make time to get this PR finalized soon. Sorry for the delay.

reinecke:
Updated with documentation and tests. Let me know what y'all think!

setup.py
```diff
@@ -151,8 +151,11 @@ def test_otio():
             'coverage==4.5',
         ]
     },
+
     test_suite='setup.test_otio',
 
+    tests_require=['mock;python_version<"3.3"'],
```
reinecke: The mock library became standard in Python 3.3, let me know if it's okay to introduce this dependency for testing only. If not, I can try and work out a testing strategy that omits it, but it may be pretty tricky.

examples/sample_plugin/otio_counter/__init__.py
```diff
@@ -0,0 +1,47 @@
...
+        "OTIO_SCHEMA" : "MediaLinker.1",
+        "name" : "linker_example",
+        "execution_scope" : "in process",
+        "filepath" : "linker.py"
+    }
+
+"""
+
+
+def plugin_manifest():
```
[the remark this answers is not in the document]
reinecke: Do you mean autodetect `plugin_manifest.json`? I like that idea. let me take a look, should be doable.

opentimelineio/plugins/manifest.py
```diff
@@ -143,6 +145,19 @@ def load_manifest():
     except ImportError:
         pass
 
+    # Discover setuptools-based plugins
+    for plugin in pkg_resources.iter_entry_points("opentimelineio.plugins"):
+        plugin_name = plugin.name
+        try:
+            plugin_entry_point = plugin.load()
+            plugin_manifest = plugin_entry_point.plugin_manifest()
```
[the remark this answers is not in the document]
reinecke: I like that idea, I don't think it should be an issue.

tests/baselines/plugin_module/otio_mockplugin.egg-info/PKG-INFO
```diff
@@ -0,0 +1,11 @@
+Metadata-Version: 1.0
+Name: otio-mockplugin
+Version: 1.0.0
+Summary: Stands in for an actual plugin for testing.
+Home-page: http://opentimeline.io
+Author: Pixar Animation Studios
+Author-email: opentimelineio@pixar.com
+License: UNKNOWN
```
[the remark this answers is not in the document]
reinecke: I'll reconcile these values against the `PKG-INFO` that is generated with the main OTIO package.

tests/test_plugin_detection.py
```diff
@@ -0,0 +1,102 @@
...
+import logging
+import os
+import pkg_resources
+
+try:
+    # Python 3.3 forward includes the mock module
+    from unittest import mock
+except ImportError:
+    # Fallback for older python
+    import mock
```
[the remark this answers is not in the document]
reinecke: sounds good!

tests/test_plugin_detection.py
```diff
@@ -74,29 +73,46 @@ def setUp(self):
...
-            side_effect=Exception
-        )
+        # Make sure the linker is included in the linker list
+        linker_names = [linker.name for linker in man.media_linkers]
+        self.assertIn('mock_linker', linker_names)
 
-        # Ensure we can load the manifest, safely
+        # Make sure adapters and linkers landed in the proper place
+        for adapter in man.adapters:
+            self.assertTrue(isinstance(adapter, otio.adapters.Adapter))
```
ssteinbach: There is a `assertIsInstance` function,: 
https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertIsInstance

PR #240, opened 2018-03-13
opentime.py.to_timecode: Use given opentime object's rate for default

rogergodspeed:
If no rate is provided then use the rate of the of the given opentime
object.

### DIFF
--- a/opentimelineio/opentime.py
+++ b/opentimelineio/opentime.py
@@ -556,18 +556,20 @@ def from_timecode(timecode_str, rate):
     return RationalTime(value, nominal_fps)
 
 
-def to_timecode(time_obj, rate):
+def to_timecode(time_obj, rate=None):
     """Convert a RationalTime into a timecode string.
 
     :param time_obj: (:class:`RationalTime`) instance to express as timecode.
     :param rate: (:class:`float`) The frame-rate to calculate timecode in
-        terms of.
+        terms of. (Default time_obj.rate)
 
     :return: (:class:`str`) The timecode.
     """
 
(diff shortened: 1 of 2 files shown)


reinecke:
I like that this change makes this function in-line with the `to_frames` function.

My only concern with this would be that it could encourage people to assume `rate` on `RationalTime` objects will always be FPS, but I think that's a separate issue.

I think these frame/time/timecode conversions being really simple to pick up are a great way to get people using opentime and create a great deal of utility!

jminor:
I share the concern that making the 2nd argument optional might encourage too many assumptions about the rate (e.g. accidentally mixing rates, or using non-standard timecode rates) but I think the added convenience and readability is worth it.

ssteinbach:
@reinecke @jminor To address those concerns, we could go the other direction - remove the rate argument altogether.  In that case, you'd have to use `.rescaled_to` to get a RationalTime in the desired rate, and then convert that to frames or timecode or whatever... would that be better or clearer?

jminor:
That's an interesting suggestion... if you're converting to a different rate you're going to have to take care with resampling/rounding issues anyways, but I worry that would just make it less obvious how to get what you want.
We should really optimize for making the API easy to learn and use. This PR as-is does a good job of making it easy and obvious how to do the simple and correct thing: `opentime.to_timecode(start_time)` I think this is a fine change. It just needs the unit test that @ssteinbach suggested.

rogergodspeed:
OK, guys, thanks for the review! Please have a look at the latest version here.

Issue #239, opened 2018-03-13
Put OTIO on PyPi


swallitsch:
It'd be great if we could install OTIO from pypi with pip.

This will require `setup.py` to have some additional fields such as `license`, `classifiers`, `keywords`, `platforms`. Because we're using a `README.md` markdown file, we'll also need a `setup.cfg` file to tell pypi where our long description is.

Once the setup.py work is complete, we can start uploading to PyPi.

[This is a good general article on getting your package ready for uploading to PyPi](http://peterdowns.com/posts/first-time-with-pypi.html) but it's upload section should not be used!

[This how to is good for everything after Register for PyPi](https://blog.jetbrains.com/pycharm/2017/05/how-to-publish-your-package-on-pypi/)

swallitsch:
I'll be submitting a PR for updating setup.py

swallitsch:
I've opened a PR (#241) which should contain the ground work needed to begin putting OTIO on PyPi. 

One thing to note as we move forward with this is that [Warehouse, the all new pypi, is now in Beta](http://pyfound.blogspot.com/2018/03/warehouse-all-new-pypi-is-now-in-beta.html), so PyPi deployment steps should be tested against that as well.

ssteinbach:
This is now done:
https://pypi.org/project/OpenTimelineIO/

We've also updated our internal release process to include PyPI.

PR #241, opened 2018-04-02
PyPi updates (#239)

swallitsch:
This PR does some groundwork to make #239 (Put OTIO on PyPi)  possible:

* Add a `setup.cfg` file which points to the `README.md` file for the description
* Update `setup.py` with:
    * Classifiers
    * License
    * Keywords
    * `install_requires` placeholder
    * support for running the main test suite with `python setup.py test`

In addition, it includes a commit to add the Pycharm `.idea/` folder to `.gitignore`. This commit can be removed from the PR if desired.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,0 +1,2 @@
+[metadata]
+description-file = README.md
--- a/setup.py
+++ b/setup.py
@@ -25,9 +25,10 @@
 
 """ Configuration file for the OpenTimelineIO Python Package.  """
 
-import sys
-
+import os
 from setuptools import setup
+import sys
+import unittest
 
 
 # check the python version first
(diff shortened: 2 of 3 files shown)


setup.cfg
```diff
@@ -0,0 +1,2 @@
+[metadata]
+description-file = README.md
```
mikemahony: Looks like we're missing a newline here.

swallitsch: Added a newline.

setup.py
```diff
@@ -43,13 +44,51 @@
...
+        'Development Status :: 4 - Beta',
+        'Topic :: Multimedia :: Graphics',
+        'Topic :: Multimedia :: Video',
+        'Topic :: Multimedia :: Video :: Display',
+        'Topic :: Multimedia :: Video :: Non-Linear Editor',
+        'Topic :: Software Development :: Libraries :: Python Modules',
+        'License :: Other/Proprietary License',
+        'Programming Language :: Python :: 2',
+        'Programming Language :: Python :: 2.7',
+        'Programming Language :: Python :: 3',
```
ssteinbach: I'm not sure how the tagging is supposed to be set up.  Officially we support python 2.7 and 3.5, those are the  only two versions we test against.  It probably works in the versions you list, but I'm not sure how strict this is meant to be.

swallitsch: I'll remove the other specified versions of 3 from the list.

ssteinbach:
Perfect, thanks again.

reinecke:
This is great +1

PR #243, opened 2018-04-10
Tweak metadata in setup.py for pypi.

ssteinbach:


swallitsch:
Since the setup.cfg description config proved ineffective, this PR should include deleting that file, as the only purpose of setup.cfg is to try and get the README onto pypi.

ssteinbach:
I want to leave that in there - multiple tutorials indicated that that should work, so maybe it just isn't an immediate thing?  Other projects also have better looking things in their sidebar, and I'm not sure how they do it, so I think it just might take a bit more research down the road to get that looking nice.

PR #247, opened 2018-04-19
Pass adapter arguments into otioview.

ssteinbach:
haven't tested this w/ pyside, just a proposal


ssteinbach:
Idea here is to allow EDLs of non 24 fps media to be viewed by Otioview.

PR #249, opened 2018-05-02
We should just use  `python` here instead of `python2.7` in hashbangs

jchen9:


ssteinbach:
I'm not sure why it was that way, good catch.

PR #250, opened 2018-05-09
Corrected handling of top-level Mobs in AAF compositions.

jminor:
If there are multiple compositions in an AAF, the top-level Mob(s) now correctly end up at the root of the OTIO hierarchy.


jminor:
I am adding some tests - which found some other bugs, so let me get those fixed and update this PR.

jminor:
I added more tests and fixed the adapter so it understands nested compositions correctly.
As a result, we now also get audio tracks! (Fixes #216)

PR #251, opened 2018-05-09
'Freeze frame' and 'LinearTimeWarp' time effect to OTIO w/ EDL adapter support.

ssteinbach:
This is stubbing out classes for attaching timing effects and specifically `FreezeFrame` and `LinearTimeWarp` effects to clips in OTIO.

* This only includes CMX EDL adapter support.
* EDL adapter only supports one timing effect / clip.
* This also doesn't include any support in the time functions (still chewing on the best way to expose this), likely will tie into #232 
* Also improves the 'assertJsonEqual' unit test w/ the logic to strip trailing `.0`s (different behavior in python3 vs python2) and makes the unit tests in `test_filter_functions` actually use this properly.


opentimelineio/adapters/cmx_3600.py
```diff
@@ -858,6 +861,11 @@ def __init__(
         line.reel = _reel_from_clip(clip)
         line.source_in = clip.source_range.start_time
         line.source_out = clip.source_range.end_time_exclusive()
+        if clip.effects and clip.effects[0].effect_name == "FreezeFrame":
```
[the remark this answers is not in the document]
ssteinbach: Sure, good note.

opentimelineio/schema/effect.py
```diff
@@ -83,3 +83,45 @@ def __repr__(self):
...
+            self,
+            name=name,
+            effect_name="LinearTimeWarp",
+            metadata=metadata
+        )
+        self.time_scalar = time_scalar
+
+    time_scalar = core.serializable_field(
+        "time_scalar",
+        doc="Linear time scalar applied to clip."
```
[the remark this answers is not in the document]
ssteinbach: I'll add that to the doc!

tests/test_filter_algorithms.py
```diff
@@ -327,18 +316,14 @@ def test_copy(self):
...
-        known = otio.adapters.write_to_string(tl, 'otio_json')
-        test = otio.adapters.write_to_string(
-            otio.algorithms.filtered_with_sequence_context(
-                tl,
-                lambda __, _, ___: _
-            ),
-            'otio_json'
+        test = otio.algorithms.filtered_with_sequence_context(
+            tl,
+            lambda __, _, ___: _
```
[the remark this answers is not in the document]
ssteinbach: The underscores indicate that the arguments are ignored but required to fit the function signature.  I can make it or args_ignored* or add a comment.  I'll poke at it.

opentimelineio/adapters/cmx_3600.py
```diff
@@ -858,6 +893,27 @@ def __init__(
         line.reel = _reel_from_clip(clip)
         line.source_in = clip.source_range.start_time
         line.source_out = clip.source_range.end_time_exclusive()
+
+        timing_effect = _relevant_timing_effect(clip)
+
+        if timing_effect:
+            if timing_effect.effect_name == "FreezeFrame":
```
jminor: Maybe isinstance would be better than checking by name?

opentimelineio/adapters/cmx_3600.py
```diff
@@ -845,6 +859,27 @@ def get_content_for_track_at_index(self, idx, title):
         return content
 
 
+def _timing_effects(clip):
+    return [
+        fx for fx in clip.effects
+        if isinstance(fx, otio.schema.LinearTimeWarp)
```
jminor: Should this be otio.schema.TimingEffect to catch other things in the future?

opentimelineio/adapters/cmx_3600.py
ssteinbach: In this case, `isinstance` would catch any derived classes as well, and I want `FreezeFrame` effects and _only_ `FreezeFrame` effects.

opentimelineio/adapters/cmx_3600.py
ssteinbach: We chatted offline; I'll address this!

PR #261, opened 2018-05-16
otioview's --adapter_arg can now handle Python literals, 

jminor:
You can pass True, False, etc. as real values, instead of everything coming in as a string.

### DIFF
--- a/bin/otioview.py
+++ b/bin/otioview.py
@@ -28,6 +28,7 @@
 import os
 import sys
 import argparse
+import ast
 from PySide import QtGui
 
 import opentimelineio as otio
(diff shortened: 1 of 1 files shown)


bin/otioview.py
```diff
@@ -181,8 +183,14 @@ def main():
     argument_map = {}
     for pair in args.adapter_arg:
         if '=' in pair:
-            key, val = pair.split('=')
-            argument_map[key] = val
+            key, val = pair.split('=', 1)  # only split on the 1st '='
+            try:
+                # Sometimes we need to pass a bool, int, list, etc.
+                parsed_value = ast.literal_eval(val)
+            except (ValueError, SyntaxError):
```
ssteinbach: Wondering if these should fall back on strings - it could hide unexpected behavior from people, but I don't feel strongly about it.

jminor: Without the fallback, you would have to put `-a key=\"value\"` on the command line to get a python literal string.

Issue #253, opened 2018-05-10
Add Position, Scale, Rotate (Transform) per Clip


backhausmartin:
Hello, I'm new to OTIO, I hope a can contribute in some way, maybe later also with code.

I need to parse the Position, Scale and Rotation Data on a per-clip basis from an Offline NLE project (in my case Adobe Premiere Pro 2018) - and re-use this information in an Online/Mastering/Comp project (currently Foundry Nuke Studio or just normal Nuke).

I already managed to parse the Premiere Pro Data from an FCP XML via Python xml.etree and figure out how the clipitem.filter.effect.parameter scale, rotation and center tags/ information work in general.
There are five `<parameter authoringApp="PremierePro">` tags `Scale, Rotation, Center, Anchor Point, Anti-flicker Filter` in the XML below the `xmeml.sequence.media.video.track.clipitem.filter.effect` tree. 

Now I wonder if I can use OTIO for this and maybe built it into it - instead of writing a custom script.

How mature is the integration of this transform data to OTIO as of today? I recently checked the code and I think there are only effect placeholders in OTIO? It looks that it is planned, see here, "Audio/Video Effects": https://github.com/PixarAnimationStudios/OpenTimelineIO/wiki/Feature-Matrix

Some Background:
I do an exhibition film project with multiple projectors in a 360° setup. I have a base canvas/box size with a resolution of more than 32.000px x 1.800px (which will be split/cropped to single projectors later). Multiple clips need to be positioned side by side, at the same time in the edit, using the x-position. It's a 10 min linear film.
My goal is to go from the Quarter-Res offline edit project to the Full-Res Online project (semi-)automatically.

If this is interesting I can provide a FCP XML sample data and also do some testing going from Premiere to Nuke via OTIO. But I understand that this is a special case, going away from standard 2K or 4K resolution.

First step: Basic x,y position should be parsed from FCP XML and AAF and saved in OTIO format. Result: A basic reframing of camera footage will be preserved from Editorial to Online/DI.

jminor:
Hi @backhausmartin, that sounds like a great use case for OTIO. You're right that our support for handling effects in OTIO is not complete yet. If you export your timeline from Premiere as XML and then run it through `otiocat` or `otioconvert` you can see what, if any, of that comes through at the moment. 

[...]

PR #264, opened 2018-05-30
fcp_xml tweaks to support XML's created by other applications than Premiere

apetrynet:
Hi!

I attempted to parse some XML's I had lying around and they both failed due some minor differences in "dialect" or implementation of FCP7 XML than Premiere. I know the one came from Hiero and I think the other one came from Resolve. 
I'll create some new XML's that are safe to be added to the example files and add some tests if that's of interest.


opentimelineio/adapters/fcp_xml.py
```diff
@@ -238,11 +248,11 @@ def _parse_clip_item_without_media(clip_item, track_rate,
     ]
 
     in_frame = (
-        int(clip_item.find('./in').text)
+        int(float(clip_item.find('./in').text))
```
jminor: Why `int(float(x))`? Would you ever want to keep the fractional part? Also in the line below there's an `int(round(x))` Is the rounding important? It seems odd that they don't match.

jminor:
Thanks! Having tests for files from other applications is super helpful.

Does Hiero read the FCPXML that OTIO outputs?

opentimelineio/adapters/fcp_xml.py
apetrynet: Hi again!
I did some additional testing and to avoid `ValueError` when receiving `'0.0'` from `clip_item.find('./in').text` it needs to be converted to a float first. Same behavior in python 2 and 3.
```
>>> int('0.0')
ValueError: invalid literal for int() with base 10: '0.0'
>>> 
```
Found someone else stumbling across the same dilemma: https://stackoverflow.com/questions/17142088/python-2-7-and-3-3-2-why-int0-0-does-not-work

apetrynet:
@jminor A quick test confirms that Hiero reads OTIO outputted XML just fine.

bashesenaxis:
It looks like Hiero is not following the FCP7 XML spec on this:
https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/FinalCutPro_XML/Elements/Elements.html#//apple_ref/doc/uid/TP30001156-TPXREF106
Timing elements are supposed to be ints. 
But if this fix means increased compatibility with Hiero, I am all for it.

opentimelineio/adapters/fcp_xml.py
```diff
@@ -210,7 +210,17 @@ def _parse_media_reference(file_e, element_map):
...
     file_rate = _parse_rate(file_e, element_map)
     timecode_rate = _parse_rate(file_e.find('./timecode'), element_map)
-    timecode_frame = int(file_e.find('./timecode/frame').text)
+
+    frame_e = file_e.find('./timecode/frame')
+    if frame_e is not None:
+        timecode_frame = int(frame_e.text)
+
+    else:
+        timecode_frame = otio.opentime.from_timecode(
```
bashesenaxis: Catching the two cases looks sound to me. According to the FCP7 XML spec timecode must contain either a string sub-element, or a frame sub-element. 
But perhaps it would make more sense to create RationalTime objects for each case?
```Python
if frame_e is None:
    start_time = otio.opentime.RationalTime(
        int(frame_e.text), timecode_rate
    )
else:
    start_time = otio.opentime.from_timecode(
        file_e.find('./timecode/string).text, timecode_rate
    )
```
Then later for available_range you can do
```Python
available_range = otio.opentime.TimeRange(
    start_time=start_time, 
    duration=otio.opentime.RationalTime(duration, file_rate
)
```

apetrynet: Sure! Sounds good.
Should I do the same for the duration perhaps?

apetrynet:
Sorry about the failed test. I'll fix it. 
Working behind tons of firewalls so don't have direct access to github. Therefore used online commit as test suite. 
Won't repeat this procedure. Will do it the correct way moving forward.

opentimelineio/adapters/fcp_xml.py
bashesenaxis: Duration should be fine. I mainly mentioned the start_time because after otio.opentime.from_timecode, you already have a rationalTime object. So in that case you were creating the same object twice.

opentimelineio/adapters/fcp_xml.py
```diff
@@ -412,6 +422,10 @@ def _parse_top_level_track(track_e, kind, rate, element_map):
     if not track_items:
         return track
 
+    trackname_e = track_e.find('./name')
```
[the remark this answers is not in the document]
apetrynet: Yeah, I noticed the <name> tag didn't refer to track in the link you posted. 
I agree with you concern, but find it strange that track names are omitted from the official XML export in the first place.
I'll let you guys make the decision. I'm happy to kill that "darling" if it's for the greater good :)

apetrynet:
@bashesenaxis What are your thoughts on parsing and storing elements from `<media> -> <video> -> <format> -> <samplecharacteristics>`? I think it would be useful to store the `<with>, <height> and <rate>` elements that define the intended resolution etc. of the source timeline. 
The data could live in `otio.schema.Timeline.metadata` (?)

bashesenaxis:
I tried to use metadata as little as possible, sticking closely to the otio feature set. This has come up before, where there is a desire to have a certain XML tag represented as metadata (#253). Because XML is so feature rich, I feel it warrants a wider discussion about the use of metadata before moving forward with any particular addition.

apetrynet:
Yeah, I see your point, but still feel that width, height and rate are fairly universal for most timelines. I'll move my future input on the topic to another thread. Thanks.

jminor:
For some of the other adapters, we take extra information like this and put it into adapter-specific sub-dictionaries of metadata. This lets us round-trip that to and from one format without losing it entirely, but we let it fall off when converting to some other format. So, for example, the EDL and ALE adapters both carry format specific stuff, like EDL headers, or ALE column ordering in metadata, but neither of those are relevant to FCP.

This strategy makes it clear which information is part of the OTIO schema and which is adapter-specific, and also helps to provoke conversation around which data should be promoted to be part of the schema. As an example, the EDL adapter currently understands CDL color correction as metadata, but we plan to promote that to be part of the OTIO Effect schema after we compare the way it is represented across other formats.

opentimelineio/adapters/fcp_xml.py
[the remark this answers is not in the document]
ssteinbach: I'm ok with this change, actually.  As long as FCP gracefully reads XMLs with the 'name' field I think its fine to include support for it.  Furthermore, in the EDL adapter we added a 'Flavor' switch to distinguish between EDLs (Avid and NuCoda), so I think that pattern is ok for this in the long run.  If we find its a problem as we scale up, we can disable behavior and fall back on lowest common denominator.

Since our ambition in the limit is that tools directly support OTIO rather than use OTIO to translate for interchange, we're trying to tease out these kinds of things that should be represented in the schema.

tests/sample_data/hiero_xml_export.xml
```diff
@@ -0,0 +1,132 @@
...
+        <format>
+          <samplecharacteristics>
+            <depth>16</depth>
+            <samplerate>48000</samplerate>
+          </samplecharacteristics>
+        </format>
+      </audio>
+    </media>
+  </sequence>
+</xmeml>
```
ssteinbach: Would you mind adding a unit test that verifies that this file reads correctly? It just needs to read the file in and check for a few properties.

thanks!

apetrynet: Sure!
I'll look at similar tests and create one for the Hiero XML.

jminor:
Thanks for the contribution!

apetrynet:
Thank you for this awesome project! Happy to contribute.

PR #269, opened 2018-06-14
Add development requirements to package and instructions to README

alexwidener:
This does _not_ add PySide2 or PySide1, but it does add the dependencies for flake8 and coverage, as well as documentation for how to install the dependencies to the README.

PySide2 Snapshot wheels are available through the following: 

`pip install --index-url=http://download.qt.io/snapshots/ci/pyside/5.9/latest/ pyside2 --trusted-host download.qt.io` 

and are hosted at http://download.qt.io/snapshots/ci/pyside/5.9/latest/pyside2/

Since you may be about to merge in #160 I chose not to add PySide2 for 2 reasons. 

1 ) To not get out of date immediately :)
2 ) The above are _snapshots_ and therefore not versioned. I think until PySide2 enters into versioned tagging, it should be up to the person managing the distribution of OTIO in their facility to handle which PySide2 version they have available. 

Something to ponder about once they do enter into a versioned tag schema is whether to actually add PySide2 as an installation dependency. I know that in my facility (Laika) we would have to fork and block this requirement in the setup.py as we build our own PySide/PyQt.

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,6 +175,8 @@ with the pep8 style.  We ask that before you submit a pull request, you:
 
 PEP8: https://www.python.org/dev/peps/pep-0008/
 
+You can install development dependencies with `pip install -e .[dev]`
+
 Contact
 -------
 
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,24 @@
 .PHONY: coverage test test_first_fail clean autopep8 lint doc-html \
 	python-version
 
+# Special definition to handle Make from stripping newlines
+define newline
+
+
+endef
+
+# Color highlighting for the shell
+ccred = $(shell echo "\033[0;31m")
+ccgreen = $(shell echo "\033[0;32m")
+ccblue = $(shell echo "\033[0;34m")
+ccend = $(shell echo "\033[0m")
+
+# Helpful link to install development dependencies declared in setup.py
+define dev_deps_message
+$(ccred)You can install this and other development dependencies with$(newline)$(ccend)\
+$(ccblue)	pip install -e .[dev]$(newline)$(ccend)
+endef
+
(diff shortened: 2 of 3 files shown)


alexwidener:
Of course, after I write that, I come across this blog post that states that they had a release earlier today: 

https://blog.qt.io/blog/2018/06/13/qt-python-5-11-released/

http://download.qt.io/official_releases/QtForPython/pyside2/

I've checked and it's not in pypi yet, but additionally, the versioning scheme is still a little odd - I wouldn't expect to see PySide2-5.11 in a setup.py.

alexwidener:
Yes, you are correct. We should be able to add a key called 'view' that will only install PySide and then users can optionally call it with `pip install .[view]` if they choose to install PySide2. This fulfills the condition of making the PySide install optional for studios as well.

ssteinbach:
@alexwidener did you want to add that [.view] target to this PR as well?  Or make that a separate PR?

alexwidener:
We should add it as a separate PR once PySide2 is publicly available on pypi.org

README.md
```diff
@@ -175,6 +175,8 @@ with the pep8 style.  We ask that before you submit a pull request, you:
 
 PEP8: https://www.python.org/dev/peps/pep-0008/
 
+You can install development dependencies with `pip install -e .[dev]`
```
ssteinbach: This is great!  Would you mind also adding this note to the Makefile, which has some code that detects if flake8 isn't installed and instructs you to download it in an error message?  That seems like another good oppurtunity to point people at this new feature.

alexwidener: Added the requested message, but was messing around and did some formatting to make the messages look better, hopefully that's okay too. 

screen shot 2018-06-20 at 8 53 18 pm
screen shot 2018-06-20 at 8 52 28 pm

Issue #270, opened 2018-06-19
Cannot read adapter manifest when installed via `python setup.py install` (compressed egg)


ssteinbach:
When OTIO is installed via `python setup.py install`, it is installed as a compressed egg.  This breaks the `open()` function in the adapter manifest system because it is no longer files on the file system but stored inside a zipfile.

Installing via `pip install` (either via pypi or `.`) fixes this problem.

```Z:\dist\third_party\apps\python\2.7\Scripts>python otioview.py test.xml
Traceback (most recent call last):
  File "otioview.py", line 4, in <module>
    __import__('pkg_resources').run_script('OpenTimelineIO==0.8.0.dev1', 'otioview.py')
  File "Z:\dist\third_party\apps\python\2.7\lib\site-packages\pkg_resources\__init__.py", line 743, in run_script
    self.require(requires)[0].run_script(script_name, ns)
  File "Z:\dist\third_party\apps\python\2.7\lib\site-packages\pkg_resources\__init__.py", line 1505, in run_script
    exec(script_code, namespace, namespace)
  File "Z:\dist\third_party\apps\python\2.7\lib\site-packages\opentimelineio-0.8.0.dev1-py2.7.egg\EGG-INFO\scripts\otioview.py", line 223, in <module>

  File "Z:\dist\third_party\apps\python\2.7\lib\site-packages\opentimelineio-0.8.0.dev1-py2.7.egg\EGG-INFO\scripts\otioview.py", line 214, in main

  File "Z:\dist\third_party\apps\python\2.7\lib\site-packages\opentimelineio-0.8.0.dev1-py2.7.egg\EGG-INFO\scripts\otioview.py", line 154, in load

  File "Z:\dist\third_party\apps\python\2.7\lib\site-packages\opentimelineio-0.8.0.dev1-py2.7.egg\opentimelineio\adapters\__init__.py", line 135, in read_from_file
  File "Z:\dist\third_party\apps\python\2.7\lib\site-packages\opentimelineio-0.8.0.dev1-py2.7.egg\opentimelineio\adapters\__init__.py", line 82, in _from_filepath_or_name
  File "Z:\dist\third_party\apps\python\2.7\lib\site-packages\opentimelineio-0.8.0.dev1-py2.7.egg\opentimelineio\adapters\__init__.py", line 95, in from_filepath
[...]

PR #273, opened 2018-06-19
Add zip_safe = false flag to address #270.

ssteinbach:
Sets the `zip_safe = False` because we use open() to fetch files from OTIO - which means OTIO can't be used with compressed eggs.


ssteinbach:
@bpeck does this address #270 for you?  This should prevent python from installing as a zip file.

bpeck:
Yeah thanks!

PR #274, opened 2018-06-20
192 code coverage badge

### LINKED ISSUE #192: Code coverage badge in README.
We have build status badges already, but it would be great to have a code coverage badge as well. We currently have more than 90% code coverage but that fact is hard to find in the test results.

alexwidener:
Closes #192 

Process:

1 ) Connect the PixarAnimationStudios Github account to codecov.io (ignore the token stuff on the Overview page after you sign in, it's not necessary since OTIO is OSS)
2 ) Select the OpenTimelineIO repository - you can see the reports after I connected it [here](https://codecov.io/gh/alexwidener/OpenTimelineIO/commits)
3 ) Run a build for Travis, which causes a coverage report to get kicked off which updates the badge (for the master branch).
4) [There's a Github integration for Codecov if you want to check it out](https://github.com/integration/codecov)

Final result:

screen shot 2018-06-19 at 8 10 56 pm

Additionally, check out the cool graphs: https://docs.codecov.io/docs/graphs#section-sunburst

Sorry for a bunch of dumb commits, I was just testing stuff - you can squash the commits into one commit whenever you merge it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,3 +24,6 @@ script:
     - make lint
     - make test
     - make coverage
+
+after_success:
+    - bash <(curl -s https://codecov.io/bash)
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,10 @@ install:
 script:
     - make lint
     - make test
+    # Build coverage which will get uploaded to codecov
     - make coverage
 
 after_success:
+    # Documentation for codecov uploader
+    # https://docs.codecov.io/docs/about-the-codecov-bash-uploader
     - bash <(curl -s https://codecov.io/bash)
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,5 +1,4 @@
 codecov:
-  branch: 192-code-coverage-badge
   notify:
     require_ci_to_pass: yes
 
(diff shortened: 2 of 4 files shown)


alexwidener:
I think that reported from the integration that I connected to my account which I am now going to disconnect because I tested it out...

Anyway the reason it's saying " No coverage uploaded for pull request base master" is because it's not finding a coverage report for master yet because it's set up to pull from master.

.travis.yml
```diff
@@ -24,3 +24,6 @@ script:
     - make lint
     - make test
     - make coverage
+
+after_success:
+    - bash <(curl -s https://codecov.io/bash)
```
ssteinbach: Can you put a comment on this to indicate that this is how codecov.io prefers reports be uploaded?  or a link to their documentation page or something?  Thanks!

.travis.yml
```diff
@@ -24,3 +24,6 @@ script:
     - make lint
     - make test
     - make coverage
```
ssteinbach: Do we still need to run `make coverage` in travis if we're having codecov.io build reports?

alexwidener: Codecov does not build reports - codecov just presents the reports. You still need to run coverage on your CI machine, so this does need to stay.

.travis.yml
alexwidener: Done.

ssteinbach:
I made a PR on your PR to combine the coverage reports (it looks like the old version just reported results from the contrib adapters).

PR #283, opened 2018-07-04
Experimental ffprobe-based media adapter

jminor:
This adapter will read common video and audio file formats, via ffprobe, and create single-clip timelines that point to the given media file. The new clip's duration and start_time match the media file.

This is meant as a convenience for quickly making new OTIO files with just one clip.


alatdneg:
That'll be handy, I implemented something like that myself, one thing to watch out for is don't use the float durations etc, use the nb_frames, duration_ts and timebase.

You'll get odd rounding errors if you  use the float values (or at least I did :))

reinecke:
This is pretty interesting. @alatdneg suggestion of using ffmpeg's exposed `AVRational` values for duration is a good call.

A media-robust way to get `RationalTime` from ffprobe's output would be something like:
```
duration_timestamp = stream['duration_ts']
time_base = fractions.Fraction(stream['time_base'])
duration = RationalTime(duration_timestamp, time_base.denominator)
```

This prevents accuracy loss over long editorial assemblies. Also, this allows you to find `RationalTime` durations in a media type independent way. Using `nb_frames` should be avoided as that it won't reflect true duration if there are frames that were dropped from the stream (common when iPhone recompresses videos).

opentimelineio_contrib/adapters/media_file.py
```diff
@@ -0,0 +1,202 @@
...
+            'ffprobe',
+            '-v', 'quiet',
+            '-print_format', 'json',
+            '-show_format',
+            '-show_streams',
+            filepath
+        ],
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+        stdin=subprocess.PIPE,
```
ssteinbach: I don't think you need to map `stdin` if you're not going to use it in your communicate.

opentimelineio_contrib/adapters/media_file.py
```diff
@@ -0,0 +1,202 @@
...
+        env=os.environ
+    )
+    out, err = proc.communicate()
+    metadata = {}
+
+    metadata = json.loads(out)
```
ssteinbach: Looks like because of this line, you can delete line 52

opentimelineio_contrib/adapters/tests/test_media_adapter.py
```diff
@@ -0,0 +1,108 @@
...
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the Apache License with the above modification is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied. See the Apache License for the specific
+# language governing permissions and limitations under the Apache License.
+#
+
+"""Test the Media File adapter."""
```
ssteinbach: Could this file be renamed to `test_media_file_adapter.py`?  Just to make sure it connects w/ `media_file.py`.

opentimelineio_contrib/adapters/media_file.py
```diff
@@ -0,0 +1,202 @@
...
+    elif has_audio:
+        return otio.schema.TrackKind.Audio
+    else:
+        return None
+
+
+def read_from_file(filepath, default_fps=24):
+
+    if not os.path.exists(filepath):
+        raise otio.exceptions.CouldNotReadFileError(
```
ssteinbach: I wonder if we should refactor this kind of check up into `adapter.read_from_file` -- presumably all the `read_from_file` calls should be checking this?

opentimelineio_contrib/adapters/media_file.py
ssteinbach: Also, might want to catch `RuntimeError` or `ValueError` in case the the json parser fails (and map them into `OTIOError`).

ssteinbach: ...Then you could include the `err` stderr output in the exception error message.

opentimelineio_contrib/adapters/media_file.py
```diff
@@ -0,0 +1,202 @@
...
+
+    for stream in streams:
+        codec_type = stream.get('codec_type')
+
+        if codec_type == 'video':
+            # trust this over anything from non-video tracks
+            time_base = stream.get('time_base', time_base)
+            r_frame_rate = stream.get('r_frame_rate', r_frame_rate)
+
+        elif codec_type == 'audio':
```
ssteinbach: If you get a video stream and then an audio stream, will the time base information from the audio track stomp the video track?   Or should it break if it hits a video track instead of continuing over all the other tracks?

opentimelineio_contrib/adapters/media_file.py
```diff
@@ -0,0 +1,202 @@
...
+
+    if timecode_str:
+        # overwrite start with timecode if we have it
+        try:
+            start = otio.opentime.from_timecode(timecode_str, fps)
+        except ValueError:
+            # ignore problems with the timecode value
+            pass
+
+    return otio.opentime.TimeRange(
```
ssteinbach: So if it can't parse information, it'll return a time range of 0, 0?

opentimelineio_contrib/adapters/media_file.py
jminor: Let's see how the experiments with reading non-files (URIs, shorthand, etc.) pan out before we move that up.

opentimelineio_contrib/adapters/media_file.py
jminor: The video track will overrule the audio track. If there are multiple video tracks, then it just picks one (the last one I think).

opentimelineio_contrib/adapters/media_file.py
jminor: Yeah, I figure a 0-length clip with a media reference is better than nothing. Do you think it should raise an exception instead?

opentimelineio_contrib/adapters/tests/test_media_adapter.py
jminor: Will do.

jminor:
@reinecke and @alatdneg I'm confused about the `duration_ts` values I'm seeing. I see `"duration_ts": 56000` and `"time_base": "1/24000"`, when I expect 56 and 1/24. Where is the extra 1000x coming from?

Here's the full dictionary from the test file I'm using.
```
{
    "avg_frame_rate": "24/1", 
    "bit_rate": "36175872", 
    "bits_per_raw_sample": "8", 
    "codec_long_name": "VC3/DNxHD", 
    "codec_name": "dnxhd", 
    "codec_tag": "0x6e645641", 
    "codec_tag_string": "AVdn", 
    "codec_time_base": "1/24", 
    "codec_type": "video", 
    "coded_height": 1088, 
    "coded_width": 1920, 
    "color_primaries": "bt709", 
    "color_range": "tv", 
    "color_space": "bt709", 
    "color_transfer": "bt709", 
    "display_aspect_ratio": "16:9", 
    "disposition": {
        "attached_pic": 0, 
        "clean_effects": 0, 
        "comment": 0, 
        "default": 1, 
        "dub": 0, 
        "forced": 0, 
        "hearing_impaired": 0, 
        "karaoke": 0, 
        "lyrics": 0, 
        "original": 0, 
        "timed_thumbnails": 0, 
        "visual_impaired": 0
    }, 
    "duration": "2.333333", 
    "duration_ts": 56000, 
    "has_b_frames": 0, 
    "height": 1080, 
    "index": 0, 
    "level": -99, 
    "nb_frames": "56", 
    "pix_fmt": "yuv422p", 
    "profile": "DNXHD", 
    "r_frame_rate": "24/1", 
    "refs": 1, 
    "sample_aspect_ratio": "1:1", 
    "side_data_list": [
        {
            "displaymatrix": "\n00000000:        49152           0           0\n00000001:            0       65536           0\n00000002:            0           0  1073741824\n", 
            "rotation": 0, 
            "side_data_type": "Display Matrix"
        }
    ], 
    "start_pts": 0, 
    "start_time": "0.000000", 
    "tags": {
        "creation_time": "2018-06-01T00:06:47.000000Z", 
        "encoder": "Avid DNxHD Codec", 
        "handler_name": "Apple Alias Data Handler", 
        "language": "eng", 
        "rotate": "0"
    }, 
    "time_base": "1/24000", 
    "width": 1920
}, 
```

reinecke:
Closing for now, if a use case emerges we can re-open.

Issue #226, opened 2018-02-07
Convenience for creating/adding a Gap, black, etc.


jminor:
Making a new Gap, or adding one to a Track, should be simpler. Same for adding a solid black clip.

Maybe something like: `gap = otio.schema.Gap(duration=otio.opentime.RationalTime(10, 24))`
or `slug = otio.make_solid_clip(color='black', duration=otio.opentime.RationalTime(10, 24))`

ssteinbach:
Addressed by #286

PR #286, opened 2018-07-09
Add convienence constructor for Gap

ssteinbach:
Addresses #226

### DIFF
--- a/opentimelineio/schema/gap.py
+++ b/opentimelineio/schema/gap.py
@@ -49,16 +49,16 @@ class Gap(core.Item):
             raise RuntimeError(
                 "Cannot instantiate with both a source range and a duration."
             )
-        if duration is None and source_range is None:
-            raise RuntimeError(
-                "Requires either a source range or a duration to construct."
-            )
 
         if duration:
             source_range = opentime.TimeRange(
                 opentime.RationalTime(0, duration.rate),
                 duration
             )
+        elif source_range is None:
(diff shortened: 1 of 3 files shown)


opentimelineio/schema/gap.py
```diff
@@ -32,6 +35,36 @@ class Gap(core.Item):
     _serializable_label = "Gap.1"
     _class_path = "schema.Gap"
 
+    def __init__(
+        self,
+        name=None,
+        # note - only one of the following is accepted
```
avrata: Super super minor, but perhaps add "two" ("only one of the following two.."). At first read I thought this applied to effects/markers/etc as well.

jminor: I agree.

opentimelineio/schema/gap.py
```diff
@@ -32,6 +35,36 @@ class Gap(core.Item):
...
+        duration=None,
+        source_range=None,
+        effects=None,
+        markers=None,
+        metadata=None,
+    ):
+        if duration and source_range:
+            raise RuntimeError(
+                "Cannot instantiate with both a source range and a duration."
+            )
```
jminor: We should also check for neither and complain.

ssteinbach: Unfortunately, the deserializer requires that we have the ability to `__init__` classes without any arguments, so we can't make this an error case.

PR #291, opened 2018-07-23
Update build for Python 3.6

alexwidener:
I've been running Python 3.6 for OTIO since last year and have never run into a problem with it and since [vfxplatform states we should be targeting 3.6](http://www.vfxplatform.com/), I figured we should all just skip forward. 

[Here is my test.](https://travis-ci.org/alexwidener/OpenTimelineIO/jobs/406978135)


reinecke:
+1
We’ve been on 3.6 for quite a while now. We haven’t run into any issues in our usage.

jminor:
Can you update the badge in the README.md to show 3.6 also?
I think the URL format is this: `https://img.shields.io/badge/python-2.7%2C%203.5%2C%203.6-blue.svg`

alexwidener:
Changed @jminor

jminor:
Thanks!

alexwidener:
:( Wrong lines were edited and now there are two sets of badges. I am not a fan of using Github's editor to do changes. Submitting separate PR now.

PR #299, opened 2018-07-30
Rearrange the commandline files into a submodule.

ssteinbach:
This takes the files from the `bin` subdirectory and moves them into `console` modules inside the python packages.  This lets the `setup.py` be cleaner in terms of generating the extensionless executable files, and also makes their functions (argument parser, etc) available for reuse by others.

@alexwidener this rearranges some of the stuff you had done previously - what do you think?


alexwidener:
Yeah, I like it better now, since `/bin` is typically reserved for binaries. Makes more sense this way.

Issue #217, opened 2018-01-29
AAF adapter doesn't read clip timecode from AAF


jminor:

jminor:
This explains how to get the timecode for a clip: https://github.com/markreidvfx/pyaaf/issues/44#issuecomment-357487131

ssteinbach:
Addressed by #301

PR #301, opened 2018-08-03
Add support for mobs with timecode slots to the aaf adapter.

ssteinbach:
Also adds unit tests.

Addresses issue #217


opentimelineio_contrib/adapters/advanced_authoring_format.py
```diff
@@ -103,6 +103,28 @@ def _transcribe_property(prop):
...
+    tc_list = [
+        s.segment['Start'].value for s in mob.slots()
+        if s.segment.media_kind == 'Timecode'
+    ]
+
+    if len(tc_list) == 1:
+        return tc_list[0]
+    elif len(tc_list) > 1:
+        raise otio.exceptions.NotSupportedError(
+            "Error: mob has more than one timecode track, this is not"
```
jminor: Could you include the mob name in the error message? That can really help when debugging if/when this comes up.

opentimelineio_contrib/adapters/advanced_authoring_format.py
```diff
@@ -164,11 +186,23 @@ def _transcribe(item, parent=None, editRate=24, masterMobs=None):
...
+            if mob:
+                mobs.append(mob)
+
+        # Evidently the last mob is the one with timecode
+        timecode = _extract_timecode(mobs[-1])
 
         length = item.length
         startTime = int(metadata.get("StartTime", "0"))
+        if timecode:
+            startTime += timecode
```
jminor: Is `timecode` an int?

opentimelineio_contrib/adapters/tests/test_aaf_adapter.py
```diff
@@ -555,10 +571,7 @@ def test_aaf_nesting(self):
         )
         self.assertEqual(
             otio.opentime.TimeRange(
-                start_time=otio.opentime.RationalTime(32, 24),
-                # TODO: should actually be this, but we're not getting the
-                # media timecode offset correctly from the AAF...
-                # start_time=otio.opentime.RationalTime(86432, 24),
+                start_time=otio.opentime.RationalTime(86432, 24),
```
jminor: I liked the way you wrote `86400+32` in the other cases, it makes it clear where this number came from.

opentimelineio_contrib/adapters/tests/test_aaf_adapter.py
```diff
@@ -573,10 +586,10 @@ def test_aaf_nesting(self):
         )
         self.assertEqual(
             otio.opentime.TimeRange(
-                start_time=otio.opentime.RationalTime(24, 24),
+                # start_time=otio.opentime.RationalTime(24, 24),
                 # TODO: should actually be this, but we're not getting the
                 # media timecode offset correctly from the AAF...
```
jminor: You could remove this TODO, right?

opentimelineio_contrib/adapters/advanced_authoring_format.py
ssteinbach: Seems to be, yes.

jminor: So maybe `media_start` or something would be a more clear name?

opentimelineio_contrib/adapters/advanced_authoring_format.py
stefanschulze: I think we should call it timecode slot instead of track to stay in the AAF terminology. 'track' in the context of AAF confuses me :)

opentimelineio_contrib/adapters/advanced_authoring_format.py
```diff
@@ -164,11 +186,23 @@ def _transcribe(item, parent=None, editRate=24, masterMobs=None):
...
+        start = item.start_time
+
+        for c in item.walk():
+            mob = c.resolve_ref()
+            start += c.start_time
+            if mob:
+                mobs.append(mob)
+
+        # Evidently the last mob is the one with timecode
+        timecode = _extract_timecode(mobs[-1])
```
stefanschulze: I think we should check the mobs list size before accessing it via [-1], I believe an empty list will create an index error in that case.

jminor:
So looking at `simple.aaf` in Media Composer, I do see the 1-hour timecode start offset on each clip, so the updated tests are accurate. At first I was surprised, since the original media for those clips *doesn't* have an intentional offset, but I suspect this offset is coming from the "default timecode" setting in Media Composer. Let me poke at that a bit more just so we don't miss something potentially important here.

opentimelineio_contrib/adapters/advanced_authoring_format.py
ssteinbach: mobs is initialized to `[item.resolve_ref()]` which I would imagine would always be valid?  But I can add a check, good note.

opentimelineio_contrib/adapters/advanced_authoring_format.py
ssteinbach: You mean for the timecode variable?  or for startTime?  I figured timecode would be good since its what is getting pulled out of the timecode track.

Issue #278, opened 2018-06-28
otioview keyboard navigation


jminor:
It would be great if you could use the arrow keys to select adjacent clips left/right or up/down in the tracks above/below the current selection.

jminor:
Thanks for adding this feature @bonpix !

PR #302, opened 2018-08-03
Keyboard shortcuts for OTIOview

bonpix:
Addresses issue #278

### DIFF
--- a/opentimelineview/timeline_widget.py
+++ b/opentimelineview/timeline_widget.py
@@ -256,6 +256,13 @@ class NestedItem(_BaseItem):
         super(_BaseItem, self).mouseDoubleClickEvent(event)
         self.scene().views()[0].open_stack.emit(self.item)
 
+    def keyPressEvent(self, key_event):
+        super(_BaseItem, self).keyPressEvent(key_event)
+        key = key_event.key()
+
+        if key == QtCore.Qt.Key_Return:
+            self.scene().views()[0].open_stack.emit(self.item)
+
 
 class Track(QtWidgets.QGraphicsRectItem):
     def __init__(self, track, *args, **kwargs):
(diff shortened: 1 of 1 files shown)


boredstiff:
I would recommend changing the `camelCasing` to fit with the `snake_casing` of the rest of the library.

bonpix:
Newest version cleans up the code and adds function of going into nested item when the "Enter" key is pressed. Also most Markers and Transitions react to the arrow keys now, with the exception of the Markers without parentItems at the very top. 

Items are also centered when selected.

opentimelineview/timeline_widget.py
```diff
@@ -537,6 +544,186 @@ def wheelEvent(self, event):
...
+
+        # Validate new item for edge cases
+        # If valid, set selected
+        if (
+            not isinstance(newSelectedItem, Track)
+            and newSelectedItem
+        ):
+            if self.scene().selectedItems():
+                for selectedItem in self.scene().selectedItems():
+                    selectedItem.setSelected(False)
```
jminor: Could you either put this into a `_deselect_all()` method or just comment that you're deselecting everything?

opentimelineview/timeline_widget.py
```diff
@@ -537,6 +544,186 @@ def wheelEvent(self, event):
...
+                else:
+                    newSelectedItem = self._get_new_item(
+                                                         key_event,
+                                                         curSelectedItem
+                                                         )
+            else:
+                newSelectedItem = self._get_new_item(
+                                                     key_event,
+                                                     curSelectedItem
+                                                     )
```
jminor: The `newSelectedItem = ...` part seems the same in all three code paths, so if you pull it out of the if/else block then you'll only need it once.

opentimelineview/timeline_widget.py
bonpix: np! _deselect_all() sounds good to me.

opentimelineview/timeline_widget.py
bonpix: Yup, can take care of that as well.

opentimelineview/timeline_widget.py
```diff
@@ -537,6 +544,186 @@ def wheelEvent(self, event):
...
+        # If valid, set selected
+        if (
+            not isinstance(newSelectedItem, Track)
+            and newSelectedItem
+        ):
+            if self.scene().selectedItems():
+                for selectedItem in self.scene().selectedItems():
+                    selectedItem.setSelected(False)
+            newSelectedItem.setSelected(True)
+            self.centerOn(newSelectedItem)
```
avrata: Hi, this is great and helps a lot, however I noticed this centerOn is being called on any keypress (even something like "control"). I wonder if `_get_new_item()` could return `None` for when a new item is not found and only call into `_select_new_item()` if a valid new item was returned. This would limit calls into this `_select_new_item()` method so we could skip the `centerOn()` and `_deselect_all()` code. .

bonpix: Good catch; didn't realize it was it was calling centerOn so frequently. Thanks!

jminor: You could consider `ensureVisible()` instead of `centerOn()` but in trying it just now, the centering actually works pretty well.

PR #310, opened 2018-08-15
[WIP] added indicator patch if there's effects in the clip or gap item and more detailed tooltip

bonpix:


avrata:
Cool! This is just my opinion though so take it with a huge grain of salt, but I wonder if it would be nice to add something to the tooltip for these clips. I sometimes get confused in UIs where I see some marker or widgetry I don't understand or explained anywhere (and often look for a tooltip to help explain it). I wonder if the details widget would make this clear, or if adding something to the tooltip like "Has Effects" would help.

I have wanted to propose adding a lot more detail to the tooltip popup for a while (and perhaps this could wait till then), but just thought I would throw that out there.

opentimelineview/timeline_widget.py
```diff
@@ -145,24 +148,55 @@ def _set_labels_timecode(self):
...
     def _set_labels(self):
         self._set_labels_rational_time()
         self.source_name_label.setText('PLACEHOLDER')
         self._position_labels()
 
     def _set_tooltip(self):
-        self.setToolTip(self.item.name)
+        toolTipStr = self.item.name
+
+        if isinstance(self.item, otio.schema.Clip) or \
```
ssteinbach: Could this be `otio.core.Item` instead?  That way it would catch sequences as well?

opentimelineview/timeline_widget.py
```diff
@@ -145,24 +148,55 @@ def _set_labels_timecode(self):
...
+           isinstance(self.item, otio.schema.Stack):
+            duration = "%0.2f" % otio.opentime.to_seconds(
```
ssteinbach: We've been using the `.format()` syntax to stay python 3 compatible, but it works almost the same:

`"{0:.2f}".format(otio.opentime.to_seconds(...)`

Should do it.

opentimelineview/timeline_widget.py
```diff
@@ -145,24 +148,55 @@ def _set_labels_timecode(self):
...
+                                self.timeline_range.duration
+                                )
+
+            toolTipStr = toolTipStr + "\nLength (secs): " + str(duration)
+
+        if isinstance(self.item, otio.core.Item) and self.item.effects:
+            if len(self.item.effects) == 1:
+                toolTipStr = toolTipStr + \
+                            "\n" + str(len(self.item.effects)) + " effect"
```
ssteinbach: Might be more scannable if you start with the word effects, then you also don't need the branch.  So the line is:

`"\nEffects: {}".format(len(self.item.effects))`

PR #312, opened 2018-08-23
308 documentation

boredstiff:
Closes #308, #304

I'm not finished and have more things I'd like to do to make it better, but I'd like to get some feedback. If you'd like to see what the documentation looks like currently, I've attached a zip file of the build, just run the index.html in your browser. 

It's not perfect, I'm still working on a few things, but it's got full API generation, and the Wiki is completely migrated over. I'd like to know your thoughts before getting too far along. I'd say it's around 90% finished, though. :)

I've still got to do things like:
 - work it into travis
 - ensure codecov is still working
 - move the flake8 over to tox, it's commented out
 - add the doc testing
 - update readme with information

Additionally, in the commits, I've written in-depth descriptions so if you have any questions as to 'why', hopefully they're answered there and if not, let me know if you have more questions. 

It'll probably be a couple of days before I get back to this, so take your time with feedback. I probably won't hit it again until Saturday or Sunday.

[html.zip](https://github.com/PixarAnimationStudios/OpenTimelineIO/files/2313337/html.zip)

Additionally, it's going to fail right now. I have no clue why, because in my Travis build history, every commit up until the last two were fine. My initial guess is that the flake8 settings in the tox.ini file are interfering with what Travis sees, but I'll look into that whenever I get back to working on this this weekend.

https://travis-ci.org/boredstiff/OpenTimelineIO/builds

PR #321, opened 2018-09-07
Dark Mode

### LINKED ISSUE #311: Dark Mode
screen shot 2018-08-19 at 11 13 43 pm

screen shot 2018-08-19 at 11 15 15 pm

What do you all think about having a dark mode for otioview?

I could either have a dark mode be the default, or have it not be a default and be enabled through an Action in the File menu.

boredstiff:
Closes #311 

Went with as minimal of an approach as possible. 

screen shot 2018-09-06 at 9 12 04 pm

Open to change requests.

### DIFF
--- a/opentimelineview/console.py
+++ b/opentimelineview/console.py
@@ -33,6 +33,7 @@ from PySide2 import QtWidgets, QtGui
 
 import opentimelineio as otio
 import opentimelineview as otioViewWidget
+from opentimelineview import settings
 
 
 def _parsed_args():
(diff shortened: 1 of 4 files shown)


ssteinbach:
Really small annoying request, but can you also update the screenshot in the readme with the new color scheme?  Thanks!

boredstiff:
Where would you like the image? It currently points to an image in the wiki, which I don't have access to edit. 

You could optionally just change the image at this address: https://github.com/PixarAnimationStudios/OpenTimelineIO/wiki/images/otioview.png

to be this image

screen shot 2018-09-25 at 7 57 03 pm

And you wouldn't have to update anything. However, once #312 is merged into master, and the wiki isn't updated anymore, you can just use docs/_static and store the image in there.

Issue #306, opened 2018-08-12
"Hooks" plugin system


ssteinbach:
It would be nice to have general plugins more general than media linkers and adapters that can run at predefined points in the API.  Initially I think we could support:

- PostImport (after `read_from_*` functions)
- PreWrite (before `write_to_*` functions)

These would be single function plugins that take a timeline and return a timeline and let you manipulate data.  I think we could extend the plugin API to have another environment variable `OTIO_HOOK_POSTIMPORT` with a `:` separated list of plugins to be run.

ssteinbach:
See #330

PR #330, opened 2018-09-20
New plugin type: HookScripts

ssteinbach:
HookScripts are plugins that run at defined points ('Hooks').

They expose a hook_function with signature:
    `hook_function :: otio.schema.Timeline, Dict -> otio.schema.Timeline`

Both hook scripts and the hooks they attach to are defined in the plugin manifest.

You can attach multiple hook scripts to a hook.  They will be executed in list order, first to last.

They are defined by the manifests HookScripts and hooks areas.  For example:
```json
{
    "OTIO_SCHEMA" : "PluginManifest.1",
    "hook_scripts" : [
        {
            "OTIO_SCHEMA" : "HookScript.1",
            "name" : "example hook",
            "execution_scope" : "in process",
            "filepath" : "example.py"
        }
    ],
    "hooks" : {
        "pre_adapter_write" : ["example hook"],
        "post_adapter_read" : []
    }
}
```

The 'hook_scripts' area loads the python modules with the 'hook_function's to call in them.  The 'hooks' area defines the hooks (and any associated scripts).  You can further query and modify these from python.

In other words:

```python
import opentimelineio as otio
hook_list = otio.hooks.scripts_attached_to("some_hook") # -> ['a','b','c']

# to run the hook scripts:
otio.hooks.run("some_hook", some_timeline, optional_argument_dict)

# this will pass (some_timeline, optional_argument_dict) to 'a', which will
# a new timeline that will get passed into 'b' with optional_argument_dict,
# etc.

# To Edit the order, change the order in the list:
hook_list[0], hook_list[2] = hook_list[2], hook_list[0]
print hook_list # ['c','b','a']

# Now c will run, then b, then a.

# To delete a function the list:
del hook_list[1]
```

This also sets up three default hooks you can attach scripts to:
- post_adapter_read
- post_media_linker
- pre_adapter_write


meshula:
This facility seems cool, but it wouldn't be portable to domains that don't know the hookscript's language. Are plugins scoped to the standard viewer? The sample says scoped "in process", but I'm not sure which process ;)

ssteinbach:
@meshula Good question!  *Currently* our conversation around this is that we will likely either leave plugins in python or allow python plugins specifically.  The thought was that as long as we are handing JSON between the plugins they can be run in subprocesses or called through the python/C API.

ssteinbach:
Also the 'in process' thing should probably be removed, it was something we haven't built out in the current version of the library.

meshula:
My main concern is that ultimately I expect to be able to read and interpret OTIO files per se in runtime environments where Python is not available. If plugins are not part of the overall OTIO specification that I'd expect to be able to read in any arbitrary DCC, then it's not an issue. (Plugins aren't documented yet in the wiki or google group, I think, hence my questions)

ssteinbach:
Post this PR, there are three kinds of plugins to OTIO:
- adapters (documentation: https://github.com/PixarAnimationStudios/OpenTimelineIO/wiki/How-to-Write-an-OpenTimelineIO-Adapter)
- media linkers (documentation: https://github.com/PixarAnimationStudios/OpenTimelineIO/wiki/How-to-write-an-OpenTimelineIO-Media-Linker)
- hooks/hookscripts (documentation page not done yet, see the comment in the PR)

The C/C++ API implementation effort is still in early days (thanks for your notes on that, btw!).  In our discussions, what we've said is that we want to be able to allow writing plugins in python, using the JSON format to go between python and the C/C++ API.  The JSON serialization system will definitely be ported to C/C++, so that will always be available.  This means at the very least you'll always be able to read/write and manipulate .otio files with C/C++, even if python isn't available.

What we haven't decided yet is if we should also allow C/C++ plugins in addition to python, or if we plan on porting any of the core adapters down to C/C++.

When you talk about the specification for OTIO, I _think_, at least as I write this, that plugins are not a part of the OTIO *format* specification but rather a feature of the OTIO API/library.  The intent of media linkers and hooks is to allow studios to build integration to their pipeline in such a way that anything calling into the OTIO library can leverage them.

So for example, a third party player application could call into OTIO and the studios' media linker could run and link media correctly.

meshula:
On the surface of it, it seems natural that a format specification, and a library to exercise it are two separate things. That works well as long as there isn't an implicit or procedural component. It seems appropriate, to me ;) , that OTIO be two things, as I can see it being useful in a web app, or an iOS app, for example. A contrasting example, where format and library are not separable, would be USD. USD implements deep evaluative heuristics for such things as overrides, and the files are therefore interpretable only by execution of the USD library. So, the USD format and library are not usefully separable, except for a trivial subset of features (such as is currently emerging for the USDZ sub-format).

ssteinbach:
Addesses #306

ssteinbach:
@meshula We have no intent in making the OTIO format implement any kind of deep evaluative hueristics, but we would like to make it convenient for users of the API to inject their own.  If you have an application like you describe, that depends on data that is set up by a plugin, you might have to generate an OTIO file on another machine with access to the plugins to send into the more constrained environment.  That seems like a reasonable trade off to us.

alatdneg:
Looks good, one question how do you deal with ordering when there are multiple plugin manifest files?

ssteinbach:
@alatdneg Great question.  If you look here:
https://github.com/PixarAnimationStudios/OpenTimelineIO/blob/59a60bbf965cff7347336a76d899aaea55fd9c32/opentimelineio/plugins/manifest.py#L129

What happens is that a series of manifests get layered in, in order.  The first one is the core plugins one, then contrib, then any installed via pip entry points, then off of `OTIO_PLUGIN_MANIFEST_PATH`.

It then calls `extend` on the plugin lists.  Looking at it now, this could be wrapped up in a better way by the `Manifest` Class.  I'll add a note for this.  Thanks!

PR #335, opened 2018-09-29
Add support for muted clips to AAF adapter.

ssteinbach:
- Handles the 'selector' aaf object, which specific code to handle muted clips (a selector with one item).
- Also adds a check to otiostat to look for tracks of unequal length (we've found this is a good indicator for something is wrong in the conversion).


opentimelineio_contrib/adapters/tests/test_aaf_adapter.py
```diff
@@ -727,6 +731,15 @@ def test_read_misc_speed_effects(self):
         # TODO: We don't yet support non-linear time warps, but when we
         # do then this effect is a "Speed Bump" from 166% to 44% to 166%
 
+    def test_muted_clip(self):
+        sc = otio.adapters.read_from_file(MUTED_CLIP_PATH, simplify=False)
+        gp = sc[0].tracks[8][0][0]
+        gp.metadata['AAF']['muted_clip']
```
jminor: this line is superfluous

opentimelineio_contrib/adapters/tests/test_aaf_adapter.py
```diff
@@ -727,6 +731,15 @@ def test_read_misc_speed_effects(self):
...
+
+        self.assertTrue(gp.metadata['AAF']['muted_clip'])
+        self.assertIsNotNone(gp)
```
jminor: Can you swap the previous two lines? As-is, the first would already fail if gp was None.

PR #342, opened 2018-10-11
new SchemaDef plugin type to define site-specific schemas

peachey:
This PR adds a new SchemaDef plugin type.  A SchemaDef plugin can define one or more new OTIO schemas (or schemata if you prefer) which define first-class serializable OTIO objects and can be upgraded based on changes to the schema version number just like other OTIO schema-based objects.  

One possible use case is to define new schemas for studio-specific metadata objects which can be included in the clip "metadata" field.  For instance, our internal media linker plugin uses a few Pixar-specific schemas to provide metadata values which can be automatically upgraded when the schema version number increases.  Those Pixar-specific metadata schemas are defined in a single SchemaDef Python plugin and included by our plugin_manifest.json.

### DIFF
--- a/tests/test_schemadef_plugin.py
+++ b/tests/test_schemadef_plugin.py
@@ -55,18 +55,17 @@ class TestPluginSchemadefs(unittest.TestCase):
     def test_plugin_schemadef(self):
         # Our test manifest should have been loaded, including
         # the example_schemadef.
+        # Try creating a schema object using the instance_from_schema method.
         peculiar_value = "something One-derful"
-        try:
-            example = otio.core.instance_from_schema("exampleSchemaDef", 1, {
-                EXAMPLE_ARG: peculiar_value
-            })
-        except otio.exceptions.NotSupportedError:
-            self.fail("raised NotSupportedError (new schema type was undefined)")
+        example = otio.core.instance_from_schema("exampleSchemaDef", 1, {
+            EXAMPLE_ARG: peculiar_value
+        })
 
         self.assertEqual(str(type(example)), EXCLASS)
         self.assertEqual(example.exampleArg, peculiar_value)
 
-        # Repeat test using the direct class definition method:
+    def test_plugin_schemadef_namespace(self):
+        # Try creating schema object with the direct class definition method:
         peculiar_value = "something Two-derful"
         example = otio.schemadef.example_schemadef.exampleSchemaDef(peculiar_value)
         self.assertEqual(str(type(example)), EXCLASS)
--- a/tests/test_schemadef_plugin.py
+++ b/tests/test_schemadef_plugin.py
@@ -1,0 +1,69 @@
+#
+# Copyright 2017 Pixar Animation Studios
+#
+# Licensed under the Apache License, Version 2.0 (the "Apache License")
+# with the following modification; you may not use this file except in
+# compliance with the Apache License and the following modification to it:
+# Section 6. Trademarks. is deleted and replaced with:
+#
+# 6. Trademarks. This License does not grant permission to use the trade
+#    names, trademarks, service marks, or product names of the Licensor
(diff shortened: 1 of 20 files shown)


peachey:
I think the test failures from Travis CI are actually the ones that will be fixed when ssteinbach's PR #347 is merged.

tests/test_schemadef_plugin.py
```diff
@@ -0,0 +1,77 @@
...
+            os.environ['OTIO_PLUGIN_MANIFEST_PATH'] = self.save_manifest_path
+        else:
+            del os.environ['OTIO_PLUGIN_MANIFEST_PATH']
+        otio.plugins.manifest._MANIFEST = self.save_manifest
+
+    def test_plugin_schemadef(self):
+        # Our test manifest should have been loaded, including
+        # the example_schemadef.
+        peculiar_value = "something One-derful"
+        try:
```
ssteinbach: You can use the `with assertRaises` syntax to do this, see:
https://github.com/PixarAnimationStudios/OpenTimelineIO/blob/f5a4480d7418cfbfea6b7604e9d28b46e4c683aa/tests/test_opentime.py#L88

tests/test_schemadef_plugin.py
```diff
@@ -0,0 +1,77 @@
...
+            example = otio.core.instance_from_schema("exampleSchemaDef", 1, {
+                EXAMPLE_ARG: peculiar_value
+            })
+        except otio.exceptions.NotSupportedError:
+            self.fail("raised NotSupportedError (new schema type was undefined)")
+
+        self.assertEqual(str(type(example)), EXCLASS)
+        self.assertEqual(example.exampleArg, peculiar_value)
+
+        # Repeat test using the direct class definition method:
```
ssteinbach: Could make this a second test method, since you've already set up the values.

tests/test_unknown_schema.py
```diff
@@ -0,0 +1,81 @@
...
+"""
+
+class UnknownSchemaTests(unittest.TestCase, otio.test_utils.OTIOAssertions):
+    def setUp(self):
+        # make an OTIO data structure containing an undefined schema object
+        self.orig = otio.adapters.otio_json.read_from_string(has_undefined_schema)
+
+    def test_serialize_deserialize(self):
+        serialized = otio.adapters.otio_json.write_to_string(self.orig)
+        test_otio = otio.adapters.otio_json.read_from_string(serialized)
```
ssteinbach: We also usually do a second test where we serialize/deserialize it again just to make sure there isn't some slightly different state if the object starts in memory vs starts from the deserializer.

tests/test_schemadef_plugin.py
peachey: The intent here was to check that it DOESN'T raise NotSupportedError, whereas self.assertRaises checks that it DOES raise the exception, I think.  But this condition likely could be removed anyway, since I took out the code in instance_from_schema that could raise this exception.  Instead now it will return an UnknownSchema object, so I should check for that instead, I guess.

peachey: Oh, and it already does check that by verifying the type of the created class, so if it is UnknownSchema rather than exampleSchemaDef, the type check will fail.

tests/test_schemadef_plugin.py
peachey: Yes, good idea.

tests/test_unknown_schema.py
peachey: I'm not sure I follow that.  In this case, self.orig was deserialized from a string constant, so it's in memory.  Then we serialize and deserialize it just to verify that the OTIO is the same, ie: that it preserved the undefined schema object.  I really wanted to check that the "serialized" string is the same string as the original string, but that's brittle.  So the second read_from_string is just to get something that can be compared more reliably.

PR #344, opened 2018-10-17
More examples & documentation

jminor:


docs/tutorials/time-ranges.md
```diff
@@ -0,0 +1,127 @@
...
+    This will return the Clip's `source_range` if it is set, otherwise it will
+    return the `available_range()`. This tells you how long the Clip is meant to
+    be in its parent Track or Stack.
+
+    - `clip.visible_range()`
+
+    This will return the same thing as `trimmed_range()` but also takes any 
+    adjacent Transitions into account. For example, a Clip that is trimmed to 
+    end at frame 10, but is followed by a symmetric 10 frame cross-dissolve, 
+    will have a `visible_range()` that ends at frame 15.
```
avrata: Should this say "ends at frame 20"?  I thought visible_range took into account the entire length of the dissolve (ie all frames shown, even if only a little).

avrata: Oh, nevermind. I think I understand in this case the 10 frame dissolve overlaps -5 frames back into the trimmed_range.

docs/tutorials/time-ranges.md
```diff
@@ -0,0 +1,127 @@
+# Time Ranges
+
+Overview
+----------
+
+A Timeline and all of the Tracks and Stacks with in it work together to place
```
ssteinbach:
```suggestion
A Timeline and all of the Tracks and Stacks it contains work together to place
```

docs/tutorials/time-ranges.md
```diff
@@ -0,0 +1,127 @@
...
+Note: You probably also want to refer to [Timeline Structure](otio-timeline-structure.html).
+
+Clips
+-------
+
+There are several ranges you might want from a Clip.
+
+- Ranges within the Clip and its media:
+
+    - `clip.available_range()`
```
ssteinbach:
```suggestion
    - `Clip.available_range()`
```

I think the class is upper case and the module is lower case?

docs/tutorials/time-ranges.md
jminor: I'll clarify that. I meant the dissolve was 5 frames on either side of the cut.

docs/tutorials/time-ranges.md
```diff
@@ -0,0 +1,127 @@
...
+    Note that `source_range` may be `None` indicating that the Clip should use
+    the full `available_range()` whatever that may be. If both `source_range`
+    and `available_range()` are None, then the Clip is invalid. You need at
+    least one of those.
+
+    Usually this will be a shorter segment than the `available_range()` but this
+    is not a hard constraint. Some use cases will intentionally ask for
+    a Clip that is longer (or starts earlier) than the available media as a way
+    to request that new media (a newly animated take, or newly recorded audio) 
+    be made to fill the requested `source_range`.
```
ssteinbach: You may want to note that `source_range` is in the coordinate frame of the `available_range`.  In other words, if the `available_range` is from frame 5 to 20, and the source range is from 0 to 10, that doesn't mean frame 5-15 in the available range, but rather frames 0 to 10, of which only 5 are in the available range (which may indicate a problem).

ssteinbach: Do we define what that coordinate frame is anywhere?  Or how it relates to the coordinate frame of the media reference?  Regardless, sounds good.

ssteinbach: That stuff is also just looking ahead to when we start back in on the time stuff.

PR #347, opened 2018-10-18
Fix subtle bug with several unit tests

ssteinbach:
There was a bug in a couple of the tests that was causing test failures in certain environments where the OTIO environment variables were set but plugins were loaded via pip.  The tests weren't doing a good enough job of bookkeeping the state of the central `ActiveManifest`.

### DIFF
--- a/tests/test_adapter_plugin.py
+++ b/tests/test_adapter_plugin.py
@@ -190,13 +190,30 @@ class TestPluginManifest(unittest.TestCase):
         )
 
     def test_find_manifest_by_environment_variable(self):
-        # write the file into a temp dir
         suffix = ".plugin_manifest.json"
+
+        # back up existing manifest
+        bak = otio.plugins.manifest._MANIFEST
+        bak_env = os.environ.get('OTIO_PLUGIN_MANIFEST_PATH')
+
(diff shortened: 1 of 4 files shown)


tests/test_adapter_plugin.py
```diff
@@ -190,13 +190,33 @@ def test_find_adapter_by_name(self):
...
+
+            # Rather than try and remove any other setuptools based plugins
+            # that might be installed, this check is made more permissive to
+            # see if the known unit test linker is being loaded by the manifest
+            self.assertTrue(len(result.media_linkers) > 0)
+            self.assertIn("example", (ml.name for ml in result.media_linkers))
+
+        otio.plugins.manifest._MANIFEST = bak
+        if bak_env:
+            os.environ['OTIO_PLUGIN_MANIFEST_PATH'] = bak_env
```
peachey: Don't you need to add an else clause to restore original state if this envvar was not originally set?
```suggestion
            os.environ['OTIO_PLUGIN_MANIFEST_PATH'] = bak_env
```
```suggestion
        else:
            os.environ['OTIO_PLUGIN_MANIFEST_PATH'] = bak_env
```

PR #358, opened 2018-11-02
Optimizations for flatten (range_of_all_children)

ssteinbach:
Flattening for timelines with many tracks/items was slow because `range_in_parent` was computed by walking through all previous items and summing their times.  This adds a function (`range_of_all_children`) that computes a map of item to range_in_parent, so you can do a look up rather than a compute to get the value.  This speeds up one particular case from 1600s to flatten down to 6s.

### DIFF
--- a/opentimelineio/algorithms/stack_algo.py
+++ b/opentimelineio/algorithms/stack_algo.py
@@ -27,7 +27,7 @@ __doc__ = """ Algorithms for stack objects. """
 import copy
 
 from .. import (
-    schema
+    schema,
 )
 from . import (
     track_algo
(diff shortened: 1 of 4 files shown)


jminor:
Hey @alatdneg does this address the speed issues you encountered?

alatdneg:
Does this create a baked track ? It'll help, but I think my code was modifying the track at the same time. So This probably will not help a huge amount.
I've got my bisect code anyway, which I add to my branch, which removes the issue. (more or less)

ssteinbach:
It doesn't create any kind of new object beyond a dictionary.  If you modified the track and knew the index of the first thing past what you modified + the duration of the previously modified thing it would be trivial (and still very fast) to iterate through this dictionary and update the ranges, if you needed to.  I would have to look more closely at your use case though.

Issue #349, opened 2018-10-23
RV Adapter should create a copy of the environment instead of directly modify it


hasielhassan:
The RV adapter is directly modifying the environment and specifically the PYTHONPATH variable in order to pass the proper paths to the subprocess, but this can cause issues in an scenario where you actually require other modules named RV.py like when running inside RV and you need to import rv module @_@

The affected module is:
opentimelineio_contrib/adapters/rv.py

Issue #341, opened 2018-10-11
"SchemaDef" Plugins


ssteinbach:
- A plugin which lets you inject schemas into the schema map, or override existing schemas with custom definitions (including versioning/upgrade functionality)
- When the serializer encounters a type from a plugin, it loads the python file associated with the plugin (if not already loaded), and then deserializes the type
- If it encounters a type it doesn't know, instead of throwing an exception (current behavior) it should instead create an "UnknownSchema" type object (which is probably just a descendent of SerializeableObject with no extra functionality) that allows users to still read the data.

This lets studios add their own types (for example, for metadata) and hand OTIO files with those schemas around without requiring them to be directly stripped out.

Issue #338, opened 2018-10-02
Better Manifest layering


ssteinbach:
Right now Manifests are layered by manually extending the adapter and media linker lists.  This should be extended to cover the new hooks plugins and should probably be wrapped up in a higher level function in the `Manifest` class.

https://github.com/PixarAnimationStudios/OpenTimelineIO/blob/59a60bbf965cff7347336a76d899aaea55fd9c32/opentimelineio/plugins/manifest.py#L203

Issue #320, opened 2018-09-05
Constructors should copy arguments ?


alatdneg:
Currently they just take references, this can give unexpected results if you're not careful.
If we're taking references then the reference object should really be immutable ?

```
t = TimeRange(RationalTime(1,1),RationalTime(1,1))
g = Gap(source_range=t)
g2 = Gap(source_range=t)
g.source_range.duration.value = 3

t
otio.opentime.TimeRange(start_time=otio.opentime.RationalTime(value=1, rate=1), duration=otio.opentime.RationalTime(value=3, rate=1))

g2
otio.schema.Gap(name=None, source_range=otio.opentime.TimeRange(start_time=otio.opentime.RationalTime(value=1, rate=1), duration=otio.opentime.RationalTime(value=3, rate=1)), effects=[], markers=[], metadata={})


```

alatdneg:
Another option is to make value/rate read only..

alatdneg:
Ack += RationalTime modifies the source .value, I suspect this has introduced some bugs in my code.. 
Which means I've already fallen in to this trap, drat..

alatdneg:
It gets worse.. 

This function can either return a reference or a new object.. 

```
   def end_time_inclusive(self):
        """The time of the last sample that contains data in the TimeRange.

        If the TimeRange goes from (0, 24) w/ duration (10, 24), this will be
        (9, 24)

        If the TimeRange goes from (0, 24) w/ duration (10.5, 24):
        (10, 24)

        In other words, the last frame with data (however fractional).
        """

        if (
            self.end_time_exclusive()
            - self.start_time.rescaled_to(self.duration)
        ).value > 1:

            result = (
                self.end_time_exclusive() - RationalTime(1, self.duration.rate)
            )

            # if the duration's value has a fractional component
            if self.duration.value != math.floor(self.duration.value):
                result = self.end_time_exclusive()
                result.value = math.floor(result.value)

            return result
        else:
            return self.start_time
```
Safe.. Constructs new obj

 ```
   def end_time_exclusive(self):
        """"Time of the first sample outside the time range.

        If Start Frame is 10 and duration is 5, then end_time_exclusive is 15,
        even though the last time with data in this range is 14.

        If Start Frame is 10 and duration is 5.5, then end_time_exclusive is
        15.5, even though the last time with data in this range is 15.
        """

        return self.duration + self.start_time.rescaled_to(self.duration)


```

jminor:
[...]

PR #359, opened 2018-11-05
Bugfixes: Issue 320, 338, and 349 (biggest feature: constructors copy values)

ssteinbach:
This PR should address:

- #320 -- constructors now copy arguments, and `TimeRange.end_time_inclusive` returns a copy
- #338 -- (partially addressed by #341) -- Upgrade functions are sorted to ensure that they run in the right order
- Removed some unnecessary code in the JSON deserializer
- #349 -- RV adapter creates a copy of the environment rather than directly modifying os.environ.

### DIFF
--- a/opentimelineio/opentime.py
+++ b/opentimelineio/opentime.py
@@ -235,8 +235,8 @@ class TimeTransform(object):
 
     def __init__(self, offset=RationalTime(), scale=1.0, rate=None):
         self.offset = copy.copy(offset)
-        self.scale = copy.copy(scale)
-        self.rate = copy.copy(rate)
+        self.scale = scale
+        self.rate = rate
 
     def applied_to(self, other):
         if isinstance(other, TimeRange):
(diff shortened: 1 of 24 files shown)


reinecke:
Good to see this change exposed some programming errors in the adaptors!

opentimelineio/opentime.py
```diff
@@ -233,9 +234,9 @@ class TimeTransform(object):
     """1D Transform for RationalTime.  Has offset and scale."""
 
     def __init__(self, offset=RationalTime(), scale=1.0, rate=None):
-        self.offset = offset
-        self.scale = scale
-        self.rate = rate
+        self.offset = copy.copy(offset)
```
alatdneg: You don't need to copy primitives ? i.e. int , float etc.

ssteinbach: Offset is a RationalTime.  Scale and Rate are primitives though, I'll fix that.

PR #361, opened 2018-11-07
AAF EssenceGroup fix

jminor:
This adds minimal support for reading AAF EssenceGroups.


jminor:
@ssteinbach this is rebased onto your AAF tests fix & I think I addressed your notes. Can you take another look?

PR #364, opened 2018-11-08
Use a weakref for _parent to prevent cycles.

ssteinbach:
The _parent pointer should be a weakref.  This also enforces that nothing outside of item should use the _parent pointer directly, things should go through `.parent()`

### DIFF
--- a/opentimelineio/core/composable.py
+++ b/opentimelineio/core/composable.py
@@ -27,6 +27,8 @@
 An object that can be composed by tracks.
 """
 
+import weakref
+
 from . import serializable_object
 from . import type_registry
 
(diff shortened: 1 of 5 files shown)


reinecke:
Nice catch! I wonder if we might be able to work some profiling visibility into our CI pipeline. There are some interesting puzzle pieces available:

- https://pypi.org/project/pytest-leaks/
- https://pypi.org/project/pytest-profiling/

We'd still need to do something with the data, however.

ssteinbach:
@reinecke if you can find a service (like codecov for coverage) that can present the data and is free for open source projects, we'd be happy to discuss integrating more metrics.  As we move to C++ having static analysis be part of the PR would be pretty cool.

PR #365, opened 2018-11-09
[WIP] Open-ended time ranges

reinecke:
This relates to issue #367

Added support for _some_ `TimeRange` operations on `TimeRangeOpenEnd`, I didn't make `TimeRangeOpenFront` but this implementation should support it. The idea is that `TimeRangeOpenFront` would have `start_time` be its end time (the fact that it sounds that awkward maybe indicates some weakness in this approach).

One alternate approach to this would be to instead add `RationalTimeInfinity` and `RationalTimeNegativeInfinity`. Additionally `TimeRange` would be refactored with `start_time` and `end_time` (exclusive) as its primary data attributes. We could then make the constructor still accept duration for backward-compatibility and update the duration property implementations to compute duration rather than return a value.

This is just an early sketch to use as a talking point and the code isn't properly formatted or fleshed out. The big stopping point is that proper implementations of `clamped` and `extended_by` would require knowledge of the `TimeRange` subclasses so that if invoked on a `TimeRange` and passed a `TimeRangeOpen(Front|End)` it can return the correct instance type. This felt weird to me.

Some notes:

- I made `duration` and the `end_time` methods return `None` for this, do we think that's correct?
- [ ] We'll need to have `TimeRangeOpenEnd`, `TimeRangeOpenFront`, and something like  `TimeRangeAllTimes` to support the union of `TimeRangeOpenFront` and `TimeRangeOpenEnd`
- [ ] What should we be using for the `hash` of the open-bounded ranges?
- [ ] Serialization unimplemented

### DIFF
--- a/opentimelineio/opentime.py
+++ b/opentimelineio/opentime.py
@@ -317,7 +317,7 @@ class TimeRange(object):
 
     def __init__(self, start_time=RationalTime(), duration=RationalTime()):
         self.start_time = copy.copy(start_time)
-        self.duration = copy.copy(duration)
+        self._duration = copy.copy(duration)
 
     def __copy__(self, memodict=None):
         # Construct a new one directly to avoid the overhead of deepcopy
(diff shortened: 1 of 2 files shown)


reinecke:
Closing this PR. Issue #367 has a lot of discussion around this. Any solution we decided to pursue should be implemented in the C++ code.

PR #373, opened 2018-11-13
use bisect in each_child

alatdneg:


alatdneg:
Note sure this code is still valid, as the compostion code has probably moved on a bit since I wrote this.
The core-tests worked though.

ssteinbach:
This is interesting.  I did a quick test using `range_of_each_child`:
```python    
        child_time_map = self.range_of_all_children() if search_range else None
        for i, child in enumerate(self._children):
            # filter out children who are not in the search range
            if search_range and not child_time_map[child].overlaps(search_range):
                continue
```

It *seems* to be pretty fast.  On our canonical "big file" (20mb) test, we tried it with the trivial loop:
```python
    tl = otio.adapters.read_from_file(filepath)
    tr = tl.tracks.trimmed_range()
    for cl in tl.each_clip(tr):
        cl.name
```

For that file, processing time went from 500s to 2.1s. (including parse time)... but you can't modify the track as you go with this approach.

ssteinbach:
In order to do the above test, I also had to add a `range_of_all_children` function for stacks.  So that might have to be elevated up a bit if we end up wanting to go with that approach.

ssteinbach:
For comparison the bisect case takes 3.2s on this file.

alatdneg:
Cool glad to be of help, I hate slow code ;)

ssteinbach:
@alatdneg I made a PR to the branch in your personal repo with the merged version + tweaks to make it pass the linter.  If you accept it there, those changes should get put into this PR and then we can accept it.  Let me know if you have any questions!

ssteinbach:
Merged in PR #404

Issue #379, opened 2018-11-14
documentation wrong for SchemaDef plugin


peachey:
frederick.tombs@autodesk.com reported trouble in making a SchemaDef plugin on the
[google group](https://groups.google.com/d/msgid/open-timeline-io/6d4384c8-f9a9-4317-a4fb-227b501c46f0%40googlegroups.com?utm_medium=email&utm_source=footer).

It turns out that the mechanism works, but I documented it incorrectly in:

[Writing an OTIO SchemaDef Plugin — OpenTimelineIO 0.10.0.dev1 documentation](https://opentimelineio.readthedocs.io/en/latest/tutorials/write-a-schemadef.html)

The plugin manifest entry for a SchemaDef plugin should have an OTIO_SCHEMA value of `"SchemaDef.1"`
but I showed the wrong schema tag in my example, i.e.: tagging it `"MyThing.1"` rather than `"SchemaDef.1"`.

Apologies to everyone who has been misled by this documentation error.  We'll get it fixed up ASAP.

-- Darwyn

ssteinbach:
Addressed by #380

PR #380, opened 2018-11-14
fix an error in the schemadef plugin tutorial

peachey:
This fixes issue #379.  A critical change to one string in the plugin manifest example.


ssteinbach:
Thanks!

PR #386, opened 2018-11-20
Hide deserialization methods of SerizableObject

ssteinbach:
We exposed the `data` dictionary (really the backing of `SerializableObject`), but since that is likely to change in the C++ version, this hides that.  We're also hiding the `.update()` method, since it only seems to be used by internal implementation details that are also likely to change in the C++ version.


reinecke:
:+1:

PR #388, opened 2018-11-28
Small general updates

shahbazk8194:


docs/tutorials/architecture.md
```diff
@@ -61,13 +58,13 @@ for clip in tl.each_clip():
 
 ## Time on otio.schema.Clip
 
-A clip may set its timing information (which is used to compute its `duration()` or its `trimmed_range()`) by configuring either its:
-- `.media_reference.available_range`
+A clip may set it's timing information (which is used to compute it's `duration()` or it's `trimmed_range()`) by configuring either it's:
```
jminor: Actually "its" is correct. https://www.dictionary.com/e/its-vs-its/

docs/tutorials/architecture.md
```diff
@@ -79,12 +76,12 @@ cl.trimmed_range()
 cl.available_range() # == cl.media_reference.available_range
 ```
 
-Generally, if you want to know the range of a clip, we recommend using the `.trimmed_range()` method, since this takes both the `media_reference.available_range` and the `.source_range` into consideration.
+Generally, if you want to know the range of a clip, we recommend using the `trimmed_range()` method, since this takes both the `media_reference.available_range` and the `source_range` into consideration.
```
jminor: I like that you're removing the `.` It reads much better with your change :)

PR #390, opened 2018-11-28
Now passing track_kind to transition/dissolve creation

avrata:
As well as a quick variable rename to make it more clear what's used in `_create_media_reference()`


avrata:
@apetrynet, while going through this code for something else, I noticed what I think is a a super small oversight in that the new `track_kind` which was not being passed into the transition code. Can you take a quick look and see what you think?

apetrynet:
@avrata Good catch! That was indeed missed. I never tested with any transitions unfortunately.

Issue #168, opened 2017-10-08
EDL clip names as reel names


tin2tin:
Clip names are often written/read(ex. Lwks, Flowblade, ShotCut and Blender) as reel names in EDL files like this(case sensitive):
```
001 4388aaBB V C 05:04:46:13 05:04:56:03 01:03:16:04 01:03:25:19
* FROM CLIP NAME: 4388aaBB.dpx
```
[NB. spacing is truncated in example. File name as Reel name is case sensitive(in Lwks), 8 characters long, no empty space characters in file name(Blender) and with no extention]
Looking through the OTIO EDL code, this practice seems to be ignored i/o in EDL files. 

[Btw. really great project. Hopefully, the open source A/V projects will embrace it. If ex. MLT was added as a format then Kdenlive, ShotCut, and Flowblade could all benefit from it. It would also be great to see Blender, Ardour, Audacity, Synfig, FFmpeg, OpenToonz, Storyboarder etc. sharing projects with Lwks, Resolve, Premiere, Final Cut and Avid and vice versa. ]

jminor:
You're right, the EDL adapter doesn't really pay much attention to the reel, which is an oversight. In the case you mention, it appears to be also covered by `FROM CLIP NAME`. If the EDL output from the adapter put the truncated/abbreviated clip name into the reel would that be sufficient, or does it need to read the reel name and carry that around too?

tin2tin:
"* From clip name: " is an addition to the original CMX 3600 EDL format. Ex. Lwks and Blender can't parse these comment lines, so the only way is to use the reel name to identify the files in those softwares. But I think most pro NLE's will look in the comments for a file-name with extension, after "* From clip name:". I know Resolve will do that if you check the right boxes for the import settings. I don't know which NLEs will use the full path "* From file:", though that would be the logic - I haven't tested those. 

If you can provide a short OTIO edl file and a video clip with cuts at the same frames as there are cuts in the video clip I can test it against Resolve, Blender and Lwks. [Like I did for ShotCut: https://forum.shotcut.org/t/mlt2edl-script-error-fix/55/40]

mikemahony:
Does https://github.com/PixarAnimationStudios/OpenTimelineIO/pull/200 address this issue? We set `clip.metadata['cmx_3600']['reel']` to the clip name before exporting an EDL so that it may be read and displayed by our 3rd party mastering suite.

tin2tin:
[...]

PR #392, opened 2018-11-29
Reel name updates for the cmx_3600 adapter

apetrynet:
Hi!
As discussed in #168 here is a suggestion for handling reel names in EDL export. With these changes we no longer default reel names to _AX_ but the best suited name based on clip information.
 
I added a `trunc_reelname=True` argument to the `write_to_string` function of the adapter which toggles whether or not to truncate the reel name to 8 characters. When truncating is enabled the adapter adds a comment line to the EDL indicating the original source name for the reel. 
Reel names get stripped of non-alphanumeric characters.

In addition I added storing of incoming reel name in a clips' metadata when reading EDLs.

Please note that one of the unit tests is broken


opentimelineio/adapters/cmx_3600.py
```diff
@@ -721,7 +729,7 @@ def read_from_string(input_str, rate=24, ignore_timecode_mismatch=False):
     return result
 
 
-def write_to_string(input_otio, rate=None, style='avid'):
+def write_to_string(input_otio, rate=None, style='avid', trunc_reelname=True):
```
ssteinbach: Instead of passing in `trunc_reelname` you could pass in `reelname_len` (by default 8), which if you pass in `None` would not truncate at all?  Or is it _always_ 8?

apetrynet: Good idea. The standard is 8, but a lot of apps use longer reelnames. This is a better way of supporting that. 
Do you have any thought on the compare test of the two jsons that I mentioned? I can always create/modify the source json to what i expect and compare that?

jminor: Could you use the optional `flavor` argument to include the new stuff, keeping the old stuff as-is?
Also, it looks like the tests are failing in Travis due to some linter issues. Once that goes further to the EDL tests, then it will be easier for us to see the diffs you are talking about.

apetrynet: When you say "the new stuff", do you mean the new comment and metadata or do you mean everything? The 8 character reelname length I believe is in the cmx3600 specs and should perhaps be part of that flavor? As for the new comment line that was just an idea to inform users about the truncated reelname as discussed in issue #168.
I'll fix the lint errors asap. Thought I caught all of them. Sorry about that.

apetrynet: @jminor lint fixes are done. It now fails on the compare test mentioned above. 
@ssteinbach I'm working on applying your comments, but have one question. When passing `None` to reelname_len, should I still strip the extension of the source file in the reelname?

ssteinbach: @apetrynet Those diffs in the test just look like you need to update the baselines.  You're adding more metadata, correct?

apetrynet: @ssteinbach Yes, I add the reelname to metadata when parsing an EDL.
Should I just add the expected metadata to the clips in `test_edl_round_trip_mem2disk2mem`?

ssteinbach:
Yeah if you remove the [wip] we will review it. Thanks!

apetrynet:
Done! Please let me know if I need to pull in the latest changes form master into this branch.

ssteinbach:
If you don’t mind rebasing that would be great!

ssteinbach:
Sorry that it took us a while to circle back to this.  We just landed #434, would you mind rebasing this change given the changes from #434?  Then we can merge this in.  Really sorry about the delay!

apetrynet:
Sure! On it now!

opentimelineio/adapters/cmx_3600.py
```diff
@@ -104,23 +104,18 @@ def add_clip(self, line, comments, rate=24):
...
-
         if comment_handler.unhandled:
             clip.metadata.setdefault("cmx_3600", {})
             clip.metadata['cmx_3600'].setdefault("comments", [])
             clip.metadata['cmx_3600']['comments'] += (
                 comment_handler.unhandled
             )
 
+        # Add reel name to metadata
+        if clip_handler.reel:
```
ssteinbach: So this is preserving even the 'AX' cases that @mikekoetter had pruned.  Do you think the strategy to ignore the AX ones is better?  (see the comment in the deleted block).

apetrynet: Yeah, when I look at it I see no reason for adding AX to the metadata. It should still be there in the event line if no clip is present.
I'll add the test for it.

ssteinbach: If you add the test, can you add back the comment?  I thought that was a good explanation.

apetrynet: @ssteinbach I added the pruning of 'AX', but and issue with the `test_cdl.test_cdl_round_trip` might have reviled some poor design choices from my end. I'm going to dig a bit deeper to come up with a solution, but I'm afraid I need until next week as I need to sleep now and I'm away from my computer this weekend. Sorry about that.

apetrynet: > If you add the test, can you add back the comment? I thought that was a good explanation.

Sure, no problem.

ssteinbach: No rush!  Definitely get your rest!

ssteinbach:
Thanks!

PR #395, opened 2018-12-03
[WIP] Example in-RV otio reader plugin

avrata:
For discussion. This is a copy and modify of the existing `extern_rv.py` session file based rv adapter and turns it into an "in-RV" plugin that'll support reading OTIO files directly into the current session.

### DIFF
--- a/contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py
+++ b/contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py
@@ -298,13 +298,20 @@ def _transcribe(item, parents, editRate, masterMobs):
         mastermob = child_mastermob or parent_mastermob or None
 
         if mastermob:
-            media = otio.schema.MissingReference()
+            mastermob_child = masterMobs.get(str(mastermob.mob_id))
+            unc_path = (mastermob_child.metadata.get("AAF", {})
+                                                .get("UserComments", {})
+                                                .get("UNC Path"))
+            if unc_path:
+                media = otio.schema.ExternalReference()
+                media.target_url = "file://" + unc_path.replace("\\", "/")
+            else:
+                media = otio.schema.MissingReference()
             media.available_range = otio.opentime.TimeRange(
                 otio.opentime.RationalTime(media_start, editRate),
                 otio.opentime.RationalTime(media_length, editRate)
             )
             # copy the metadata from the master into the media_reference
-            mastermob_child = masterMobs.get(str(mastermob.mob_id))
             media.metadata["AAF"] = mastermob_child.metadata.get("AAF", {})
             result.media_reference = media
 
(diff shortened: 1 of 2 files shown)


jminor:
I like this strategy better than our existing RV adapter. We can certainly have both, but doing this from within RV itself has several advantages:
- You can inspect the actual media as it loads via RV's API.
- Easier support for both read and write. (The RV adapter doesn't support reading because the rvSession module only supports writing.)
- Is mostly transparent to the user. They could open any OTIO-supported file format directly in RV (e.g. an EDL, FCPXML, AAF, etc.)

Ideally we could share code between this and the RV adapter, but that seems tricky since the RV in-memory API and the rvSession API are independent.
Other applications with their own app-specific file formats will likely have similar tradeoffs, so this can be a good point of reference.

Also, I like that this gets us closer to a general purpose OTIO playback tool, which would be really really handy.

opentimelineio_contrib/application_plugins/rv/example_otio_reader/otio_reader.py
```diff
@@ -0,0 +1,263 @@
...
+        new_media = [
+                        "{},start={},end={},fps={}.movieproc".format(
+                            kind,
+                            range_to_read.start_time.value,
+                            range_to_read.end_time_inclusive().value,
+                            range_to_read.duration.rate
+                        )
+                    ]
+
+    src = commands.addSourceVerbose(new_media)
```
JoshBurnell: Is there possibly a way to add the Sources outside of this adapter and, if they've already been added, have this adapter connect the existing sources into the Sequence?

I ask because, in our experience, while addSourceVerbose does add the source (verbosely), but it doesn't actively update the UI, so if you have several sources in your OTIO (as one would expect for a feature sequence), the user doesn't get any feedback until all the sources have been added. Maybe there's a world where you add all the sources for the OTIO in one step using rv.commands.addSources() (which provides a "Loading X of X" dialog) and then assemble the Sequence from the existing sources. (rv.commands.addSources() fires after-progressive-loading after it completes, so this second step could take place there.)

For an example OTIO reader, that may be overkill, but for production, it's something to consider.

Have you guys tried anything like that?

opentimelineio_contrib/application_plugins/rv/example_otio_reader/example_otio_reader_plugin.py
```diff
@@ -0,0 +1,168 @@
...
+
+import opentimelineio as otio
+
+import otio_reader
+
+
+class Mode(object):
+    sleeping = 1
+    loading = 2
+    processing = 3
```
JoshBurnell: This is super clever.

opentimelineio_contrib/application_plugins/rv/example_otio_reader/otio_reader.py
GatesARC: I've tried using QtGui.QApplication.processEvents() to solve this problem, but the nature of processEvents can reverse the order of items loaded and cause other unintended consequences.  You get a nice live loading feel though.  Maybe a more specific processEvents command could be a solution.  

I'm not sure which events to include though. IE
app.processEvents(QtCore.QEventLoop.ExcludeUserInputEvents, stepSize)

jminor:
Oops. This was closed by accident. We'll get it re-opened shortly...

avrata:
I was unable to re-opening this unfortunately, but recreated it as #565.

PR #396, opened 2018-12-04
Initial port of aaf adapter from using pyaaf to pyaaf2

freesonluxo:
This is an initial (and somewhat messy, but working!) port in advanced_authoring_formay.py of pyaaf to pyaaf2. All tests pass locally.

### DIFF
--- a/opentimelineio_contrib/adapters/advanced_authoring_format.py
+++ b/opentimelineio_contrib/adapters/advanced_authoring_format.py
@@ -30,25 +30,25 @@ Depending on if/where PyAAF is installed, you may need to set this env var:
 
 import os
 import sys
-import opentimelineio as otio
 from collections import Iterable
+import opentimelineio as otio
 
 lib_path = os.environ.get("OTIO_AAF_PYTHON_LIB")
 if lib_path and lib_path not in sys.path:
-    sys.path += [lib_path]
+    sys.path.insert(0, lib_path)
 
-import aaf2
-import aaf2.content
-import aaf2.mobs
-import aaf2.components
-import aaf2.core
+import aaf2  # noqa: E731
+import aaf2.content  # noqa: E731
+import aaf2.mobs  # noqa: E731
+import aaf2.components  # noqa: E731
+import aaf2.core  # noqa: E731
 
 debug = False
 __names = set()
 
 
 def _get_parameter(item, parameter_name):
-    values = dict([(value.name, value) for value in item.parameters.value])
+    values = dict((value.name, value) for value in item.parameters.value)
     return values.get(parameter_name)
 
 
@@ -68,13 +68,16 @@ def _walk_item(thing):
     #         for item in clip._walk_item():
     #             yield item
     #     else:
-    #         raise NotImplementedError("Sequence returned %s not implemented" %  str(type(segment)))
+    #         raise NotImplementedError(
+    #            "Sequence returned {} not implemented".format(type(segment))
     elif isinstance(segment, aaf2.components.EssenceGroup):
         yield segment
     elif isinstance(segment, aaf2.components.Filler):
         yield segment
     else:
-        raise NotImplementedError("walking %s not implemented" % str(type(segment)))
+        raise NotImplementedError(
+            "walking {} not implemented".format(type(segment))
+        )
 
 
 def _get_name(item):
@@ -159,12 +162,12 @@ def _extract_start_timecode(mob):
     """Given a mob with a single timecode slot, return the timecode in that
     slot or None if no timecode slots could be found.
     """
-    
+
     tc_list = []
     for s in mob.slots:
         if s.segment.media_kind != 'Timecode':
             continue
-        
+
         if s.segment.get('Start'):
             tc_list.append(s.segment.get('Start').value)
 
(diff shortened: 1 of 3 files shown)


jminor:
After this lands into OTIO, lets try to get the `_walk_item` function pushed upstream into pyaaf2.

opentimelineio_contrib/adapters/advanced_authoring_format.py
```diff
@@ -31,35 +31,64 @@
...
+import aaf2.mobs
+import aaf2.components
+import aaf2.core
 
 debug = False
 __names = set()
 
 
+def _get_parameter(item, parameter_name):
+    values = dict([(value.name, value) for value in item.parameters.value])
```
ssteinbach: fwiw I don't think you need the [ ], this should work too:
```suggestion
    values = dict((value.name, value) for value in item.parameters.value)
```

opentimelineio_contrib/adapters/advanced_authoring_format.py
```diff
@@ -31,35 +31,64 @@
...
 if lib_path and lib_path not in sys.path:
     sys.path += [lib_path]
 
-import aaf  # noqa (E402 module level import not at top of file)
-import aaf.storage  # noqa
-import aaf.mob  # noqa
-import aaf.define  # noqa
-import aaf.component  # noqa
-import aaf.base  # noqa
+import aaf2
```
ssteinbach: You might need 
```suggestion
import aaf2 # noqa: E402
```

On these lines so that the linter doesn't complain about the imports not being at the top.

opentimelineio_contrib/adapters/advanced_authoring_format.py
```diff
@@ -31,35 +31,64 @@
 import os
 import sys
 import opentimelineio as otio
+from collections import Iterable
 
 lib_path = os.environ.get("OTIO_AAF_PYTHON_LIB")
 if lib_path and lib_path not in sys.path:
     sys.path += [lib_path]
```
ssteinbach: As noted by @shahbazk8194 earlier today, this should be a prepend instead of an append.

opentimelineio_contrib/adapters/advanced_authoring_format.py
```diff
@@ -31,35 +31,64 @@
...
+    #         for item in clip._walk_item():
+    #             yield item
+    #     else:
+    #         raise NotImplementedError("Sequence returned %s not implemented" %  str(type(segment)))
+    elif isinstance(segment, aaf2.components.EssenceGroup):
+        yield segment
+    elif isinstance(segment, aaf2.components.Filler):
+        yield segment
+    else:
+        raise NotImplementedError("walking %s not implemented" % str(type(segment)))
```
ssteinbach: We're using the python3-style `format` method for string formatting in OTIO.

```suggestion
        raise NotImplementedError("walking {} not implemented".format(type(segment))
```

tox.ini
```diff
@@ -33,6 +28,11 @@ commands =
     coverage run -a --source=opentimelineio_contrib/adapters -m unittest discover opentimelineio_contrib/adapters/tests -vvv
     coverage report --include=* -m
 
+    # Install pyaaf2
```
ssteinbach: This needs to be installed before the coverage suite runs the tests.

jminor:
Something isn't quite right about the pyaaf install. The AAF tests are being skipped - which is why the coverage dropped so much. If you click through to the travis logs, you'll see a bunch of this:
`test_aaf_flatten_tracks (test_aaf_adapter.AAFAdapterTest) ... skipped 'AAF module not found. You might need to set OTIO_AAF_PYTHON_LIB'`

jminor:
Actually, it is just on mention of the old pyaaf in the test itself. See line 89 here:
https://github.com/PixarAnimationStudios/OpenTimelineIO/blob/master/opentimelineio_contrib/adapters/tests/test_aaf_adapter.py#L89

opentimelineio_contrib/adapters/advanced_authoring_format.py
```diff
@@ -30,36 +30,68 @@
 
 import os
 import sys
+from collections import Iterable
 import opentimelineio as otio
 
 lib_path = os.environ.get("OTIO_AAF_PYTHON_LIB")
 if lib_path and lib_path not in sys.path:
-    sys.path += [lib_path]
+    sys.path.insert(0, lib_path)
```
swallitsch: If we're moving to pyaaf2 with pure python, then we should put the pip library in the setup.py, and trust that it will be installed. We should no longer need lines ~35-39 at all.

tox.ini
```diff
@@ -20,19 +20,17 @@ deps =
...
     Pillow
     mock
-; TODO: This will ONLY work on linux systems.  If we could get this from PyPI,
-;       then the platform resolution could be handled by pip instead of here.
-    py27: https://github.com/markreidvfx/pyaaf/releases/download/v1.0.0/PyAAF-1.0.0-cp27-cp27mu-manylinux1_x86_64.whl
-    py35: https://github.com/markreidvfx/pyaaf/releases/download/v1.0.0/PyAAF-1.0.0-cp35-cp35m-manylinux1_x86_64.whl
-    py36: https://github.com/markreidvfx/pyaaf/releases/download/v1.0.0/PyAAF-1.0.0-cp36-cp36m-manylinux1_x86_64.whl
 
 commands =
+    git clone https://github.com/markreidvfx/pyaaf2
```
swallitsch: Unless we're trying to get a specific branch here, we should rely on the new line 26 below to pull pyaaf2 from [pypi](https://pypi.org/project/pyaaf2/). Right now this git clone line will float as new commits come in on master, which is probably not desirable for testing.

opentimelineio_contrib/adapters/tests/test_aaf_adapter.py
```diff
@@ -85,8 +85,8 @@
 try:
     lib_path = os.environ.get("OTIO_AAF_PYTHON_LIB")
     if lib_path and lib_path not in sys.path:
-        sys.path += [lib_path]
-    import aaf # noqa
+        sys.path.insert(0, lib_path)
```
swallitsch: Same comment as above, trust that pyaaf2 has been installed as part of setup.py and we can skip all of this.

PR #397, opened 2018-12-06
Better dependency reporting for `pip` and `setuptools`

ssteinbach:
We depend on relatively modern features inside the pip and setuptools libraries.  Ironically, `pip` and `setuptools` don't provide great tools for catching this and reporting this.  

- Require `pip` version of at least 6.0
- Require `setuptools` version of at least 20.5.0
- Also make the dev dependencies floating (`>=` rather than pinned to a specific version `==`)

Definitely open to comments on this.  Is this a harsh requirement on folks?  VFX Platform doesn't specify a `pip` or `setuptools` version, but it probably should.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -239,7 +239,7 @@ setup(
             'tox>=3.0',
         ],
         'view': [
-            'PySide2>=5.11'
+            'PySide2~=5.11'
         ]
     },
 
--- a/setup.py
+++ b/setup.py
@@ -48,7 +48,7 @@ if (
     print(
         "Your setuptools version is: '{}', OpenTimelineIO requires at least "
         "version '{}'.  Please update setuptools by running: "
-        "pip install -u setuptools".format(
+        "pip install -U setuptools".format(
             SETUPTOOLS_VERSION,
             REQUIRED_SETUPTOOLS_VERSION,
         )
--- a/setup.py
+++ b/setup.py
@@ -41,10 +41,10 @@ if (
         distutils.version.StrictVersion(PIP_VERSION)
         <= distutils.version.StrictVersion(REQUIRED_PIP_VERSION)
 ):
-    print(
+    sys.stderr.write(
         "Your pip version is: '{}', OpenTimelineIO requires at least "
-        "version '{}'.  Please update setuptools by running: "
-        "pip install -U pip".format(
+        "version '{}'.  Please update pip by running:\n"
+        "pip install -U pip\n".format(
             PIP_VERSION,
             REQUIRED_PIP_VERSION,
         )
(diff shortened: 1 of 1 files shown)


davidbaraff:
Do we have a minimum cmake version?  Could it be bumped up somewhat (into the 3’s?)

setup.py
```diff
@@ -30,6 +30,49 @@
...
+):
+    print(
+        "Your setuptools version is: '{}', OpenTimelineIO requires at least "
+        "version '{}'.  Please update setuptools by running: "
+        "pip install -U setuptools".format(
+            SETUPTOOLS_VERSION,
+            REQUIRED_SETUPTOOLS_VERSION,
+        )
+    )
+    sys.exit(1)
```
reinecke: Could the pip and setuptools version requirements be designated by `install_requires`? What's the reason for special handling these two?

setup.py
```diff
@@ -30,6 +30,49 @@
...
+        <= distutils.version.StrictVersion(REQUIRED_PIP_VERSION)
+):
+    print(
+        "Your pip version is: '{}', OpenTimelineIO requires at least "
+        "version '{}'.  Please update setuptools by running: "
+        "pip install -U pip".format(
+            PIP_VERSION,
+            REQUIRED_PIP_VERSION,
+        )
+    )
```
reinecke: If it still works nicely, using `logging.error` instead of `print` or specify `file=sys.stderr` might be good for this. That way if this is running in build toolchains it will land on standard error and be more likely to escalate.

setup.py
```diff
@@ -191,18 +234,20 @@ def test_otio():
...
-            'flake8==3.5',
-            'coverage==4.5',
-            'tox==3.0',
+            'flake8>=3.5',
+            'coverage>=4.5',
+            'tox>=3.0',
         ],
         'view': [
-            'PySide2==5.11'
+            'PySide2>=5.11'
```
reinecke: My inclination for floating versions is to use the [compatible release](https://www.python.org/dev/peps/pep-0440/#compatible-release) version specifier (`~=`). This way we can opt into minor and patch release without opting in to majors that could have breaking changes.

reinecke: Theoretically, in semantic versioning, if a major version is upped, it would be a breaking change. So if dependency developers are observing that rule a major version bump means work needs to be done for that dependency to keep working.
Whether or not that's observed is another question. And I would agree, for the things in dev my inclination is to get those to break as soon as possible.

So maybe I want to alter my recommendation. For `view` `~=` might make sense, for `dev` `>=` pushes us to stay up to date, and that's a good thing.

setup.py
reinecke: Ah, thought it may be something like that. In that case this makes total sense!

setup.py
ssteinbach: Ok fair enough.  I'll switch view to be ~=.  I think that would be more in line with the VFX platform.

PR #399, opened 2018-12-13
Moving already parented item is an error

ssteinbach:
If you have an item whose `.parent()` is not `None`, it is now a ValueError to add it to another composition.  You need to remove it from the other composition before adding it to the new one.  This is to anticipate an implementation detail of the C++ API.

### DIFF
--- a/opentimelineio/core/composition.py
+++ b/opentimelineio/core/composition.py
@@ -371,7 +371,7 @@ class Composition(item.Item, collections.MutableSequence):
             result_range.start_time += parent_range.start_time
             current = parent
 
-        if not self.source_range or not result_range:
+        if not self.source_range:
             return result_range
 
         new_start_time = max(
--- a/opentimelineio/core/composition.py
+++ b/opentimelineio/core/composition.py
@@ -88,17 +88,6 @@ class Composition(item.Item, collections.MutableSequence):
         "Items contained by this composition."
     )
 
-    def __del__(self):
-        # remove parent pointer on all the children
-        for c in self._children:
-            c._set_parent(None)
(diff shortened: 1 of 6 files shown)


opentimelineio/core/composition.py
```diff
@@ -371,7 +371,7 @@ def trimmed_range_of_child(self, child, reference_space=None):
             result_range.start_time += parent_range.start_time
             current = parent
 
-        if not self.source_range:
+        if not self.source_range or not result_range:
```
ssteinbach: whoops, that bug fix wasn't meant to go into this one.  I can split it out if you want @jminor

jminor: Yeah, if you could split it out that would keep this clean. Thanks.

tests/test_composition.py
```diff
@@ -647,13 +693,21 @@ def test_instancing(self):
...
+        self.assertIsNone(it.parent())
+
     def test_range(self):
         length = otio.opentime.RationalTime(5, 1)
         tr = otio.opentime.TimeRange(otio.opentime.RationalTime(), length)
         it = otio.core.Item(source_range=tr)
         sq = otio.schema.Track(children=[it])
         self.assertEqual(sq.range_of_child_at_index(0), tr)
 
+        del sq
```
jminor: This feels a little odd. Maybe the test should use sq, sq2, etc. instead of reusing the same variable?
Or just putting a comment explaining why you need to `del sq` would help someone reading this & wondering why you need to do that.

jminor:
This looks pretty good. I wonder if we want a `Composable.remove_from_parent()` convenience method?

PR #400, opened 2018-12-13
Remove the copy() method and make __copy__ an error.

ssteinbach:
Again, anticipating the C++ implementation, we are removing the shallow copy function.  It doesn't look like anyone in the code base was using it.


tests/test_composition.py
```diff
@@ -893,11 +893,7 @@ def test_range_nested(self):
             ]
         )
 
-        #  Subtle point, but copy() of a Track returns a new Track with a copy
-        #  of all of its metadata, but not of its children.  To get that you
-        #  need to deepcopy().
```
jminor: Yeah, that was pretty odd. The new behavior is clear and simple.

PR #402, opened 2018-12-17
Force RationalTime to use floats.

ssteinbach:
As written today, OTIO will preserve the number type that you pass into `value` and `rate` of `RationalTime`.  In certain places, it will cast to float before doing computations on those numbers, but if you just add and subract you have a good shot of sticking with ints.

This changes proposes forcing those fields to be floats to match what the C++ implementation is doing.  We want to put this to the community to see what people think of this.

This was originally part of #372, but it was pointed out that this was a separate feature so we split it out.


ssteinbach:
Going to address in the next version in the C++ release and potential future changes to the spec for `RationalTime`.

PR #403, opened 2018-12-18
Fix for hook manifest issue

timlehr:
Hi guys,

I had trouble getting the automatic hook calls to work. After digging through the code, I realized that some parts of the manifest class are missing code to keep track of parsed hooks / hook_scripts. I also added a missing hook to the tutorial.


ssteinbach:
Hi @timlehr!  Thanks for catching this.   It looks like the linter is failing in CI - you can check that locally with:
`pip install -e . \[dev]`
and then:
`make lint`

Or you can look at the log in the CI - looks like one of the lines is too long.

But the change looks good!  Good catch!

jminor:
Thanks @timlehr ! In order to accept this, we will need a CLA from you. Here are the details: https://opentimelineio.readthedocs.io/en/latest/tutorials/contributing.html

PR #406, opened 2019-01-08
Use walk() method provided by PyAAF2 classes

jchen9:
Now what Shahbaz’s pull request https://github.com/markreidvfx/pyaaf2/pull/21 was merged to PyAAF2, we could simplify this.

TODO: in PyAAF2, `Sequence` needs to implement `component_at_time` method as suggested in https://github.com/markreidvfx/pyaaf2/pull/21#issuecomment-448089635


ssteinbach:
@markreidvfx Is this in a release yet?  That way we can put a minimum version into the setup.py as a dependency.

PR #410, opened 2019-01-11
Extracting Timecode start and length attributes into available_range

shahbazk8194:


opentimelineio_contrib/adapters/advanced_authoring_format.py
```diff
@@ -268,6 +261,10 @@ def _transcribe(item, parent=None, editRate=24, masterMobs=None):
             masterMob = masterMobs.get(mobID)
             if masterMob:
                 media = otio.schema.MissingReference()
+                media.available_range = otio.opentime.TimeRange(
+                    otio.opentime.RationalTime(timecode_start, editRate),
+                    otio.opentime.RationalTime(timecode_length, editRate)
+                )
```
jminor: If the timecode wasn't found it would be better to leave the available_range as None instead of (0,0). Maybe using a None sentinel instead of (0, 0) in the code earlier would help to distinguish that case. See the discussion on available_range here: https://opentimelineio.readthedocs.io/en/latest/tutorials/time-ranges.html

PR #411, opened 2019-01-15
[WIP] AAF adapter to support writing AAF compositions

shahbazk8194:


shahbazk8194:
sounds good, will do!

PR #412, opened 2019-01-23
GStreamer Editing Services XML adapter

thiblahute:
This implement the basic for a "xges" adapter, the native [GStreamer Editing Services] serialization format.

## Implemented features:

- [x] Clip serialization/de-serialization
- [x] Transitions

## Missing features

- [ ] Stacks inside Tracks, we are missing [nested timelines](https://gitlab.freedesktop.org/gstreamer/gst-editing-services/issues/60) in **GES**
- [ ] [GESGroups](https://gstreamer.freedesktop.org/data/doc/gstreamer/head/gstreamer-editing-services/html/GESGroup.html), basically I can't find any way to represent them in **otio**. Groups are simply a way to represent the fact that several clips are meant to be manipulated together in a timeline.
- [ ] Effects, **otio** doesn't seem to have that feature yet
- [ ] Keyframes, **otio** doesn't seem to have that feature yet

I have already implemented a [GESFormatter](https://gstreamer.freedesktop.org/data/doc/gstreamer/head/gstreamer-editing-services/html/GESFormatter.html) to be able to take advantage of the otio formatter in any project using GES. I am not sure if it belongs to otio itself or GES, what would you prefer?

Also note that in GES each and every timestamp is a GstClockTime (a.k.a `long unsigned int`) representing times in nanoseconds, this lead to big approximation because of the nature of times representation in otio. Note that this "restriction" in GES is being worked on and we plan on introducing a mode where we would work with rational time exclusively (see [this issue](https://gitlab.freedesktop.org/gstreamer/gst-editing-services/issues/61)). 

[GStreamer Editing Services]: https://gstreamer.freedesktop.org/data/doc/gstreamer/head/gstreamer-editing-services/html/ch01.html

### DIFF
--- a/opentimelineio_contrib/adapters/tests/test_xges_adapter.py
+++ b/opentimelineio_contrib/adapters/tests/test_xges_adapter.py
@@ -1,0 +1,75 @@
+#
+# Copyright (C) 2019 Igalia S.L
+#
+# Licensed under the Apache License, Version 2.0 (the "Apache License")
+# with the following modification; you may not use this file except in
+# compliance with the Apache License and the following modification to it:
+# Section 6. Trademarks. is deleted and replaced with:
+#
+# 6. Trademarks. This License does not grant permission to use the trade
+#    names, trademarks, service marks, or product names of the Licensor
+#    and its affiliates, except as required to comply with Section 4(c) of
+#    the License and to reproduce the content of the NOTICE file.
+#
+# You may obtain a copy of the Apache License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the Apache License with the above modification is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied. See the Apache License for the specific
+# language governing permissions and limitations under the Apache License.
+#
+
+import os
+import tempfile
+import unittest
+
+import opentimelineio as otio
+
+SAMPLE_DATA_DIR = os.path.join(os.path.dirname(__file__), "sample_data")
+XGES_EXAMPLE_PATH = os.path.join(SAMPLE_DATA_DIR, "xges_example.xges")
+
+
+class AdaptersXGESTest(unittest.TestCase, otio.test_utils.OTIOAssertions):
+
+    def test_read(self):
+        timeline = otio.adapters.read_from_file(XGES_EXAMPLE_PATH)[0]
+        self.assertIsNotNone(timeline)
+        self.assertEqual(len(timeline.tracks), 6)
+
+        video_tracks = [
+            t for t in timeline.tracks
+            if t.kind == otio.schema.TrackKind.Video
+        ]
+        audio_tracks = [
+            t for t in timeline.tracks
+            if t.kind == otio.schema.TrackKind.Audio
+        ]
+
+        self.assertEqual(len(video_tracks), 3)
+        self.assertEqual(len(audio_tracks), 3)
+
+    def test_roundtrip_disk2mem2disk(self):
+        self.maxDiff = None
+        timeline = otio.adapters.read_from_file(XGES_EXAMPLE_PATH)
+        tmp_path = tempfile.mkstemp(suffix=".xges", text=True)[1]
+
+        otio.adapters.write_to_file(timeline, tmp_path)
+        result = otio.adapters.read_from_file(tmp_path)
+
+        original_json = otio.adapters.write_to_string(timeline, 'otio_json')
+        output_json = otio.adapters.write_to_string(result, 'otio_json')
+        self.assertMultiLineEqual(original_json, output_json)
+
+        self.assertIsOTIOEquivalentTo(timeline, result)
+
+        # But the xml text on disk is not identical because otio has a subset
+        # of features to xges and we drop all the nle specific preferences.
+        with open(XGES_EXAMPLE_PATH, "r") as original_file:
+            with open(tmp_path, "r") as output_file:
+                self.assertNotEqual(original_file.read(), output_file.read())
+
+if __name__ == '__main__':
+    unittest.main()
--- a/opentimelineio_contrib/adapters/xges.py
+++ b/opentimelineio_contrib/adapters/xges.py
@@ -1,0 +1,787 @@
+#
+# Copyright (C) 2019 Igalia S.L
+#
+# Licensed under the Apache License, Version 2.0 (the "Apache License")
+# with the following modification; you may not use this file except in
+# compliance with the Apache License and the following modification to it:
+# Section 6. Trademarks. is deleted and replaced with:
+#
+# 6. Trademarks. This License does not grant permission to use the trade
+#    names, trademarks, service marks, or product names of the Licensor
+#    and its affiliates, except as required to comply with Section 4(c) of
+#    the License and to reproduce the content of the NOTICE file.
+#
+# You may obtain a copy of the Apache License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the Apache License with the above modification is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied. See the Apache License for the specific
+# language governing permissions and limitations under the Apache License.
+#
+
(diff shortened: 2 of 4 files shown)


thiblahute:
Anything missing to get a review and get that merged?

opentimelineio_contrib/adapters/tests/test_xges_adapter.py
```diff
@@ -0,0 +1,68 @@
...
+#
+# You may obtain a copy of the Apache License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the Apache License with the above modification is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied. See the Apache License for the specific
+# language governing permissions and limitations under the Apache License.
```
jminor: This is a slightly different license text than used in the rest of OTIO. Are you able to use the exact same license as in the other files?

opentimelineio_contrib/adapters/xges.py
```diff
@@ -0,0 +1,774 @@
...
+            p = s.index('"')
+            l += p + 1
+            if s[p - 1] != '\\':
+                return l
+            s = s[(p + 1):]
+        return -1
+
+    @staticmethod
+    def escape(s):
+        if not isinstance(s, str):
```
jminor: Should unicode be supported also?

opentimelineio_contrib/adapters/xges.py
```diff
@@ -0,0 +1,774 @@
...
+    CUSTOM = 1 << 4
+
+    @staticmethod
+    def to_otio_type(_type):
+        if _type == GESTrackType.AUDIO:
+            return otio.schema.TrackKind.Audio
+        elif _type == GESTrackType.VIDEO:
+            return otio.schema.TrackKind.Video
+
+        assert(("Can't translate track type %s" % _type) is None)
```
jminor: To match the error reporting pattern used elsewhere in OTIO, we recommend making a subclass of Exception, and using raise instead of assert.

Something like this:
```
class GSTParseError(otio.exceptions.OTIOError):
    pass
```
and then
`raise GSTParseError("Can't translate track type %s" % _type)`

opentimelineio_contrib/adapters/xges.py
```diff
@@ -0,0 +1,774 @@
...
+
+    def _create_otio_transition(self, clip, all_names):
+        start = self.to_rationnal_time(clip.attrib["start"])
+        end = start + self.to_rationnal_time(clip.attrib["duration"])
+        cut_point = otio.opentime.RationalTime((end.value - start.value) / 2, start.rate)
+
+        return otio.schema.Transition(
+            name=self._get_clip_name(clip, all_names),
+            transition_type=TRANSITION_MAP.get(
+                clip.attrib["asset-id"], otio.schema.TransitionTypes.Custom
```
jminor: It looks like GStreamer has a nice set of transition types. I like how you've made a map here so this can be easily extended once OTIO supports more types. Does GStreamer support anything beyond the SMPTE transition types?

opentimelineio_contrib/adapters/xges.py
```diff
@@ -0,0 +1,774 @@
...
+        if rate is None:
+            return
+
+        self.rate = float(Fraction(rate))
+        if self.rate == int(self.rate):
+            self.rate = int(self.rate)
+        else:
+            self.rate = float(round(Decimal(self.rate), 2))
+
+    def to_rationnal_time(self, ns_timestamp):
```
jminor: Misspelled "rational" as "rationnal"

opentimelineio_contrib/adapters/tests/test_xges_adapter.py
thiblahute: Sorry about that, I don't know what happened here :-)

opentimelineio_contrib/adapters/xges.py
thiblahute: Done with the same trick as in the aaf adapter to handle py2 and 3 in the same code path.

opentimelineio_contrib/adapters/xges.py
thiblahute: Fixed.

opentimelineio_contrib/adapters/xges.py
thiblahute: Currently the GES.VideoTransition only supports SMPTE but we could use effects (and the huge database of effects elements from GStreamer) to basically create other types.

opentimelineio_contrib/adapters/xges.py
thiblahute: Fixed.

opentimelineio_contrib/adapters/tests/test_xges_adapter.py
thiblahute: I have a doubt, I am still allowed to use:

```
# Copyright (C) 2019 Igalia S.L
# Copyright 2019 Thibault Saunier <tsaunier@igalia.com>
```

right?

Also I copied the previous version from somewhere in the codebase (not sure where to be honnest), so I guess there is a place where it is not correct :-)

thiblahute:
> Also, a heads up that some time before we get to 1.0 we will likely move the entire contrib area into a separate repo, but you don't need to worry about that until much later.

A bit sad from my point of view, but I understand.

Also, ooc, how do you choose what adapters deserve to be in "core" ?

thiblahute:
I do not get how that patch could lead to those travis failures.

jminor:
Something changed upstream in pyaaf2. One other PR is also affected. We're in-progress on a patch to pin to a specific version of that package which will avoid this. Sorry for the hassle.

opentimelineio_contrib/adapters/tests/test_xges_adapter.py
jminor: Let me check about the copyright...

jminor:
#419 is fixed now. If you rebase onto the latest master, then the build problem will go away.

thiblahute:
> #419 is fixed now. If you rebase onto the latest master, then the build problem will go away.

Thanks! It is done, let's see how it goes.

opentimelineio_contrib/adapters/xges.py
```diff
@@ -0,0 +1,788 @@
...
+TRANSITION_MAP.update(dict([(v, k) for k, v in TRANSITION_MAP.items()]))
+
+
+class GstParseError(otio.exceptions.OTIOError):
+    pass
+
+
+class GstStructure(object):
+    """
+    GstStructure parser with a "dictionnary" like API.
```
aleb:
```suggestion
    GstStructure parser with a "dictionary" like API.
```

opentimelineio_contrib/adapters/xges.py
```diff
@@ -0,0 +1,788 @@
...
+GST_CLOCK_TIME_NONE = 18446744073709551615
+GST_SECOND = 1000000000
+
+
+def to_gstclocktime(rational_time):
+    """
+    This converts a RationalTime object to a GstClockTime
+
+    Args:
+        rational_time (str): This is a RationalTime object
```
aleb:
```suggestion
        rational_time (RationalTime): This is a RationalTime object
```

opentimelineio_contrib/adapters/xges.py
```diff
@@ -0,0 +1,788 @@
...
+
+    Returns:
+        GstClockTime: A time in nanosecond
```
aleb:
```suggestion
        int: A time in nanosecond
```

thiblahute:
I think I addressed all your comments, is anything else missing?

jminor:
Yes, the code all looks good. I'm checking on the copyright question - the person I need to check with is out until 2/15. That's the only thing holding this up. Sorry about that.

If you're willing to use the Pixar copyright, we can accept this today. We make sure to keep the commit attributed to you & to mention you in release notes.

jminor:
Oops, I just heard back. It is fine to have a non-Pixar copyright in the new file(s), but there should be just one per file, and since your CLA is on behalf of Igalia, it should be that one. Does that work for you?

thiblahute:
> Oops, I just heard back. It is fine to have a non-Pixar copyright in the new file(s), but there should be just one per file, and since your CLA is on behalf of Igalia, it should be that one. Does that work for you?

Done just that.

jminor:
Thanks for the contribution to OTIO :)

PR #413, opened 2019-01-28
Avoid NotImplementedError exception for an OperationGroup or Pulldown…

shahbazk8194:


ssteinbach:
@shahbazk8194 one small request - if you have a test file handy, could you add a unit test for this?

ssteinbach:
> @shahbazk8194 one small request - if you have a test file handy, could you add a unit test for this?

Nevermind - I forgot that this whole block will go away once the functions exist inside of PyAAF2.  Thanks!

PR #416, opened 2019-02-01
Adding long to list of flat property types

freesonluxo:


ssteinbach:
@freesonluxo if you rebase onto the latest master, these build errors should go away.  Thanks!

PR #417, opened 2019-02-06
otioconvert fixes output adapter arguments

ssteinbach:
- fixes the track selection in `otioconvert`, which was broken when we switched the parenting rules
- adds support for -A to send arguments to the output adapter as well.

### DIFF
--- a/opentimelineio/console/otioconvert.py
+++ b/opentimelineio/console/otioconvert.py
@@ -101,6 +101,16 @@ def _parsed_args():
         'key=value. Values are strings, numbers or Python literals: True, '
         'False, etc. Can be used multiple times: -a burrito="bar" -a taco=12.'
     )
+    parser.add_argument(
+        '-A',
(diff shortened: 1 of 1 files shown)


opentimelineio/console/otioconvert.py
```diff
@@ -100,6 +101,16 @@ def _parsed_args():
...
     )
+    parser.add_argument(
+        '-A',
+        '--output-adapter-arg',
+        type=str,
+        default=[],
+        action='append',
+        help='Extra arguments to be passed to output adapter in the form of '
+        'key=value. Values are strings, numbers or Python literals: True, '
+        'False, etc. Can be used multiple times: -a burrito="bar" -a taco=12.'
```
jminor: The example should use `-A` instead of `-a`
```suggestion
        'False, etc. Can be used multiple times: -A burrito="bar" -A taco=12.'
```

PR #420, opened 2019-02-07
Guard against `timecode_start` is None

jchen9:
With the change in https://github.com/PixarAnimationStudios/OpenTimelineIO/pull/410 `_extract_timecode_info` could potential return startTime as None. 

It'll trigger error like this:
File "/Users/jchen/scratch/gitlab/transmogrifier2/lib/OpenTimelineIO/opentimelineio_contrib/adapters/advanced_authoring_format.py", line 262, in _transcribe startTime += timecode_start TypeError: unsupported operand type(s) for +=: 'int' and 'NoneType'

I ran into this several days ago. I will upload an example somewhere accessible.

### DIFF
--- a/opentimelineio_contrib/adapters/advanced_authoring_format.py
+++ b/opentimelineio_contrib/adapters/advanced_authoring_format.py
@@ -260,7 +260,9 @@ def _transcribe(item, parent=None, editRate=24, masterMobs=None):
         startTime = int(metadata.get("StartTime", "0"))
         if timecode_info:
             timecode_start, timecode_length = timecode_info
-            startTime += timecode_start
+
+            if timecode_start is not None:
+                startTime += timecode_start
 
         result.source_range = otio.opentime.TimeRange(
             otio.opentime.RationalTime(startTime, editRate),
--- a/opentimelineio_contrib/adapters/advanced_authoring_format.py
+++ b/opentimelineio_contrib/adapters/advanced_authoring_format.py
@@ -179,6 +179,13 @@ def _extract_timecode_info(mob):
         timecode = timecodes[0]
         timecode_start = timecode.getvalue('Start')
         timecode_length = timecode.getvalue('Length')
+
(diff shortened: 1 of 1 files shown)


shahbazk8194:
I remember seeing this as well. I think this is now fixed with https://github.com/PixarAnimationStudios/OpenTimelineIO/commit/f9ae17cda3e93eb4753e957e82801b06f07ed547

The reason was because checking the media_kind of a segment turned out not to be the most reliable way of ensuring we have a Timecode object. So in my experience, I had an AAF file return a Pulldown with its media_kind being "Timecode". We now check if it's an instance of a Timecode object.

Did you see this error with the new check in place?

jchen9:
That's it! Bringing your change fixes the issue.

Now we expect we will always get timecode component.
Maybe now we should raise a different exception in `_extract_timecode_info()` (similar to when `len(timecode) > 1`) if either `timecode_start` or `timecode_length` is None?

shahbazk8194:
Yeah that would be a good idea because I don't think a Timecode object should have its length or start attributes ever be None, it should default to zero. So if we ever get into that case, it should raise an exception

opentimelineio_contrib/adapters/advanced_authoring_format.py
```diff
@@ -179,6 +179,13 @@ def _extract_timecode_info(mob):
         timecode = timecodes[0]
         timecode_start = timecode.getvalue('Start')
         timecode_length = timecode.getvalue('Length')
+
+        if timecode_start is None or timecode_length is None:
+            raise otio.exceptions.NotSupportedError(
+                "Unexpected timecode value(s) in mob name: `{}`."
```
jminor: I was confused by the phrasing of this. Could you say "in mob named" or "in mob with name" so it is clear the timecode isn't in the name itself. Thanks!

jchen9: Done.

PR #428, opened 2019-02-19
AAF Markers

shahbazk8194:


jminor:
It looks like github is confused about the merge. Could you rebase onto the latest master?

opentimelineio_contrib/adapters/advanced_authoring_format.py
```diff
@@ -814,6 +811,31 @@ def _contains_something_valuable(thing):
...
 
+def attach_markers(timeline):
+    '''
+    Search for markers and attach them to their corresponding track
+    '''
+    markers = []
+
+    # Get all markers in the given unsimplified timeline. Markers come in as non
+    # audio or video tracks
+    for child in list(timeline.each_child()):
```
jminor: This would be better as:
`for track in each_child(descended_from_type=otio.schema.Track):`
then you'll only need to check for markers.

opentimelineio_contrib/adapters/advanced_authoring_format.py
```diff
@@ -814,6 +811,31 @@ def _contains_something_valuable(thing):
...
+
+    # Add marker(s) to corresponding video or audio track
+    for child in timeline.each_child():
+        if (isinstance(child, otio.schema.Track) and child.kind in
+                [otio.schema.TrackKind.Audio, otio.schema.TrackKind.Video]):
+            slot_id = child.metadata.get('AAF').get('SlotID')
+            for m in markers:
+                if str(slot_id) in m.metadata.get('AAF').get('DescribedSlots'):
+                    child.markers.append(m)
+
```
jminor: We might want to remove markers from `markers` as you go, and then check for any left-over markers that didn't get assigned at the end to make sure this is doing what we expect.

opentimelineio_contrib/adapters/advanced_authoring_format.py
```diff
@@ -814,6 +811,31 @@ def _contains_something_valuable(thing):
...
+    # Get all markers in the given unsimplified timeline. Markers come in as non
+    # audio or video tracks
+    for child in list(timeline.each_child()):
+        if isinstance(child, otio.schema.Track) and child.markers:
+            markers.extend(child.markers)
+            child.parent().remove(child)
+
+    # Add marker(s) to corresponding video or audio track
+    for child in timeline.each_child():
```
ssteinbach: Same thing here w/ `descended_from_type`

opentimelineio_contrib/adapters/advanced_authoring_format.py
jminor: Also, we should look at how markers work in Media Composer - I believe they act like they are attached to the clips within the track, instead of the track itself. If that's the case, we can push them down to the child of the track at the matching time (and adjust the marker's `marked_range` to match).

opentimelineio_contrib/adapters/advanced_authoring_format.py
```diff
@@ -814,6 +811,31 @@ def _contains_something_valuable(thing):
...
+        if (isinstance(child, otio.schema.Track) and child.kind in
+                [otio.schema.TrackKind.Audio, otio.schema.TrackKind.Video]):
+            slot_id = child.metadata.get('AAF').get('SlotID')
+            for m in markers:
+                if str(slot_id) in m.metadata.get('AAF').get('DescribedSlots'):
```
ssteinbach: Could you build a dictionary mapping the marker described slots to a list of markers?  That way you could iterate through the tracks and just add all the relevant markers?

opentimelineio_contrib/adapters/tests/test_aaf_adapter.py
```diff
@@ -755,6 +759,21 @@ def test_essence_group(self):
...
+        self.assertEqual(type(timeline), otio.schema.Timeline)
+        self.assertEqual(timeline.name, "Just one Marker.Exported.01")
+        self.assertEqual(len(timeline.tracks), 3)
+        self.assertEqual(len(timeline.tracks[0].markers), 1)
+        self.assertEqual(len(timeline.tracks[1].markers), 0)
+        self.assertEqual(len(timeline.tracks[2].markers), 0)
+        marker = timeline.tracks[0].markers[0]
+        self.assertTrue(type(marker), otio.schema.Marker)
+        self.assertEqual(marker.name, 'DescriptiveMarker')
+        self.assertEqual(marker.color, 'RED')
```
jminor: We should verify `marked_range` also.

opentimelineio_contrib/adapters/advanced_authoring_format.py
```diff
@@ -862,6 +861,34 @@ def _contains_something_valuable(thing):
...
+    markers_dict = {}
+
+    # Get all markers in the given unsimplified timeline. Markers come in as non
+    # audio or video tracks
+    for track in timeline.each_child(descended_from_type=otio.schema.Track):
+        if track.markers:
+            for m in track.markers:
+                # Using pop() will remove the value from the set causing
+                # incomplete metadata for DescribedSlots
+                d_slot = next(iter(m.metadata.get('AAF').get('DescribedSlots')))
```
ssteinbach: if `metadata.get('AAF').get('DescribedSlots')` is `None` this will raise an exception.  Does this always get populated by the adapter?

opentimelineio_contrib/adapters/advanced_authoring_format.py
```diff
@@ -862,6 +861,34 @@ def _contains_something_valuable(thing):
...
+def attach_markers(timeline):
+    '''
+    Search for markers and attach them to their corresponding track
+    '''
+    markers_dict = {}
+
+    # Get all markers in the given unsimplified timeline. Markers come in as non
+    # audio or video tracks
+    for track in timeline.each_child(descended_from_type=otio.schema.Track):
+        if track.markers:
```
ssteinbach: If you invert this logic, you can keep to the left:

```
if not track.markers:
  continue
```

opentimelineio_contrib/adapters/advanced_authoring_format.py
shahbazk8194: Yeah every marker should have DescribedSlots in the metadata. This indicates which slot the marker corresponds to. If that isn't there, then I think raising an exception is appropriate. Instead of the NoneType is not iterable exception, I could check for this case and raise another exception with a more meaningful message?

ssteinbach: That sounds good.

PR #430, opened 2019-02-20
Adding Transition PointList metadata for creating AAF Transition objects

shahbazk8194:
Co-authored-by: Freeson Wang <freeson@pixar.com>

### DIFF
--- a/opentimelineio_contrib/adapters/advanced_authoring_format.py
+++ b/opentimelineio_contrib/adapters/advanced_authoring_format.py
@@ -273,7 +273,7 @@ def _transcribe(item, parent=None, editRate=24, masterMobs=None):
                 varying_value = param
                 break
 
-        if varying_value:
+        if varying_value is not None:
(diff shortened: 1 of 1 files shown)


opentimelineio_contrib/adapters/advanced_authoring_format.py
```diff
@@ -263,6 +265,21 @@ def _transcribe(item, parent=None, editRate=24, masterMobs=None):
...
 
+        # Extract value and time attributes of both ControlPoints used for
+        # creating AAF Transition objects
+        varying_value = None
+        for param in item.getvalue('OperationGroup').parameters:
+            if isinstance(param, aaf2.misc.VaryingValue):
+                varying_value = param
+                break
+
+        if varying_value:
```
ssteinbach: You might want to make this an `is None` check in case the truthiness of param ever gets built out in PyAAF2.

```suggestion
        if varying_value is None:
```

PR #434, opened 2019-02-27
Store the REEL variable in the cmx_3600 metadata when not `AX`

mikekoetter:
Stashes the clip_handlers `reel` in the cmx_3600 metadata so it can be recalled when written back out by the cmx_3600 adapter

Also added a test event that `round trips` the sample nucoda edl while maintaining the REEL(SOURCE_ID) field

~BREAKING CHANGE: had to update the nucoda sample edl to match the output format, which includes a new line after Title.  May or may not effect others using that sample~

### DIFF
--- a/opentimelineio/adapters/cmx_3600.py
+++ b/opentimelineio/adapters/cmx_3600.py
@@ -111,7 +111,7 @@ class EDLParser(object):
             clip.metadata['cmx_3600']['reel'] = reel
 
         if comment_handler.unhandled:
-            if clip.metadata.get('cmx_3600', None) is None:
+            if clip.metadata.get('cmx_3600') is None:
                 clip.metadata.setdefault("cmx_3600", {})
             clip.metadata['cmx_3600'].setdefault("comments", [])
             clip.metadata['cmx_3600']['comments'] += (
--- a/opentimelineio/adapters/cmx_3600.py
+++ b/opentimelineio/adapters/cmx_3600.py
@@ -105,10 +105,10 @@ class EDLParser(object):
         clip_handler = ClipHandler(line, comment_handler.handled, rate=rate)
         clip = clip_handler.clip
         reel = clip_handler.reel
-        
+
         # A reel name of `AX` represents an unknown or auxilary source
         # We don't currently track these sources outside of this adapter
-        # So lets skip adding AX reels as metadata for now, 
+        # So lets skip adding AX reels as metadata for now,
         # as that would dirty json outputs with non-relevant infomartion
         if reel != 'AX':
             clip.metadata.setdefault("cmx_3600", {})
--- a/opentimelineio/adapters/cmx_3600.py
+++ b/opentimelineio/adapters/cmx_3600.py
@@ -105,14 +105,17 @@ class EDLParser(object):
         clip_handler = ClipHandler(line, comment_handler.handled, rate=rate)
         clip = clip_handler.clip
         reel = clip_handler.reel
+        
+        # A reel name of `AX` represents an unknown or auxilary source
+        # We don't currently track these sources outside of this adapter
+        # So lets skip adding AX reels as metadata for now, 
+        # as that would dirty json outputs with non-relevant infomartion
         if reel != 'AX':
-            if not clip.metadata.get('cmx_3600'):
-                clip.metadata.setdefault("cmx_3600", {})
+            clip.metadata.setdefault("cmx_3600", {})
             clip.metadata['cmx_3600']['reel'] = reel
 
         if comment_handler.unhandled:
-            if clip.metadata.get('cmx_3600') is None:
-                clip.metadata.setdefault("cmx_3600", {})
+            clip.metadata.setdefault("cmx_3600", {})
             clip.metadata['cmx_3600'].setdefault("comments", [])
             clip.metadata['cmx_3600']['comments'] += (
                 comment_handler.unhandled
--- a/opentimelineio/adapters/cmx_3600.py
+++ b/opentimelineio/adapters/cmx_3600.py
@@ -104,8 +104,15 @@ class EDLParser(object):
         comment_handler = CommentHandler(comments)
         clip_handler = ClipHandler(line, comment_handler.handled, rate=rate)
         clip = clip_handler.clip
+        reel = clip_handler.reel
+        if reel != 'AX':
+            if clip.metadata.get('cmx_3600',None) is None:
+                clip.metadata.setdefault("cmx_3600", {})
+            clip.metadata['cmx_3600']['reel'] = reel
+
         if comment_handler.unhandled:
-            clip.metadata.setdefault("cmx_3600", {})
+            if clip.metadata.get('cmx_3600',None) is None:
+                clip.metadata.setdefault("cmx_3600", {})
             clip.metadata['cmx_3600'].setdefault("comments", [])
             clip.metadata['cmx_3600']['comments'] += (
                 comment_handler.unhandled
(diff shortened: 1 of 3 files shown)


opentimelineio/adapters/cmx_3600.py
```diff
@@ -104,8 +104,15 @@ def add_clip(self, line, comments, rate=24):
         comment_handler = CommentHandler(comments)
         clip_handler = ClipHandler(line, comment_handler.handled, rate=rate)
         clip = clip_handler.clip
+        reel = clip_handler.reel
+        if reel != 'AX':
+            if clip.metadata.get('cmx_3600', None) is None:
```
mikemahony: `dict.get` will default to `None`, so this can be rewritten:
```suggestion
            if clip.metadata.get('cmx_3600') is None:
```
`None` is also falsy so you _could_ write it:
```suggestion
            if not clip.metadata.get('cmx_3600'):
```
...however it looks like there's already `is None` precedent in the file so  ¯\\\_(ツ)\_/¯.

opentimelineio/adapters/cmx_3600.py
```diff
@@ -104,8 +104,15 @@ def add_clip(self, line, comments, rate=24):
...
+                clip.metadata.setdefault("cmx_3600", {})
+            clip.metadata['cmx_3600']['reel'] = reel
+
         if comment_handler.unhandled:
-            clip.metadata.setdefault("cmx_3600", {})
+            if clip.metadata.get('cmx_3600', None) is None:
```
mikemahony:
```suggestion
            if clip.metadata.get('cmx_3600') is None:
```

ssteinbach:
Sorry - just a question about your "breaking change" comment - do you mean you're matching the output format of the CMX EDL adapter or of NuCoda itself?

mikekoetter:
It is updated to match the adapter output, which adds a new line after the TITLE: entry
I'm not sure why the adapter adds a new line in the format but I wasn't looking at that in the scope of this PR, instead I usurped the Nucoda example.
I can revert that change if we like and perhaps pop off the lines we know wont match before doing the string compare

mikekoetter:
This is what an EDL out of our 2018.3 Nucoda looks like
```bash
TITLE: Untitled
FCM: FILM
001  NOTAPE   V     C        00:00:00:01 00:01:10:00 00:00:13:07 00:01:23:06
* FROM CLIP NAME:  
* FROM FILE: Z:\color\tmp\for_nucoda\WIR2__244.0__011.00__364__main.0001.exr
* FROM COMMENT: 
```

mikekoetter:
Ok, I reverted the sample, and updated the test by inlining the sample data.

tests/test_cmx_3600_adapter.py
```diff
@@ -504,6 +504,23 @@ def test_nucoda_edl_write(self):
...
+* FROM FILE: S:\path\to\ZZ100_501.take_1.0001.exr
+002  ZZ100_50 V     C        01:00:06:13 01:00:08:15 00:59:54:18 00:59:56:20
+* FROM CLIP NAME:  take_2
+* FROM FILE: S:\path\to\ZZ100_502A.take_2.0101.exr
+'''
+        
+        timeline = otio.adapters.read_from_string(sample_data, adapter_name="cmx_3600")
+        otio_data = otio.adapters.write_to_string(timeline, adapter_name="cmx_3600", style="nucoda")
+        
+        self.assertMultiLineEqual(sample_data,otio_data)
```
mikemahony: Does `make lint` catch the missing space after the comma here?
```suggestion
        self.assertMultiLineEqual(sample_data, otio_data)
```

opentimelineio/adapters/cmx_3600.py
```diff
@@ -104,8 +104,15 @@ def add_clip(self, line, comments, rate=24):
...
         clip = clip_handler.clip
+        reel = clip_handler.reel
+        if reel != 'AX':
+            if not clip.metadata.get('cmx_3600'):
+                clip.metadata.setdefault("cmx_3600", {})
+            clip.metadata['cmx_3600']['reel'] = reel
+
         if comment_handler.unhandled:
-            clip.metadata.setdefault("cmx_3600", {})
+            if clip.metadata.get('cmx_3600') is None:
```
mikemahony: Now that I'm looking at this closer, `setdefault` won't kill any initial value.  So I _believe_ that this line and line 109 are unnecessary.

opentimelineio/adapters/cmx_3600.py
```diff
@@ -104,8 +104,15 @@ def add_clip(self, line, comments, rate=24):
         comment_handler = CommentHandler(comments)
         clip_handler = ClipHandler(line, comment_handler.handled, rate=rate)
         clip = clip_handler.clip
+        reel = clip_handler.reel
+        if reel != 'AX':
```
mikemahony: Is there any harm in setting `clip.metadata['cmx_3600']['reel']` to `AX` (is this `if`necessary)?

mikemahony:
Good to see you here @mikekoetter! 🎉🎉

opentimelineio/adapters/cmx_3600.py
mikekoetter: I was trying to use a light touch.  Creating the cmx_3600 meta object and setting the AX reel breaks a number of the existing tests.  
I personally don't see any _harm_ in setting it.  But if this is the only piece of cmx metadata and we are using AX as  the equivalent of undefined or unknown I can see the argument to omit adding it

tests/test_cmx_3600_adapter.py
mikekoetter: doh!

opentimelineio/adapters/cmx_3600.py
ssteinbach: Can you add a comment to that effect?  That would help clarify why its being skipped.

opentimelineio/adapters/cmx_3600.py
mikekoetter: We will need to initialize the cmx_3600 dict _somewhere_.
The default in this adapter was to not have that metadata entry unless needed.
I'm adding by default now when we have defined SourceID/Reels, but leaving an exception in to ignore it for AX sources, to preserve some existing assumptions.

I am happy to add it by default for all entries; we will just need to adjust any breaking tests

ssteinbach: I think what Mike was saying was that `setdefault` is the same as saying: `if clip.metadata.get('cmx_3600'): clip.metadata['cmx_3600'] = {}`

ssteinbach: so 109/110 can be replaced with the setdefault call, and 114 and 115 can be replaced the same way.  see: https://docs.python.org/2/library/stdtypes.html#dict.setdefault

mikekoetter: ah, my apologies.  I misread that.
Good catch, I will fix that

mikekoetter:
@ssteinbach Oh, I had added a comment in issue #168 , suggesting that if we stash the reel names in the metadata you could add a hook that replaces clipnames with reel names

opentimelineio/adapters/cmx_3600.py
```diff
@@ -104,6 +104,16 @@ def add_clip(self, line, comments, rate=24):
         comment_handler = CommentHandler(comments)
         clip_handler = ClipHandler(line, comment_handler.handled, rate=rate)
         clip = clip_handler.clip
+        reel = clip_handler.reel
+
+        # A reel name of `AX` represents an unknown or auxilary source
+        # We don't currently track these sources outside of this adapter
+        # So lets skip adding AX reels as metadata for now,
+        # as that would dirty json outputs with non-relevant infomartion
```
mikemahony: ¯\\\_(ツ)\_/¯
```suggestion
        # as that would dirty json outputs with non-relevant information
```

ssteinbach:
Thank You!

PR #450, opened 2019-03-06
Update the adapter.py to pass Argument Maps to Hooks

mikekoetter:
This MR passes through the  argument maps in the adapter.py read/write systems, passing them through to the hooks.

The existing #TODO items suggested that we perhaps should be doing this...
I can't find any reason not to.

An example use-case would be to create a pre-write adapter hook that checks the argument map for `style` being identified as _nucoda_ and then preforms a path replacement on the reference url

```pyhthon
def hook_function(in_timeline,argument_map=None):
    if argument_map.get('style') == 'nucoda':
        for in_clip in in_timeline.each_clip():
            ''' Change the Path to use windows drive letters ( Nucoda is not otherwise forward slash sensative ) '''
            if in_clip.media_reference:
                in_clip.media_reference.target_url = in_clip.media_reference.target_url.replace(r'/linux/media/path','S:')
```

### DIFF
--- a/docs/tutorials/write-a-hookscript.md
+++ b/docs/tutorials/write-a-hookscript.md
@@ -6,12 +6,12 @@ To write a new hook script, you create a python source file that defines a
 a function named ``hook_function`` with signature:
 ``hook_function :: otio.schema.Timeline, Dict => otio.schema.Timeline``
 
-The first argument is the timeline to process, and the second one is a dictionary of arguments that can be passed to it.  Only one hook function can be defined per python file.
+The first argument is the timeline to process, and the second one is a dictionary of arguments that can be passed to it by the adapter or media linker.  Only one hook function can be defined per python file.
 
 For example:
 ```python
 
-def hook_function(tl, arg_dict):
+def hook_function(tl, argument_map=None):
     for cl in tl.each_clip():
         cl.metadata['example_hook_function_was_here'] = True
     return tl
@@ -78,4 +78,19 @@ Now c will run, then b, then a.
 To delete a function the list:
 ```
 del hook_list[1]
+```
+
+## Example Hooks
+
+### Replacing part of a path for drive mapping
+
+An example use-case would be to create a pre-write adapter hook that checks the argument map for style being identified as nucoda and then preforms a path replacement on the reference url
+
+```python
+def hook_function(in_timeline,argument_map=None):
+    if argument_map.get('style') == 'nucoda':
+        for in_clip in in_timeline.each_clip():
+            ''' Change the Path to use windows drive letters ( Nucoda is not otherwise forward slash sensative ) '''
+            if in_clip.media_reference:
+                in_clip.media_reference.target_url = in_clip.media_reference.target_url.replace(r'/linux/media/path','S:')
 ```
--- a/docs/tutorials/write-a-hookscript.md
+++ b/docs/tutorials/write-a-hookscript.md
@@ -88,7 +88,8 @@ An example use-case would be to create a pre-write adapter hook that checks the 
 
 ```python
 def hook_function(in_timeline,argument_map=None):
-    if argument_map.get('style') == 'nucoda':
+    adapter_args = argument_map.get('adapter_arguments')
+    if if adapter_args and adapter_args.get('style') == 'nucoda':
         for in_clip in in_timeline.each_clip():
             ''' Change the Path to use windows drive letters ( Nucoda is not otherwise forward slash sensative ) '''
             if in_clip.media_reference:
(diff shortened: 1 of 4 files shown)


ssteinbach:
@mikekoetter Your example is great - if it isn't too much to ask, do you think you could add an "Example Hooks"  or "Example Use Case" section to this doc:
https://opentimelineio.readthedocs.io/en/latest/tutorials/write-a-hookscript.html

(docs/tutorials/write-a-hookscript.md) at the bottom and include that snippet?

ssteinbach:
I think we had originally envisioned having a separate set of arguments for hook functions, but its clever to have them receive the adapter arguments as well.  My only concern is that if you pass an argument that isn't defined by either the adapter or the hook function you'll get a `TypeError`, right?

mikekoetter:
That is a concern I had as well, however I think that **is** what happens now in the adapter functions.

If we try and pass `style='nucoca'` to anything other than the cmx_3600 adapter we will get a TypeError
```python
>>> otio.adapters.write_to_string(mytimeline,style='nucoda')
  File "/usr/lib/python2.7/site-packages/OpenTimelineIO-0.10.0.dev1-py2.7.egg/opentimelineio/adapters/__init__.py", line 212, in write_to_string
    **adapter_argument_map
  File "/usr/lib/python2.7/site-packages/OpenTimelineIO-0.10.0.dev1-py2.7.egg/opentimelineio/adapters/adapter.py", line 223, in write_to_string
    **adapter_argument_map
  File "/usr/lib/python2.7/site-packages/OpenTimelineIO-0.10.0.dev1-py2.7.egg/opentimelineio/plugins/python_plugin.py", line 128, in _execute_function
    return (getattr(self.module(), func_name)(**kwargs))
TypeError: write_to_string() got an unexpected keyword argument 'style
```
--

For the hooks, it was a little unclear to me from the docs what the argument list should be.
So I followed the calling structure in [hooks.py](https://github.com/PixarAnimationStudios/OpenTimelineIO/blob/master/opentimelineio/hooks.py#L107) and the hook used in the test [example.py](https://github.com/PixarAnimationStudios/OpenTimelineIO/blob/master/tests/baselines/example.py#L56)

```python
def hook_function(in_timeline,argument_map=None):
```

If we are counting on that as being the structure of a hook ( in which case the doc example needs updating ) then I would not expect to see a TypeError in the hooks section.  Or rather it would be incumbent upon a hook's author to catch KeyErrors from the incoming argument_map dict

ssteinbach:
I see, so this works as long as the signature of a hook function is always:
`(in_timeline, argument_map)`

I've been chewing on this and I think that a slightly deeper (potentially breaking) fix might be better.

The adapters take arguments expanded out, but the media linkers take them packed into a dictionary.  I feel like we should probably make that consistent... 
Adapters:
https://github.com/PixarAnimationStudios/OpenTimelineIO/blob/e164549fe1951c43aa0b836fd134ef9b807845d8/opentimelineio/adapters/adapter.py#L193
Media Linkers:
https://github.com/PixarAnimationStudios/OpenTimelineIO/blob/e164549fe1951c43aa0b836fd134ef9b807845d8/opentimelineio/media_linker.py#L148

It looks like the hook functions were modeled after the media linkers, with an argument dictionary instead of expanded arguments.

I think that is a design mistake on our part - it makes more sense to me that arguments would be passed to EVERYTHING the adapter way - expanded out.  That way in the default case, the interface to the adapter is externally inspectable and python throws meaningful errors.

I actually think the better solve is to go in and fix both hooks and media linkers to take arguments in the adapter way.

Furthermore, I think each plugin (adapter, media linker, hooks) should get their own argument dictionaries.  If you think its valuable context for hook functions to also get the arguments that were passed to adapters, I could totally see also providing a 'context' dictionary as an argument to the hook functions, but I think the best thing would be for us to change hook functions and media linkers to match the way that adapters work.

ssteinbach:
Ok, after thinking about it some more and looking through the code, I actually am backtracking on my previous comment.  Because hook functions typically fire off as adapters run, any arguments would need to be passed to `read_from_file` etc directly, like media linker arguments currently are.  I think the solve is to add a `hook_function_argument_map` to this:
https://github.com/PixarAnimationStudios/OpenTimelineIO/blob/e164549fe1951c43aa0b836fd134ef9b807845d8/opentimelineio/adapters/adapter.py#L181

And pass it through the same way the media linker arguments are being passed through:
https://github.com/PixarAnimationStudios/OpenTimelineIO/blob/e164549fe1951c43aa0b836fd134ef9b807845d8/opentimelineio/adapters/adapter.py#L197

Your change is really close to this, in other words - if you add `hook_function_argument_map` to the function signature of the adapters read/write functions, then pass that into the hook function instead of the adapter argument map I think its what we want.

Sorry about the flip flop on this!  Hopefully this feedback makes sense.

ssteinbach:
Oh and I think it could make sense to pass the adapter arguments in to the hook function in the way you note, but I would just wrap it in the dictionary:
```python
hook_function_argument_map['adapter_arguments'] = adapter_args
hook_function_argument_map['media_linker_argument_map'] = media_linker_argument_map
```

Then we should also note that in the docs.

mikekoetter:
ok, @ssteinbach  I think I follow. 

I took a swing at what I thought you were describing.  let me know how close to the mark I got

PR #451, opened 2019-03-06
Ruler

tdervieux:
This is a pull request about the "ruler feature".
-show frame in "media space" or the frame number of the clip (right click on the ruler to show contextual menu)
-Ctrl-left and Ctrl-right to snap the ruler to the end or beginning an item
-Show previous and next frames when the rule is at the beginning of a clip.
-fix bug where markers in the TimeSlider would disappear when the marker was selected and ctrl was pressed.
Some methods will need to be adapted once the functionalities about Time Coordinate Space will be available.

### DIFF
--- a/opentimelineview/timeline_widget.py
+++ b/opentimelineview/timeline_widget.py
@@ -35,6 +35,7 @@ TIME_MULTIPLIER = 25
 LABEL_MARGIN = 5
 MARKER_SIZE = 10
 RULER_SIZE = 10
+EFFECT_HEIGHT = 0.25 * TRACK_HEIGHT
 
 
 class _BaseItem(QtWidgets.QGraphicsRectItem):
@@ -60,6 +61,7 @@ class _BaseItem(QtWidgets.QGraphicsRectItem):
         self._add_markers()
         self._set_labels()
         self._set_tooltip()
+        self._add_effects()       
 
     def paint(self, *args, **kwargs):
         new_args = [args[0],
@@ -98,6 +100,14 @@ class _BaseItem(QtWidgets.QGraphicsRectItem):
                 ) * TIME_MULTIPLIER
             )
             marker.setParentItem(self)
+
+    def _add_effects(self):
+        if not hasattr(self.item,"effects") :
+            return
+        if not self.item.effects :
+            return
+        effect = Effect(self.item.effects, self.timeline_range, self.rect())
+        effect.setParentItem(self)
 
     def _position_labels(self):
         self.source_in_label.setY(LABEL_MARGIN)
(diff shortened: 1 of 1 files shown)


opentimelineview/timeline_widget.py
```diff
@@ -344,24 +353,262 @@ def counteract_zoom(self, zoom_level=1.0):
...
+                self.setX(2)
+        else:
+            self.hide()
+
+
+class Ruler(QtWidgets.QGraphicsPolygonItem):
+
+    time_space = OrderedDict ([ # @TODO pending on Global Space implementation
+                                #("external_space", "External Space"),
+                                ("media_space", "Media Space"),
```
ssteinbach: @jminor what do you think these labels should be?

opentimelineview/timeline_widget.py
```diff
@@ -344,24 +353,262 @@ def counteract_zoom(self, zoom_level=1.0):
...
+        bounded_data = namedtuple("bounded_data",["f","is_bounded","is_tail","is_head"])
+        return bounded_data(f, is_bounded, is_tail, is_head)
+
+    def update_frame(self):
+        for track, frameNumber_tail, frameNumber_head in self.labels:
+            f_tail = ""
+            f_head = ""
+            highlight_head = False
+            for item in track.childItems():
+                if isinstance(item, ClipItem) or isinstance(item, NestedItem):
```
ssteinbach: if you invert this if statement:
`if not (isinstance...)`
then you can dedent this block

opentimelineview/timeline_widget.py
```diff
@@ -344,24 +353,262 @@ def counteract_zoom(self, zoom_level=1.0):
...
+                    bounded_data = self.map_to_time_space(item)     
+                    # check if ruler is within an item boundary
+                    # in other word, start_frame <= ruler <= end_frame                
+                    if bounded_data.is_bounded :
```
ssteinbach: same here with the indentation

opentimelineview/timeline_widget.py
```diff
@@ -511,11 +772,15 @@ def __init__(self, stack, *args, **kwargs):
...
         self.setStyleSheet('border: 0px;')
         self.scene().selectionChanged.connect(self.parse_selection_change)
-
+    
     def parse_selection_change(self):
-        selection = self.scene().selectedItems()
-        if selection:
-            self.selection_changed.emit(selection[-1].item)
+        selection = self.scene().selectedItems()                
+        if selection :
```
ssteinbach: could invert this if block

opentimelineview/timeline_widget.py
```diff
@@ -795,7 +1081,7 @@ def add_stack(self, stack):
         new_stack.open_stack.connect(self.add_stack)
         new_stack.selection_changed.connect(self.selection_changed)
         self.setCurrentIndex(self.count() - 1)
-        self.currentWidget().frame_all()
+        #self.currentWidget().frame_all()
```
ssteinbach: dead code

opentimelineview/timeline_widget.py
ssteinbach: I talked to josh - we concluded following the naming convention in the time-scopes document makes sense for now at  least.  That would be "Media Space" and "Trimmed Space".  We can adapt them later if we decide to change the names.

PR #454, opened 2019-03-09
AAF writer: Adding support for audio transitions

shahbazk8194:
Creating abstract method _transition_parameters that is called in _TrackTranscriber.aaf_transition to support both video and audio transitions

side note: Using metadata clip length instead of duration() in aaf_sourceclip. We'll fix this later

Co-authored-by: Freeson Wang <freeson@pixar.com>

### DIFF
--- a/opentimelineio_contrib/adapters/aaf_adapter/aaf_writer.py
+++ b/opentimelineio_contrib/adapters/aaf_adapter/aaf_writer.py
@@ -41,6 +41,7 @@ AAF_PARAMETERDEF_AVIDEFFECTID = uuid.UUID(
     "93994bd6-a81d-11d3-a05b-006094eb75cb")
 AAF_PARAMETERDEF_AFX_FG_KEY_OPACITY_U = uuid.UUID(
     "8d56813d-847e-11d5-935a-50f857c10000")
+AAF_PARAMETERDEF_LEVEL = uuid.UUID("e4962320-2267-11d3-8a4c-0050040ef7d2")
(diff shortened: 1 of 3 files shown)


opentimelineio_contrib/adapters/aaf_adapter/aaf_writer.py
```diff
@@ -316,24 +291,24 @@ def aaf_transition(self, otio_transition):
             "OperationGroup"
         ]["Operation"]["Name"]
 
+        # Create OperationDefinition
         op_def = self.aaf_file.create.OperationDef(uuid.UUID(effect_id), op_def_name)
         self.aaf_file.dictionary.register_def(op_def)
         op_def.media_kind = self.media_kind
```
stefanschulze: Can you remind me, when/where is self.media_kind set?

jminor: It is a method overridden to return "picture" or "sound" like this:
https://github.com/PixarAnimationStudios/OpenTimelineIO/blob/2ebba59d872f60b65e434a5fa534674081faadf9/opentimelineio_contrib/adapters/aaf_adapter/aaf_writer.py#L385

shahbazk8194:
@jminor We don't know what the issue is regarding why the duration().value. There is an issue for that. It is issue https://github.com/PixarAnimationStudios/OpenTimelineIO/issues/438

I imagine it would be a small change to the reader. We plan on tackling that along side the work that's happening to reduce metadata coupling throughout the writer

PR #457, opened 2019-03-14
Add Effect Widget.

tdervieux:
1. Add an **Effect** Widget to easily spot clips with effects in the *otioviewer*.
2. Center the *source_name_label* in the track. Before it was sitting at 1/3 of its height and would have partially overlap with the **Effect** widget.
3. Display the latest *inclusive* frame rather than the latest *exclusive* frame. Now the last frame number on a clip matches the frame number from the *ruler*.

### DIFF
--- a/opentimelineview/timeline_widget.py
+++ b/opentimelineview/timeline_widget.py
@@ -107,7 +107,7 @@ class _BaseItem(QtWidgets.QGraphicsRectItem):
             return
         if not self.item.effects:
             return
-        effect = Effect(self.item.effects, self.rect())
+        effect = EffectItem(self.item.effects, self.rect())
         effect.setParentItem(self)
 
     def _position_labels(self):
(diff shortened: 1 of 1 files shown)


opentimelineview/timeline_widget.py
```diff
@@ -205,6 +214,67 @@ def __init__(self, *args, **kwargs):
         self.source_name_label.setText('GAP')
 
 
+class Effect(QtWidgets.QGraphicsRectItem):
```
ssteinbach: what if this was called `EffectsBadge` or `EffectsWidget` or something else like that to disambiguate it from `otio...Effect`

opentimelineview/timeline_widget.py
```diff
@@ -205,6 +214,67 @@ def __init__(self, *args, **kwargs):
...
+        pen.setColor(QtGui.QColor(0, 0, 0, 80))
+        pen.setWidth(0)
+        self.setPen(pen)
+
+    def _set_tooltip(self):
+        tool_tips = list()
+        for effect in self.item:
+            name = effect.name if effect.name else ""
+            effect_name = effect.effect_name if effect.effect_name else ""
+            tool_tips.append("%s %s" % (name, effect_name))
```
ssteinbach: We're trying to use the newer (and python3 friendlier) `.format` style of string formatting in python:

```python
"{} {}".format(name, effect_name)
```

opentimelineview/timeline_widget.py
tdervieux: what about EffecfItem ? The other pysideWidget are like ClipItem, GapItem, TransitionItem , NestedItem.

ssteinbach: Sounds great to me!

ssteinbach:
Thanks @tdervieux!

PR #459, opened 2019-03-15
[WIP] Time scopes for translating time up and down the hierarchy

ssteinbach:
This PR adds a substantial amount of functionality aimed at allowing robust queries up and down the hierarchy of the nature "given a frame in the top level timeline, what frame number of media is playing".

Documentation:
https://github.com/ssteinbach/OpenTimelineIO/blob/impl_time_scopes/docs/tutorials/time-scopes.md

- Adds a `TimeTransform` class, a one dimensional homogenous coordinates transformation matrix allowing scale and offset
- Adds `CoordinateSpaceReference` class which allows programmers to refer to a specific coordinate space on an object (for example, `some_clip.trimmed_space()` or `some_timeline.global_space()`
- Adds methods on `Item` and `Timeline` that return `CoordinateSpaceReferences` as outlined by the document noted above
- Adds `otio.algorithms.transform_time()` function which transforms time from one coordinate space reference to another
- Refactors the base `Effect` and `TimeEffect` classes into `core` from `schema`
- Adds functions to `LinearTimeWarp` that returns a transformation matrix for that class.
- Adds unit tests for the above

*DOES NOT IMPLEMENT*:
- This does not support trims because of the way that transitions are currently implemented.  Going to start a separate thread on that topic.

TODO:
- media/occupied range


ssteinbach:
Having spent a substantial amount of time rethinking the underlying principles behind this PR, we're going to close this and re-frame the work based on insights we've had since doing this initial proposal.

Turns out time is difficult!  Thanks for sticking with us!

PR #462, opened 2019-03-20
Add VIDEO_FORMAT header when writing to string/file from ALE adpater

mikekoetter:
So, the Media Composer seems to be a bit picky about ALE headers and it will complain if we don't set a `VIDEO_FORMAT`.   This MR looks to do that.

```
The first thing it does is add a `video_format` kwarg to the adapter signature, allowing you to set the format on output.

If we don't set that variable it will try to preserve the existing header column (if set).
This would likely only be set if the time-line was read in from a previous ALE

When the header column is not found, it searches each clips ALE metadata for an "Image Size" setting to infer a project format. 

If none of that is found it defaults to the adapters DEFAULT_VIDEO_FORMAT (1080)
```

1080p is probably the most popular format being used right now, but I am open to suggestions if anyone has other thoughts on default settings


opentimelineio_contrib/adapters/ale.py
```diff
@@ -109,6 +111,20 @@ def _parse_data_line(line, columns, fps):
         ))
 
 
+def get_video_format(width, height):
```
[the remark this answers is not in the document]
mikekoetter: I'm not sure what 'top level' means in an adapter file, but I added a brief comment and made it "private"

opentimelineio_contrib/adapters/ale.py
```diff
@@ -109,6 +111,45 @@ def _parse_data_line(line, columns, fps):
         ))
 
 
+def _video_format(width, height):
+    """Utility function to map a width and hight to an Avid Project Format"""
+
+    format_map = {
```
ssteinbach: I meant that this map could be moved outside the function as a module level constant -- `AVID_PROJECT_HEIGHT_TO_WIDTH_MAP = {...}` or something.  We're trying to call out these kinds of constant definitions to make them easier to find in the future.

mikekoetter: ok, I _think_ I understand the request, however I don't think that dict would be very useful unless you are applying the width check for resolutions wider than 1920, which avid (currently) considers CUSTOM

 I can move the _video_format(...) function to the top of the file and make give that a "public" syntax if you think that would be helpful

mikekoetter: @ssteinbach - I moved and renamed it.  please let me know if that is more what you are thinking

ssteinbach: Yup, looks great.  Thanks!

PR #463, opened 2019-03-22
AAF Writer: Making tests more robust

freesonluxo:
Refactoring the aaf writer tests into more reusable functions. Now also with duration() value testing.

Co-authored-by: Shahbaz Khan <shahbazkhan8194@gmail.com>


ssteinbach:
Already folded into #460

Issue #464, opened 2019-03-25
AAF reader hard coded to 24fps?


andrewmoore-nz:
The AAF adapter appears to currently be hard coded to read everything at 24fps. 

Line 184 in `/opentimelineio_contrib/adapters/advanced_authoring_format.py`:

`def _transcribe(item, parent=None, editRate=24, masterMobs=None):`

When this function is called from within `read_from_file()` on lines 877 & 883, there is no editRate value passed in as an argument, so it defaults to 24. 

The AAF I am testing this on is from a 25fps project in Premiere, and I can see from reading the raw OTIO JSON string that this information is coming through in the metadata:

> "metadata": {
>     "AAF": {
>         "ClassName": "TimelineMobSlot", 
>         **"EditRate": "25",** 
>         "MediaKind": "Picture", 
>         "Name": "TimelineMobSlot", 
>         "Origin": 0, 
>         "PhysicalTrackNumber": 1, 

...

Is there a way that the correct frame rate can be deduced from the AAF metadata when parsing to create the OTIO model? We have currently hacked our own version of this to allowing for calling `read_from_file()` with a specified editRate value, but not really a solution to the problem.

thecargocultnz:
I would add that if and when a default is needed for framerate, it should be some nonsense value like -1.

It's better that it completely blows up than return a seemingly complete sequence which is actually full of unseen issues.

jminor:
I'll take a look.

jminor:
Oops. Yeah, I can see that this is totally busted for non-24 fps AAFs. I see how to fix this.

andrewmoore-nz:
Great... that'll be super handy if it can pull in the correct editRate from the AAF itself. 

On that note, I've been doing some testing today with AAFs coming out of DaVinci Resolve and they don't seem to be carrying the same metadata through in to the OTIO model i.e. there is no metadata object that comes with the track specifying the frame rate. Might there be somewhere else that this gets specified? Or any idea what might be happening there? Can provide one to look if it helps.

jminor:
My fix #465 is almost ready.

For your other issue: Here is a simple idiom for finding the edit rate of a timeline: `edit_rate = timeline.duration().rate` Give that a try instead of peeking into AAF metadata and see if it works for you. If not, we can make some API to explicitly detect the edit rate.

andrewmoore-nz:
Great... I can pull that fix down to test with as soon as it's available.

[...]

PR #465, opened 2019-03-26
Added support for non-24 fps AAFs

jminor:
including sample AAFs at 30 and 29.97fps.


jminor:
Addresses #464

opentimelineio_contrib/adapters/advanced_authoring_format.py
```diff
@@ -190,6 +190,15 @@ def _transcribe(item, parent=None, editRate=24, masterMobs=None):
     metadata["Name"] = _get_name(item)
     metadata["ClassName"] = _get_class_name(item)
 
+    # Some AAF objects (like TimelineMobSlot) have an edit rate
+    # which should be used for all of the object's children.
+    # We will pass it on to any recursive calls to _transcribe()
+    if hasattr(item, "edit_rate"):
+        editRate = float(item.edit_rate)
+        # Only use a float if really needed.
```
freesonluxo: Could there be a comment here on why this is needed? (pyaaf2 borks for some reason on floats). I think I've seen this borking before. Maybe we should talk to Mark about it.

jminor: I'll file a separate issue to track that down.

jminor: Actually, I think I found the problem in the writer and fixed it.

jminor:
@freesonluxo @shahbazk8194 @ssteinbach I think I addressed all of your notes. Also, I think I fixed the writer to support non-integer rates.

PR #466, opened 2019-03-26
Add media linker arguments to commandline tools

ssteinbach:
Adds:
- support for `-M foo=bar` media linker arguments to `otiocat`, `otioconvert`, `otioview`
- refactors code to turn `foo=bar` arguments into a `console_utils.py` module
- cleanup & linting.

### DIFF
--- a/opentimelineio/console/console_utils.py
+++ b/opentimelineio/console/console_utils.py
@@ -23,7 +23,6 @@
 #
 
 import ast
-import sys
 
 """Utilities for OpenTimelineIO commandline modules."""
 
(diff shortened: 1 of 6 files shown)


opentimelineio/console/console_utils.py
```diff
@@ -0,0 +1,54 @@
...
+            except (ValueError, SyntaxError):
+                # Fall back to a simple string
+                parsed_value = val
+            argument_map[key] = parsed_value
+        else:
+            print(
+                "error: {} arguments must be in the form key=value"
+                " got: {}".format(label, pair)
+            )
+            sys.exit(1)
```
jminor: This should throw an exception and leave the sys.exit() behavior to the caller.

opentimelineview/console.py
```diff
@@ -217,27 +234,18 @@ def main():
...
+    )
+    media_linker_argument_map = otio.console.console_utils.arg_list_to_map(
+        args.media_linker_arg,
+        "media linker"
+    )
 
     application = QtWidgets.QApplication(sys.argv)
 
-    window = Main(argument_map, ml)
+    window = Main(read_adapter_arg_map, ml, media_linker_argument_map)
```
jminor: Maybe `ml` could be `media_linker` or `media_linker_name`?

Issue #468, opened 2019-03-27
Clip rate conversion putting entire track out of sync


andrewmoore-nz:
In doing some testing with a small scale but real world project last week, I came across a specific issue with the way frame rates are parsed from XMLs, but also a more fundamental issue with how these small bugs can effect the entire system. 

This project _(an FCP7 XML exported from a 29.97fps Premiere timeline)_ has two clips at the end of the timeline (on separate tracks) that were copied in from another timeline. Due to the way this is represented in the XML, these clips become 'Clip' objects within OTIO with a source_range duration and start_time rate of 23.976.

When it then comes to calculating the master in/out values for each clip on their respective track, a conversion has to happen to get the required 29.97fps duration from the Clip's stored 23.976 duration value, resulting in a rounding error that puts the entire track 1 frame out of sync. For example:

```
"duration": {
    "OTIO_SCHEMA": "RationalTime.1", 
    "rate": 23.976, 
    "value": 48
```
When this value is converted to 29.97fps, resulting in a value of 59.8801797602996404, this is rounded up to **60** in OTIO (or the conversion math is being over-simplified), when in Premiere it is clipped at **59**, which is what is specified in the XML:

```
<masterclipid>masterclip-11</masterclipid>
<name>TCC_VideoCard_End3.png</name>
<enabled>TRUE</enabled>
<duration>1035764</duration>
<rate>
	<timebase>24</timebase>
	<ntsc>TRUE</ntsc>
</rate>
<start>1971</start>
<end>2030</end>
<in>86312</in>
<out>86360</out>
```

This means that every clip that comes after it on the same track will be 1 frame out of sync, due the way OTIO approaches constructing a timeline based on the cumulative duration values of every clip/gap that comes before it. This can be seen very clearly here:

![OTIO_FramerateConversionSlippingWholeTrackError](https://user-images.githubusercontent.com/19657812/55040683-4b467e80-508e-11e9-9a94-17c267fc69d2.png)

This clip has shifted 1 frame in relation to its surrounding tracks, however it has not been altered or effected in Premiere at all, and its XML `<start>` & `<end>` values have not changed. It is simply that a copy of the clip in question has been placed prior to it on the timeline.

[...]

PR #469, opened 2019-03-27
WIP: Fix NTSC support in FCP XML adapter.

jminor:


jminor:
Attempts to address #468

ssteinbach:
Nice catch.

bashesenaxis:
In the xml spec, the **start** and **end** tags behave a bit funny. When there is a transition at the start of the clip, the **start** value will be -1. Similarly, if there is a transition at the end of the clip, the **end** value will be -1. So you also cannot assume to use the ratio between **in** / **out** and **start** / **end** to see if the clip is sped up or slowed down.

Specs here:
https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/FinalCutPro_XML/Elements/Elements.html#//apple_ref/doc/uid/TP30001156-TPXREF105

Hope this helps

jminor:
@bashesenaxis is there an easy way to get the adjacent transitions from inside that function, or do we have to pass them in? It looks like `context_transition_offsets` could be expanded to include the full start and end also?

bashesenaxis:
I don't think an explicit link exists between the clips and the neighbouring transition elements. Also, etree only has api calls to get the children of an element, not the parents.
So I think the best way to get the information that you need, is to expand on the way that `context_transition_offsets` gets passed, as you suggested.

ssteinbach:
This still has WIP in the title, pushing back a release.

reinecke:
Was fixed in #494 instead.

PR #475, opened 2019-03-29
Allow non-ASCII names in AAFs

jminor:
Adds better handling of AAF clips with missing mobs.
Fixed a bug in otioview when viewing an OTIO file with a SerializableCollection at the top.


jminor:
Something odd is going on with this SerializableCollection. The order of the children comes out different on Python 2 vs 3. I made a separate unit test just for that, but it passes fine. It must be something deeper in the AAF adapter, or PyAAF2. I think I'll make a simpler AAF to test this PR with and come back to that as a separate issue.

opentimelineio_contrib/adapters/aaf_adapter/aaf_writer.py
```diff
@@ -217,10 +217,10 @@ def _from_aaf_file(clip):
...
         if os.path.isfile(target_url) and target_url.endswith("aaf"):
-            aaf_file = aaf2.open(clip.media_reference.target_url)
-            mastermobs = list(aaf_file.content.mastermobs())
-            if len(mastermobs) == 1:
-                mob_id = mastermobs[0].mob_id
+            with aaf2.open(clip.media_reference.target_url) as aaf_file:
+                mastermobs = list(aaf_file.content.mastermobs())
+                if len(mastermobs) == 1:
+                    mob_id = mastermobs[0].mob_id
         return mob_id
```
jminor: FYI, switching to this "with ..." syntax avoids holding the file open. I was seeing some warnings about unclosed file handles when running the tests.

jminor:
Oops, I just realized that the new tests I added to the end of the file are in the `SimplifyTests` class. @ssteinbach do you think I should split the `test_aaf_adapter.py` into some separate files? It is getting pretty long.

jminor:
Fixed.
I think this is ready.

PR #477, opened 2019-04-04
Add global_start_time as a serializable_field.

ssteinbach:
- Also add the ability to layer over baselines in tests
- global_start_time has a default value with a rate of 24

### DIFF
--- a/opentimelineio/schema/timeline.py
+++ b/opentimelineio/schema/timeline.py
@@ -68,6 +68,11 @@ class Timeline(core.SerializableObject):
         dict,
         "Metadata dictionary."
     )
+    global_start_time = core.serializable_field(
+        "global_start_time",
+        opentime.RationalTime,
+        doc="Global starting time value and rate of the timeline."
+    )
 
     def __str__(self):
         return 'Timeline("{}", {})'.format(str(self.name), str(self.tracks))
--- a/opentimelineio/schema/timeline.py
+++ b/opentimelineio/schema/timeline.py
@@ -47,8 +47,6 @@ class Timeline(core.SerializableObject):
     ):
         super(Timeline, self).__init__()
         self.name = name
-        if global_start_time is None:
-            global_start_time = opentime.RationalTime(0, 24)
         self.global_start_time = copy.deepcopy(global_start_time)
 
         if tracks is None:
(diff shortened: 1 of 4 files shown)


opentimelineio/schema/timeline.py
```diff
@@ -68,6 +68,11 @@ def __init__(
         dict,
         "Metadata dictionary."
     )
+    global_start_time = core.serializable_field(
+        "global_start_time",
+        opentime.RationalTime,
+        doc="Global starting time value and rate of the timeline."
```
jminor: Hmm.. interesting that this also implies the edit rate of the timeline.

ssteinbach: Yeah.  What do you think? Really it gives the timecode of the first time in the timeline.

ssteinbach: ...but that implies the rate

jminor: I don't mind that so much in principle, but given that there's a default rate of 24 I think it will be a problem in practice.

jminor: Is there a clean way to remove the default rate?

ssteinbach:
The default value of the global_time_offset can be `None`.  That would imply a value of 0 with a rate of… something?

davidbaraff:
I did just notice the other day that global_start_time is not being serialized, and wondered if that was intentional or not.
are we going to start actually serializing it?  if so, i’ll adjust the C++ implementation to follow.

jminor:
If we can change the default to be None that would be better. If anyone is doing arithmetic with it, they'll have to check for None first though.

Currently the fcp_xml and maya adapters put real values in there, and I'm about to update the AAF adapter to do the same.

We might have to bump the schema version to Timeline.2. Lets discuss more before we commit to this change.

ssteinbach:
Either way I would argue this change should land in beta 11 and not beta 10.  Thoughts?

opentimelineio/schema/timeline.py
ssteinbach: I'm not sure that there is a clean way to remove the default.  `RationalTime` objects are immutable, so it can be `None` or not set AFAIK.  we should discuss further.

ssteinbach:
Addresses #480

ssteinbach:
We are going to land this in beta 10 after all.  This change should be a sooner-not-later change.

PR #486, opened 2019-04-09
C++ implementation of OTIO core

davidbaraff:
Port of the OpenTimelineIO core into C++, wrapped with pybind11 python bindings.  Only port of the core, all adapters and unit tests still run in python.

Also includes _very_ preliminary swift bindings.


meshula:
This is very exciting, I'm looking forward to playing with it.

My preliminary comments are that

1.  I think it would make sense to start with cmake 3 (3.5 is a safe bet)
2. The AnyVector and AnyMap are subclassed, with using, from std containers. This will be a road block when it comes time to make this library available as a shared library for the msvc ABI, because the std containers are not decorated for shared visibility, and the ABI has rules about decorations on bases when subclasses are exported. A simple fix is to use composition instead of inheritance. That doesn't change the implementation much, as inlines are no more expensive than using hoists.
3. Again, very excited to see this!

davidbaraff:
> 
>  A simple fix is to use composition instead of inheritance. That doesn't change the implementation much, as inlines are no more expensive than using hoists.
It’s just a lot more typing, because you have to respecify all the arguments and parameters and then make the call as an inline.  If anything is ever updated about the signatures, you’d have to make changes, while the “using” just says, yeah, use that function, whatever the heck it’s signature was.

But if we have to, we have to.

meshula:
Subclassing Iex::BaseExc from std::string caused crashes for years on Windows OpenEXR, we only just finally were able to merge the fixes, so it's kind of a once bitten twice shy sort of thing.

meshula:
Thinking.... There might be a pattern that gives you what you need without a mess of forwarding. Maybe don't change anything until I have a play with it?

davidbaraff:
Take your time.  there’s no rush.

ssteinbach:
@meshula I'm going to land this into the Cxx branch and then open a pull request from the cxx branch to master to leave open for a while, if you have any comments for this branch you can post them there or contact us directly!

Issue #476, opened 2019-04-02
OpenTime: Non-Drop-frame Timecode not supported at some rates


reinecke:
Given the ground truth values for drop-frame timecode, non-drop timecode, and frame number at 29.97fps:
```
DF_TC = "01;00;02;05"
NDF_TC = "00:59:58:17"
frames = 107957
```

Converting the time corresponding to the frame count gives the correct Drop-Frame value:
```
>>> opentime.to_timecode(opentime.RationalTime(frames, 29.97))
'01:00:02;05'
```

Converting that same time, but forcing the rate arg to 30 give the correct non-drop value:
```
>>> opentime.to_timecode(opentime.RationalTime(frames, 29.97), 30)
'00:59:58:17'
```

Converting from Drop-Frame timecode gives a value that matches frame count:
```
>>> opentime.from_timecode(DF_TC, 29.97)
otio.opentime.RationalTime(value=107957, rate=29.97)
```

Converting from non-drop does not:
```
>>> opentime.from_timecode(NDF_TC, 29.97)
otio.opentime.RationalTime(value=107849, rate=29.97)
```

If you attempt to use 30fps with the Non drop-frame to force correct value and override rate, you get an error:
```
>>> t = opentime.from_timecode(NDF_TC, 30)
>>> t
otio.opentime.RationalTime(value=107957, rate=30)
>>> t.rate = 29.97
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/ereinecke/Projects/github/OpenTimelineIO/opentimelineio/opentime.py", line 72, in __setattr__
    raise AttributeError("RationalTime is Immutable.")
AttributeError: RationalTime is Immutable.
```

Here is an example from Premiere of valid drop-frame, non drop-frame, and frame count values from a 29.97fps sequence:
![Premiere TC examples](https://user-images.githubusercontent.com/1649087/55438361-35393200-5556-11e9-9fc1-e4b47344b134.jpg)

ssteinbach:
@reinecke @apetrynet this has been landed, can we consider this issue 'closed'?

apetrynet:
I would assume so.

ssteinbach:
Thanks!  I've confirmed it.

PR #488, opened 2019-04-12
Suggested fix for #476

apetrynet:
Hey!
This PR should produce the result @reinecke is after in #476 

When passing an NDF timecode with a DF frame Rate to `opentime.from_timecode` calculation respects the NDF TC and does not remove frames.

### DIFF
--- a/opentimelineio/opentime.py
+++ b/opentimelineio/opentime.py
@@ -623,7 +623,7 @@ def from_timecode(timecode_str, rate):
                 timecode_str, nominal_fps - 1))
 
     dropframes = 0
-    if rate_is_dropframe and treat_as_df:
+    if treat_as_df:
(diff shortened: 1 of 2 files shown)


opentimelineio/opentime.py
```diff
@@ -620,7 +623,7 @@ def from_timecode(timecode_str, rate):
                 timecode_str, nominal_fps - 1))
 
     dropframes = 0
-    if rate_is_dropframe:
+    if rate_is_dropframe and treat_as_df:
```
ssteinbach: `rate_is_dropframe` is already factored into `treat_as_df` in line 599, I think this can just be:
```suggestion
    if treat_as_df:
```

apetrynet: Agree :)

Issue #490, opened 2019-04-15
CXX branch build failure in python3


ssteinbach:

ssteinbach:
Fixed as of #492.  Still seeing occasional failures in python3.5, but we're deprecating support for that anyway.

PR #492, opened 2019-04-15
Python3 contrib area unit test cleanups.

ssteinbach:
Ideally this gets contrib area tests passing for the CXX branch.  We shouldn't need to cherry pick this change to master.


ssteinbach:
Addresses #490

PR #493, opened 2019-04-16
AAF Writer: Fixes the crash seen when running otio -> aaf,  in some AAF files due to nesting

freesonluxo:
This makes sure that all nested groups are in stacks, then proceeds to do the nesting correctly. Roughly

otio.schema.Stack = AAF Operation Group
otio.schema.Track = AAF Sequence


opentimelineio_contrib/adapters/aaf_adapter/aaf_writer.py
```diff
@@ -75,6 +75,7 @@ def __init__(self, input_otio, aaf_file, **kwargs):
         self._unique_mastermobs = {}
         self._unique_tapemobs = {}
         self._clip_mob_ids_map = _gather_clip_mob_ids(input_otio, **kwargs)
+        self._stackify_nested_groups(input_otio)
```
jminor: You should make a copy of input_otio before modifying it. Someone calling `write_to_file` would be surprised if the timeline they passed in got modified.

opentimelineio_contrib/adapters/aaf_adapter/aaf_writer.py
```diff
@@ -116,6 +117,17 @@ def _unique_tapemob(self, otio_clip):
             self._unique_tapemobs[mob_id] = tapemob
         return tapemob
 
+    def _stackify_nested_groups(self, timeline):
+        for track in timeline.tracks:
```
jminor: A nice comment here explaining why this is needed would be helpful.

opentimelineio_contrib/adapters/aaf_adapter/aaf_writer.py
```diff
@@ -291,20 +303,24 @@ def transcribe(self, otio_child):
...
-
             sequence.length = length
+            return sequence
+        elif isinstance(otio_child, otio.schema.Stack):
+            # TODO: Move this to function
+            operation_group = self.nesting_operation_group()
+            length = 0
+            for nested_otio_child in otio_child:
+                result = self.transcribe(nested_otio_child)
+                length = length + result.length
```
jminor:
```suggestion
                length += result.length
```

jminor:
How does the AAF NestedScope object relate to this? I would have expected to see it in there somewhere.

Issue #265, opened 2018-06-01
fcp_xml adapter doesn't seem to export offline/unreferenced clips correctly


apetrynet:
When exporting an XML with offline/unreferenced clips, OTIO should provide all tags needed to recreate those clips in an application even though some tags remain "empty". Currently the `_build_clip_item_without_media` function doesn't include the `<name>`, `<duration>`, `<pathurl>` and `<timecode>` tags in the `<clipitem> -> <file>` tag which results in the clip not getting created.

I like the explicitness of a call to `_build_clip_item_without_media`, but is it really necessary to differentiate the creation of an "online" or "offline" clip other than the `<pathurl>` being empty?  All other data should be in place if it's a clip placed in a timeline should it not?

Please note that I only tested import of such an XML in Hiero.

ssteinbach:
@bashesenaxis Do you have any thoughts on this?

bashesenaxis:
The idea behind `_build_clip_item_without_media` is that it deals with title objects. So they would never have any media.
For example:
https://github.com/PixarAnimationStudios/OpenTimelineIO/blob/master/tests/sample_data/premiere_example.xml#L688
In this case, `<clipitem> -> <file> -> <name>` is identical to `<clipitem> -> <name>`, the `<duration>` wouldn't have any meaningful information and there would not be a `<parthurl>` associated with it either.

I am not entirely sure whether this is a premiere specific way of doing things.

apetrynet:
Closing this one as it's really old and I haven't had use for this since the hiero plugin came to life.
Feel free to reopen if someone finds it useful.

Issue #485, opened 2019-04-08
XML writer not writing correct Timeline/Sequence names


wrosecrans:
I am trying to write an FCP7 XML file with multiple sequences, but all the sequences get written with the name "tracks."

Here is a simple script to demonstrate:

```
import opentimelineio

names = ["Episode One", "Dubious Sequel", "Franchise Sellout"]
def write_xml(timelines):
    otio_timelines = opentimelineio.schema.serializable_collection.SerializableCollection()
    for t in timelines:
        tl = opentimelineio.schema.timeline.Timeline(name=t)
        print(tl.name)
        otio_timelines.append(tl)
    opentimelineio.adapters.write_to_file(otio_timelines, "result.xml")
    
write_xml(names)

```
Here is a snippet of the result XML, where I would expect the name tag to be "Episode 1" rather than "tracks"
`            <sequence id="sequence-1">
                <name>tracks</name>`

And here is the full output XML file:  https://gist.github.com/wrosecrans/6746479675cf4dbec34699b2f7d603d3
I am expecting that the names would wind up in the XML.  Since the Sequences have the same name, Premiere will only be able to load one of them from the XML, which isn't very useful.  Is there some other way to set the timeline name that I should be using in the itio API, or is this just a limitation of the XML writer?

Issue #334, opened 2018-09-28
odd clip rate from xml adaptor


alatdneg:
Is this right ? As it's breaking other code in otio as it's an invalid time rate ?

23.976023976

Should it not round it to 23.976?

This is triggered by this in the xml

<rate>
<timebase>24</timebase>
<ntsc>TRUE</ntsc>
</rate>

alatdneg:
ValueError: 23.976023976 is not a valid frame rate, Please use one of these: (1, 12, 23.976, 23.98, 24, 25, 30, 48, 50, 60, 29.97, 59.94)
Should this validation be smarter ?

reinecke:
The FCP XML adapter may be using sub-optimal parsing.
In the raw XML, 23.976 is specified as:
```
<rate>
    <timebase>24</timebase>
    <ntsc>TRUE</ntsc>
</rate>
```

And the adapter parses it using this function:
```
def _parse_rate(elem, element_map):
    elem = _resolved_backreference(elem, 'all_elements', element_map)

    rate = elem.find('./rate')
    # rate is encoded as a timebase (int) which can be drop-frame
    base = float(rate.find('./timebase').text)
    if rate.find('./ntsc').text == 'TRUE':
        base *= .999
    return base
```

This table from apple lays out some expectations: https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/FinalCutPro_XML/FrameRate/FrameRate.html#//apple_ref/doc/uid/TP30001158-BCIHDFGD

But I think it's imperfect because it's still saying the "actual rate" is a decimal rather than a fraction like 24000/1001. However using this kind of lookup would likely be a bit more internally consistent with the OTIO timecode handling and if the same lookup is used on export it will be symmetric.

thecargocultnz:
Yes, that "0.999" is an approximation, and will definitely cause rounding issues and drift over time.
The only way to correctly calculate NSTC values is using *1000/1001 (or  the inverse).

Issue #287, opened 2018-07-10
FCP xml adapter and invalid "out" values (premiere)


alatdneg:
Hi,
We've seen some clip duration issues, looks like the start-end / in-out values can be exported oddly from Premiere.

I'm guessing that most apps ignore the "out" attribute and just add the duration from the start-end properties.

In this example the in/out duration is a frame short, has anyone come across this before ?
 
```
<masterclipid>masterclip-26</masterclipid>
<duration>51</duration>
<rate>
<timebase>24</timebase>
<ntsc>FALSE</ntsc>
</rate>
<start>1574</start>
<end>1618</end>
<in>7</in>
<out>50</out>
```

ssteinbach:
@alatdneg if you have a second, can you turn this into a unit test case we can add to the test suite?  I think that might help debug it.  @bashesenaxis might know the specifics of what you're asking, though.

alatdneg:
Yeah, I've not had the contributors form signed off yet, not sure how you'd do a unit test for this, I think is a bug in how premiere writes it's FCP7 XML, it seems most apps ignore the <out> values and calculate it from <end>-<start>

bashesenaxis:
Hmm, that does indeed look interesting.
```
duration = 51
out - in = 50 - 7 = 43
end - start = 1618 - 1574 = 44
```
The way that I would interpret that, is as follows
- there are 51 frames available in the media
- we are taking frames 7 to 50 (43 frames)
- and displaying that on the timeline from frame 1574 to 1618 (44 frames)

Therefor, the clip would be stretched in time ever so slightly. Or perhaps you are displaying a 24 fps clip on a 25 fps timeline? @alatdneg, can you confirm whether this is the case?

@ssteinbach, @jminor  I`m not entirely sure if OTIO supports scaled clips at the moment?

alexwidener:
Tagging @bbb999 to try and get some eyes from Premiere on it.

ssteinbach:
@bashesenaxis we have time effects including a linear time warp that we're starting to introduce to support scaled clips.  for an example in the EDL adapter, see:
https://github.com/PixarAnimationStudios/OpenTimelineIO/blob/33ad026c7d46f106eb079f77324ffae6060691d5/opentimelineio/adapters/cmx_3600.py#L150

ssteinbach:
Pinging @alatdneg just to put this back on your radar per our earlier conversation.  See the question from @bbb999.

Thanks!

[...]

PR #494, opened 2019-04-16
Refactor FCP XML adapter - add generator and effect support

reinecke:
This is a refactor of the FCP XML adapter adapter. The goal of this refactor was to:
- Address some timing accuracy issues that could be encountered
- Add support for handling effects and generators
- Add support for parsing and roundtripping FCP XML metadata

This is expected to address the following issues (I would love confirmation from the relevant parties):
- #468 
- #265 
- #334 
- #287 
- #485

### DIFF
--- a/opentimelineio/adapters/fcp_xml.py
+++ b/opentimelineio/adapters/fcp_xml.py
@@ -253,7 +253,6 @@ def _time_from_timecode_element(tc_element, context=None):
     rate = _rate_from_context(local_context)
 
     # Try using the display format and frame number
-    display_format = tc_element.find("./displayformat")
     frame = tc_element.find("./frame")
 
     # Use frame number, if available
@@ -267,16 +266,6 @@ def _time_from_timecode_element(tc_element, context=None):
         raise ValueError("Timecode element missing required elements")
 
     tc_string = tc_string_element.text
-
-    # Determine if it's drop, non-drop, or unspecified
-    is_non_drop = None
-    if display_format is not None:
-        is_non_drop = (display_format.text.lower() == "ndf")
-
-    # There is currently a bug where OTIO does not properly support NDF for
-    # fractional framerates
-    if round(rate) != rate and is_non_drop:
-        raise ValueError("Only drop TC supported: {}".format(tc_string))
 
     return opentime.from_timecode(tc_string, rate)
 
@@ -1344,19 +1333,23 @@ def _build_timecode(time, fps, drop_frame=False, additional_metadata=None):
         tc_element = cElementTree.Element("timecode")
 
     tc_element.append(_build_rate(fps))
+    rate_is_not_ntsc = (tc_element.find('./rate/ntsc').text == "FALSE")
+    if drop_frame and rate_is_not_ntsc:
+        tc_fps = fps * (1000 / 1001.0)
+    else:
+        tc_fps = fps
 
     # Get the time values
-    tc_time = time.rescaled_to(fps)
-    tc_string = opentime.to_timecode(time, fps)
-    frame_number = int(round(tc_time.value))
+    tc_time = opentime.RationalTime(time.value_rescaled_to(fps), tc_fps)
+    tc_string = opentime.to_timecode(tc_time, tc_fps, drop_frame)
 
     _append_new_sub_element(tc_element, "string", text=tc_string)
+
+    frame_number = int(round(time.value))
     _append_new_sub_element(
         tc_element, "frame", text="{:.0f}".format(frame_number)
     )
 
-    # TODO: In the future, this should be driven by drop_frame passed in, not
-    #       inferred from the TC string.
     drop_frame = (";" in tc_string)
     display_format = "DF" if drop_frame else "NDF"
     _append_new_sub_element(tc_element, "displayformat", text=display_format)
--- a/opentimelineio/adapters/fcp_xml.py
+++ b/opentimelineio/adapters/fcp_xml.py
@@ -973,6 +973,7 @@ class FCP7XMLParser:
             "file",
             "marker",
             "effect",
+            "rate",
             "sequence",
         }
         metadata_dict = _xml_tree_to_dict(
@@ -1571,7 +1572,6 @@ def _build_clip_item_without_media(
         )
 
     _append_new_sub_element(clip_item_e, 'name', text=clip_item.name)
-    clip_item_e.append(_build_rate(media_start_time.rate))
     clip_item_e.append(
         _build_empty_file(
             clip_item.media_reference, timeline_range, br_map
--- a/opentimelineio/adapters/fcp_xml.py
+++ b/opentimelineio/adapters/fcp_xml.py
@@ -1838,6 +1838,11 @@ def _build_sequence_for_timeline(timeline, timeline_range, br_map):
         timeline.tracks, sequence_e, timeline_range, br_map
     )
 
+    # In the case of timelines, use the timeline name rather than the stack
+    # name.
+    if timeline.name:
+        sequence_e.find('./name').text = timeline.name
+
     # Add the sequence global start
     if timeline.global_start_time is not None:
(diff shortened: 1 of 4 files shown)


apetrynet:
Hey!
I'm afraid this doesn't resolve #265 

Running `otioconvert xml->xml` on a file with offline clips the resulting xml yields an empty timeline when re-importing in Hiero.
I tried looking at the output file and found some differences, but uncertain what exactly is causing the different behavior. Timecode and In/Out points differ for instance, but manually tweaking these still don't produce a timeline containing clips. I don't know if this is true for other applications.
Here's the [file](https://gist.github.com/apetrynet/6eb4f92fc1dfc0bfb9454f6b0acd5249) I tested with if you want to check it out.

reinecke:
Thanks @apetrynet for taking a look! I'll take a quick pass and see if I can resolve.

reinecke:
@apetrynet I definitely see the start timecode on file elements not roundtripping, I'll take a look a that. In terms of not loading, would you mind trying this file: https://gist.github.com/reinecke/5b4bbb89ab6dc3717f8d3f053708fd58

It's a wild hunch I'd like to validate but I don't have Hiero on my machine to try with :/

apetrynet:
@reinecke I actually tried removing the project tags earlier today. No go I'm afraid. Unfortunately I won't be able to double check your file until Tuesday when I'm back at work.

reinecke:
All I did was try removing the project tags, so I'm guessing that file won't work then :/
I noticed there were cases where Resolve had an issue but this file seemed to work fine there. I'll keep looking.

One thing I notice in the file you provided is that while in and out values are floats, the start and end values are ints. There could be an issue arising there.

reinecke:
@apetrynet If you get a chance, I updated the gist. In fixing the start timecode issue in the file elements for missing references I also discovered duration elements were being omit. With luck, this may help Heiro out a little.

apetrynet:
@reinecke sounds good! I'll check when I get back to work and report back.

apetrynet:
@reinecke Your latest gist works like a charm! Good job!

opentimelineio/adapters/fcp_xml.py
```diff
@@ -562,95 +1375,129 @@ def _build_item_timings(
...
+        file_media_e = _get_or_create_subelement(file_e, "media")
 
-    for kind in content_types:
-        _insert_new_sub_element(file_media_e, kind)
+        audio_exts = {'.wav', '.aac', '.mp3', '.aif', '.aiff', '.m4a'}
+        has_video = (os.path.splitext(url_path)[1].lower() not in audio_exts)
+        if has_video and file_media_e.find("video") is None:
+            _append_new_sub_element(file_media_e, "video")
+
+        # All files have audio?
```
ssteinbach: I don't understand this comment

opentimelineio/adapters/fcp_xml.py
```diff
@@ -562,95 +1375,129 @@ def _build_item_timings(
...
-        content_types.append('video')
-    content_types.append('audio')
+    # TODO: We should use a better method for this. Perhaps pre-walk the
+    #       timeline and find all the track kinds this media is present in?
+    if not file_e.find("media"):
+        file_media_e = _get_or_create_subelement(file_e, "media")
 
-    for kind in content_types:
-        _insert_new_sub_element(file_media_e, kind)
+        audio_exts = {'.wav', '.aac', '.mp3', '.aif', '.aiff', '.m4a'}
```
ssteinbach: Just so I understand, FCP doesn't explicitly mark a track as audio or video?  Its implicit based on what the thing is pointed at?

opentimelineio/adapters/fcp_xml.py
```diff
@@ -24,156 +24,316 @@
...
+        is_non_drop = (display_format.text.lower() == "ndf")
 
-def _insert_new_sub_element(into_parent, tag, attrib=None, text=''):
-    elem = cElementTree.SubElement(into_parent, tag, **attrib or {})
-    elem.text = text
-    return elem
+    # There is currently a bug where OTIO does not properly support NDF for
+    # fractional framerates
+    if round(rate) != rate and is_non_drop:
+        raise ValueError("Only drop TC supported: {}".format(tc_string))
```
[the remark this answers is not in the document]
reinecke: Yes, Between #497 and #488 there were a couple of DF/NDF issues that should have been resolved. I'm working to address locally, however this particular oddity in our test file has be a touch puzzled: https://github.com/PixarAnimationStudios/OpenTimelineIO/blob/929149e97333edb61a18e53de2cb4ac40bc6a10d/tests/sample_data/premiere_example.xml#L720

That is 30fps (not 29.97) but using Drop-frame timecode. opentime does not currently support this, so I'm trying to sort out the desired behavior.

opentimelineio/adapters/fcp_xml.py
wrosecrans: In the Final Cut XML, the structure of the tags for tracks basically look like 
```
<project> <children> <sequence> 
    <media> <video>
            <track/> <track/> ...
    </video>
    <audio>
            <track/> <track/> ...
    </audio> </media>
</sequence> </children> </project>
```

where "sequence" == "a timeline", and a particular track element is in either a "video" or "audio" element, and only has one kind of media one it.  (But one _file_ can wind up referenced in both a video and audio track, linked so you can drag them as a unified thing in the UI of something like Premiere.)  If you use OTIO to populate only the video track with a quicktime file, it shows up as silent.  Likewise, if you put a .wav file in a video track, it won't work right, so needs to be filtered out from video tracks.

All that said, this spot in fcp_xml.py is for making a file, which can potentially exist outside of a track, and sort of has its own semantics, because xmeml wants to rot your brain.

When a video <track> has a <clipitem>, the clipitem has a <file> which has its own <media> element that has to have both <video> and <audio> elements describing the format if the actual file does -- even if the track that contains the element is only a video track.  I have seen Premiere fail to link a Quicktime file if the XML omits the the <audio> tag, even if it is only in a <video> track, because in sees the file has two audio channels, but it thinks the file referenced in the timeline is supposed to have none, and thinks there is a mismatch.

I _think_ that answers your question, but I fear I may have answered your question by making it even less clear.  :)

opentimelineio/adapters/fcp_xml.py
```diff
@@ -24,156 +24,316 @@
...
 
-META_NAMESPACE = 'fcp_xml'  # namespace to use for metadata
+# Regex to match identifiers like clipitem-22
+ID_RE = re.compile(r"^(?P<tag>[a-zA-Z]*)-(?P<id>\d*)$")
 
 
+"""
+Adapter TODOs:
+    - Support start timecode on sequences
+    - Support top-level objects other than sequences: project, bin, clip
```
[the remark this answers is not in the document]
reinecke: Done. see #514

opentimelineio/adapters/fcp_xml.py
reinecke: In this pass I did some refactoring for clarity on the output side, but I was trying to minimally impact some of the structural decisions made. This comment was an attempt to call out that this piece of logic may not be sustainable. Will update with a better comment.

opentimelineio/adapters/fcp_xml.py
reinecke: Also, this whole area of logic feels a bit CBB to me. However, the solve in unclear to me at the moment.

opentimelineio/adapters/fcp_xml.py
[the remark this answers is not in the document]
ssteinbach: It may have crept in and been ignored for a while. @bashesenaxis do you have any insight on this?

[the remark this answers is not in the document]
reinecke: After digging a little deeper, I was able to resolve that in this case the start time is the `frame` over the `rate`, and the timecode is frame over the ntsc version of the rate when `displayformat` is `DF

reinecke:
Updated based on review notes. If it checks out, the PR is ready to merge from my POV.

PR #496, opened 2019-04-17
Use warnings module to handle overlapping transitions case during flatten

jminor:
When flattening tracks or trimming a track, we sometimes encounter a transition that overlaps the edge of the trim. Instead of throwing an exception, we now produce a warning. The caller can choose how to handle the warning via the standard warnings module.

### DIFF
--- a/opentimelineio/algorithms/track_algo.py
+++ b/opentimelineio/algorithms/track_algo.py
@@ -25,6 +25,7 @@
 """Algorithms for track objects."""
 
 import copy
+import warnings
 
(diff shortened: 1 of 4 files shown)


jminor:
(from discussion with @davidbaraff) There's a concern here that the warnings module in Python doesn't have an equivalent in C++, so introducing this here makes the upcoming C++ port more difficult.

jminor:
Closing this as no longer relevant since the core of OTIO is now in C++

PR #497, opened 2019-04-17
Timecode rate enhancements

reinecke:
- Added support for more accurate ntsc rates (24000/1001, 30000/1001, and 60000/1001)
- Ability to explicitly control drop-frame and non-drop-frame timecodes (this allows creation of 29.97 non-drop timecode)


opentimelineio/opentime.py
```diff
@@ -641,12 +650,15 @@ def from_timecode(timecode_str, rate):
...
-def to_timecode(time_obj, rate=None):
+def to_timecode(time_obj, rate=None, drop_frame=None):
     """Convert a RationalTime into a timecode string.
 
     :param time_obj: (:class:`RationalTime`) instance to express as timecode.
     :param rate: (:class:`float`) The frame-rate to calculate timecode in
         terms of. (Default time_obj.rate)
+    :param drop_frame: (:class:`bool`) ``True`` to make drop-frame timecode,
+        ``False`` for non-drop. If left ``None``, a format will be guessed
+        based on rate.
```
jminor: Using None to mean "auto" might be a little confusing. When this is ported to C++ we might need to use an enum. Maybe just defaulting to False would be better?

PR #509, opened 2019-04-24
WIP: Controlled instancing

jminor:
This is a proposal for conversation: What if we were able to strictly control which schema types could be instanced, and which could not?

### DIFF
--- a/src/opentimelineio/mediaReference.cpp
+++ b/src/opentimelineio/mediaReference.cpp
@@ -17,6 +17,10 @@ bool MediaReference::is_missing_reference() const {
     return false;
 }
 
+bool MediaReference::is_instanceable() const {
+    return true;
+}
+
 bool MediaReference::read_from(Reader& reader) {
     return reader.read("available_range", &_available_range) &&
         Parent::read_from(reader);
--- a/src/opentimelineio/serializableObject.cpp
+++ b/src/opentimelineio/serializableObject.cpp
@@ -66,6 +66,10 @@ bool SerializableObject::is_unknown_schema() const {
     return false;
 }
 
+bool SerializableObject::is_instanceable() const {
+    return false;
+}
+
 std::string SerializableObject::to_json_string(ErrorStatus* error_status, int indent) const {
     return serialize_json_to_string(any(Retainer<>(this)), error_status, indent);
 }
--- a/src/opentimelineio/serializableObject.h
+++ b/src/opentimelineio/serializableObject.h
@@ -433,6 +433,8 @@ public:
     virtual void write_to(Writer&) const;
 
     virtual bool is_unknown_schema() const;
+
+    virtual bool is_instanceable() const;
     
     std::string const& schema_name() const {
         return _type_record()->schema_name;
--- a/src/opentimelineio/serialization.cpp
+++ b/src/opentimelineio/serialization.cpp
@@ -556,8 +556,13 @@ void SerializableObject::Writer::write(std::string const& key, SerializableObjec
         /*
          * We've already written this value.
          */
-        _encoder.write_value(SerializableObject::ReferenceId { e->second });
-        return;
+
+        if (value.in_instanceable()) {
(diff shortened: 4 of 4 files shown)


davidbaraff:
I can see that one might choose not to allow instancing of certain types of objects, but might it not depend on where the instancing is?  Specifically, you might want to say a SerializableCollection has to hold distinct objects.  But maybe it's useful to allow metadata, which is unstructured, to be able to point at a particular object?   In which case, controlling instancing based on the type, rather than whom the referer is, is not what you want.   Should metadata continue to be allowed to be the wild wild west?  Offhand, I'd say yeah, that's what metadata is for...

davidbaraff:
Actually, I'd like to see a case of where instancing needs to be held in check, because allowing it creates badness.  Understanding the reasons for needing to control it (if in fact we do) might suggest a targeted path to how to do it.

jminor:
Yeah, I'm not sure if restricting it at the type level is the right place. We'll have to think about that more.

Here are use cases where it probably makes sense:

MediaReferences: Several people have asked for instanced MediaReferences. This makes sense in that all of the clips that point to the same media could share the same MediaReference in the OTIO data structure and you don't have to worry about divergent metadata on each one.

Inputs to video effects: When we eventually add support for video effects that perform compositing, we may want an Effect to point to other Clips or MediaReferences that are used as elements of the composite (e.g. keys, channels, etc.) This might be a case where the source of the reference, not the type pointed to is important. Modeling effects like this is on the roadmap, but we haven't done the research to propose anything specific yet.

Here are cases where it would be problematic:

Composable: For example a Clip that appears in a Track twice, or a single Clip that appears in multiple Tracks. Since a Clip has a reference to its parent, the first case might sort of work (they both have the same parent), but the second does not. If we allowed multiple parents, then it could be made to work, but a variety of common use cases become more difficult to handle generally. For example, making a Clip longer will make the other instances longer - something that many users of OTIO might not expect. It also complicates any questions we ask about what range of time a Clip occupies in a parent Composition since we might have to disambiguate the two (or more) instances.

Instanced RationalTime or TimeRange: They are immutable, so this is relatively safe, but what would this mean and why would you want them instanced?

Instanced Effects or Markers: What would this mean and why would you want them instanced?

The argument for most of these is just "it would be complicated and confusing". Adding the extra burden of "what if this object is used in multiple places?" to many run of the mill use cases would make working with OTIO more complex - something we're trying to avoid unless there is a clear gain.

davidbaraff:
> Composable: For example a Clip that appears in a Track twice, or a single Clip that appears in multiple Tracks.
> 
The composition mechanism currently screens this out, right?  I can’t put the same clip in more than one track because it’s an error at the Composition level already when I go to insert it as a child.

I’m saying that the explicit parent/child mechanism we have for Composition catches many of the cases you want to rule out.  What’s a case not caught by Compsition that also needs to be disallowed?


> Since a Clip has a reference to its parent, the first case might sort of work (they both have the same parent), but the second does not. If we allowed multiple parents, then it could be made to work, but a variety of common use cases become more difficult to handle generally. For example, making a Clip longer will make the other instances longer - something that many users of OTIO might not expect. It also complicates any questions we ask about what range of time a Clip occupies in a parent Composition since we might have to disambiguate the two (or more) instances.
> 
> Instanced RationalTime or TimeRange: They are immutable, so this is relatively safe, but what would this mean and why would you want them instanced?
> 
> 

FYI, instancing is only for SerializableObjects.  There is no instancing of RationalTime, TimeTransform, or TimeRange because these are value types now (in C++).  Instancing them would be plain dumb, and unsupportable in general.  You’ll notice that these types don’t pump out a REF_ID field.

Instanced Effects or Markers: What would this mean and why would you want them instanced?

not sure.  if effects and markers must be “owned” by someone, maybe they want an owner field, so they can’t be given to multiple owners?
If I’ve some sort of custom effect, and I want to use it multiple times, but not keep track of changing it multiple times because I had to copy it, is that a valid use case?

PR #511, opened 2019-04-25
Add function to explicitly load SchemaDef plugins

ssteinbach:
- Currently, schemadefs are only loaded implicitly, either by reading a file that contains the schemadef or by calling `ActiveManifest()` or `otio.schema.schemadef.from_name().module()`.
- This PR adds a function `otio.schema.schemadef.module_from_name()` that makes it easy to explicitly load a plugin.
- This also clarifies the unit tests around how schemadefs get loaded and adds some documentation about how to use this function.

### DIFF
--- a/docs/tutorials/write-a-schemadef.md
+++ b/docs/tutorials/write-a-schemadef.md
@@ -91,8 +91,20 @@ Then you need to add this manifest to your `$OTIO_PLUGIN_MANIFEST_PATH` environm
 
 Now that we've defined a new otio schema, how can we create an instance of the
 schema class in our code (for instance, in an adapter or media linker)?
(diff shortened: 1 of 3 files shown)


opentimelineio/schemadef/__init__.py
```diff
@@ -1,5 +1,52 @@
...
-    ns[name] = mod
+
+def _load_schema_module():
+    from .. import schema
+    return schema
+
+
+class DeferredPluginLoaderModuleWrapper(object):
+    """Replaces the otio.schemadef module so that __getattr__ can be overridden
+    in python2 on modules.  In python3.7+ this module wrapper isn't necesarry.
```
mikemahony:
```suggestion
    in python2 on modules.  In python3.7+ this module wrapper isn't necessary.
```

Issue #484, opened 2019-04-08
AAF Reader: File can no longer be parsed after #460


andrewmoore-nz:
Have just come across a file I'd been using for testing that was previously being parsed (mostly) correctly by OTIO but is now throwing an error:

**CannotComputeAvailableRangeError: No available_range set on media reference on clip**
```
WARNING:root:fat sector count missmatch difat: 236 header: 138
Traceback (most recent call last):
  File "/usr/local/bin/otiocat", line 11, in <module>
    load_entry_point('OpenTimelineIO==0.10.0.dev1', 'console_scripts', 'otiocat')()
  File "/Users/dev/BitBucket/Project/Submodules/OpenTimelineIO/opentimelineio/console/otiocat.py", line 132, in main
    read_adapter_arg_map
  File "/Users/dev/BitBucket/Project/Submodules/OpenTimelineIO/opentimelineio/console/otiocat.py", line 99, in _otio_compatible_file_to_json_string
    **adapter_argument_map
  File "/Users/dev/BitBucket/Project/Submodules/OpenTimelineIO/opentimelineio/adapters/__init__.py", line 144, in read_from_file
    **adapter_argument_map
  File "/Users/dev/BitBucket/Project/Submodules/OpenTimelineIO/opentimelineio/adapters/adapter.py", line 136, in read_from_file
    **adapter_argument_map
  File "/Users/dev/BitBucket/Project/Submodules/OpenTimelineIO/opentimelineio/plugins/python_plugin.py", line 128, in _execute_function
    return (getattr(self.module(), func_name)(**kwargs))
  File "/Users/dev/BitBucket/Project/Submodules/OpenTimelineIO/opentimelineio_contrib/adapters/advanced_authoring_format.py", line 908, in read_from_file
    _fix_transitions(result)
  File "/Users/dev/BitBucket/Project/Submodules/OpenTimelineIO/opentimelineio_contrib/adapters/advanced_authoring_format.py", line 656, in _fix_transitions
    _fix_transitions(thing.tracks)
  File "/Users/dev/BitBucket/Project/Submodules/OpenTimelineIO/opentimelineio_contrib/adapters/advanced_authoring_format.py", line 700, in _fix_transitions
    _fix_transitions(child)
  File "/Users/dev/BitBucket/Project/Submodules/OpenTimelineIO/opentimelineio_contrib/adapters/advanced_authoring_format.py", line 692, in _fix_transitions
[...]

PR #512, opened 2019-05-02
AAF Writer: Making precheck() more robust

freesonluxo:
This addresses #484 (not fixing it, but making the failure more elegant).

The validate_metadata() function requires access and verification of certain fields in objects that may or not be there. I thought that using a "promise" like pattern led to a much more elegant, robust result. Hopefully this isn't too much overkill, it does allow our precheck to be a lot more expressive.


freesonluxo:
Perhaps I should add a negative test that actually fails the precheck here.

freesonluxo:
Hm. I'd like to check some more things before pushing this. Can we pause on this for now?

freesonluxo:
Actually, I think I might have jumped the gun thinking my code was causing issues. It was user error. This is ready to be pushed!

PR #515, opened 2019-05-08
Add support for more elements and XMLs without libraries

eric-with-a-c:
This merge adds support for the following elements:
* User Metadata
* Keywords
* Notes

Additionally it adds support for XMLs that don't have a library element. When using the drag-and-drop features of FCP X the XML that is generated doesn't necessarily contain a library. This is the case when you drag an Event, a Project, or a Clip from FCP X to outside the app.


opentimelineio_contrib/adapters/fcpx_xml.py
```diff
@@ -162,13 +161,24 @@ def __init__(self, otio_timeline):
...
-            "event",
-            {"name": self.otio_timeline.name}
-        )
-        self.format_dictionary = {}
-        self.resource_dictionary = {}
+        if self.otio_timeline.schema_name() == "Timeline":
+            self.timelines = [self.otio_timeline]
+        else:
+            self.timelines = [
+                timeline for timeline in self.otio_timeline.each_child(
```
ssteinbach: I _think_ (not 100% though) you could replace this block with something like:
`self.timelines = list(self.otio_timeline.each_child(descended_from_type=otio.schema.Timeline)))`

eric-with-a-c: Yeah, that's much better! The fix has been submitted.

PR #517, opened 2019-05-10
AAF Reader: simplify source_range and available_range in transcribe() of SourceClip

freesonluxo:
During _transcribe(), we store just one parent. This is great, but it would be even better to have the entire parent chain available - both for debugging and perhaps for logic. I believe there is at least one if statement which really wants to say "am I a clip in the CompositionMob", which would now be made a lot easier by this change.


ssteinbach:
Awesome, thanks!

PR #518, opened 2019-05-21
AAF Writer: To get the edit rate, look at the first "child" not the first "clip".

freesonluxo:
This simple change is to address cases where the track has all gaps and does not have proper "clip"s


ssteinbach:
That unit test you added could be useful in the future to poke at simplification stuff as well. Thanks!

PR #521, opened 2019-05-28
Autogenerate documentation of the serialized data model

ssteinbach:
Adds a script that autogenerates a markdown file containing the serialized classes and their serialized properties only.  Also adds a unit test that checks to see if that document needs to be updated.

*TODO*:

- [x] Add schema version numbers to document
- [x] Add a reference from the file format schema doc to this doc
- [x] Add documentation to the unit test on how to update the doc using the make target
- [x] Makefile target is out of date and needs to point at the correct path
- [x] Add `opentime` classes as well
- [x] Remove existing notes on schema
- [x] Add second file, without documentation, just the fields to the `docs` folder.

Resulting page can be viewed here: https://github.com/ssteinbach/OpenTimelineIO/blob/document_serialized_data_model/docs/tutorials/otio-serialized-schema.md

Schema only page (no documentation):
https://github.com/ssteinbach/OpenTimelineIO/blob/document_serialized_data_model/docs/tutorials/otio-serialized-schema-only-fields.md


opentimelineio/console/autogen_serialized_datamodel.py
```diff
@@ -0,0 +1,233 @@
...
+*full module path*: {}
+
+*documentation*:
+
+```
+{}
+```
+
+parameters:
+"""
```
reinecke: nit: I'd prefer to use the field names and keyword args to the `format` method when the templates are located away from where `format` is being called.

opentimelineio/console/autogen_serialized_datamodel.py
```diff
@@ -0,0 +1,233 @@
...
+ regenerated.
+
+# Classes
+
+"""
+
+CLASS_HEADER = """
+### {}
+
+*full module path*: {}
```
reinecke: The doc may read a little better if the module path is visually differentiated using code tick (`) marks

ssteinbach:
@reinecke interesting suggestion.  Where would you put a document like that?  (just the schema, no docstrings).

reinecke:
Knee-jerk, `tests/sample_data` feels like the current best fit. I don't have too strong of opinions however.

Also this suggestion may live a bit in the uncanny valley between documenting the data model and having a more formal spec, so I'd understand if it felt out-of-scope for this PR.

jminor:
I think this is a good idea as long as it is clearly labelled as auto-generated. This can be a good evolutionary step towards the formal spec.

ssteinbach:
`tests/sample_data` doesn't feel correct.  I feel like it should go somewhere in the documentation area if anywhere.  Do you think the unit test should cover the version w/ documentation?  or just the bare one that you're proposing?

jminor:
I think the output should go into `docs/` and it should include the doc strings.

ssteinbach:
@reinecke @jminor just to clarify:

1. In addition to the version of this document _with_ docstrings, there should be a second version _without_ docstrings so that just changes to the serialized format are visible.
2. The file with docstrings goes into `docs/tutorials` currently.  The file without docstrings would go into `docs`
3. The file without docstrings would *not* get referenced by the documentation pages, which instead reference the file with docstrings
4. The existing schema notes in: https://opentimelineio.readthedocs.io/en/latest/tutorials/otio-file-format-specification.html would get removed (replaced by the auto-generated, up to date version this PR makes)
5. The unit test is currently running against the version _with_ docstrings... for keeping the documentation up to date that is probably nice, but if all we care about is correctness, then we might want to run it against the version _without_ docstrings

ssteinbach:
Alright, I created a second file, that isn't linked by the docs but is in that directory.  The unit test runs against the exhaustive one (with docstrings) to ensure that document is always up to date.

reinecke:
Overall, that seems to hit the mark.
Ultimately we're talking about two docs with specific purposes here:
- version _without_ docstrings: meant to trigger a test failure when there are schema updates and highlight PRs that may affect the schema.
- version _with_ docstrings: provides helpful documentation for humans and should be kept up to date with the current state of affairs.

So it seems that test failures looking for diffs in these have separate goals:
- The _without_ docstrings test failure indicates that the change may be potentially breaking to existing serialized files
- The _with_ docstring test failure indicates that the change has left the documentation out of sync with the current code

All this is a long way of saying, would it make sense to have the _with_ docstring version auto-generated as part of the documentation publishing rather than checking it into the repo? Then the _without_ one is checked into the repo and triggers test failure when out of sync.

PR #524, opened 2019-06-06
FCP XML: Adds a check on the NTSC element before reading its value

BadSingleton:
Some softwares don't add the `<NTSC>` element to the `<rate>` element if the framerate
is not NTSC. Check if the element can be found before trying to
check it's value.


BadSingleton:
@ssteinbach Sorry for the late reply, yes there is a CLA covering me on the behalf of Unity; @LauraJeffords is the legal person overseeing this.

ssteinbach:
Excellent.  Sorry for the confusion!

Issue #69, opened 2017-01-23
Image sequence support


jminor:
Sometimes the media for a clip is a sequence of images.

repsac:
We've only been using image sequences. I have been simply storing the root directory in "target_url" and then handle loading the sequence in my own logic.

jminor:
I've seen someone else use a `%d` or `*` in place of frame number, like many VFX apps do. These all work fine for the moment in a closed workflow, but some of the adapters, media linkers and cross-company workflows will require that we formalize this more. We should formalize the syntax of the `target_url` to make it clear how this works.

bashesenaxis:
fcp xml 7 urls for image sequences are simply a url to the first frame. Though...that doesn't really work all that well, because if I do premiere > xml > premiere, the result is that I only see that first frame. It seems to have forgotten that it was an image sequence. I guess that`ll also make it a bit harder to reliably discern whether we`re dealing with a single image or an image sequence.

But on the topic of frame wildcard, our preference for otio would be in the form of "%04d".

apetrynet:
Has this been discussed any further? I'm used to and like variations of the "%04d" format my self. 

Some alternative ways of indicating sequences (just throwing them out there):
```
target_url='/path/to/sequence.1001.ext[1001-1050]'
target_url='/path/to/sequence.%04d.ext[1001-1050]'
```

The following would probably fit in the best to the current code:
```
target_url='/path/to/sequence.%04d.ext'
start_time=1001
duration=50
```
And perhaps add an attribute in the `ExternalReference` class that indicates an image sequence or not. In my work on the Hiero plugin I looked at the MediaSource and MediaFileInfo documentation which gives you an indicator if source is a video container or an image sequence. I would assume other applications supporting image sequences have similar logic so OTIO media_linkers, adapters and plugins could set the `"is_image_sequence"` flag to `True` or `False` accordingly.

In addition to this there's always the option to write functions/methods that analyze the media_reference and determine whether or not it's dealing with a sequence, but not sure if that should be a part of creating a timeline or not. Might slow things down a bit and demands robust I/O routines for all supported platforms.

apetrynet:
Thought about this some more and threw together a quick suggestion on one way of doing it. It's based on known regex patterns for frame substitutions. So when asking if a media_reference is a sequence it checks for a known pattern in `target_url`.
It's still up to each user to actually convert the target_url into a valid path based on `available_range` for instance.
[...]

PR #536, opened 2019-07-11
[WIP] Image reference schema suggestion

apetrynet:
Hi!

I thought I'd put together a suggestion for how to implement `ImageReference` as mentioned in #69 
Also added a test case to illustrate usage.

Let me know what you think.

### DIFF
--- a/opentimelineio/schema/__init__.py
+++ b/opentimelineio/schema/__init__.py
@@ -32,6 +32,9 @@ from .missing_reference import (
 from .external_reference import (
     ExternalReference
 )
+from .image_reference import (
+    ImageReference
+)
 from .clip import (
     Clip,
 )
--- a/opentimelineio/schema/clip.py
+++ b/opentimelineio/schema/clip.py
@@ -31,7 +31,8 @@ from .. import (
     exceptions,
 )
 from . import (
-    missing_reference
+    missing_reference,
+    image_reference
 )
 
 
@@ -100,6 +101,24 @@ class Clip(core.Item):
             )
 
         return copy.copy(self.media_reference.available_range)
+
+    def trimmed_range(self):
+        """The range after applying the source range.
+         This overrides `core.Item`'s `trimmed_range` to support
+         image sequences.
+        """
+        if self.source_range is not None:
+            if isinstance(
+                self._media_reference,
+                image_reference.ImageReference
+            ):
+                return self._media_reference.map_source_range_to_frame_range(
+                    self.source_range
+                )
+
+            return copy.copy(self.source_range)
+
+        return self.available_range()
 
     def __str__(self):
         return 'Clip("{}", {}, {}, {})'.format(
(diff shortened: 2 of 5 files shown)


apetrynet:
So, I realize the timing of this PR was quite bad (with the move to C++, SIGGRAPH and all), but has anyone had the chance to look at it and made any thoughts on it? Is it worth reproducing in C++?

jminor:
Hi @apetrynet! Yes, we did see this and a few people at SIGGRAPH mentioned the topic. Sorry for the delay in responding. As you mentioned, we were busy with the C++ port, and now we're beginning to move the project over to the Academy Software Foundation.

That will involve forming a technical steering committee who can help us decide, as a community, about the roadmap. One of the key things that the roadmap needs to include is schema improvements like this.

In effect that means that we're not quite ready to move forward with this yet, but we do hope to in the coming months.

As for the specifics of your proposal, here are a few things that come to mind: 

- Could you explain more about why frame_range could be different from available_range?
- Could you add an example `.otio` file with an ImageReference in it?
- Maybe ImageSequenceReference might be a better name to avoid confusion versus a single held image?
- I'm sure people will have opinions about the `%04d` syntax.
- We often see cases where people render every 2nd or 4th frame. Should that be represented as well?

apetrynet:
Hi! 

> Could you explain more about why frame_range could be different from available_range?

In my experience, almost all image based plates in a VFX pipeline have frame numbers starting at something else than what the TC would imply. Usually 1001-1200 etc. I don't know all the history for this, but it makes life easier using the plate in software like Maya, Nuke etc. to keep the timeline at a standard range. I seem to recall Maya actually having an issue with timelines that had really high numbered ranges. 
By keeping the `available_range` in "sync" with the source TC and the `frame_range` with the renumbered frames on disk, we are able to maintain the metadata throughout the pipeline.

> Maybe ImageSequenceReference might be a better name to avoid confusion versus a single held image?

Sure. Went back and forth on this myself. I haven't thought it through well enough yet, but this could also be used for held single frames though.

> I'm sure people will have opinions about the %04d syntax.

Oh, yes. The everlasting debate on frame representation :) 
There's no enforcing of the "%04d" in the code. Just used it as an example in the docstring since I find it the most common. It would be up to the developers using OTIO to decide which format suits them best I guess. 

> We often see cases where people render every 2nd or 4th frame. Should that be represented as well?

True. I haven't thought about that yet. That's a tough one in any case. The TC duration could be "full length", but the data won't. How would this be reflected in other media? I think in a video container, this would remap the frame numbers to create a shortened video file. 
However, I think they should be treated as missing frames in the application reading the files. So in OTIO there shouldn't be a need to write any logic to support it(?)

> Could you add an example .otio file with an ImageReference in it?

Sure. Will add it to the PR

apetrynet:
@jminor @reinecke I'll just close this PR then right?

reinecke:
@apetrynet As long as you're happy with #602 :)

apetrynet:
I'm confident we'll land a good solution with that one, yes :)
Thanks!

PR #538, opened 2019-07-12
xges: Fix tracks ordering

thiblahute:
The tracks should be loaded in reverse order for the XGES format.
    
Also in GES, layers (tracks in otio terms) do not have a track type
(tracks are global to the timeline which formalizes the outputed
types), thus we should try to keep audio and video tracks in the same
order during de-serialization.

### DIFF
--- a/contrib/opentimelineio_contrib/adapters/xges.py
+++ b/contrib/opentimelineio_contrib/adapters/xges.py
@@ -307,7 +307,8 @@ class XGES:
     def _add_layers(self, timeline, otio_timeline, all_names):
         for layer in timeline.findall("./layer"):
             tracks = self._build_tracks_from_layer_clips(layer, all_names)
-            otio_timeline.tracks.extend(tracks)
+            for track in tracks:
+                otio_timeline.tracks.insert(0, track)
 
     def _get_clips_for_type(self, clips, track_type):
         if not clips:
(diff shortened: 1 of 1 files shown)


opentimelineio_contrib/adapters/xges.py
```diff
@@ -307,7 +307,8 @@ def to_otio(self):
     def _add_layers(self, timeline, otio_timeline, all_names):
         for layer in timeline.findall("./layer"):
             tracks = self._build_tracks_from_layer_clips(layer, all_names)
-            otio_timeline.tracks.extend(tracks)
+            for track in tracks:
+                otio_timeline.tracks.insert(0, track)
```
ssteinbach:
```python
    otio_timeline.tracks.extend(reversed(tracks))
```

Seems more readable than `insert(0, ...)`

opentimelineio_contrib/adapters/xges.py
```diff
@@ -695,7 +696,28 @@ def _serialize_timeline(self, project, ressources, otio_timeline):
...
+        audio_tracks = [
+            t for t in otio_timeline.tracks
+            if t.kind == otio.schema.TrackKind.Audio and list(t)
+        ]
+        audio_tracks.reverse()
+
+        all_tracks = []
+        for i, otio_track in enumerate(video_tracks):
+            all_tracks.append(otio_track)
+            if len(audio_tracks) -1 >= i:
```
ssteinbach: I think you can also here:
```python
try:
  all_tracks.append(audio_tracks[i])
except IndexError:
  pass
```

opentimelineio_contrib/adapters/xges.py
```diff
@@ -695,7 +696,28 @@ def _serialize_timeline(self, project, ressources, otio_timeline):
...
+                all_tracks.append(audio_tracks[i])
+
+        if len(audio_tracks) > i:
```
ssteinbach: similarly, I think a bit more explicit here:

```python
if len(audio_tracks) > len(video_tracks):
  all_tracks.extend(audio_tracks[len(video_tracks):])
```

opentimelineio_contrib/adapters/xges.py
thiblahute: That wouldn't add the tracks to the start of the `otio_timeline.tracks` list but it would add them to the end.

PR #539, opened 2019-07-13
Include <functional> to fix the build on fedora 30 (gcc 9.1.1)

thiblahute:


davidbaraff:
On what systems (i.e. OS and compiler) is <functional> needed?

thiblahute:
Fedora 30 gcc `9.1.1`

thiblahute:
Also happens on ubuntu with gcc 7.4.0...

PR #545, opened 2019-07-23
Fix installing in user prefix when --user is specified

thiblahute:
When the user specifies `--user` we have to make sure to use
the usersite package directory instead of letting cmake install
lib systemwide.

Without that I am getting a permission issue:

```
-- Install configuration: "Release"
CMake Error at src/opentime/cmake_install.cmake:47 (file):
  file cannot create directory:
  /usr/lib/python3.7/site-packages/opentimelineio/cxx-libs/lib.  Maybe need
  administrative privileges.
Call Stack (most recent call first):
  src/cmake_install.cmake:43 (include)
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -20,6 +20,7 @@ from setuptools import (
 import setuptools.command.build_ext
 import setuptools.command.build_py
 from setuptools.command.install import install
+from distutils.sysconfig import get_python_lib
 from distutils.version import LooseVersion
 import distutils
 
@@ -83,7 +84,7 @@ def compute_cmake_args():
                 )
             else:
                 cxxLibDir = os.path.abspath(
-                    os.path.join(setuptools.__file__, "../../opentimelineio/cxx-libs")
+                    os.path.join(get_python_lib(), "opentimelineio", "cxx-libs")
                 )
             cmake_args += ['-DCMAKE_INSTALL_PREFIX=' + cxxLibDir,
                            '-DOTIO_CXX_NOINSTALL:BOOL=ON']
--- a/setup.py
+++ b/setup.py
@@ -35,6 +35,7 @@ _ctx.build_temp_dir = None
 _ctx.installed = False
 _ctx.ext_dir = None
 _ctx.source_dir = os.path.abspath(os.path.dirname(__file__))
+_ctx.install_usersite = ''
 _ctx.debug = False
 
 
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -76,9 +77,14 @@ def compute_cmake_args():
             cmake_args += ['-DCMAKE_INSTALL_PREFIX=' + _ctx.cxx_install_root]
 
         else:
-            cxxLibDir = os.path.abspath(
-                os.path.join(setuptools.__file__, "../../opentimelineio/cxx-libs")
-            )
+            if "--user" in sys.argv:
```
thiblahute: I know this is far from being elegant, but it looks like setuptools is not providing the information in a clean way.

setup.py
```diff
@@ -76,9 +77,14 @@ def compute_cmake_args():
...
+                cxxLibDir = os.path.abspath(
+                    os.path.join(_ctx.install_usersite, "opentimelineio/cxx-libs")
```
reinecke: To keep platform agnostic, this line should change to:
`os.path.join(_ctx.install_usersite, "opentimelineio", "cxx-libs")`

setup.py
ssteinbach: do these slashes still work on windows?  or do we need to switch to an `os.path.sep.join` kind of thing?

setup.py
```diff
@@ -76,9 +77,14 @@ def compute_cmake_args():
...
+                )
+            else:
+                cxxLibDir = os.path.abspath(
+                    os.path.join(setuptools.__file__, "../../opentimelineio/cxx-libs")
+                )
```
reinecke: I know this is from the existing code, but these lines should probably be something like:
```
setuptools_parent_dir = os.path.dirname(os.path.dirname(setuptools.__file__))
cxxLibDir = os.path.abspath(os.path.join(setuptools_parent_dir, "opentimelineio", "cxx-libs"))
```

thiblahute: I ended up using `distutils.sysconfig.get_python_lib()` in another commit as it seems to be the correct solution. Let me know if you prefer me to get back to using that "hack" :-)

reinecke: That distutils solution certainly does feel much better!

ssteinbach: Yes, that is much cleaner.  Thanks!

PR #546, opened 2019-07-23
Add `meson` as buid system

thiblahute:
[Meson] is a build system very wildely used in the [open source
community] and gainning traction.

This aims at replacing setuptools completely as it builds the c++
code as well as handling setuptools/python files.

We can keep both build system for a while to see if meson works for
the community (hoping you like the idea of using Meson :-))

This port supports usage of otio in an 'uninstalled' environment only
requiring to set the `PYTHONPATH`, as explained at the end of the
configuration. We could add an helper in the future to make it even
simpler.

The difference in term of build compared to setuptools is that it
statically links libopentime[lineio] into the python bindings when
the libs are not installed, making the setup simpler.

There still is a `OTIO_CXX_NOINSTALL` option which works the same way
as with CMake. Two more options are provided:

- `python`: A [feature] to enable or disable python bindings
- `python_version`: The python version to use, valid values are
  ['2', '3']

This build also generate `.pc` files for consumption with `pkg-config`
so libopentimelineio can be more easily consumed by other open source
components.

This port builds everything but the swift bindings.

The port as only been tested on linux right now, but meson has been
proven to work on all platforms (for example we build GStreamer on
all platforms and with all major C compilers without any problem) so
I do not forsee any major issue for that port to work on other
targeted platforms.

[Meson]: http://mesonbuild.com/
[open source community]: http://mesonbuild.com/Users.html#page-description
[feature]: http://mesonbuild.com/Build-options.html#features


thiblahute:
Tested on mac OSX and it works there too.

thiblahute:
> Meson looks appealing, particularly in that it simplifies the way that the python and C++ build system interoperate, but given that CMake appears to be the build system for most of the projects that ASWF has adopted, we'd like to stick with CMake for now. Thanks for the suggestion!

Well, you are not only using CMake right now, but setuptools and some CMake, meson could also consume the CMake part as a subproject and only replace setuptools, but to me it doesn't make so much sense as meson can replace it all in a clean integrated way.

jminor:
We can revisit this when we switch to the ASWF CI system, as that will likely require some changes as well.

PR #549, opened 2019-07-31
otioview: Implement rendering support with GES

thiblahute:
This is really an MVP where we simply render the timeline in
jpeg+opus in matroska.

Once rendering is done, the file is opened with the default video
player from the system

### DIFF
--- a/src/opentimelineview/console.py
+++ b/src/opentimelineview/console.py
@@ -34,6 +34,7 @@ import opentimelineio as otio
 import opentimelineio.console as otio_console
 import opentimelineview as otioViewWidget
 from opentimelineview import settings
+from . import ges_renderer
 
 
 def _parsed_args():
@@ -147,6 +148,10 @@ class Main(QtWidgets.QMainWindow):
         file_menu = menubar.addMenu('File')
         file_menu.addAction(file_load)
         file_menu.addSeparator()
+        if ges_renderer.Renderer is not None:
+            render = QtWidgets.QAction('Render...', menubar)
+            render.triggered.connect(self._render)
+            file_menu.addAction(render)
         file_menu.addAction(exit_action)
 
         # navigation menu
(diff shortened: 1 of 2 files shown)


src/opentimelineview/console.py
```diff
@@ -147,6 +148,10 @@ def __init__(
         file_menu = menubar.addMenu('File')
         file_menu.addAction(file_load)
         file_menu.addSeparator()
+        if Renderer:
```
ssteinbach: `is not None` is a more explicit check

src/opentimelineview/console.py
```diff
@@ -34,6 +34,7 @@
 import opentimelineio.console as otio_console
 import opentimelineview as otioViewWidget
 from opentimelineview import settings
+from ges_renderer import Renderer
```
ssteinbach: I'd prefer not to do a `from` import for `Renderer`, I think it would read more clearly below if the branch read:

`if ges_renderer.Renderer is not None`

thiblahute: Done.

src/opentimelineview/console.py
thiblahute: Done.

thiblahute:
> This looks cool to me, but I think I want to consider GStreamer a bit more (and whether this is how we'd want to integrate a viewer capability into otioview). It would definitely need some docs so that people know this is a capability of the system, and we'd probably want to include an optional PIP depedency on the `gi` module so that people can easily install it into virtual environments if they're so inclined.
> Let us talk it over a bit and get back to you. Thanks!

Great, let me know what you decide and I will add all the required documentation.

Also as discussed this is enabling rendering the full timeline, I could also work on adding support for playback with GStreamer, that wouldn't be much work and could also add up quite a lot to `otioview`, just let me know.

reinecke:
We think we want to minimize dependencies in OpenTimelineIO to keep it accessible and easy to maintain. The distros used at many studios does not have a new enough gstreamer to support this work, so merging it in could cause issues for them.
Perhaps this work could form the basis of a separate project?

Issue #552, opened 2019-08-07
PyAAF documentation requirement needs update


swallitsch:
When #383 was completed the [documentation](https://opentimelineio.readthedocs.io/en/latest/tutorials/adapters.html?highlight=pyaaf#avid-aaf) wasn't updated to reference that the requirement is now pyaaf2

ssteinbach:
Should be addressed by #554

PR #554, opened 2019-08-07
Update docs for pyaaf2

ssteinbach:
Just fixing an oversight - when we moved to pyaaf2 we forgot to update the docs with new instructions on how to get the library and use it with OpenTimelineIO


ssteinbach:
Addresses #552

Issue #553, opened 2019-08-07
setup.py should check if src/deps are empty


swallitsch:
src/deps are nested git repositories referencing specific commits for the required libraries. When missing, CMake gives a decent error but this could be improved with direct checking.

If 'git' is found in the path, setup.py could even attempt a git checkout during the build.

ssteinbach:
good call.  there are already a bunch of checks in there, I can add one for this.

ssteinbach:
I think it makes more sense to put that check in the cmake build system rather than setup.py -- let me know if this makes sense! #556

swallitsch:
Yes absolutely. Sorry I was just going off what we briefly discussed at siggraph

PR #556, opened 2019-08-07
Detect submodules not synced

ssteinbach:
Add code in the CMake build system to automatically run `git submodule update`.  

Addresses #553


ssteinbach:
@swallitsch great note!  I added an extra check in the CMakeLists.txt in the src/deps folder that explicitly checks the submodules to make sure they exist with a more exact message of what might be going on.

PR #558, opened 2019-08-14
AAF reader: global_start_timecode fixes

### LINKED ISSUE #522: AAF Reader: Global Start Time not being set for various files
Still coming across many scenarios where the `global_start_time` field is not being set when it should be. I see that this was supposedly addressed in #478 , and the test file included (.../sample_data/simple.aaf) seems to generate a valid `global_start_time` RationalTime object, but perhaps this being done in a too-specific way? 

An example here of an AAF from Premiere that starts at `05:00:00:00`, but when read in to an OTIO, the global_start_time comes through as `null`. For reference, the .xml that was exported from the same project/timeline comes through with the correct field set of:

```
    "global_start_time": {
        "rate": 23.976023976023978, 
        "value": 432000
    }, 
```

Files attached here:

[FullTestSeq_v02_TimeshiftExport_5hours_AAF_XML.zip](https://github.com/PixarAnimationStudios/OpenTimelineIO/files/3245575/FullTestSeq_v02_TimeshiftExport_5hours_AAF_XML.zip)

Also seeing this with AAFs coming from Media Composer as well. This following example should start at `01:00:00:00` but is coming through with a `global_start_time` of `null`. 

[ClipMutingTest_1hours_AAF.zip](https://github.com/PixarAnimationStudios/OpenTimelineIO/files/3245600/ClipMutingTest_1hours_AAF.zip)

reinecke:
Fixes #522  

1. Added code to look for LegacyTimecode (Premiere AAFs) in addition to Timecode
2. Updated to support fps fractions
3. Timecode candidate selected for global_start_timecode is now selected using PhysicalTrackNumber


contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py
```diff
@@ -537,18 +538,20 @@ def _find_timecode_track_start(track):
...
-        edit_rate = aaf_metadata.get("EditRate", "0")
-        fps = aaf_metadata.get("Segment", {}).get("FPS", 0)
+    if aaf_metadata.get("MediaKind") in {"Timecode", "LegacyTimecode"}:
+        edit_rate = Fraction(aaf_metadata.get("EditRate", "0"))
+        if edit_rate.denominator == 1:
+            rate = edit_rate.numerator
+        else:
+            rate = float(edit_rate)
         start = aaf_metadata.get("Segment", {}).get("Start", "0")
+        physical_track_number = aaf_metadata.get("PhysicalTrackNumber", 1)
```
jminor: Is the default value here needed? Would PhysicalTrackNumber ever be missing?

reinecke: I don't expect it would be. We should probably be able to assume the same for `EditRate`, `Segment`, and `Start` too, right? I also noticed that defaulting `Start` to a string `"0"` didn't match the files I looked like that had integer values.

I can do something like:
```try:
    edit_rate = Fraction(aaf_metadata.get["EditRate"])
    start = aaf_metadata["Segment"]["Start"]
    physical_track_number = aaf_metadata["PhysicalTrackNumber"]
except KeyError:
    continue```

That way we'll do a best-effort without totally dying if we're missing something.

Comment (jminor) [reply contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py]: Since we still have a ways to go in understanding AAF deeply, I think we should throw an exception if we find something we don't expect. That isn't the best thing for the user at that moment, but it forces us to stop, figure out what we misunderstood, and then fix it. So maybe
```
...
except KeyError:
  raise AAFAdapterError("Timecode missing PhysicalTrackNumber")
```

File: contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py [hunk 2]
```diff
@@ -33,6 +33,7 @@
 import numbers
 import copy
 from collections import Iterable
+from fractions import Fraction
```
ssteinbach: I'm guessing the style in this file is to do the "from" style imports but I think I'd prefer it if moving forward we could leave the namespaces in... I still think thats more readable, especially for stuff like this.

contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py
```diff
@@ -537,22 +542,36 @@ def _find_timecode_track_start(track):
...
+
+    # Edit Protocol section 3.6 specifies PhysicalTrackNumber of 1 as the
+    # Primary timecode
+    try:
+        physical_track_number = aaf_metadata["PhysicalTrackNumber"]
+    except KeyError:
+        raise AAFAdapterError("Timecode missing 'PhysicalTrackNumber'")
+
+    if physical_track_number != 1:
+        return
```
ssteinbach: Just so I understand, there can be multiple timecode tracks with physical track number != 1 and they're not what we're looking for, but *if* there is one with the number 1 then its the edit rate for the track?

reinecke: More or less. I wouldn't say it's the edit rate for the track, it's that it's the primary timecode offset.

I think this addresses the idea that in production cases you could have multiple timecodes associated with something. For instance when working with telecine dailies, you have the tape timecode you've captured (usually starting at a specific hour offset based on shoot day) but there would also be a film timecode which is expressed by a numbering system burned into the edge of the film when it was manufactured. For the post-production use case the tape timecode is wheat you work with most the time. Then at the very end of the process you'd want to use that secondary timecode to map back to the original negative.

The language from the AAF spec says:

> A top-level composition shall contain one or more timecode tracks and include a Primary timecode track.

And then a table below shows that "Primary" will have a Physical track number of 1 and alternative timecodes will have other physical track numbers. My interpretation of this is that there will always be a "Primary" and that it is most likely the timecode a user would be interested in.

contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py
reinecke: updated code to conform to style guidelines.

contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py
ssteinbach: Is it possible to have more than one with a physical track number of 1?  Or does the spec prohibit this?

reinecke: The text is a little ambigous but it is strongly implied that this is the case. Anecdotally from some aaf examples I saw it does seem to be the case.

ssteinbach: Sounds good to me.  Should be obvious if we ever hit this case.

PR #559, opened 2019-08-19
Windows build fixes

darbyjohnston:
Various fixes for building on Windows 10 / Visual Studio 2019:
* Use Windows specific compiler flags
* Only build pybind11 if OTIO_PYTHON_INSTALL=TRUE
* Use WINDOWS_EXPORT_ALL_SYMBOLS to allow building as shared libraries
* Windows doesn't have cxxabi.h so use std::type_info.name() instead; on Windows this returns a human readable string
* A couple of other miscellaneous code fixes


src/opentime/rationalTime.cpp
```diff
@@ -125,10 +125,10 @@ RationalTime::from_timecode(std::string const& timecode, double rate, ErrorStatu
 
     int dropframes = 0;
     if (rate_is_dropframe) {
-        if ((rate == 29.97) or (rate == 30000/1001.0)) {
+        if ((rate == 29.97) || (rate == 30000/1001.0)) {
             dropframes = 2;
         }
-        else if ((rate == 59.94) or (rate == 60000/1001.0)) {
+        else if ((rate == 59.94) || (rate == 60000/1001.0)) {
```
ssteinbach: Does windows not support the `or` and `and` keywords, only `||` and `&&` operators?  I'd prefer we use the `or` and `and` form if compatible.

src/opentime/rationalTime.cpp
```diff
@@ -204,7 +204,7 @@ RationalTime::to_timecode(
     }
 
     bool rate_is_dropframe = is_dropframe_rate(rate);
-    if (drop_frame == IsDropFrameRate::ForceYes and not rate_is_dropframe) {
+    if (drop_frame == IsDropFrameRate::ForceYes && !rate_is_dropframe) {
```
ssteinbach: same here, prefer `and not` to `&& !`

src/opentime/rationalTime.cpp
```diff
@@ -230,7 +230,7 @@ RationalTime::to_timecode(
         }
     }
     else {
-        if ((rate == 29.97) or (rate == 30000/1001.0)) {
+        if ((rate == 29.97) || (rate == 30000/1001.0)) {
```
ssteinbach: and here

src/opentimelineio/stringUtils.h
```diff
@@ -11,7 +11,6 @@ namespace opentimelineio { namespace OPENTIMELINEIO_VERSION  {
     
 void fatal_error(std::string const& errMsg);
 
-std::string demangled_type_name(const char* name);
```
ssteinbach: Does this also need to be bracketed by `#if defined(__GNUC__) or defined(__clang__)` as in the .cpp file?

ssteinbach: ...Oh I see that version became private instead of public.  We might want to change the name in that case so that its clear that that overloaded version of `demangled_type_name` is for wrapping in the `GNUC`/`__clang__`  environment.

davidbaraff:
I originally coded NOT using or and and, and prefer not to use them in c++ code.
Just a preference though.

Yes, I’m old.  ;)

src/opentime/rationalTime.cpp
darbyjohnston: They are not defined in Visual Studio 2019, no idea why. I hadn't seem them before and thought they were GCC/clang specific, but according to cppreference they are standard:
https://en.cppreference.com/w/cpp/language/operator_alternative

This Microsoft doc shows they are missing:
https://docs.microsoft.com/en-us/cpp/cpp/keywords-cpp?view=vs-2019

I guess we could define them as macros on Windows?

darbyjohnston:
OK, I've made the changes and also did a pass to clean up the compiler warnings; now there's just a single warning that looks like it's coming from the pybind11 headers. I've also successfully run "make test" on OSX and Linux, though not on Windows yet since it looks like there is something broken in "setup.py/pip install .".

src/opentime/keywords.h
```diff
@@ -0,0 +1,14 @@
+#pragma once
```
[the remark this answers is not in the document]
darbyjohnston: Good to see you as well! Thanks for checking out the changes.

> #include <ciso646>

Great, did not know about that, definitely much better than defining the macros manually.

src/opentimelineio/deserialization.cpp
```diff
@@ -592,7 +592,12 @@ bool deserialize_json_from_string(std::string const& input, any* destination, Er
 }
 
 bool deserialize_json_from_file(std::string const& file_name, any* destination, ErrorStatus* error_status) {
-    FILE* fp = fopen(file_name.c_str(), "r");
+    FILE* fp = nullptr;
+#if defined(_WIN32)
+    fopen_s(&fp, file_name.c_str(), "r");
```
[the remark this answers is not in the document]
darbyjohnston: From my reading the MSDN docs say that fp will not be changed in the event of an error, but I agree, it would be more clear to test for an error and then set fp to nullptr just in case.

src/py-opentimelineio/opentimelineio-bindings/otio_utils.cpp
```diff
@@ -40,17 +40,17 @@ py::object plain_int(int i) {
 
 py::object plain_int(int64_t i) {
     #if PY_MAJOR_VERSION >= 3
-        PyObject *p = PyLong_FromLong(i);
+        PyObject *p = PyLong_FromLongLong(i);
     #else
-        PyObject *p = PyInt_FromLong(i);
+        PyObject *p = PyLong_FromLongLong(i);
```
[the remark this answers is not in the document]
darbyjohnston: No, thanks, good catch!

meshula:
@darbyjohnston Do you think a rebase against head needs to be done to resolve the conflicts? I'm puzzling through what the problem is...

darbyjohnston:
Sorry about that, let me check...

darbyjohnston:
Hi Nick,

Sorry about that, I forgot to merge the latest changes in. The branch should be up to date now though I wasn't sure if I was supposed to "squash" the merge commit? Also it looks like the Python 2.7 build failed on Travis CI, taking a look now...

meshula:
Since this change enforces UTF8 encoding, the last thing it's hung up on is a non-utf8 character somewhere.

```
UnicodeEncodeError: 'ascii' codec can't encode character u'\u2018' in position 17701: ordinal not in range(128)
317The command "export PYTHONIOENCODING=UTF-8 && tox -e $TOX_ENV" exited with 1.
```

Somewhere a file has got a back tick. 

https://www.fileformat.info/info/unicode/char/2018/index.htm

meshula:
Travis is hung up on is a non-utf8 character somewhere, not sure why.
```
UnicodeEncodeError: 'ascii' codec can't encode character u'\u2018' in position 17701: ordinal not in range(128)
317The command "export PYTHONIOENCODING=UTF-8 && tox -e $TOX_ENV" exited with 1.
```

darbyjohnston:
Hi Nick,

Sorry I should have given you a heads-up; the changes to the .travis.yml file were just an attempt to troubleshoot the failed Python 2.7 Travis build. It started failing for some unknown reason when I merged the latest changes from master with my branch. Oddly enough, it is only failing on the Travis machine, it works locally, and the master branch also works OK on Travis, so I'm not sure what's going on. The env vars were an attempt to get tox to spit out some more useful error information, but that didn't work so I will revert the changes. The only other thing I can think of is to back out my individual changes and see which one is triggering the error, unless you have any other ideas?

Thanks, Darby

meshula:
There's three(?) things going on in this PR; straight up compilation fixes for windows, CMake fixes, and warning fixes, for windows. What would you think about introducing each category, one at a time as new possibly independent PRs? We might sidestep the Travis issue entirely. Also, it sounds like the consensus is landing around non-keyword operators...

darbyjohnston:
Sure, that's fine, I'll close this one and open up the new ones.

darbyjohnston:
Here's the first one:
https://github.com/PixarAnimationStudios/OpenTimelineIO/pull/576

It just passed the CI build. :)

meshula:
Great! One down :)

darbyjohnston:
You probably saw it already, but just in case here's the one for the warnings:
https://github.com/PixarAnimationStudios/OpenTimelineIO/pull/577

Unfortunately it seems to be causing the Travis CI Python 2.7 failure.

PR #561, opened 2019-08-23
Add filebundle (otioz and otiod) adapters

ssteinbach:
# overview

Introduces the idea of an OTIO "file bundle".  This has two flavors: .otioz and .otiod.  This PR includes documentation and implements adapters that can convert into those formats.

# otioz
- Adds .otioz suffix file
- Bundles media and an otio file together into a libzip compatible zipfile (using python's `zipfile` module)
- _only_ works with files that can be found through the filesystem (`file:` protocol `target_url` fields, as parsed by `urllib` in python)
- media files must have unique basenames (because they get put into a flat namespace)
- zipfile structure (for foo.otioz):
   - foo.otioz/
     - version.txt
     - content.otio
     - media/
        - file1
        - file2
        - file3
- the `content.otio` encodes the structure of the timeline and exclusively references media which is present in the `media` subdirectory.
- the `content.otio` is compressed, but the rest of the media files are not
- `version.txt` is only present in otioz files and exclusively encodes a file bundle version string into the file in case of future changes to the layout of file bundles

# otiod

- identical to the otioz in layout, except expressed without the zipfile container as files and directories in the filesystem.  
- This way a `.otioz` can be expanded using `otioconvert` through the adapter system without needing to use zip directly, into a form that any system that reads `otio` from the filesystem can consume.

TODO:

- [x] make the `file://` vs `file:` consistent when generating absolute paths
- ~[ ] add a `file:///path/to/otioz?file=media/somefile.mov` form of `ExternalReference.target_url` when reading OTIOZ files... this allows the otiod adapter the ability to directly decompress `otioz` files, and `otioconvert` could detect it when going from otioz to otiod.~ -- going to do this in a future PR/issue


jminor:
I like this a lot. It seems tremendously useful and is very simple in concept.

If you require the internal otio to have the same name as the zip file, then it will break if/when someone renames the zip file. Using a consistent name, like `content.otio` would avoid this.

On the reading side, how would someone extract the media files? Maybe there could be a helper function or option on the reader to unzip the bundle? That would also allow for a future switch from zip to some other bundle mechanism, if needed.

ssteinbach:
Great note!  I like content.otioz.

As far as a command to unzip the bundle, do you mean some kind of api function you can call?  `otio.adapters.expand_bundle()` or something?

I like the idea of putting an API call that encapsulates the schema though.  Also I wonder if we can put an extra magic number into the file somehow to detect which 'version' of OTIOZ it is.

src/py-opentimelineio/opentimelineio/adapters/filebundle.py
```diff
@@ -0,0 +1,123 @@
...
+
+    # gather the files up in the staging_dir
+    for fn in referenced_files:
+        target = os.path.join(otio_basename, os.path.basename(fn))
+        fmapping[fn] = target
+
+    # update the media references
+    for cl in input_otio.each_clip():
+        try:
+            source_fpath = cl.metadata.target_url
```
reinecke: This looks like it's setting a target_url property on the clip metadata. Shouldn't this be operating on the media_reference?

jminor:
You could put a version in OTIOZ metadata on the top level object, or wrap the content in a `Bundle.1` schema type or something. I bet the IMF and DCP folks would have advice on what else would be worth tracking, though those formats tend to have a bunch of manifest details, checksums, etc. that might make this more complicated than it needs to be.

src/py-opentimelineio/opentimelineio/adapters/filebundle.py
```diff
@@ -0,0 +1,123 @@
...
+    for cl in input_otio.each_clip():
+        try:
+            target_file = cl.media_reference.target_url
+        except AttributeError:
+            continue
+
+        if target_file in referenced_files:
+            continue
+
+        if not os.path.exists(target_file) or not os.path.isfile(target_file):
```
thiblahute: `target_file` is `media_reference.target_url` which is suposdely an 'url' not a path, shouldn't you check that the protocol is `file` (make the assumption it is file if for some reason the url is a path) and then work with the URL path?

tests/test_otioz.py
```diff
@@ -0,0 +1,93 @@
+#!/usr/bin/env python
+#
+# Copyright 2017 Pixar Animation Studios
```
thiblahute: 2019?

tests/test_otioz.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+class OTIOZTester(unittest.TestCase, otio_test_utils.OTIOAssertions):
+    def setUp(self):
+        tl = otio.adapters.read_from_file(SCREENING_EXAMPLE_PATH)
+
+        # convert to contrived local reference
+        for cl in tl.each_clip():
+            cl.media_reference = otio.schema.ExternalReference(
+                target_url=MEDIA_EXAMPLE_PATH
```
thiblahute: Should use a `file://` url instead (as per some recent conversation on slack where we seemed to all agree about that).

src/py-opentimelineio/opentimelineio/adapters/otiod.py
```diff
@@ -0,0 +1,146 @@
...
+# distributed under the Apache License with the above modification is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied. See the Apache License for the specific
+# language governing permissions and limitations under the Apache License.
+#
+
+"""OTIOD adapter - bundles otio files linked to local media in a directory
+
+Takes as input an OTIO file that has media references which are all relative
+local paths (ie file:///foo.mov) and bundles those files and the otio file into
```
reinecke: `file:///foo.mov` is actually an absolute path (which should also work fine). The current behavior will work with files with paths that are relative to cwd. Would it be better to treat relative paths as relative to the input file?

src/py-opentimelineio/opentimelineio/adapters/otiod.py
```diff
@@ -0,0 +1,146 @@
...
+    for fn in manifest:
+        target = os.path.join(
+            filepath,
+            utils.BUNDLE_DIR_NAME,
+            os.path.basename(fn)
+        )
+        fmapping[fn] = target
+
+    # so we don't edit the incoming file
+    input_otio = copy.deepcopy(input_otio)
```
reinecke: This duplicates the behavior of line 80, right?

src/py-opentimelineio/opentimelineio/adapters/otiod.py
```diff
@@ -0,0 +1,146 @@
...
+            cl.media_reference = schema.MissingReference()
+            continue
+
+        try:
+            source_fpath = cl.media_reference.target_url
+        except AttributeError:
+            continue
+
+        basename = os.path.basename(source_fpath.split("file://")[1])
+        newpath = os.path.join("media", basename)
```
reinecke: should this be using `utils.BUNDLE_DIR_NAME`?

src/py-opentimelineio/opentimelineio/adapters/otioz.py
```diff
@@ -0,0 +1,140 @@
...
+                )
+
+            )
+
+    with zipfile.ZipFile(filepath, 'r') as zi:
+        result = otio_json.read_from_string(zi.read(utils.BUNDLE_PLAYLIST_PATH))
+
+        if extract_to_directory:
+            zi.extractall(extract_to_directory)
+
```
reinecke: Would it make sense to relink all the media_reference `target_url` attributes here when `extract_to_directory` is set?

ssteinbach: you mean with an absolute path?  I think they get set to relative paths as is.  But thats a good question.

jminor: I like the idea of using relative paths to point to stuff in the media folder. This will help to move forward some strictness and uses of the target_url which are too loose at the moment.

src/py-opentimelineio/opentimelineio/adapters/otiod.py
```diff
@@ -0,0 +1,146 @@
...
+    # write the otioz file to the temp directory
+    otio_json.write_to_file(
+        input_otio,
+        os.path.join(filepath, utils.BUNDLE_PLAYLIST_PATH)
+    )
+
+    # write the media files
+    os.mkdir(os.path.join(filepath, utils.BUNDLE_DIR_NAME))
+    for src, dst in fmapping.items():
+        shutil.copyfile(src, dst)
```
apetrynet: Would it be an idea to abstract the file copy into a function that attempts to use hard links on supported OS and when files exist on the same server? It would speed up the process and save disk space.

src/py-opentimelineio/opentimelineio/adapters/otioz.py
reinecke: Makes sense. As long as the paths are relative to the `.otio` file then you can move the `.otiod` around on your filesystem and have everything still be properly pathed.

src/py-opentimelineio/opentimelineio/adapters/otioz.py
```diff
@@ -0,0 +1,140 @@
...
+    # write the otioz file to the temp directory
+    otio_str = otio_json.write_to_string(input_otio)
+
+    with zipfile.ZipFile(filepath, mode='w') as target:
+        # write the media
+        for src, dst in fmapping.items():
+            target.write(src, dst)
+
+        # write the OTIO
+        target.writestr(utils.BUNDLE_PLAYLIST_PATH, otio_str)
```
ssteinbach: We should pass in `compress_type=ZIP_LZMA` for the json part since its going to be parsed anyway.  Then we get the benefits of the zip file format for people concerned with large otio file size.

src/py-opentimelineio/opentimelineio/adapters/otioz.py
```diff
@@ -0,0 +1,140 @@
...
+            fmapping[source_fpath.split("file://")[1]]
+        )
+
+    # write the otioz file to the temp directory
+    otio_str = otio_json.write_to_string(input_otio)
+
+    with zipfile.ZipFile(filepath, mode='w') as target:
+        # write the media
+        for src, dst in fmapping.items():
+            target.write(src, dst)
```
ssteinbach: explicitly pass in `compress_type=ZIP_STORED` so that it is clear that media is written into the archive as uncompressed bits (like usdz)

docs/tutorials/otio-filebundles.md
```diff
@@ -0,0 +1,63 @@
...
+- MissingIfNotFile: will replace any media references that meet the above condition with a `MissingReference`, preserving the original media reference in the metadata of the new `MissingReference`.
+- AllMissing: will replace all media references with `MissingReference`, preserving the original media reference in metadata on the new object.
+
+## OTIOD
+
+The OTIOD adapter will build a bundle in a directory stucture on disk.  The adapter will gather up all the files it can and copy them to the destination directory, and then build the `.otio` file with local relative path references into that directory.
+
+## OTIOZ
+
+The OTIOD adapter will build a bundle into a zipfile (using the zipfile library).
```
ssteinbach: Add some notes about the compression strategies (that media is uncompressed vs the timeline is compressed)

ssteinbach: Also OTIOD on this line is a typo

src/py-opentimelineio/opentimelineio/adapters/otiod.py
apetrynet: Could quite simply be done like this as well:

``` python
try:
    os.link(src, dst)

except (OSError, AttributeError):
    # OSError - Perhaps file resides on another storage device
    # AttributeError - Windows does not support hard links and thus don't supply os.link()
    shutil.copyfile(src, dst)

```

src/py-opentimelineio/opentimelineio/adapters/file_bundle_utils.py
```diff
@@ -0,0 +1,138 @@
...
+                    name=md.name,
+                    metadata={
+                        adapter_name: {
+                            'original_reference': md,
+                            'missing_reference_because': (
+                                "target_url does not start with 'file://'"
+                            )
+                        }
+                    }
+                )
```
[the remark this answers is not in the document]
ssteinbach: @reinecke what about something like this?

```            if media_policy is MediaReferencePolicy.MissingIfNotFile:
                md = copy.deepcopy(cl.media_reference)
                cl.media_reference = schema.MissingReference()
                cl.media_reference.__dict__ = md.__dict__
                cl.media_reference.metadata['missing_reference_because'] = (
                    "target_url is not a file scheme url (start with url:)"
                )
                cl.media_reference.media_reference['original_target_url'] = (
                    md.target_url
                )
                continue
```

src/py-opentimelineio/opentimelineio/adapters/file_bundle_utils.py
```diff
@@ -0,0 +1,138 @@
...
+
+    referenced_files = set()
+
+    for cl in input_otio.each_clip():
+        try:
+            target_url = cl.media_reference.target_url
+        except AttributeError:
+            continue
+
+        if not target_url.startswith("file://"):
```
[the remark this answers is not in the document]
ssteinbach: that is way better, good call.

src/py-opentimelineio/opentimelineio/adapters/otiod.py
ssteinbach: I like the idea but I'd want to make this an enhancement on the basic thing rather than do this out of the gate.  Lets keep it simple for now and see what kinds of options people need/want.  What you're proposing should be straightforward to add in a later PR.  Thanks!

src/py-opentimelineio/opentimelineio/adapters/otioz.py
ssteinbach: turns out `ZIP_LZMA` is python 3 only, so we're using regular zip compression for now.

docs/tutorials/otio-filebundles.md
```diff
@@ -4,6 +4,18 @@
...
+
+For creating otio bundles, an OTIO file is used as input, whose media references are composed only of `ExternalReference` that have a target_url field pointing at a media file with a unique basename, because file bundles have a flat namespace for media.  For example, if there are media references that point at:
+
+`/project_a/academy_leader.mov`
+
+and:
+
+`/project_b/academy_leader.mov`
+
+Because the basename of both files is `academy_leader.mov`, this will be an error.  The adapters have different policies for how to handle media references.  See below for more information.
```
meshula: Perhaps a otioconvert could have an option to flatten, by substituting an escaped sequence for the slash?

ssteinbach: You mean flatten the file name in?

meshula: yes, something like project_b___academy_leader.mov

ssteinbach: Yeah, I see what you're saying.  But we might also want to preserve the folder hierarchy or make some other guidelines for how files get added to the repo.  Also, this might imply that when the file is unpacked, the files would get put back somewhere new, or that the names won't be legible.  Since this adapter is speculative (at the moment), I'm inclined to keep the current (simple) restriction until we hear from use cases that call for something else.  Both this kind of a scheme and the idea of having hierarchy under the bundle wouldn't be too challenging to build.

ssteinbach:
We'll version in the OTIO file rather than come up with a separate versioning scheme specifically for otioz.

docs/tutorials/otio-filebundles.md
meshula: Works for me.

ssteinbach:
Rebased onto current master branch.

src/py-opentimelineio/opentimelineio/adapters/file_bundle_utils.py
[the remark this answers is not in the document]
ssteinbach: Just a ping on this, @reinecke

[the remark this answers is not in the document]
reinecke: That looks really clear. I like it!

ssteinbach:
Because we don't have a concrete need for this at the moment, and there is concern that this might expand the scope of OTIO (because this is a file format that includes media), we've decided to close this for now.  The branch will remain in my fork so that if there is a desire for this in the future, this work can be brought forward and re-applied.  Please let us know if you have any questions!

jminor:
I came across this SMPTE spec for AXF which attempts to address a similar problem:
https://ieeexplore.ieee.org/stamp/stamp.jsp?tp=&arnumber=7879152

It looks like AXF attempts to address issues of very large data sets, as well as a bunch of extra metadata about each item in the archive.

I'm just posting this here for future reference, since OTIOZ is likely to come up again in the future.

ssteinbach:
Following discussion in the community, in particular some use cases surfaced, this PR has been reopened, rebased and updated to pass tests on all platforms.

meshula (approved):
I think this is a good starting point, and that we should consider landing this as a live proof of concept/RFC.

src/py-opentimelineio/opentimelineio/adapters/otiod.py
```diff
@@ -0,0 +1,149 @@
...
+        return result
+
+    for cl in result.each_clip():
+        try:
+            source_fpath = cl.media_reference.target_url
+        except AttributeError:
+            continue
+
+        rel_path = urlparse.urlparse(source_fpath).path
+        new_fpath = "file:{}".format(os.path.join(filepath, rel_path))
```
reinecke, line 69: In light of some recent discussions we've had about what the form of `target_url` should be, since these are built as relative URLs, maybe we can drop the `file:` off the front to match the types of relative URLs you'd use in HTML.
so instead of:
`file:media/MyMovieFile.mp4`

It comes out:
`media/MyMovieFile.mp4`

reinecke (changes requested):
Looks really great, one small nitpick.

meshula (approved):
using six here is nice

reinecke (approved):

ssteinbach:
Thank you to the community for the thorough review.

PR #577, opened 2019-09-18
Fix warnings for Visual Studio 2019

darbyjohnston:
These changes fix all warnings when compiling with Visual Studio 2019. Note that I bumped PyBind11 to C++14 on Windows, since it looks like C++11 isn't supported by PyBind11 on Windows (I was getting "/std:c++11" argument not recognized warnings when compiling the Python bindings):

https://pybind11.readthedocs.io/en/stable/compiling.html#configuration-variables

### DIFF
--- a/src/py-opentimelineio/opentimelineio-bindings/otio_utils.cpp
+++ b/src/py-opentimelineio/opentimelineio-bindings/otio_utils.cpp
@@ -30,20 +30,12 @@ py::object plain_string(std::string const& s) {
 }
 
 py::object plain_int(int i) {
-    #if PY_MAJOR_VERSION >= 3
-        PyObject *p = PyLong_FromLong(i);
-    #else
-        PyObject *p = PyInt_FromLong(i);
-    #endif
+    PyObject *p = PyLong_FromLong(i);
     return py::reinterpret_steal<py::object>(p);
 }
 
 py::object plain_int(int64_t i) {
-    #if PY_MAJOR_VERSION >= 3
-        PyObject *p = PyLong_FromLong(i);
-    #else
-        PyObject *p = PyInt_FromLong(i);
-    #endif
+    PyObject *p = PyLong_FromLongLong(i);
     return py::reinterpret_steal<py::object>(p);
 }
 
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -2,7 +2,7 @@ cmake_minimum_required(VERSION 2.8.12)
 
 set(CMAKE_CXX_STANDARD 11)
 if(MSVC)
-	set(PYBIND11_CPP_STANDARD /std:c++11)
+	set(PYBIND11_CPP_STANDARD /std:c++14)
 	set(CMAKE_CXX_FLAGS "/W4 /EHsc")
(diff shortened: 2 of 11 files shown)


darbyjohnston:
Drat, looks like these changes are what is causing the Travis CI Python 2.7 build to fail...

meshula:
The /std flag only supports specifying c++14 and C++17. https://docs.microsoft.com/en-us/cpp/build/reference/std-specify-language-standard-version?view=vs-2019 

Pretty sure MSVC has been C++11 compliant since VS2012.

Perhaps simply deleting the /std line completely resolves the issue?

darbyjohnston:
Yeah, I realized that and changed it to C++14 though unfortunately that's not what is causing the Travis CI failure. I'm going to try backing out the changes one by one to try and pinpoint what's causing the build breakage.

darbyjohnston:
OK, it looks like the problem was here (otio_utils.cpp):

    #if PY_MAJOR_VERSION >= 3
        PyObject *p = PyLong_FromLongLong(i);
    #else
        PyObject *p = PyInt_FromLongLong(i);
    #endif

There is no function PyInt_FromLongLong() so it was causing a compile error. My bad for not testing with Python 2, though I'm surprised the output from the Travis CI build wasn't more helpful. :)

I think the solution is to use PyLong_FromLongLong() which fixes the Travis CI build and doesn't cause any warnings.

src/py-opentimelineio/opentimelineio-bindings/otio_utils.cpp
```diff
@@ -40,9 +40,9 @@ py::object plain_int(int i) {
 
 py::object plain_int(int64_t i) {
     #if PY_MAJOR_VERSION >= 3
```
meshula: looking good! can you remove the version check as it is now moot?

darbyjohnston: Good point; this is maybe getting a bit off track but the other plain_int(int) function above could probably also use the same change since apparently PyInt and PyLong seem to be interchangeable?
https://www.python.org/dev/peps/pep-0237/

meshula: yes, that would be consistent. I think that would put a nice bow on this commit, and then I can go ahead and merge. Thanks for the detective work on the merge failure, that was mighty mysterious given the lack of clear output from CI.

ssteinbach: Sorry about tox eating the message like that.  We've been talking offline about what we might be able to do to improve the CI results.

darbyjohnston:
OK, all set. Hopefully I got the git rebase correct, I notice that this pull request has some of Stephan's commits which I guess came from when I merged upstream?

darbyjohnston:
woot! :)
OTIO VS2019

I wonder how difficult it would be to enable Windows for Travis CI...

jminor:
Yay!

FYI, when we switch to the ASWF CI system (Azure) we'll get Windows, Mac, and Linux CI.

PR #584, opened 2019-09-24
RV application plugin updates

avrata:
Small updates I missed in #565

### DIFF
--- a/contrib/opentimelineio_contrib/application_plugins/rv/example_otio_reader/otio_reader.py
+++ b/contrib/opentimelineio_contrib/application_plugins/rv/example_otio_reader/otio_reader.py
@@ -1,0 +1,263 @@
+#
+# Copyright 2019 Pixar Animation Studios
+#
+# Licensed under the Apache License, Version 2.0 (the "Apache License")
+# with the following modification; you may not use this file except in
+# compliance with the Apache License and the following modification to it:
+# Section 6. Trademarks. is deleted and replaced with:
+#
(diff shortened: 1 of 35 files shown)


contrib/opentimelineio_contrib/application_plugins/rv/example_otio_reader/otio_reader.py
```diff
@@ -231,13 +230,17 @@ def _create_item(it, track_kind=None):
...
-            range_to_read.end_time_inclusive().value,
-            range_to_read.duration.rate)]
+        new_media = [_create_movieproc(range_to_read, kind)]
+
+    try:
+        src = commands.addSourceVerbose(new_media)
+    except Exception as e:
+        # Perhaps the media was missing, if so, lets load an error
+        # source
+        print(e)
```
ssteinbach: just a naive RV question, does this need to be printed through some kind of error handler so that RV knows its an error?

avrata: Ok, From what I can tell if I have "ERROR" in the string, the RV "Console" will raise. This seems to be the pattern in other python plugins that ship with RV. Updated code for that.

ssteinbach:
How does this look, @JoshBurnell @eric-desruisseaux-adsk?

@JoshBurnell can you rebase your writer plugin onto this once we land this PR?  

Thanks!

PR #585, opened 2019-09-25
Fix tests for Visual Studio 2019

darbyjohnston:
A number of the core tests (OTIO_DISABLE_SHELLOUT_TESTS=true) were failing on Windows, this pull request has two changes to fix that.

Tests were crashing on exit with the error message, "d:\agent\_work\2\s\src\vctools\crt\crtw32\stdcpp\thr\mutex.c(64): mutex destroyed while busy". To fix this I unlocked the mutex in serializableObject.cpp before it is deleted.

Tests with temporary files were failing since the C++ code could not reopen the files opened by Python (this is mentioned as a caveat in the docs for tempfile.NamedTemporaryFile()). Since the Python tests don't actually need the file object, just the file name, I replaced using NamedTemporaryFile() with temporary directories that Python then passes to C++ along with a file name.

### DIFF
--- a/src/opentimelineio/serializableObject.cpp
+++ b/src/opentimelineio/serializableObject.cpp
@@ -133,6 +133,7 @@ void SerializableObject::_managed_release() {
     _mutex.lock();
 
     if (--_managed_ref_count == 0) {
+        _mutex.unlock();
         delete this;
         return;
     }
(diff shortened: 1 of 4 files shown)


darbyjohnston:
PS. I have some separate changes to actually run the tests on Windows, but since that is all CMake stuff it seemed better to separate it from the bug fixes.

darbyjohnston:
Whoops, looks like Linux didn't like my changes too much, taking a look...

Issue #587, opened 2019-09-25
AAF Reader: Ignoring Essence Locators in SourceMobs


JoshBurnell:
Hey, gang,

We're attempting to use the AAF Adapter to convert our composition AAF's to OTIO's, however the resulting OTIO doesn't contain references to much of the media. We do get some of it in the MasterMob metadata (#551), but we're not getting any of the media that was contained in SourceMob Essences.

The [advanced_authoring_format.py](https://github.com/PixarAnimationStudios/OpenTimelineIO/blob/master/contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py) format does manage to get the relevant MasterMob, but it doesn't dissect it.

This may be related to #482 in some way. This also may be related to how we're exportting our AAF's from AVID, so if someone has already worked through this issue, lemme know.

I have a PR that I'll attach after I have this Issue number.

JoshBurnell:
Here's how I've fixed the issue in the short-term (#588), but I wonder if the [advanced_authoring_format.py](https://github.com/PixarAnimationStudios/OpenTimelineIO/blob/master/contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py) adapter shouldn't have a more comprehensive fix for SourceMobs and Essences, either a sub-def that it feeds Essences into, or a part of the ```_transcribe()``` decision tree that handles MasterMobs with SourceMob slots and creates a MediaReference.

Issue #551, opened 2019-08-06
AAF Writer: Provide media references to UNC Paths


JoshBurnell:
When using the AAF adapter to create an OTIO from an AAF, the resulting OTIO Clips contain MissingReference schemas that have a "UNC Path" path in the "UserComments."

```
"media_reference": {
    "OTIO_SCHEMA": "MissingReference.1",
    "metadata": {
    "AAF": {
        "MobID": "64d3faed",
        "Name": "2",
        "UserComments": {
            "Name": "1",
            "UNC Path": "\\tmp\\jburnell\\500_2349\\2.aaf"
        }
    }
}
```

It would be great if the AAF adapter created an ExternalReference instead that references the "UNC Path."

jminor:
Addressed by #560 
Thanks @JoshBurnell !

Issue #482, opened 2019-04-08
AAF Reader: Sequence with nested scopes unable to be reconstructed


andrewmoore-nz:
Now that we are able to process an AAF containing nested sequences and start analysing the resulting OTIO, I've come across this first issue, which is a pretty big one really. 

The clip inside the sequence that has been exported simply has the following information:

```
"OTIO_SCHEMA": "Clip.1", 
"effects": [], 
"markers": [], 
"media_reference": {
    "OTIO_SCHEMA": "MissingReference.1", 
    "available_range": null, 
    "metadata": {}, 
    "name": null
}, 
"metadata": {
    "AAF": {
        "ClassName": "SourceClip", 
        "DataDefinition": {
            "Description": "Picture data", 
            "Identification": "01030202-0100-0000-060e-2b3404010101", 
            "Name": "DataDef_Picture"
        }, 
        "Length": 120, 
        "Name": "SourceClip Missing Mob?", 
        "SourceID": "urn:smpte:umid:00000000.00000000.00000000.00000000.00000000.00000000.00000000.00000000", 
        "SourceMobSlotID": 0, 
        "StartTime": 0
    }
}, 
"name": "SourceClip Missing Mob?", 
"source_range": null
```

There is clearly an issue, what with the name showing up as `"SourceClip Missing Mob?"` so worth looking in to what's happening there. Inspecting the OTIO object itself, the two nested sequences in question are present, existing as their own `Timeline` structures inside a top-level `SerializableCollection` structure, which would make it possible to parse and recreate, but without any info in the main timeline's clip, this obviously isn't able to be done.

Given the way that this nested structure is exported out to AAF, is it possible for it to be treated the same way as with a nested XML e.g. each nested scope existing as a `Stack` on a `Track` and having it's own internal hierarchy going down the levels?

Zip attached with example AAF & XML: 

[SpeedChange_MstrSeq_withNesting_Examples_190408.zip](https://github.com/PixarAnimationStudios/OpenTimelineIO/files/3052794/SpeedChange_MstrSeq_withNesting_Examples_190408.zip)

freesonluxo:
Confirming that I can repro this error.

I don't believe that this is related to nesting or having a nested scope. It looks like the raw AAF file itself has three clips in the CompositionMob that have no Mob. So, from that perspective, the code is doing the right thing - the three clips themselves have no Mob, so OTIO says so and moves on.

[...]

PR #588, opened 2019-09-25
Extract target_url from Essence locator in SourceMob

JoshBurnell:
From Issue #587 ...

We're attempting to use the AAF Adapter to convert our composition AAF's to OTIO's, however the resulting OTIO doesn't contain references to much of the media. We do get some of it in the MasterMob metadata (#551), but we're not getting any of the media that was contained in SourceMob Essences.

The advanced_authoring_format.py format does manage to get the relevant MasterMob, but it doesn't dissect it.

This may be related to Issue #482 in some way. This also may be related to how we're exportting our AAF's from AVID, so if someone has already worked through this issue, lemme know.


jminor:
@JoshBurnell here's the existing AAF samples, which all get tested when the unit tests run: https://github.com/PixarAnimationStudios/OpenTimelineIO/tree/master/contrib/opentimelineio_contrib/adapters/tests/sample_data
That's a good place to add a new one to cover this PR.

JoshBurnell:
Updated with test AAF and unittest (and adapters/tests/__init__.py for convenience)

contrib/opentimelineio_contrib/adapters/tests/test_aaf_adapter.py
```diff
@@ -848,6 +852,20 @@ def test_external_reference_from_unc_path(self):
             unc_path
         )
 
+    def test_external_reference_paths(self):
+        timeline = otio.adapters.read_from_file(COMPOSITE_PATH)
+        for video_track in timeline.video_tracks():
+            for clip in video_track:
+                self.assertIsInstance(clip.media_reference,
+                                      otio.schema.ExternalReference)
+                self.assertIsNotNone(clip.media_reference.target_url)
```
jminor: If you compare the actual contents of `target_url` then this test will cover the handling of backslashes as well. I have a slight worry that there might be a platform dependency on the path separator in these UNC paths, so it would be good to check that.

jminor: @JoshBurnell would you mind adding a check for the correct `target_url` here?

Issue #590, opened 2019-09-26
media_linker.load_manifest clobbers local Adapters of similar suffix


JoshBurnell:
For testing (and customizability), it would be useful to supersede the Adapters in the OpenTimelineIO distribution with local adapters of the same suffix.

media_linker.load_manifest doesn't allow this, because it processes the Adapter manifests in this order:
1. opentimelineio/adapters/builtin_adapters.plugin_manifest.json
2. opentimelineio_contrib/adapters/contrib_adapters.plugin_manifest.json
3. $OTIO_PLUGIN_MANIFEST_PATH

It would allow more customizability to process $OTIO_PLUGIN_MANIFEST_PATH first.
1. $OTIO_PLUGIN_MANIFEST_PATH
2. opentimelineio/adapters/builtin_adapters.plugin_manifest.json
3. opentimelineio_contrib/adapters/contrib_adapters.plugin_manifest.json

JoshBurnell:
Submitted #591 to resolve.

ssteinbach:
Fixed by #591

PR #591, opened 2019-09-26
#590: Process local adapter manifests before internal manifests

JoshBurnell:
Reordered the processing of adapter manifests. Also fixed bug in Manifest.extend() which missed hook types if they weren't included at Manifest creation.


tests/test_adapter_plugin.py
```diff
@@ -221,6 +221,54 @@ def test_find_manifest_by_environment_variable(self):
...
+            # Rather than try and remove any other setuptools based plugins
+            # that might be installed, this check is made more permissive to
+            # see if the known unit test linker is being loaded by the manifest
+            self.assertTrue(len(result.adapters) > 0)
+            self.assertIn("otio_json", (ml.name for ml in result.adapters))
+            self.assertIn("local_json", (ml.name for ml in result.adapters))
+            self.assertLess([ml.name for ml in result.adapters].index("local_json"),
+                            [ml.name for ml in result.adapters].index("otio_json"))
+
+
```
ssteinbach:
```suggestion
```

ssteinbach: I think deleting this line will make the lint pass.

src/py-opentimelineio/opentimelineio/plugins/manifest.py
```diff
@@ -130,8 +130,9 @@ def extend(self, another_manifest):
             self.hook_scripts.extend(another_manifest.hook_scripts)
 
             for trigger_name, hooks in another_manifest.hooks.items():
-                if trigger_name in self.hooks:
-                    self.hooks[trigger_name].extend(hooks)
+                if trigger_name not in self.hooks:
+                    self.hooks[trigger_name] = []
+                self.hooks[trigger_name].extend(hooks)
```
ssteinbach: oof good catch

tests/test_adapter_plugin.py
```diff
@@ -221,6 +221,54 @@ def test_find_manifest_by_environment_variable(self):
...
+            otio.adapters.write_to_file(local_manifest, otio_path.name, 'otio_json')
+
+            result = otio.plugins.manifest.load_manifest()
+            self.assertTrue(len(result.adapters) > 0)
+            self.assertIn("otio_json", (ml.name for ml in result.adapters))
+            self.assertNotIn("local_otio", (ml.name for ml in result.adapters))
+
+            # set where to find the new manifest
+            os.environ['OTIO_PLUGIN_MANIFEST_PATH'] = otio_path.name
+            result = otio.plugins.manifest.load_manifest()
```
ssteinbach: This unit test is great.  But a note to self is that it might be nice to have a way to at run time add manifests on the path rather than doing it by manipulating the environment variable and calling `load_manifest()`.

ssteinbach:
It looks like there is a conflict in manifest.py.  Can you rebase on top of the latest master to resolve the conflict?  Sorry about that.  I suspect that is also why the unit test suite hasn't been run on this.

JoshBurnell:
Resolved.

JoshBurnell:
Or maybe not... hang on.

JoshBurnell:
There we go.

ssteinbach:
Great catch @JoshBurnell, thanks for fixing this!

Issue #592, opened 2019-09-27
Markers: AttributeError when calling __repr__ / __str__


elabrosseRodeofx:
The `otio.schema.Marker` object's `__repr__` and `__str__` methods are referencing a 'media_reference' and a 'source_range' attribute.
however Markers do not have those attributes. They do however have a 'marked_range' attribute that is not referenced in the `__repr__` / `__str__`

Printing a Marker instance results in an AttributeError:
```
>>> otio.schema.Marker()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/devLocal/elabrosse/rdoenv/auto_lineup/otio/lib/python2.7/site-packages/opentimelineio/schema/marker.py", line 25, in __repr__
    repr(self.media_reference),
AttributeError: 'opentimelineio._otio.Marker' object has no attribute 'media_reference'
```

From what I see it seems that the content of opentimelineio/schema/clip.py was copy/pasted almost as is in opentimelineio/schema/marker.py.

My corporation is waiting for confirmation that our CLA was received. We will submit a PR as soon as possible.

jminor:
Yep, we got the CLA. Thanks for finding this oversight.

elabrosseRodeofx:
Made a PR:

https://github.com/PixarAnimationStudios/OpenTimelineIO/pull/596

reinecke:
I accidentally resolved this in #740 after it'd already been resolved in #596 (Oops!)

PR #596, opened 2019-09-27
[RodeoFX] Fix Marker __repr__ and __str__ #592

elabrosseRodeofx:
Fixes: #592

This PR fixes an AttributeError when calling
`repr(otio.schema.Marker())` or `str(otio.schema.Marker())`

Without the fix:
```
>>> repr(otio.schema.Marker())
  File "/devLocal/elabrosse/rdoenv/auto_lineup/otio/lib/python2.7/site-packages/opentimelineio/schema/marker.py", line 25, in __repr__
    repr(self.media_reference),
AttributeError: 'opentimelineio._otio.Marker' object has no attribute 'media_reference'

>>> str(otio.schema.Marker())
  File "/devLocal/elabrosse/rdoenv/auto_lineup/otio/lib/python2.7/site-packages/opentimelineio/schema/marker.py", line 9, in __str__
    self.media_reference,
AttributeError: 'opentimelineio._otio.Marker' object has no attribute 'media_reference'
```

With the fix:
```
>>> repr(otio.schema.Marker())
"otio.schema.Marker(name='', marked_range=otio.opentime.TimeRange(start_time=otio.opentime.RationalTime(value=0, rate=1), duration=otio.opentime.RationalTime(value=0, rate=1)), metadata={})"

>>> str(otio.schema.Marker())
'Marker("", TimeRange(RationalTime(0, 1), RationalTime(0, 1)), {})'
```

The otio.schema.Marker object's `__repr__` and `__str__` methods are referencing a 'media_reference' and a 'source_range' attribute.
however Markers do not have those attributes. They do however have a 'marked_range' attribute that is not referenced in the `__repr__` / `__str__`

### DIFF
--- a/tests/test_marker.py
+++ b/tests/test_marker.py
@@ -88,6 +88,40 @@ class MarkerTest(unittest.TestCase, otio_test_utils.OTIOAssertions):
         self.assertNotEqual(m, bo)
         self.assertNotEqual(bo, m)
 
+    def test_string(self):
+        tr = otio.opentime.TimeRange(
+            otio.opentime.RationalTime(5, 24),
+            otio.opentime.RationalTime(10, 24)
+        )
+        m = otio.schema.Marker(
+            name="marker_1",
+            marked_range=tr,
+            color=otio.schema.MarkerColor.GREEN,
+            metadata={'foo': 'bar'}
+        )
+
+        self.assertEqual(
+            repr(m),
+            "otio.schema.Marker("
+            "name='marker_1', "
+            "color=u'GREEN', "
+            "marked_range=otio.opentime.TimeRange("
+            "start_time=otio.opentime.RationalTime(value=5, rate=24), "
+            "duration=otio.opentime.RationalTime(value=10, rate=24)"
+            "), "
+            "metadata={'foo': u'bar'}"
+            ")"
+        )
+        self.assertEqual(
+            str(m),
+            'Marker('
+            '"marker_1", '
+            '"GREEN", '
+            'TimeRange(RationalTime(5, 24), RationalTime(10, 24)), '
+            '{\'foo\': u\'bar\'}'
+            ')'
+        )
+
 
 if __name__ == '__main__':
     unittest.main()
--- a/src/py-opentimelineio/opentimelineio/schema/marker.py
+++ b/src/py-opentimelineio/opentimelineio/schema/marker.py
@@ -4,10 +4,9 @@ from .. import _otio
 
 @add_method(_otio.Marker)
 def __str__(self):
-    return 'Marker("{}", {}, {}, {})'.format(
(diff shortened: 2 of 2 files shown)


ssteinbach:
Awesome, thanks!

If you say:
Fixes: #592 
instead of
Issue: #592

Then when this PR is accepted github will automatically close the issue for us, too.

tests/test_marker.py
```diff
@@ -88,6 +88,40 @@ def test_equality(self):
...
+            marked_range=tr,
+            color=otio.schema.MarkerColor.GREEN,
+            metadata={'foo': 'bar'}
+        )
+
+        self.assertEqual(
+            repr(m),
+            "otio.schema.Marker("
+            "name='marker_1', "
+            "color=u'GREEN', "
```
jminor: It looks like the Python 3 tests are failing because there's a subtle difference in how unicode is handled between Python 2 and 3. You can try just removing the `u` from this line to see if that avoids the issue.

jminor: @reinecke do you have any advice about how to make this work in both Python 2 and 3? Marker colors will likely stay as plain ASCII, but metadata should allow for unicode characters.

jminor: This might help illuminate the different handling of str/unicode in Python 2 vs 3: https://portingguide.readthedocs.io/en/latest/strings.html

reinecke: My current guess on this is that it has to do with the marshaling behavior in pybind 11: https://pybind11.readthedocs.io/en/stable/advanced/cast/strings.html
Given that, it appears all strings should become unicode upon transfer back into python.

So, for python 2, it seems like we should at least be able to rely on the `u` being added consistently? Unfortunately I can't think of a clean way of branching the literals we compare based on the two python versions.

Would it make sense to just include two copies of the `repr` unittests (one marked skip in python3 and one marked skip in python2)? Then when python2 support is dropped we can remove the tests marked skip in python3 and delete the skip annotations.

An alternate approach could be to have a dedicated custom assertion that when run in python2 compares against a "patched" version of the string using something like `comp_string.replace("u'", "'").replace('u"', '"')`, but that feels pretty sketchy. The upside is that it would be a no-op in the python3 use case and the assertion could be find and replaced upon dropping python 2 support.

For reference, here is a good way to mark the unittests:
```
IS_PYTHON_2 = (sys.version_info < (3, 0))

@unittest.skipIf(IS_PYTHON_2, "In python2 string values become u-strings.")
def test_repr():
    ...do stuff...

@unittest.skipIf(not IS_PYTHON_2, "In python2 string values become u-strings.")
def test_repr_py2():
    ...do stuff...
```

reinecke:
Accidentally re-resolved in #740 (even though this was the original fix).

PR #604, opened 2019-10-23
doc: Update feature matrix

thiblahute:
Updated with the data from last SIGGRAPH talk (updating GStreamer only)


docs/tutorials/feature-matrix.rst
```diff
@@ -3,27 +3,28 @@ Feature Matrix
...
++==========================+==========+=======+========+========+=======+========+=========+====+
+|Single Track of Clips     |    ✔   |    ✔    |    ✔   |   ✔    |   ✔   | W-O    |    ✔    | ✔  |
++--------------------------+--------+---------+--------+--------+-------+--------+---------+----+
+|Multiple Video Tracks     |    ✔   |    ✖    |    ✔   |   ✔    |   ✔   | W-O    |flattened| ✔  |
++--------------------------+--------+---------+--------+--------+-------+--------+---------+----+
+|Audio Tracks & Clips      |    ✔   | testing |    ✔   |   ✔    |   ✔   | W-O    |    ✔    | ✔  |
++--------------------------+--------+---------+--------+--------+-------+--------+---------+----+
+|Gap/Filler                |    ✔   |    ✔    |    ✔   |   ✔    |   ✔   |   ✔    |skipped  | ✔  |
++--------------------------+--------+---------+--------+--------+-------+--------+---------+----+
+|Markers                   |    ✔   |    ✔    |    ✔   |contrib?|planned|contrib?|   ✖     | ✔  |
```
ssteinbach: @eric-with-a-c do you know if the FCP-X adapter supports markers as well (I see it mentioned in the code)?  And RV adapter does not support markers, but there is an open issue for that (I think we should just put an X in that box).

docs/tutorials/feature-matrix.rst
```diff
@@ -3,27 +3,28 @@ Feature Matrix
...
++--------------------------+--------+---------+--------+-----------+-------------+---------+----+
+|Nesting                   |    ✔   |    ✖    |    ✔   |   ✔    |   ✔   |   W-O  |flattened| ✔  |
++--------------------------+------------------+--------+--------+-------+--------+---------+----+
+|Transitions               |    ✔   |    ✔    |contrib?|contrib?|   ✔   |   W-O  |   ✖     | ✔  |
```
ssteinbach: Looks like FCPX doesn't have transition support (I Don't see any mention of them in the code).  Looks like FCP7 XML does have transition support based on the code.

docs/tutorials/feature-matrix.rst
```diff
@@ -3,27 +3,28 @@ Feature Matrix
...
++--------------------------+----------+-------+--------+--------+-------+--------+---------+----+
+|Feature                   | OTIO   | EDL     |FCP7 XML| FCP X  | AAF   |   RV   | ALE     |Gst |
++==========================+==========+=======+========+========+=======+========+=========+====+
+|Single Track of Clips     |    ✔   |    ✔    |    ✔   |   ✔    |   ✔   | W-O    |    ✔    | ✔  |
++--------------------------+--------+---------+--------+--------+-------+--------+---------+----+
+|Multiple Video Tracks     |    ✔   |    ✖    |    ✔   |   ✔    |   ✔   | W-O    |flattened| ✔  |
++--------------------------+--------+---------+--------+--------+-------+--------+---------+----+
+|Audio Tracks & Clips      |    ✔   | testing |    ✔   |   ✔    |   ✔   | W-O    |    ✔    | ✔  |
++--------------------------+--------+---------+--------+--------+-------+--------+---------+----+
+|Gap/Filler                |    ✔   |    ✔    |    ✔   |   ✔    |   ✔   |   ✔    |skipped  | ✔  |
```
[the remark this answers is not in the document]
jminor: If I remember correctly, "contrib?" in that slide meant that we were seeking contributions to add support for those things. This relates to @thiblahute's question about what we meant by "planned".

I think we should use a simple notation that clarifies which features are not yet implemented versus what features are not supported at all by the particular format (someday vs never?). Anything more nuanced than that probably belongs in a separate adapter-specific page.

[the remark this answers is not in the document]
thiblahute: Agreed, what symbols do you suggest? (I have no good idea and would rather avoid bikeshedding :))

docs/tutorials/feature-matrix.rst
```diff
@@ -3,27 +3,33 @@ Feature Matrix
...
-+--------------------------+----------+----------------+----------+-----------+------------+-----------+-----------+
-|Speed Effects             | yes      | yes            | planned  | read-only | planned    | no        | planned   |
-+--------------------------+----------+----------------+----------+-----------+------------+-----------+-----------+
-|Color Decision List (CDL) | metadata | yes            | planned  | planned   | planned    | planned   | no        |
-+--------------------------+----------+----------------+----------+-----------+------------+-----------+-----------+
\ No newline at end of file
+* flattened
+
+N/A: Not Applicable
+W-O: Write Only
```
jminor: There's one "RO" in the matrix. Could you make it "R-O" to match, and add it to the legend at the bottom. The rest looks great!

PR #605, opened 2019-10-23
Adding support for negative time strings when converting a negative

stefanschulze:
RationalTime object to a string.

Before this change, doing

negativeRationalTime = otio.opentime.RationalTime(-24,24)
timeString = otio.opentime.to_time_string(negativeRationalTime)

timeString would be "-1:-1:-10", which is not a useful time string.

This change introduces a 'negative' time string, i.e. generate the
same string value as a positive RationalTime value would do but
with a '-' sign in front of the string:

timeString is now: '-00:00:01.0'

This is actually a valid string format for ffmpeg and useful if you
for example want to add silence to an audio file at the beginning.

### DIFF
--- a/src/opentime/rationalTime.cpp
+++ b/src/opentime/rationalTime.cpp
@@ -282,6 +282,15 @@ RationalTime::to_timecode(
 std::string
 RationalTime::to_time_string() const {
     double total_seconds = to_seconds();
+    bool is_negative = false;
+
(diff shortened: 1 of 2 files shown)


jminor:
@stefanschulze FYI I edited the description above to use backticks so we get `-1:-1:-10` instead of -1:-1:-10 with an emoji in it :)

Issue #606, opened 2019-10-24
to_timecode ignoring RationalTime's rate


splidje:
How to reproduce:

``import opentimelineio as otio``
``time1 = otio.opentime.RationalTime(24.0, 24.0)``
``time2 = otio.opentime.RationalTime(1.0, 1.0)``

time1 and time2 are equivalent, here's proof:
``time1 - time2``
gives:
``otio.opentime.RationalTime(value=0.0, rate=24.0)``

converting time1 to a 24fps timecode:
``otio.opentime.to_timecode(time1, 24.0)``
_correctly_ gives:
``'00:00:01:00'``

however, converting time2 to a 24fps timecode:
``otio.opentime.to_timecode(time2, 24.0)``
_erroneously_ gives:
``'00:00:00:01'``

in fact, it seems the RationalTime's rate is being completely ignored by to_timecode, as converting time1 to a 12fps timecode:
``otio.opentime.to_timecode(time1, 12.0)``
_erroneously_ gives:
``'00:00:02:00'``

reinecke:
Hi @splidje, thanks for filing! This is definitely an issue. I've checked the pip version as well as the C++ version and they both show the bug.
I'm sure you figured it out already, but the workaround is to rescale the `RationalTime` to the frame rate before converting:
`otio.opentime.to_timecode(time2.rescaled_to(24))`
gives:
`'00:00:01:00'`

PR #612, opened 2019-11-06
Timecode rate is ignored

ssteinbach:
Fix: #606 

This addresses the issue as written... but because having an implicit `rescale` in `to_timecode` seems iffy, I'm also proposing another PR that drops the `rate` argument from `to_timecode` entirely in favor of explicitly calling `rescale_to` in calling code.  Thoughts?

### DIFF
--- a/src/opentime/rationalTime.cpp
+++ b/src/opentime/rationalTime.cpp
@@ -193,8 +193,13 @@ RationalTime::to_timecode(
 ) const {
 
     *error_status = ErrorStatus();
+
+    double value_in_target_rate = _value;
+    if (rate != _rate) {
+        value_in_target_rate = this->value_rescaled_to(rate);
+    }
     
-    if (_value < 0) {
+    if (value_in_target_rate < 0) {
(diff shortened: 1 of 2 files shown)


src/opentime/rationalTime.cpp
```diff
@@ -193,8 +193,13 @@ RationalTime::to_timecode(
 ) const {
 
     *error_status = ErrorStatus();
+
+    double value_in_target_rate = _value;
```
reinecke: The goal of this is to convert the time to a frame count. I think the logic would be clearer if rather than using `value` and `rate` from the input `RationalTime` in this method, we instead used `RationalTime::to_frames` to get the frame count and never let the member `_rate` from the `RationalTime` instance enter this function's calculations. This clarifies the distinction between `RationalTime` `rate` (which is an arbitrary time division used to accurately express a time in seconds), and fps (which is a scheme of recording samples).

This is a bit pedantic, but it's akin to the distinction between pixels and little squares.

reinecke: As discussed, since we'll be looking more at this more in #613 we should keep it as-is t limit impact

PR #620, opened 2019-11-19
C++ Error documentation additions

reinecke:
- Added documentation for best practices on schema read errors
- Created a table outlining C++ error types, how they map to python exceptions, and some semantic description

The error chart may or may not be a good idea. I'd also look to others to help fill in the semantic meanings for some of the errors to make sure I don't write something misleading.

This also poses the question about when it's appropriate to add additional errors. If we tackle that, then it'd be good to outline the steps one should take when doing that.

### DIFF
--- a/docs/cxx/cxx.rst
+++ b/docs/cxx/cxx.rst
@@ -127,26 +127,17 @@ When an error is encountered in reading, ``read_from`` should set the error
 on the ``Reader`` instance and return ``false``: ::
 
     bool Marker::read_from(Reader& reader) {
-        std::string color_value;
-        bool return_value = reader.read("color", &color_value);
-        if (!return_value) {
-            return false;
-        }
-
-        if (color_value == "invalid_value") {
-            ErrorStatus error_status = ErrorStatus(ErrorStatus::JSON_PARSE_ERROR,
-                        "invalid_value not allowed for color");
-            reader.error(error_status);
-            return false;
-        )
-        else {
-            _color = color_value;
-        }
-
-        return return_value &&
-            reader.read("marked_range", &_marked_range) &&
-	    Parent::read_from(reader);
-     }
+    if (!reader.read(“color”, &_color)) {
+        return false;
+    }
+    if (_color == “invalid_value”) {
+        reader.error( ErrorStatus(ErrorStatus::JSON_PARSE_ERROR,
+                                                  “invalid_value not allowed for color”));
+        return false;
+    }
+    return reader.read(“marked_range”, &_marked_range) &&
+           Parent::read_from(reader);
+    }
 
 This is a contrived example but it describes the basic mechanics. Adjust the
 details above as appropriate for your case.
@@ -187,7 +178,7 @@ Creating and manipulating schema objects is also simple: ::
              ...
         }
     }
-    
+
 
 Serializable Data
 +++++++++++++++++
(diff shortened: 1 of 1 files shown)


docs/cxx/cxx.rst
```diff
@@ -123,6 +123,34 @@ well: ::
 Even when we define more complex properties, the reading/writing code is as
 simple as shown above, in almost all cases.
 
+When an error is encountered in reading, ``read_from`` should set the error
+on on the ``Reader`` instance and return ``false``: ::
```
jminor:
```suggestion
on the ``Reader`` instance and return ``false``: ::
```

docs/cxx/cxx.rst
```diff
@@ -550,6 +578,52 @@ to allow constructors to "fail" gracefully.  Accordingly, a class like
...
+    :header: "Value", "Python Exception Type", "Meaning"
+    
+    OK, n/a, No Error
+    INVALID_TIMECODE_RATE, ``ValueError``, "Timecode rate isn't a valid SMPTE rate"
+    NON_DROPFRAME_RATE,  ``ValueError``, "Timecode rate isn't valid for SMPTE Drop-Frame Timecode"
+    INVALID_TIMECODE_STRING,  ``ValueError``, "String is not properly formatted SMPTE timecode string"
+    TIMECODE_RATE_MISMATCH,  ``ValueError``, " Timecode string has a frame number higher than the frame rate"
+    INVALID_TIME_STRING,  ``ValueError``,
+    NEGATIVE_VALUE,  ``ValueError``,
+    INVALID_RATE_FOR_DROP_FRAME_TIMECODE,  ``ValueError``, "Timecode rate isn't valid for SMPTE non Drop-Frame Timecode"
```
jminor:
```suggestion
    INVALID_RATE_FOR_DROP_FRAME_TIMECODE,  ``ValueError``, "Timecode rate isn't valid for SMPTE Drop-Frame Timecode"
```

Also, I think `INVALID_RATE_FOR_DROP_FRAME_TIMECODE` and `NON_DROPFRAME_RATE` are basically the same. Perhaps independent from this change, we could consolidate them.

docs/cxx/cxx.rst
```diff
@@ -550,6 +578,52 @@ to allow constructors to "fail" gracefully.  Accordingly, a class like
...
+.. csv-table:: OpenTimelineIO error codes
+   :header: "Value", "Python Exception Type", "Meaning"
+   
+    OK, n/a, No Error
+    NOT_IMPLEMENTED, ``NotImplementedError``, "A feature is known but deliberately unimplemented"
+    UNRESOLVED_OBJECT_REFERENCE, ``ValueError``, "An object reference is unresolved while reading"
+    DUPLICATE_OBJECT_REFERENCE, ``ValueError``, "An object reference is duplicated while reading"
+    MALFORMED_SCHEMA, ``ValueError``, "The Schema string was invalid"
+    JSON_PARSE_ERROR, ``ValueError``, "Malformed JSON encountered when parsing"
+    CHILD_ALREADY_PARENTED, ``ValueError``, "A child was attempted to be added to a collection when it's already a member of another collection instance"
```
jminor:
```suggestion
    CHILD_ALREADY_PARENTED, ``ValueError``, "Attempted to add a child to a collection when it's already a member of another collection instance"
```

docs/cxx/cxx.rst
reinecke: Good suggestion, filed issue #625 to track this.

reinecke:
Removing draft status on this one. There could be more done here, but it should still add value

reinecke:
Good call @davidbaraff, I like your version much better.

PR #621, opened 2019-11-21
Unifies ALE to the EDL adapter by storing the cdl data in the metadata

vvzen:
Minor fix to the ALE adapter so that it stores the cdl data in the `cdl` property of the metadata dictionary, in the same way that the EDL one does.
It also adds one more ALE sample and related unittest.

Related issue: https://github.com/PixarAnimationStudios/OpenTimelineIO/issues/619


contrib/opentimelineio_contrib/adapters/tests/test_ale_adapter.py
```diff
@@ -98,6 +99,37 @@ def test_ale_read2(self):
             ]
         )
 
+    def test_ale_read3(self):
```
reinecke: Would you mind adding a check in the tests to make sure the CDL data comes through correctly?

vvzen: Thanks God you actually asked me to do that! :D
I found that I had swapped the offset and power values.
I added the cdl test. Let me know.

contrib/opentimelineio_contrib/adapters/ale.py
```diff
@@ -126,6 +154,28 @@ def _parse_data_line(line, columns, fps):
             line, repr(ex)
         ))
 
+def _cdl_values_from_metadata(asc_sop_string):
+
+    if not isinstance(asc_sop_string, (str, unicode)):
```
reinecke: `unicode` and `str` are just merged to `str` in python3 so the `isinstance` check is problematic there. In what cases will this be a non-string type? Perhaps we can get away with something like an `is None` check here?

vvzen: Oh I see!
This is basically just me trying to do type checking in python, I don't know in which cases this value will not be a string, but I have no guarantee that it will always be - that's why I put that check in place as a "defence" mechanism - the idea being that the sooner and you raise an Exception, the better (and also explicit is always better than implicit).
I guess it could become a check for None, but it would also become a little bit less explicit.
The alternative would be adding a check like this  https://stackoverflow.com/a/32632271 ?
If we feel safe enough about `metadata['ASC_SOP']` and `metadata['CDL']` always being strings or `None`, we could just actually check for `None`. What do you think?

PS: linting fixed!

contrib/opentimelineio_contrib/adapters/ale.py
```diff
@@ -116,6 +117,33 @@ def _parse_data_line(line, columns, fps):
                 target_url=source
             )
 
+        # If available, collect cdl values in the same way we do for CMX EDL
+        cdl = {}
+
+        if 'CDL' in metadata:
+            cdl = _cdl_values_from_metadata(metadata['CDL'])
+            if cdl:
+                metadata.pop('CDL')
```
jminor: It is slightly more pythonic to use `del metadata['CDL']` instead of `pop`

contrib/opentimelineio_contrib/adapters/ale.py
```diff
@@ -116,6 +117,33 @@ def _parse_data_line(line, columns, fps):
...
+
+        # If we have more specific metadata, let's use them
+        if 'ASC_SOP' in metadata:
+            cdl = _cdl_values_from_metadata(metadata['ASC_SOP'])
```
jminor: Would you ever have both CDL and ASC_SOP? If so, this will clobber the earlier one.

contrib/opentimelineio_contrib/adapters/ale.py
```diff
@@ -116,6 +117,33 @@ def _parse_data_line(line, columns, fps):
...
+
+            if cdl:
+                metadata.pop('ASC_SOP')
+
+        if 'ASC_SAT' in metadata:
+            if metadata['ASC_SAT']:
```
jminor: You could combine these two ifs into one like this: `if metadata.get('ASC_SAT')` since `get` will return `None` if the key is missing.

contrib/opentimelineio_contrib/adapters/ale.py
jminor: You can use `isinstance(foo, (type(""), type(u"")))` to get around the issue.

contrib/opentimelineio_contrib/adapters/ale.py
vvzen: It can happen to have both CDL and ASC_SOP, so in this case I'd rather trust the ASC_SOP more than the generic CDL. Even though for now in all the samples I examined the values were pretty much the same.

contrib/opentimelineio_contrib/adapters/ale.py
vvzen: Yup, I generally use the `in` keyword because it should perform in constant time since it performs an hashing function, while I don't know if get() internally tries to get the key and catches a KeyError (which is generally slower). But yeah, no harm in just using `get` I guess!

contrib/opentimelineio_contrib/adapters/ale.py
vvzen: Cool! Makes sense since we don't need the returned index

contrib/opentimelineio_contrib/adapters/ale.py
vvzen: @jminor Nice trick!

vvzen:
Any updates on this one? :)

contrib/opentimelineio_contrib/adapters/ale.py
jminor: As long as it is documented that ASC_SOP takes precedence over CDL that should be fine.

jminor:
@reinecke can you look again? I think all your review notes were addressed.

PR #635, opened 2019-12-23
Detect if plugin doesn't have a docstring and return an error

ssteinbach:
… which says which plugin it is that doesn't have the docstring.


reinecke:
Like it!

PR #637, opened 2020-01-13
Added support for ImageSequenceReference to example RV plugin

avrata:
This syncs the example OTIO reader plugin with the recent changes in the RV adapter wrt support for the ImageSequenceReference schema

### DIFF
--- a/contrib/opentimelineio_contrib/application_plugins/rv/example_otio_reader/otio_reader.py
+++ b/contrib/opentimelineio_contrib/application_plugins/rv/example_otio_reader/otio_reader.py
@@ -207,6 +207,18 @@ def _create_media_reference(item, track_kind=None):
                 media.append(blank)
 
             return media
+        elif isinstance(item.media_reference,
+                        otio.schema.ImageSequenceReference):
+            frame_sub = "%0{n}d".format(
+                n=item.media_reference.frame_zero_padding
+            )
+
+            media = [
+                str(item.media_reference.abstract_target_url(symbol=frame_sub))
+            ]
+            print("ImageSequenceReference media: %s" % media)
+
+            return media
         elif isinstance(item.media_reference, otio.schema.GeneratorReference):
             if item.media_reference.generator_kind == "SMPTEBars":
                 kind = "smptebars"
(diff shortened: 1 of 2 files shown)


contrib/opentimelineio_contrib/application_plugins/rv/example_otio_reader/otio_reader.py
```diff
@@ -206,6 +206,18 @@ def _create_media_reference(item, track_kind=None):
...
+        elif isinstance(item.media_reference,
+                        otio.schema.ImageSequenceReference):
+            frame_sub = "%0{n}d".format(
+                n=item.media_reference.frame_zero_padding
+            )
+
+            media = [
+                str(item.media_reference.abstract_target_url(symbol=frame_sub))
+            ]
+            print("ImageSequenceReference media: %s" % media)
```
reinecke: Is it desirable to print from within an adapter?

contrib/opentimelineio_contrib/application_plugins/rv/example_otio_reader/otio_reader.py
```diff
@@ -250,19 +262,28 @@ def _create_item(it, track_kind=None):
...
+                    range_to_read
+                )
+
+    if not in_frame and not out_frame:
+        # because OTIO has no global concept of FPS, the rate of the duration
+        # is used as the rate for the range of the source.
+        # RationalTime.value_rescaled_to returns the time value of the object
+        # in time rate of the argument.
+        in_frame = range_to_read.start_time.value_rescaled_to(
+            range_to_read.duration
```
reinecke: I think you may want to use `range_to_read.duration.rate` here (assuming `range_to_read` is a `TimeRange`). As is, this will cause an implicit conversion of the duration rational to float and use that float as the rate.

contrib/opentimelineio_contrib/application_plugins/rv/example_otio_reader/otio_reader.py
avrata: Ack! Debug. I'll remove. Good catch.

contrib/opentimelineio_contrib/application_plugins/rv/example_otio_reader/otio_reader.py
avrata: Sure, however this was simply a copy of the original external rv adapter:

https://github.com/PixarAnimationStudios/OpenTimelineIO/blob/imagesequence/contrib/opentimelineio_contrib/adapters/extern_rv.py#L316-L320

We should just keep them in sync.

avrata:
re: unit test. Sure, I can try and set that up. It's a bit challenging as it'll need a bunch of tooling to build and deploy this plugin as well as run RV. I'll talk to @jminor to see if he has any good ideas.

contrib/opentimelineio_contrib/application_plugins/rv/example_otio_reader/otio_reader.py
apetrynet: Which I again kept from the original adapter pre ImageSequenceReference. 🙄

reinecke: Hrm, okay. I'm a little confused now. Reading in more detail, I'd expect this code to look a little more like:
```
in_frame = otio.opentime.to_frames(range_to_read.start_time, fps=range_to_read.duration.rate)
out_frame = otio.opentime.to_frames(range_to_read.end_time_inclusive(), fps=range_to_read.duration)
```

jminor: We figured out why this code looks funny. `value_rescaled_to` is overloaded to accept either a rate, or a RationalTime, in which case it will use the rate from that RationalTime. So passing `duration` does the same thing as passing `duration.rate`.

Using `to_frames` would be fine too if you want to ensure that you snap to an integer frame number.

avrata: Something like this? fdcda6771fdaaf870a511e72243a8032cc754f80

reinecke: I like it! I think it clarifies the intent of the code.

apetrynet: @jminor @reinecke Should I do this in the adapter as well?

reinecke:
Since testing became a larger project, we're going to merge this PR in and let tests for `ImageSequenceReference` follow #710 which addresses the test concerns.

PR #638, opened 2020-01-14
Prevent FD leak in RVsession tests

apetrynet:
I moved the creation and deletion of the temp files and corresponding file descriptors in `setUp()` and `tearDown()` methods.

I suspect I've done something stupid in my merging of the upstream code, thus all the previous commits are tagging along in this PR. Please let me know if I should close this PR and start over. I think I did my previous change directly in the imagesequence branch and not a feature branch based on it. That might be what's causing this.
Apologize if this causes head aches.

### DIFF
--- a/contrib/opentimelineio_contrib/adapters/tests/test_rvsession.py
+++ b/contrib/opentimelineio_contrib/adapters/tests/test_rvsession.py
@@ -45,6 +45,10 @@ TRANSITION_EXAMPLE_PATH = os.path.join(OTIO_SAMPLE_DATA_DIR, "transition_test.ot
 SAMPLE_DATA_DIR = os.path.join(os.path.dirname(__file__), "sample_data")
 BASELINE_PATH = os.path.join(SAMPLE_DATA_DIR, "screening_example.rv")
 BASELINE_TRANSITION_PATH = os.path.join(SAMPLE_DATA_DIR, "transition_test.rv")
+IMAGE_SEQUENCE_EXAMPLE_PATH = os.path.join(
+    SAMPLE_DATA_DIR,
+    "image_sequence_example.otio"
+)
 
 
 SAMPLE_DATA = """{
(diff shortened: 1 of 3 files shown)


reinecke:
I'll update my comment because I just caught up with the thread on the original PR.
That windows caveat @darbyjohnston brought up is a good one. I'd still like to see this done a little more locally though. Could using the `TemporaryDirectory` context manager work? Within that dir you may name files whatever you like then.

apetrynet:
OK, I'll give it a go.

apetrynet:
Hmm. `TemporaryDirectory` isn't available in python2.7 and for some reason the RV adapter only works with python2.
I'll still find another approach.

contrib/opentimelineio_contrib/adapters/tests/test_rvsession.py
```diff
@@ -468,14 +468,20 @@
     "RV Adapter does not work in python 3."
 )
 class RVSessionAdapterReadTest(unittest.TestCase):
+    def setUp(self):
+        self.fd, self.tmp_path = tempfile.mkstemp(suffix=".rv", text=True)
```
reinecke: I think you should be closing the fd here to avoid the issue of re-opening an already open file in windows.

contrib/opentimelineio_contrib/adapters/tests/test_rvsession.py
```diff
@@ -468,14 +468,20 @@
...
+        os.unlink(self.tmp_path)
+
     def test_basic_rvsession_read(self):
         timeline = otio.adapters.read_from_file(SCREENING_EXAMPLE_PATH)
-        tmp_path = tempfile.mkstemp(suffix=".rv", text=True)[1]
 
-        otio.adapters.write_to_file(timeline, tmp_path)
-        self.assertTrue(os.path.exists(tmp_path))
+        otio.adapters.write_to_file(timeline, self.tmp_path)
+        self.assertTrue(os.path.exists(self.tmp_path))
```
reinecke: You can omit this assertion because the `open` on the next line will exercise that for you.

contrib/opentimelineio_contrib/adapters/tests/test_rvsession.py
```diff
@@ -503,16 +508,15 @@ def test_transition_rvsession_read(self):
...
 
         # EXERCISE
-        otio.adapters.write_to_file(timeline, tmp_path)
+        otio.adapters.write_to_file(timeline, self.tmp_path)
 
         # VERIFY
-        self.assertTrue(os.path.exists(tmp_path))
-        self.assertTrue(os.path.getsize(tmp_path) > 0)
+        self.assertTrue(os.path.exists(self.tmp_path))
+        self.assertTrue(os.path.getsize(self.tmp_path) > 0)
```
reinecke: Since you're checking the content below, this zero size assertion isn't necessary.

PR #639, opened 2020-01-14
Docstring/error messaging improvements for ImageSequenceReference

reinecke:
Adding docstrings to `ImageSequenceReference` CPP bindings.

### DIFF
--- a/docs/tutorials/otio-serialized-schema.md
+++ b/docs/tutorials/otio-serialized-schema.md
@@ -361,21 +361,21 @@ parameters:
 *documentation*:
 
 ```
-None
+Refers to a sequence of image files to be played back at a constant rate.
 ```
 
 parameters:
 - *available_range*: 
-- *frame_step*: 
-- *frame_zero_padding*: 
+- *frame_step*: Step between frame numbers in the image sequence.
+- *frame_zero_padding*: Number of digits to pad zeros out to in the image names.
 - *metadata*: 
-- *missing_frame_policy*: 
+- *missing_frame_policy*: Policy application should use for handling missing images.
 - *name*: 
-- *name_prefix*: 
-- *name_suffix*: 
-- *rate*: 
-- *start_frame*: 
-- *target_url_base*: 
+- *name_prefix*: The image name leading up to the frame number section.
+- *name_suffix*: The part of the image name after the image number - including file extension.
+- *rate*: Frame rate as a float.
+- *start_frame*: The first frame number of the image sequence.
+- *target_url_base*: The base part of the url on which to append the image file name.
 
 ### LinearTimeWarp.1
 
--- a/docs/tutorials/otio-serialized-schema.md
+++ b/docs/tutorials/otio-serialized-schema.md
@@ -354,6 +354,29 @@ parameters:
 - *name*: 
 - *parameters*: 
 
+### ImageSequenceReference.1
+
+*full module path*: `opentimelineio.schema.ImageSequenceReference`
+
+*documentation*:
+
+```
+None
+```
+
+parameters:
+- *available_range*: 
+- *frame_step*: 
+- *frame_zero_padding*: 
+- *metadata*: 
+- *missing_frame_policy*: 
+- *name*: 
+- *name_prefix*: 
+- *name_suffix*: 
+- *rate*: 
+- *start_frame*: 
+- *target_url_base*: 
+
 ### LinearTimeWarp.1
 
 *full module path*: `opentimelineio.schema.LinearTimeWarp`
--- a/docs/tutorials/otio-serialized-schema.md
+++ b/docs/tutorials/otio-serialized-schema.md
@@ -361,13 +361,52 @@ parameters:
 *documentation*:
 
 ```
-Refers to a sequence of image files to be played back at a constant rate.
+Refers to a sequence of image files to be played back
+                    at a constant rate.
+
+                    The target urls for the images are determined by a concatentation of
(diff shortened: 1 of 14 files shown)


docs/tutorials/otio-serialized-schema.md
```diff
@@ -361,21 +361,21 @@ parameters:
...
-None
+Refers to a sequence of image files to be played back at a constant rate.
 ```
 
 parameters:
 - *available_range*: 
-- *frame_step*: 
-- *frame_zero_padding*: 
+- *frame_step*: Step between frame numbers in the image sequence.
+- *frame_zero_padding*: Number of digits to pad zeros out to in the image names.
```
apetrynet: Could this be simplified to something like: Number of digits in frame number

reinecke: How about: "Number of digits in image name frame number."?
Just want to make sure I'm explicit about this being for how the file name is generated. Though I could also clarify that more in the docstring for the class.

avrata: Coming from studios that don't pad our frame numbers, I worry a bit about simply "number of digits...". For example frame "foo.100.exr" should not be a pad of "3" for us since we'll want to see "99" and not "099" for the previous frame. The argument is pretty descriptive though but I would just ask that the doc string continue to keep "pad" somehow if possible.

apetrynet: `"Number of digits in image name frame number."`
is fine by me :)

reinecke: @avrata do you think having the word "padding" in the attribute name is good enough, or do you feel like the behavior should be outlined more explicitly in the docstring?

Thank you for the help hashing this out!

avrata: Yep, it's probably fine. I was going to comment on the danger of: https://github.com/PixarAnimationStudios/OpenTimelineIO/pull/633/commits/bb2bd99c3ba2825df8d672a079f9d35ad5035aad#diff-c166f6447c3991e9810480f19426caaaR255 but saw that got fixed that up before I could mention it. The attribute is descriptive enough.

apetrynet: This whole no padding point. Have we covered that in the schema? I can't remember the default value of `frame_zero_padding`. If it's set to 1 I guess we're ok?

avrata: I think so, the default is "0" https://github.com/PixarAnimationStudios/OpenTimelineIO/pull/602/files#diff-06c32a0f949112099a7b59156703bba9R29 which works great for me :) (from non-padding studios)

apetrynet: Ok, great! Was worried that some of the other methods would freak out, but @reinecke knows his stuff :)

docs/tutorials/otio-serialized-schema.md
```diff
@@ -361,13 +361,52 @@ parameters:
...
+                    ``frame_zero_padding``) and ``name_suffix``.
+                    For instance, if the following values are used::
+                        'target_url_base': 'file:///show/sequence/shot/',
+                        'name_prefix': 'sample_image_sequence.',
+                        'name_suffix': '.exr'
+                        'frame_zero_padding': 4
+
+                    Then the target url for frame 7 in the sequence would be::
+                        file:///show/sequence/shot/sample_image_sequence.0007.exr
+
```
apetrynet: The `ImageSequenceReference` doesn't have the standard `target_url()` method does it? I'm afraid referring to "target url" will mislead people into calling that method. Perhaps including the call for the url in the example will avoid confusion?
``` shell
>>> my_mediareference.target_url_for_image_number(7)
>>> 'file:///show/sequence/shot/sample_image_sequence.0007.exr'
```

docs/tutorials/otio-serialized-schema.md
```diff
@@ -361,21 +361,65 @@ parameters:
...
+                    the url will be for the first image in the sequence. For
+                    instance, if ``start_frame`` is set to ``5``, then a call
+                    ``ref.target_url_for_image_number(0)`` would return
+                    something like::
+                        file:///show/sequence/shot/sample_image_sequence.0005.exr
+
+                    Frame step sets number of frames to "skip". For instance, if
+                    ``start_frame`` is 1 and ``frame_step`` is 2, then the frame
+                    number sequence will be 1, 3, 5, 7, etc.
+
```
ssteinbach, line 408: This comment is great.  The only thing I would add to it is a (admittedly pedantic) note about how the timing is implemented.  IE played back at constant rate, but starting with the first frame in the sequence, which is on the screen for the time range [0, 1/rate), then the next image is shown for [1/rate, 2/rate) etc.  This maps image files on disk to time ranges in the media time.

ssteinbach: Also any limitations on what the rate can be set to would be good to note here

ssteinbach:
This change looks like its good to go but needs to be rebased again, (sorry!).  Then we should land it.

ssteinbach:
Hey @reinecke can you rebase this PR?  It would be great to land this.  Sorry about the delay!

reinecke:
The docstrings covered in this PR were actually addressed in this commit: https://github.com/PixarAnimationStudios/OpenTimelineIO/commit/2a1551e175a2d6012d5b895608dc0aa8d2857440#

The Error handling updates have been split into #783

PR #641, opened 2020-01-17
Expose json indent to the otio_json adapter

avrata:
Not at all critical, but this threw me for a bit and I thought it might be nice to expose here.

### DIFF
--- a/docs/tutorials/otio-plugins.md
+++ b/docs/tutorials/otio-plugins.md
@@ -116,15 +116,59 @@ This adapter lets you read and write native .otio files
 
 *Supported Features (with arguments)*:
 
-- read_from_file:
+- read_from_file: 
+```
+De-serializes an OpenTimelineIO object from a file
+
+  Args:
+      filepath (str): The path to an otio file to read from
+
+  Returns:
+      OpenTimeline: An OpenTimeline object
+```
   - filepath
-- read_from_string:
+- read_from_string: 
+```
+De-serializes an OpenTimelineIO object from a json string
+
+  Args:
+      input_str (str): A string containing json serialized otio contents
+
+  Returns:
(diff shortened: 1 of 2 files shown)


jminor:
If you want to go one step farther, you could expose a `minify=true` option, and then toggle to use `Writer` instead of `PrettyWriter` in the serializer.

avrata:
Yeah, good point. I found while reading the code you can trigger that PrettyWriter vs Writer switch by passing a negative value to `indent` that I ended up using in #640. I wonder if simply adding some documentation would help here.

jminor:
That should either be clearly documented, or perhaps minify=true could do that for you? That would be clearer for someone reading the code, and wouldn't rely on an implementation detail of RapidJSON.

avrata:
Totally, however that (< 0) code is ours: https://github.com/PixarAnimationStudios/OpenTimelineIO/blob/0a35894d72b14c7cb8373a00f0dfa594dc493751/src/opentimelineio/serialization.cpp#L707-L756

If you wanted to split out the logic into indent vs no-newlines, we should make the change there and carry the arguments along out to python.

davidbaraff:
Note that the decision of which formatter to use (the “PrettyWriter” or the other one RapidJSON supplies) is made outside of rapidJSON, i.e. by the C++ code, if that matters in this discussion.

jminor:
Ah, I see. Thanks for pointing that out. I have a slight preference for making this a toggle (pretty vs minified) since I doubt anyone would actually want `indent=3` or any other value.

davidbaraff:
Is it possible there are three values?
	-1
	4
	0 (eliminate spaces to make the file as small as possible?)

avrata:
Ok, I added some docs for the otio_json adapter. Can someone take a look and perhaps suggest improved wording?

PR #646, opened 2020-02-08
fix otioconvert for Kdenlive with python3

vpinon:
Hello,
A minor change that avoids unicode/utf-8 XML encoding failures when switching between python2 & python3.
Best regards,
Vincent


ssteinbach:
Also addresses #656

PR #663, opened 2020-03-17
Tweaks to cmake so that pip and local builds both work

ssteinbach:
Adds some checks to make sure that git is present before attempting to recursively update.


src/CMakeLists.txt
```diff
@@ -11,8 +11,16 @@ endif()
 
 # make sure that git submodules are up to date when building
 find_package(Git QUIET)
+if (GIT_FOUND)
```
meshula: nice way to deal with this issue

PR #669, opened 2020-03-24
Fix the windows build

meshula:
Add a bin target to satisfy cmake windows requirement, bump the version to cmake 3.

The cmake version for Linux is per vfx-platform 2018, the version for Windows is the latest, otherwise VS2019 can't be supported.

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,7 +102,7 @@ If you want to contribute to the project, please see: https://opentimelineio.rea
 
 You can get the latest development version via:
 
-`git clone --recursive git@github.com:PixarAnimationStudios/OpenTimelineIO.git OpenTimelineIO`
+`git clone git@github.com:PixarAnimationStudios/OpenTimelineIO.git OpenTimelineIO`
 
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -102,7 +102,7 @@ If you want to contribute to the project, please see: https://opentimelineio.rea
 
 You can get the latest development version via:
 
-`git clone git@github.com:PixarAnimationStudios/OpenTimelineIO.git OpenTimelineIO`
+`git clone --recursive git@github.com:PixarAnimationStudios/OpenTimelineIO.git OpenTimelineIO`
```
ssteinbach: This isn't wrong, but also cmake should do the recursive init for you (unless that doesn't work on windows?)  This shouldn't be necessary in other words.

meshula: I'll test that - I think you are correct that it works. I made this note at the point when the cmake files errored out and didn't run at all.

meshula: removed --recursive

PR #671, opened 2020-03-25
Do not require media_reference when deserializing Clip

mikemahony:
A potential fix for https://github.com/PixarAnimationStudios/OpenTimelineIO/issues/670.


reinecke:
We've decided we'd like for OTIO files to be explicit about null media references.

PR #672, opened 2020-03-26
extended edl metadata items for convinience getting of original EDL info

vcerundolo:
This allows you to get the original EDL fields including event_no, reel, record_tc, etc.
There is some additional custom items that can extend the edl comments to include a show name and a lut name.


reinecke:
Useful additions! Would you be able to add/update some unittests for this metadata?

mikemahony:
Hi Vince - This is the code you mentioned in your SIGGRAPH talk, yes?  :tada::tada:

Just curious... are `SHOW` and `LUT` EDL comments industry conventions or in a spec somewhere or are they specific to Marvel?

vcerundolo:
Thanks,
yes this is the code from the talk.

show and lut are not industry standard, but then again not much about the
cmx_3600 is anymore. I have used this type of tagging in previous places as
well.
But yes they are specific to Marvel.

The cmx format allows for any number of comments. 'show' might be a little
strange to have, but i think that 'lut' could be very useful, especially
with the ASC CDL info in there. although for the complete color pipeline
there should also be an input transform, but that one is harder to get. The
lut is usually readily available fromt he dailies ALE, and therefore the
AVID.

I do need to go back and fix the testing. I might remove the 'show' just to
avoid confusion. It also isn't required for most of what we are doing
currently.
I was using it when we were doing a trailer that spanned multiple shows,
and needed a way to automatically set certain parameters based on it.

More of an edge case currently.

Vince Cerundolo

jminor:
All of this EDL-specific metadata should go under the `cmx_3600` entry in the clip metadata instead of at the top level.

Also, why do you need the source/record timecode metadata when you can ask OTIO to compute these for you? If you store them as metadata they will get out of sync if you edit your timeline. I'm concerned that this will be confusing.

vcerundolo:
That is true, but that is exactly why i want them as as metadata. I want to
always be able to see what the original tc is.
For example, in some cases the source-out tc that is being returned is
being re-calculated based on the record duration, and then my cuts are
short. This happens when there are time-warp effects on my source.
There are other examples, but in some cases I just want the data that I put
in, and not the computed out...

Thanks

jminor:
Ah, that makes sense. If you have repro cases of the mismatch, can you file an issue about that? Hopefully that is easy to fix.

reinecke:
I have a use case where people are asking for the original timecode values as well as event numbers.

I've been looking deeper into how to persist this information this and there are some interesting problems, including:
- Multiple events can be responsible for having described what becomes a single clip in OTIO. For instance: Transitions happen in separate events, if the media described in the transition is continuous with the previous clip then that clip is just extended to include media used in the transition. This would mean two EDL events are actually used for describing the single OTIO clip.
- This same issue applies for the timecode values from the EDL. A clip can be extended by a duration implicit by the record timecode in a transition. This means that there are two sets of timecodes that describe what became the OTIO clip.

For a more concrete example, check out `dissolve_test_2.edl` from the unittests:

```
TITLE:   dissolve_test_2
FCM: NON-DROP FRAME
001  TST V     C        01:00:04:05 01:00:04:10 01:00:00:00 01:00:00:05 
* FROM CLIP NAME:  clip_A
002  TST V     C        01:00:04:10 01:00:04:10 01:00:00:05 01:00:00:05 
002  TST V     D    010 01:00:08:04 01:00:08:14 01:00:00:05 01:00:00:15 
* BLEND, DISSOLVE 
* FROM CLIP NAME:  clip_A
* TO CLIP NAME:  clip_B
003  TST V     C        01:00:08:14 01:00:08:19 01:00:00:15 01:00:00:20 
* FROM CLIP NAME:  clip_B
```

The EDL is read as this otio timeline:

image

Event `001` only describes the clip up to the moment just before where the transition starts. Event `002` describes (implicitly) the extension of `clip_A` for the transition duration. Moreover, `002` actually declares two different record timecodes.

For `clip_A` in the OTIO model, the possible metadata becomes:
- `event_no` - `001` or `002`
- `source_tc_in` - `01:00:04:05` or `01:00:04:10`
- `source_tc_out` - `01:00:04:10` or `01:00:04:10` (note `source_tc_in` and `source_tc_out` are the same in event `002`)
- `record_tc_in` - `01:00:00:00` or `01:00:00:05` (this value is the same in both the `002` entry record ranges)
- `record_tc_out` - `01:00:00:05` or `01:00:00:05` or `01:00:00:15` (from the second `002` event)

Given the tangly nature of this set of info, it may be best to model it as a dictionary and treat a transitions second event entry as the source of truth for `record_tc` values.

So the above example would come out like this:

```
clip.metadata["cmx_3600"].update(
    {
        "edl_timecode": {
            "001": {
                "source_tc_in": "01:00:04:05",
                "source_tc_out": "01:00:04:10",
                "record_tc_in": "01:00:00:00",
                "record_tc_out": "01:00:00:05",
            },
            "002": {
                "source_tc_in": "01:00:04:10",
                "source_tc_out": "01:00:04:10",
                "record_tc_in": "01:00:00:05",
                "record_tc_out": "01:00:00:15",
            },
        }
    }
)
```

Would love to hear everyone's thoughts on this!

vcerundolo:
Unfortunately the second event isn't always the source of truth
https://xmil.biz/EDL-X/CMX3600.pdf
Take a look at dissolve cases 3, 4, 7, etc.

But I think you are on the right path to wanting to create an event
dictionary.

PR #675, opened 2020-03-27
Not all clip’s media_reference have `target_url`

jchen9:
This came up when there’s a clip like the following in the timeline:
```JSON
{
    "OTIO_SCHEMA": "Clip.1",
    "metadata": {},
    "name": "[Black]",
    "source_range": null,
    "effects": [],
    "markers": [],
    "media_reference": {
        "OTIO_SCHEMA": "GeneratorReference.1",
        "metadata": {},
        "name": "[Black]",
        "available_range": {
            "OTIO_SCHEMA": "TimeRange.1",
            "duration": {
                "rate": 24.0,
                "value": 24.0
            },
            "start_time": {
                "rate": 24.0,
                "value": 1.0
            }
        },
        "generator_kind": "Slug",
        "parameters": {}
    }
}

```
This change will return None and let caller deciding the next mobid strategy  to use.

### DIFF
--- a/contrib/opentimelineio_contrib/adapters/aaf_adapter/aaf_writer.py
+++ b/contrib/opentimelineio_contrib/adapters/aaf_adapter/aaf_writer.py
@@ -220,14 +220,21 @@ def _gather_clip_mob_ids(input_otio,
 
     def _from_aaf_file(clip):
         """ Get the MobID from the AAF file itself."""
-        mob_id = None
-        target_url = clip.media_reference.target_url
+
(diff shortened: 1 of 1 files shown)


stefanschulze:
looks good to me

jminor:
Do we have test coverage for this?

jchen9:
After some software archaeology, it looks like this was addressed in https://github.com/PixarAnimationStudios/OpenTimelineIO/pull/628

PR #676, opened 2020-03-31
Add List of Supported Formats to Conform.py Help Text

TheBigSasha:
Add list of supported files to tooltip for the conform.py example script.

### DIFF
--- a/examples/conform.py
+++ b/examples/conform.py
@@ -59,7 +59,8 @@ def parse_args():
         type=str,
         required=True,
         help='Timeline file(s) to read. Any format supported by OTIO will'
-        ' work.'
+        ' work: fcp_xml, fcpx_xml, EDL, AAF, and more via community '
+        'developed adapters.'
     )
     parser.add_argument(
         '-f',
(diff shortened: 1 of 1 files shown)


examples/conform.py
```diff
@@ -59,7 +59,8 @@ def parse_args():
         type=str,
         required=True,
         help='Timeline file(s) to read. Any format supported by OTIO will'
-        ' work.'
+        ' work: fcp_xml, fcpx_xml, EDL, AAF, and more via community '
```
ssteinbach: You can take this a step further if you'd like and list the available adapters, using this function:
https://github.com/PixarAnimationStudios/OpenTimelineIO/blob/9d010f2966523505b5c954d803f4f982f5e4edb5/src/py-opentimelineio/opentimelineio/console/otioconvert.py#L36

And if you want to put the file suffixes in the message, you can use this function:
https://github.com/PixarAnimationStudios/OpenTimelineIO/blob/9d010f2966523505b5c954d803f4f982f5e4edb5/src/py-opentimelineio/opentimelineio/adapters/__init__.py#L52

Its a great idea to put that information into the file input argument!  That upgrade could go into the real scripts (otioconvert, otioview, etc.) if you were interested in making the fix.

TheBigSasha: On it! Thank you for the info!

PR #684, opened 2020-04-14
Otioview display

### LINKED ISSUE #280: otioview video display
It would be great if otioview had a video display that showed the media that the selected clip's media reference target_url points to.

Note: that this is *not* the same as playing back the whole timeline, which would be a much larger request.

KarthikRIyer:
Closes #280 

This PR is on top of #677 and can be merged after that.
I've implemented this by adding a display widget that displays the currently visible frame.
Only the last two commits are relevant to this PR.

image

Questions posed by @reinecke :

> - How is missing media handled in this?
> - How are clips with TimeWarp effects shown?
> - How are transitions handled?
> - What if the top track is a transparent PNG overlaying a graphic in the corner?

Some suggestions:

By @reinecke :
> Perhaps when there are effects on clips we could do something like overlay a tint color on the playback image along with text like "Effect: Sepia". In the "special" case of LinearTimeWarp effect, it may be best to just omit the clip frames altogether and put up text saying it's not supported?

By @jminor :

> One way to approach this without leading to a full video editing application is to show the display only when someone clicks on a clip. By tying the display to the inspector, we could be visually clear that you're looking at the media reference for that clip, without connecting it to the playhead position. This is a bit like the source monitor in an NLE, rather than the record monitor.

I'll make the changes as and when these questions are answered.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -442,7 +442,8 @@ setup(
             'urllib3>=1.24.3'
         ],
         'view': [
-            'PySide2~=5.11'
+            'PySide2~=5.11',
+            'opencv-python'
         ]
     },
 
--- a/src/opentimelineview/console.py
+++ b/src/opentimelineview/console.py
@@ -28,7 +28,7 @@
 import os
 import sys
 import argparse
-from PySide2 import QtWidgets, QtGui
+from PySide2 import QtWidgets, QtGui, QtCore
 
 import opentimelineio as otio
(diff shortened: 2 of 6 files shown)


ssteinbach:
This was discussed further in the latest TSC meeting, see the notes on #280

KarthikRIyer:
Now that the idea has been refined I'm closing this PR. I'll open a new one when it's ready.

PR #686, opened 2020-04-19
Implement Clip Inspector

### LINKED ISSUE #280: otioview video display

KarthikRIyer:
Closes #280 

As discussed in the TSC meeting, I've implemented a Clip Inspector that allows you to play the clip you've clicked on.

For now I show the complete range of the clip. I'll try implementing an option to view the `source_range`.

It looks something like this:
player

To show an effect in case it exists I've drawn a translucent rectangle and shown the effect-name on top. Something like this:
Screenshot from 2020-04-20 02-07-01

In case a Media Reference cannot be resolved I've show a message saying the same:
image

There are a few things to clarify/improve:
- In case there are multiple effects on a clip, what do we want to do? At the moment I've just shown the name of the first effect.
- Right now, the size of the Clip Inspector is constant. I had to do this because the size of the widget is very large at startup, which I'm not sure but might be because of the way I'm showing the error message. But this does seem to work alright.
- Also, In case I do not set a constant widget size, I'm having trouble maintaining the aspect ratio of the video upon resizing.

@ssteinbach @reinecke @jminor I look forward to hearing your thoughts on this!

### DIFF
--- a/src/opentimelineview/clip_inspector_widget.py
+++ b/src/opentimelineview/clip_inspector_widget.py
@@ -29,7 +29,7 @@ from PySide2.QtCore import Qt
 from PySide2.QtMultimediaWidgets import QGraphicsVideoItem
 
 import opentimelineio as otio
-from PySide2.QtWidgets import QWidget, QStackedLayout, QLabel,\
+from PySide2.QtWidgets import QWidget, QStackedLayout, QLabel, \
     QGraphicsScene, QGraphicsView, QGraphicsTextItem, QGraphicsRectItem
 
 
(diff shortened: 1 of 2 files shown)


KarthikRIyer:
Tiny update to effects banner:
image

reinecke:
@KarthikRIyer This is coming along really nicely!
I think if there are a stack of effects, perhaps it could say something like "Burn Effect (and 3 others)" or something? That would let the person using it know they should dive deeper into the inspection if they want to know more.

One other thing to check: What is the behavior with unsupported media types?

KarthikRIyer:
Yeah, showing the number of effects seems like a good option. For unsupported formats some error message is printed in the terminal. Qt uses GStreamer. I tried to play RAW ProRes footage and it gave some error message. (I'm sorry I don't remember the exact message) I'll post it tomorrow when I'm on my system.

reinecke:
Is it possible to detect the unsupported media error after setting the media on the player and display some sort of message in the UI similar to the "media reference can't be resolved" one?
It looks like you might be able to connect some slots to the `mediaStatusChanged` and/or `error` slots to trigger showing this message once the player has determined there is a problem (I suspect the player loads media in an async fashion, so you probably can't immediately check these properties after setting the playback item).

KarthikRIyer:
This is the error I get for RAW ProRes :
```console
Warning: "No decoder available for type 'video/x-gst-fourcc-aprh, width=(int)4112, height=(int)2176, framerate=(fraction)25/1, pixel-aspect-ratio=(fraction)1/1'."

```

> It looks like you might be able to connect some slots to the mediaStatusChanged and/or error slots to trigger showing this message once the player has determined there is a problem (I suspect the player loads media in an async fashion, so you probably can't immediately check these properties after setting the playback item).

@reinecke 
I don't think I'd need to use `mediaStatusChanged`, because as you said I can connect the error function from player to get the error. I get something like this from there:

```
Cannot play stream of type: <unknown>
```

I'll update the UI to show something like the unresolved media message.

apetrynet:
Hi 
This is going to be very useful.

Since we're not going to write a compositing engine anyway, I'd prefer if the effect indicators were placed outside of the image. Perhaps as a line of "badges"? That way we get an unrestricted view of the clip we're inspecting while informing about effects etc.
For missing or unsupported media it makes sense to have a placeholder image or text.

```
----------------------------------------------------
|                                                  |
|                                                  |
|                                                  |
|                                                  |
|                                                  |
|                      IMAGE                       |
|                                                  |
|                                                  |
|                                                  |
----------------------------------------------------
|burn|timewarp|flip|                               |
----------------------------------------------------
```

KarthikRIyer:
PR Update:

Error message in UI:
image

Effects badges instead of banner:
image

KarthikRIyer:
For the option to see the `whole clip`/`source_range` I was thinking of putting the option in the status bar instead of the clip inspector. That are already seems to be getting a bit crowded. Any thoughts?

apetrynet:
@KarthikRIyer Nice! really quick turnaround! I think this looks way cleaner. Let's hear what the others say. 
For the range toggle you could consider using a context menu over the image and/or clip in the timeline to switch between the two. 
You could also consider having a representation of the source vs available range at the end of the play bar widget. 
Something like: `(0 100-134 200)` for source range and `(0 0-200 200)` for available? May not be that readable, but might spike some ideas :)

reinecke:
@KarthikRIyer This is coming along really nicely!
To make sure I'm giving clear feedback, could you share exactly what you're calling different areas of the UI?
Perhaps there could be something like a QComboBox to the right of the effects display to select whether it's media range or source range. That will then give us an easy way to add additional viewing options.

KarthikRIyer:
image
@reinecke this is what I mean by status bar.

And by ClipInspector, I meant somewhere around ClipInspector.

But I like the QComboBox idea!

reinecke:
Oh, interesting. I think that is typically called the menu bar.
I think, in this case, keeping the control closer to the preview is better since it only applies to that section of the UI.
It's a good point that we should keep the menu bar in mind as a place to put functionality in the future though.

KarthikRIyer:
I've added a QComboBox to toggle between the ranges:

image


image

apetrynet:
Cool!
May I suggest using OTIO terms and change "Entire clip" to "Available range"?

reinecke:
I think in this case it might actually be entire clip. Available range is possibly something in addition to these two.

KarthikRIyer:
Tiny PR update.
I've added an option for available range as well. Here I get the duration of the clip from the media player after the media has been loaded.For this we need to update the clip duration on the seekbar after the media is loaded. For this I've updated the way to handle the change in clip duration and now using a callback function.
Could someone please recap when the entire clip duration and available range might be different?

KarthikRIyer:
@ssteinbach @reinecke @jminor @apetrynet are any more changes required? Any suggestions?

apetrynet:
Hi! 
I don't think I've come across situations where they differ. Perhaps the others have.

ssteinbach:
Realistically, there are three different potential ranges for a clip:

- `source_range` -- the range of the media, in the media space that the clip is "cutting into" the parent track or stack.  can be `None` or a `TimeRange`
- `available_range` -- the intended full range of media that ought to be considered available for editorial operations.  can be `None` or a `TimeRange`
- actual media on disk -- might be unreachable or not yet rendered.  OTIO doesn't store this or have a way to compute this directly, that is left up to consuming applications.  So if we provide a way of reading this its really for debugging only.

I'd argue for not including the actual media on disk in that combo box, but am curious to hear what other people think.  I would say the options for the pull down should be:
- `available_range`
- `source_range`
- `trimmed_range` (computed via the method: `source_range` if it is set, otherwise `available_range`)

Would be nice if it indicated (maybe by being grayed out) if `available_range` or `source_range` was not set.

What do folks think about that?

apetrynet:
Ok, I see.
Sounds good to me.

KarthikRIyer:
This looks good to me. I'll make the changes.

KarthikRIyer:
PR Update:

- Removed Entire clip option. Added trimmed range option.
- Unavailable ranges now greyed out.

It looks like this:
Screenshot from 2020-05-01 00-45-07

I think this is good to go. Any other suggestions or changes required?

src/opentimelineview/clip_inspector_widget.py
```diff
@@ -0,0 +1,208 @@
...
+        painter.fillRect(QRectF(0, 0, self.clipWidth, self.clipHeight), Qt.green)
+        painter.setPen(QPen(Qt.black))
+        painter.setFont(font)
+        painter.drawText(QRectF(self.clipWidth * 0.25, self.clipHeight * 0.25,
+                                self.clipWidth * 0.5, self.clipHeight * 0.6),
+                         "Unsupported Media format.")
+
+    def handle_error(self):
+        if self.player.errorString() == 'Resource not found.' or \
+                self.player.errorString() == 'Not Found':
```
ssteinbach, line 205: Its usually more readable to use `()` to do line continuation:
```suggestion
        if (
                self.player.errorString() == 'Resource not found.' 
                or self.player.errorString() == 'Not Found'
        ):
```

That way the or also comes first, showing the logical relationship of the two conditions.

KarthikRIyer:
@ssteinbach any changes required here?

jlskuz:
This feature idea looks great! What is missing here? From the conversation it looks read… @ssteinbach @KarthikRIyer

KarthikRIyer:
From a thread in the opentimelineio slack:

@jminor -
```
I think we decided that turning otioview into a video player was outside the scope of what our project wants to support. For playback, we suggest using tlRender, RV, cineSync Play, etc.
Showing just a thumbnail would certainly be useful - but also opens up considerations with format compatibility, target_url support (e.g. http, etc.) so may lead to a ton of work to make it truly useful & reliable.
If there was a way to connect otioview and one of the playback applications listed above so that the playheads were in sync - or to embed otioview into one of those applications, then that would achieve the goal of seeing the images without expanding the scope of what otioview is meant to accomplish.
```

ssteinbach:
Right. I think the right thing to do is actually to close this PR.  For OTIO itself, we'd like to keep OTIOView as simple and light as possible, and something like this would be a good downstream project that uses OTIO as a dependency.  Its still really cool though!

apetrynet:
Should we consider closing this ticket? 
We've got applications able to view OTIO files with media since this PR was written.

jminor:
Our recommendation is to use one of these player applications:
https://github.com/AcademySoftwareFoundation/OpenTimelineIO/wiki/Tools-and-Projects-Using-OpenTimelineIO

PR #687, opened 2020-04-21
cmx3600 reverse and dissolve fixes

reinecke:
- Addressed an issue where the speed change regex wouldn't properly parse reversed clips
- Fixed assumption that clips used in a transition would always be continuous with the clips before and after the transition

Additional notes:
This fix works but is a touch messy. The parser is structured in a way that OTIO Clips are created before all the metadata that might apply to them has been parsed and some of the transition handling is a bit band-aidy as a result. We may start thinking about how to refactor a bit to make the code a bit cleaner for these situations.


src/py-opentimelineio/opentimelineio/adapters/cmx_3600.py
```diff
@@ -665,14 +673,28 @@ def _expand_transitions(timeline):
                 transition_duration.rate
             )
 
+            # Because of how comments fall on transition events, some of the
```
ssteinbach, line 676: Can this be more specific? "Because comments in an EDL are listed after transition events" 
(rather than "how comments fall")

src/py-opentimelineio/opentimelineio/adapters/cmx_3600.py
```diff
@@ -725,6 +771,48 @@ def _expand_transitions(timeline):
...
+        if reelname(clip_a) == reelname(clip_b):
+            return True
+    except KeyError:
+        pass
+
+    try:
+        return clip_a.name == clip_b.metadata['previous_metadata']['source_clip_name']
+    except KeyError:
+        pass
+
```
ssteinbach, line 814: Add a `return False` just to be clear what the caller is expecting here

reinecke:
Good notes! Updated with clarifications.

PR #689, opened 2020-04-30
Fixed reading of XML files exported from Hiero or NS

Eiken:
- Previously only the first occurrence of a file with an id was properly
added.

### DIFF
--- a/src/py-opentimelineio/opentimelineio/adapters/fcp_xml.py
+++ b/src/py-opentimelineio/opentimelineio/adapters/fcp_xml.py
@@ -30,6 +30,7 @@ import itertools
 import math
 import os
 import re
+import uuid
 from xml.etree import cElementTree
 from xml.dom import minidom
 
@@ -619,12 +620,18 @@ class FCP7XMLParser:
         Given an element, dereferences it by it's id attribute if needed. If
         the element has an id attribute and it's our first time encountering
         it, store the id.
+
+        Hiero and Nuke Studio generates xml files where a clip of a file into
+        two got the same file id for both files in the clips.
+        By adding a uuid4 to the elem_id name OpenTimelineIO parses all the
+        files correctly. This is only used during read and can not be seen
+        in the converted otio file.
         """
         if element is None:
             return element
 
         try:
-            elem_id = element.attrib["id"]
+            elem_id = "{}{}".format(element.attrib["id"], uuid.uuid4())
         except KeyError:
             return element
 


Eiken:
We had issues reading XML files from Hiero and NS. Where clips magically disappeared. The issue boiled down to that otio was disregarding file elements with the same id.

This patch solves the problem but not being super familiar with XML or the code base I am not sure if it is solved in the correct way. Getting Foundry to fix their XML export seems like an impossible task so adding a workaround to otio seemed like the way to go.

reinecke:
Hi @Eiken, thanks for the submission!

It looks like this update may have broken the unittests for this adapter, we'd need that resolved before merging.
Also, do you happen to have a sample file you could provide that exhibits the issue? It would be great if we could add a test to make sure this doesn't happen.

The FCP XML spec [talks about how to properly handle](https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/FinalCutPro_XML/Basics/Basics.html#//apple_ref/doc/uid/TP30001154-TPXREF102) `id` attribute. If I'm reading what you're saying correctly, it sounds like the adapter may not be properly doing a lookup for a given id when it's reused - but I'd love to get an example EDL to validate.

Let me know if you can provide something, I'd love to help get this fixed!

Eiken:
Yeah I saw the tests failing as well. Will dig them up and see if I can resolve them and make a small example file of the output.

Eiken:
Example file. Renamed from xml to txt because github dont accept xml
[foobar.txt](https://github.com/PixarAnimationStudios/OpenTimelineIO/files/4586205/foobar.txt)

Steps to reproduce:
* Open blank project
* Import random clip
* Put in sequence
* Split in two with slice 
* Export as xml

Result:
* id for file in both clips are the same
* Reading with the fcp_xml adapter disregards one of the clips

reinecke:
@Eiken Thanks for providing the example file and the steps you used in Hiero, that kind of context is super helpful. I'll try and take a peek in the next day or two, I suspect it should be a pretty straightforward fix.

jhodges10:
@reinecke did you end up reproducing this?

Eiken:
@reinecke I'll check again tomorrow but what I was getting was just the clip before the cut. The other was thrown away by both otiocat and otioview

Eiken:
Now I cant reproduce it anymore either. Must have been some glitch in the matrix or a venv with some old version. Closing the PR

reinecke:
Thanks for following up!

Eiken:
So this caused an Issue again and I can reproduce it. If applying my fix it reads it correctly

Eiken:
The problem seems to stem from how Hiero/NS decides on the `id` for files. If you have a track in Hiero/NS that looks like attached picture then it will write out a xml with the same `id` for all files:
```
<file id="Juliet">
```

This causes all clips having mediareferences to file with id= "Juliet" to all have the same media_ref instead of unique.
```
media_ref = c.media_reference
```
media_ref will be the same for both clips from picture even tho they actually reference different files.
Screenshot from 2020-12-16 13-10-19

Eiken:
Got some days off next week so I'll check on getting the CI to complete successfully then and rebase on current main branch

src/py-opentimelineio/opentimelineio/adapters/fcp_xml.py
```diff
@@ -619,12 +620,18 @@ def _derefed_element(self, element):
...
+        By adding a uuid4 to the elem_id name OpenTimelineIO parses all the
+        files correctly. This is only used during read and can not be seen
+        in the converted otio file.
         """
         if element is None:
             return element
 
         try:
-            elem_id = element.attrib["id"]
+            elem_id = "{}{}".format(element.attrib["id"], uuid.uuid4())
```
reinecke, line 634: This fix is problematic because it makes `elem_id` is unique every time the function is invoked - even if invoked twice with the exact same element. This renders `self._id_map` useless because when `setdefault` is called below, `elem_id` will always be unique and return the element passed in.

Unfortunately, this behavior breaks files using the `id` attribute as intended. This is also why the unittests are breaking - there are `MissingReferences` getting introduced in the place of `ExternalReferences` for empty `file` elements meant to be resolved to previous occurrences.

Here is the reference in the [FCP 7 XML specification](https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/FinalCutPro_XML/Basics/Basics.html#//apple_ref/doc/uid/TP30001154-TPXREF126) providing some context around how the `id` attribute is meant to be used. 

In the sample data repo, there is an [example file](https://github.com/PixarAnimationStudios/OpenTimelineIO/blob/master/tests/sample_data/premiere_example.xml) using the `id` behavior as designed. Here is the original declaration for `file-1`:
https://github.com/PixarAnimationStudios/OpenTimelineIO/blob/master/tests/sample_data/premiere_example.xml#L67-L109

And here is a subsequent usage intended to resolve to the content of the first occurrence:
https://github.com/PixarAnimationStudios/OpenTimelineIO/blob/master/tests/sample_data/premiere_example.xml#L436

If you search that file, you can see `file-1` referenced repeatedly throughout meant to re-use that original declaration.

Maybe the pragmatic solution would be to add some logic saying that if the element has children, update the cache for that `id` and return the element. If the element doesn't have children, use the current `setdefault` behavior. For the set of test EDLs I've seen, I think that should have the expected behavior.

reinecke (changes requested):
Dealing with out-of-spec files is always tricky. I'd love for the Foundry to address this case, but it would still be hard to address all the out-of-spec files floating around. Hopefully the solution I pitched may be helpful.

When you make your update, would you be able to create a unittest to address this behavior? Perhaps you can author a simple example case in Heiro or NS that causes the issue?

src/py-opentimelineio/opentimelineio/adapters/fcp_xml.py
Eiken: I'll try and dig into the code and see if I can find some other way to fix it.

Eiken:
> Dealing with out-of-spec files is always tricky. I'd love for the Foundry to address this case, but it would still be hard to address all the out-of-spec files floating around. Hopefully the solution I pitched may be helpful.

I'll throw an email to the foundry and see if they have any plans to follow the xml standard. If they are we can probably live with just having our hacky patch applied for a while until they do.

reinecke:
> I'll throw an email to the foundry and see if they have any plans to follow the xml standard. If they are we can probably live with just having our hacky patch applied for a while until they do.

Even better, maybe they could adopt writing OTIO directly and we can be one step closer to deprecating the entire FCP XML adapter, pragmatic patch and all! 😉

Eiken:
> > I'll throw an email to the foundry and see if they have any plans to follow the xml standard. If they are we can probably live with just having our hacky patch applied for a while until they do.
> 
> Even better, maybe they could adopt writing OTIO directly and we can be one step closer to deprecating the entire FCP XML adapter, pragmatic patch and all!

I will for sure check the status on that as well. When I checked with em like a year ago it sounded like they where waiting for the C++ version but that is available now so they can't have that as an excuse anymore.

jminor:
@Eiken does the new OTIO integration in Nuke 13 mean that this PR is no longer needed? https://www.youtube.com/watch?v=autieMqcgSU

jminor:
Heiro and Nuke Studio both now have native OTIO support. I'm assuming that avoids this issue entirely, so I'm closing this PR. Feel free to re-open if this is still relevant.

PR #690, opened 2020-05-01
Toggle 'complete track name' display on double click in otioview

KarthikRIyer:
@ssteinbach as per your suggestion to have an option to view long track names in otioview, I've added an option to do that on double-clicking on the TrackName widget.

This is how it looks:

Screenshot from 2020-05-01 16-18-36

Screenshot from 2020-05-01 16-18-53

### DIFF
--- a/src/opentimelineview/track_widgets.py
+++ b/src/opentimelineview/track_widgets.py
@@ -36,6 +36,7 @@ MARKER_SIZE = 10
 EFFECT_HEIGHT = (1.0 / 3.0) * TRACK_HEIGHT
 HIGHLIGHT_WIDTH = 5
 TRACK_NAME_WIDGET_WIDTH = 100.0
+SHORT_NAME_LENGTH = 7
 CURRENT_ZOOM_LEVEL = 1.0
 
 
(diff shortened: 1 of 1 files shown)


src/opentimelineview/track_widgets.py
```diff
@@ -239,15 +241,54 @@ class TrackNameItem(BaseItem):
     def __init__(self, track, rect, *args, **kwargs):
         super(TrackNameItem, self).__init__(None, None, rect,
                                             *args, **kwargs)
-        track_name = 'Track' if track.name == '' else track.name
-        if len(track_name) > 7:
-            track_name = track_name[:7] + '...'
-        self.source_name_label.setText(track_name + '\n({})'.format(track.kind))
+        self.track = track
+        self.track_name = 'Track' if track.name == '' else track.name
```
ssteinbach, line 245:
```suggestion
        self.track_name = 'Track' if not track.name else track.name
```
Should be the same but also covers the case where `track.name` is `None`

src/opentimelineview/track_widgets.py
```diff
@@ -239,15 +241,54 @@ class TrackNameItem(BaseItem):
...
+        self.full_track_name = self.track_name
+        if len(self.track_name) > 7:
```
ssteinbach, line 247: lets break the 7 out to be a top level constant, maybe `SHORT_NAME_LENGTH`

PR #692, opened 2020-05-06
Fixed deepcopy for opentime.TimeRange

reinecke:
The bindings for `__deepcopy__` on `TimeRange` previously didn't accept the `memo` arg. This was causing the following error:

```
>>> copy.deepcopy(otio.opentime.TimeRange())
  File "/Users/ereinecke/.newt-cache/pyenv/versions/3.8.2/lib/python3.8/copy.py", line 153, in deepcopy
    y = copier(memo)
TypeError: __deepcopy__(): incompatible function arguments. The following argument types are supported:
    1. (self: opentimelineio._opentime.TimeRange) -> opentimelineio._opentime.TimeRange

Invoked with: otio.opentime.TimeRange(start_time=otio.opentime.RationalTime(value=0, rate=1), duration=otio.opentime.RationalTime(value=0, rate=1)), {}
```

I've also added unitests for copy and deepcopy on `RationalTime` and `TimeRange` to protect against regression.

### DIFF
--- a/src/py-opentimelineio/opentime-bindings/opentime_timeRange.cpp
+++ b/src/py-opentimelineio/opentime-bindings/opentime_timeRange.cpp
@@ -49,7 +49,7 @@ void opentime_timeRange_bindings(py::module m) {
         .def("__copy__", [](TimeRange tr) {
                 return tr;
             })
-        .def("__deepcopy__", [](TimeRange tr) {
+        .def("__deepcopy__", [](TimeRange tr, py::object memo) {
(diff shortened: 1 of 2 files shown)


reinecke:
I did a quick grep and the others seemed to be alright. I only ran into this because I was using a `TimeRange` in a `dataclass` and certain actions in that context trigger a deepcopy. Since `RationalTime` and `TimeRange` are effectively immutable there's no real reason for people to be doing this directly.

Issue #693, opened 2020-05-06
C language bindings


jminor:
We would like to have a C wrapper API around the C++ implementation, which allows for easier integration with C-based applications, or as an intermediate layer for other language bindings where C is better suited than C++ (e.g. Lua, etc.)

jminor:
Here are some examples of C wrapper APIs around C++ implementations:
https://github.com/bulletphysics/bullet3/blob/master/examples/SharedMemory/PhysicsClientC_API.h
https://developer.blender.org/diffusion/B/browse/master/intern/opencolorio/

If anyone knows of other examples that could be helpful here, please feel free to suggest them. There are several important choices in terms of memory management, linking, etc. that we should consider carefully.

jminor:
@meshula, @ssteinbach and @KarthikRIyer do any of you have other examples of C wrapped C++ APIs that we can use as reference?

KarthikRIyer:
I don't know about any good standard examples. The links you posted seem to be a good reference.
Here's a very simple C-CPP wrapper I made for my Swift data viz lib https://github.com/KarthikRIyer/swiftplot/tree/master/Sources/AGGRenderer/CAGGRenderer

Also, do we have tests for the CPP core (I don't think we do)? I was thinking of using [googletest](https://github.com/google/googletest) to set up tests for the C bindings. It's based on the xUnit architecture so it'll be similar to the JUnit tests I'll be using for Java bindings. And it's pretty compact.
Any thoughts?

meshula:
cimgui https://github.com/cimgui/cimgui is a particularly interesting one. They use LuaJit and clang to scrape the Dear Imgui C++ headers to automatically generate the C binding.

KarthikRIyer:
This looks interesting. 
I'm looking at this: http://samanbarghi.com/blog/2016/12/06/generate-c-interface-from-c-source-code-using-clang-libtooling/

I'll try to see if this works.
So currently building llvm-clang. The apt-get install of llvm and clang don't seem to provide the libtooling stuff.

KarthikRIyer:
@meshula could you please note here and elaborate a little on the sample you mentioned in the TSC meeting to demonstrate the bindings?

jminor:
Done! Thanks @KarthikRIyer :)

PR #696, opened 2020-05-08
Initialize C language bindings [WIP]

KarthikRIyer:
#693

I've made a prototype for the C bindings. This covers some portion of the `timeRange.h` file in opentime. I have generated these bindings using this: https://github.com/KarthikRIyer/cpp2c . This is just hacked together. I can refine it once I know how we need to proceed.

I still haven't properly mapped enums and haven't mapped structs,copy constructors or overloaded operators to C. This just maps public methods and the main constructor.

I have used googletest as the testing framework. The test for currently mapped functions pass so the bindings do work.

There's more stuff to figure out.
How can we make the cpp2c program more easy to use. Currently we need to specify the classes and enums in a vector to look for prior to building it. Also we need to specify the c++ standard include directories in the command line arguments. You'll see that if you take a look at the link above. I have at the moment appended the directories to the `**argv` according my system. 

Does anyone know any way to get all these include directories from CMake to C++.

#### What I have tried till now:
After some googling I found that `CMAKE_CXX_IMPLICIT_INCLUDE_DIRECTORIES` gives these directories. One way could be to pass this to C++ by doing this in CMake:

```
add_definition(-DCXX_LIB_PATHS=${CMAKE_CXX_IMPLICIT_INCLUDE_DIRECTORIES})
```

and then using the STRINGIFY macro to convert this to string and then append them to the command line arguments.
The problem here is that there are multiple directories in `CMAKE_CXX_IMPLICIT_INCLUDE_DIRECTORIES` and using the `add_definition` method only the first directory is passed into the C++ code. I have verified the existence of multiple directories by printing the value of the variable in CMake itself using `message(STATUS ${CMAKE_CXX_IMPLICIT_INCLUDE_DIRECTORIES})`

@jminor @ssteinbach @meshula @reinecke any thought on this and how we can proceed?


meshula:
Looking good already! I have a suggestion about structure. At the moment, you've got the includes as e.g. include/opentime.h and then you install them as e.g. include/copentimelineio/opentime.h. This is good from the point of view of the installation, but in my experience it's a better practice to have the installation structure and the source structure the same. So in this case, I'd suggest include/coptentimelineio/opentime.h. This lets you write #Include "copentimelineio/opentime.h" in your implementation, and absolutely eliminates any possible confusion about where the compiler, or the interested reader, is to find opentime.h.

meshula:
Thanks for reorganizing the headers!

KarthikRIyer:
There's one thing I'd like to discuss. Do we want just the C-bindings as the deliverable or the generator too. The reason I'm bringing this up is that I'm thinking about the viability of the generator. There will be cases that will be difficult to handle using the generator. For example templates and nested classes/structs etc. Also, we won't be to add header includes using the generator. A few functions in timeRange return rationalTime so we need to include the rationalTime header.
I have handled a few things like converting between string and char*, passing enum values etc. Just wanted to clarify the goal before moving ahead.

meshula:
This definitely needs discussion before diving in. One avenue that we should discuss as a team, is to invert the C and C++ relationship. There is a conversation right now about representation of rational time, c.f. @reinecke 's issue https://github.com/PixarAnimationStudios/OpenTimelineIO/issues/682 points out the troublesome aspect of dealing with double as the representation of rate. There is the additional need to make a distinction between continuous time representations, as are necessary for real world time measurements for mechanical devices like themepark rides, and discrete time representations for when we are dealing strictly with frame numbers. If we were to invert the relationship, such that the C++ interface is an ease-of-use gloss over a C interface, we could proceed along these lines:

```cpp
#include <stdint.h>

struct OpenTimeRatio
{
    uint32_t numerator, denominator;
};

bool OpenTimeRatio_IsValid(OpenTimeRatio const*const r)
{
    return !r || r->numerator <= 0 || r->denominator <= 0
}

OpenTimeRatio OpenTimeRatio_fromDouble(double r)
{
    // @TODO calculate minimum rational representation of r
    return {1,1};
}

struct OpenTimeContinuousTime
{
    double value;
    struct OpenTimeRatio ratio;
};

struct OpenTimeDiscreteTime
{
    uint64_t value;
    struct OpenTimeRatio ratio;
};

namespace opentime { namespace OPENTIME_VERSION  {

enum IsDropFrameRate : int {
    InferFromRate = -1,
    ForceNo = 0,
    ForceYes = 1,
};
    

template <typename TimeType>    
class RationalTime {
public:
    explicit RationalTime(typeof(TimeType>::value) value = 0, OpenTimeRatio rate = {1, 1})
    : _value {value}, _rate {rate} {}

    RationalTime(RationalTime const&) = default;
    RationalTime& operator= (RationalTime const&) = default;

    bool is_invalid_time() const {
        return OpenTimeRatio_IsValid(&_time.rate);
    }
    
    typeof(TimeType::value) value() const {
        return _time.value;
    }

    OpenTimeRatio rate() const {
        return _time.rate;
    }

```
Once again, I am not proposing this as the actual solution, but I am proposing that we should have this discussion first.

KarthikRIyer:
I'm not sure I understand what Inverting the C and C++ relationship means. Also, in otio are we not always dealing with frames? Why would we need a continuous representation?

Would it be possible to schedule a short zoom call to discuss these issues, the points you mentioned in the time range operators PR and refine the goals for the summer?

meshula:
Yes, we should have a zoom meeting with the team, as what I'm talking about here is a discussion point.

What I mean by "inverting" the relationship is this ~ C by its nature is a lowest common denominator for bindings to all languages ~ it's generally a trivial matter to create bindings to a C api. C++ by its nature is a high level language, with higher level constructs that by their nature are not trivial to bind to other languages as its concepts are in some cases inherently alien.

If we were to consider making a C interface as the base level interface which are wrapped in C++, rather than trying to shoehorn a C++ interface into C, implementations might be easier all around. That's the thought.

Currently, OpenTime deals with continuous representations. The RationalTime is defined as a double for the value, and a double for a rate. The notion of creating a C based OpenTimeContinuousTime is to preserve the current behavior of OpenTime which is continuous. The notion of creating a C based OpenTimeDiscreteTime with a uint64_t as the value is to acknowledge that OpentimelineIO should be dealing with discrete frames in many cases, and in those cases, having an explicit type that enforces discrete values could be a powerful tool for reasoning about, and implementing operations without having to fall back on ad hoc mechanisms to approximate discreteness, such as epsilon comparison parameters in functions.

meshula:
To further elaborate on my thought process here - a clear issue you've pointed out is that C functions can't return C++ objects. That's what led to my thought in the code snippet above where I've created a C structure with the useful data in it, as this struct can be passed around and consumed by both C and C++. Templating on that struct for C++ then allows all the members to be reintroduced in a way that treats the C++ class as a first class citizen with members, but dependent on the underlying C to implement details.

KarthikRIyer:
I understand. If we consider inverting the relationship that would mean reimplementing it in C and would require significant rescoping of the project. @jminor when could it be possible to schedule a meeting?

meshula:
Hi Karthik, before proceeding on the C bindings, we'd like to all get together as a team, and coordinate on how to move forward. We'll coordinate on the email thread started by @jminor!

KarthikRIyer:
I've removed a redundant test in `test_opentime.py`.

RationalTime and ErrorStatus headers are essentially done. Right now the overloaded methods are translated by adding an incrementing index like `func_1`, `func_2` and so on. I think I should change this to something more meaningful. 

I'm specifically confused about this case (b/w the 2nd and 3rd constructor):
```C
ErrorStatus *ErrorStatus_create() {
    return reinterpret_cast<ErrorStatus *>( new opentime::ErrorStatus());
}
ErrorStatus *ErrorStatus_create_1(Outcome in_outcome) {
    return reinterpret_cast<ErrorStatus *>( new opentime::ErrorStatus(
            static_cast<opentime::v1_0::ErrorStatus::Outcome>(in_outcome)));
}
ErrorStatus *ErrorStatus_create_2(Outcome in_outcome, const char *in_details) {
    return reinterpret_cast<ErrorStatus *>( new opentime::ErrorStatus(
            static_cast<opentime::v1_0::ErrorStatus::Outcome>(in_outcome), in_details));
}
```
Any ideas?

meshula:
Instead of 1, 2, I suggest somewhat reflecting the parameter list, eg.

Of course that would be tough to automate.

```cpp
ErrorStatus *ErrorStatus_create() {
    return reinterpret_cast<ErrorStatus *>( new opentime::ErrorStatus());
}
ErrorStatus *ErrorStatus_createUsingOutcome(Outcome in_outcome) {
    return reinterpret_cast<ErrorStatus *>( new opentime::ErrorStatus(
            static_cast<opentime::v1_0::ErrorStatus::Outcome>(in_outcome)));
}
ErrorStatus *ErrorStatus_createUsingDetails(Outcome in_outcome, const char *in_details) {
    return reinterpret_cast<ErrorStatus *>( new opentime::ErrorStatus(
            static_cast<opentime::v1_0::ErrorStatus::Outcome>(in_outcome), in_details));
}
```

KarthikRIyer:
Sounds good. I had thought of using the parameters in the name, but was confused about the `createUsingDetails` constructor, if it should also reflect that an outcome needed to be passed in.

meshula:
Feel free to make a proposal :) Apple uses With, so perhaps thats a precedent to follow. Also notice the rest of the API is using snake_case so

```create_with_outcome``` and ```create_outcome_and_details```

anyone have any thoughts on this?

File: src/c-opentimelineio/opentime-bindings/include/copentime/rationalTime.h [hunk 1]
```diff
@@ -1,42 +1,57 @@
 #ifdef __cplusplus
-extern "C"{
+extern "C"
+{
 #endif
-#include <stdbool.h>
 #include "copentime/errorStatus.h"
-struct      RationalTime;
-typedef     struct RationalTime RationalTime;
+#include <stdbool.h>
```
meshula, line 6: #includes shouldn't be contained within an extern "C" block, as it affects the linkage.

src/c-opentimelineio/opentime-bindings/include/copentime/timeRange.h
```diff
@@ -0,0 +1,39 @@
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+#include "rationalTime.h"
```
meshula, line 5: #Include within an extern block

src/c-opentimelineio/opentime-bindings/include/copentime/rationalTime.h
KarthikRIyer: Oh yes. I'll make the change. Negligence on my part.

KarthikRIyer:
Going from float to double should happen without any problems, right?
RationalTime stores values as double but if we pass in floats in the constructor that shouldn't that work fine?

While writing tests this didn't happen in one place.

```C
    float        d1           = 0.3;
    float        d2           = 0.4;
    RationalTime* zeroTime     = RationalTime_create(0, 1);
    RationalTime* d1Time       = RationalTime_create(d1, 1);
    RationalTime* d2Time       = RationalTime_create(d2, 1);
    RationalTime* d1Plusd2Time = RationalTime_create(d1 + d2, 1);
    TimeRange*    r1 =
        TimeRange_create_with_start_time_and_duration(zeroTime, d1Time);
    RationalTime* r1_ete = TimeRange_end_time_exclusive(r1);
    TimeRange*    r2 =
        TimeRange_create_with_start_time_and_duration(r1_ete, d2Time);
    TimeRange* full =
        TimeRange_create_with_start_time_and_duration(zeroTime, d1Plusd2Time);

    EXPECT_FALSE(TimeRange_overlaps_time_range(r1, r2));
    TimeRange*    r1Extendedr2 = TimeRange_extended_by(r1, r2);
    RationalTime* ex_start     = TimeRange_start_time(r1Extendedr2);
    RationalTime* ex_ete       = TimeRange_end_time_exclusive(r1Extendedr2);
    std::cout << RationalTime_value(ex_ete) << std::endl;
    std::cout << RationalTime_rate(ex_ete) << std::endl;
    std::cout << RationalTime_value(d1Plusd2Time) << std::endl;
    std::cout << RationalTime_rate(d1Plusd2Time) << std::endl;
    EXPECT_TRUE(RationalTime_equal(ex_start, zeroTime));
    EXPECT_TRUE(RationalTime_equal(ex_ete, d1Plusd2Time));
    EXPECT_TRUE(TimeRange_equal(r1Extendedr2, full));
```

These conditions failed:
`EXPECT_TRUE(RationalTime_equal(ex_ete, d1Plusd2Time));`
`EXPECT_TRUE(TimeRange_equal(r1Extendedr2, full));`

whereas the start time condition passed.

I printed out these values as you can see above and they matched. But still the equality condition failed. It worked like a charm as soon as I made `d1` and `d2` doubles.

Any idea why this happened?

KarthikRIyer:
I'll probably finish up with c-opentime today and begin with c-opentimelineio. I was thinking of setting up CI for this. I think GitHub Actions is a good option considering that it is easy to get started with and we might eventually move to it after the ASWF transition. Should I go ahead? Any thoughts?

KarthikRIyer:
I think c-opentime should be done. Could someone please take a closer look at the PR and suggets changes?

If you're ok with adding GitHub Actions for this I'd like to get this merged before moving on to c-opentimelineio. So that we can test it.

KarthikRIyer:
I've started with c-opentimelineio.
First I've added headers for Any and AnyDictionary. Any has only an int constructor for now. I just wanted to see if the way I've written it works or not. I'll change it as I implement more stuff.

I'd like to confirm if this way is fine, becuase I had a very brief look at the swift bindings and it seems to so it some other way with some properties being defined explicitly. Can anyone please take a closer look at this and my previous work?

I've wrapped AnyDictionary in C just like everything I did for opentime. So, I've wrapped some common methods that people might use for `map`, like `size`, `insert`, getting `begin` and `end` iterators, etc. And as I can't define custom operators and had no way to implement the ++ operator for the iterator, I wrapped the `advance` method.

Maybe we could discuss this in the meeting.

davidbaraff:
The Swift bindings are piggy-backing off the fact that we can use Objective-C++ as a bridge, and get free translation from some Objective-C routines into Swift.  I’m not sure how Swift does stuff teaches one how to do a pure C binding or not.

KarthikRIyer:
I haven't used Obj-C++ before,.but I've written swift binding with C bridging headers. So it thought maybe it might be similar.

src/c-opentimelineio/opentime-bindings/include/copentime/errorStatus.h
```diff
@@ -0,0 +1,31 @@
...
+        OK                                   = 0,
+        INVALID_TIMECODE_RATE                = 1,
+        NON_DROPFRAME_RATE                   = 2,
+        INVALID_TIMECODE_STRING              = 3,
+        INVALID_TIME_STRING                  = 4,
+        TIMECODE_RATE_MISMATCH               = 5,
+        NEGATIVE_VALUE                       = 6,
+        INVALID_RATE_FOR_DROP_FRAME_TIMECODE = 7,
+    } Outcome_;
+    typedef int  Outcome;
```
ssteinbach, line 22: Ok, so unfortunately because C doesn't have namespaces, we should include them in both the enum type name and the enumerations.   Can we use this pattern for the other enums as well?

```suggestion
    typedef enum
    {
        OTIO_ErrorStatus_Outcome_OK                                   = 0,
        OTIO_ErrorStatus_Outcome_INVALID_TIMECODE_RATE                = 1,
        OTIO_ErrorStatus_Outcome_NON_DROPFRAME_RATE                   = 2,
        OTIO_ErrorStatus_Outcome_INVALID_TIMECODE_STRING              = 3,
        OTIO_ErrorStatus_Outcome_INVALID_TIME_STRING                  = 4,
        OTIO_ErrorStatus_Outcome_TIMECODE_RATE_MISMATCH               = 5,
        OTIO_ErrorStatus_Outcome_NEGATIVE_VALUE                       = 6,
        OTIO_ErrorStatus_Outcome_INVALID_RATE_FOR_DROP_FRAME_TIMECODE = 7,
    } OTIO_ErrorStatus_Outcome;
```

tests/test_opentime.py
```diff
@@ -555,14 +555,6 @@ def test_seconds(self):
...
         self.assertAlmostEqual(otio.opentime.to_seconds(t4), s3)
 
-        v3 = 3459
-        r3 = 24
-        s3 = float(3459) / 24
-        t3 = otio.opentime.RationalTime(v3, r3)
-        t4 = otio.opentime.from_seconds(s3)
-        self.assertAlmostEqual(otio.opentime.to_seconds(t3), s3)
-        self.assertAlmostEqual(otio.opentime.to_seconds(t4), s3)
-
```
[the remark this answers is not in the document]
KarthikRIyer: I thought this was redundant. The same test is already there just above it. Is there a reason for that?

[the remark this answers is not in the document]
ssteinbach: Ah!  ok makes sense.  Can you separate this into its own PR though?  Thats a separate thing from the rest of this.

Sorry!  Just helps us keep track of everything going on with the project.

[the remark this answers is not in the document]
KarthikRIyer: Sure

src/c-opentimelineio/opentime-bindings/src/timeRange.cpp
```diff
@@ -0,0 +1,130 @@
...
+            *reinterpret_cast<opentime::RationalTime*>(start_time)));
+    }
+    TimeRange* TimeRange_create_with_start_time_and_duration(
+        RationalTime* start_time, RationalTime* duration)
+    {
+        return reinterpret_cast<TimeRange*>(new opentime::TimeRange(
+            *reinterpret_cast<opentime::RationalTime*>(start_time),
+            *reinterpret_cast<opentime::RationalTime*>(duration)));
+    }
+    RationalTime* TimeRange_start_time(TimeRange* self)
```
jminor, line 25: For efficiency, we could also include functions like this, where there is no new allocation, as long as the caller passes in a pre-allocated RationalTime.

ErrorStatus TimeRange_copy_start_time(TimeRange* self, RationalTime* output)
{
    *output = self->start_time();
    return SUCCESS;
}

The error checking part is a separate issue to the memory allocation part...

src/c-opentimelineio/opentimelineio-bindings/include/copentimelineio/typeRegistry.h
```diff
@@ -0,0 +1,48 @@
...
+#endif
+    struct TypeRegistry;
+    typedef struct TypeRegistry TypeRegistry;
+    TypeRegistry*               TypeRegistry_instance();
+    _Bool                       TypeRegistry_register_type(
+                              TypeRegistry* self,
+                              const char*   schema_name,
+                              int           schema_version,
+                              TypeInfo*     type,
+                              function*     create,/*how to pass such functions*/
```
KarthikRIyer, line 19: Even if we wrap std::function like we've done with the other classes, is there a way we can pass it from C to C++? I don't think we can create such function pointers in C.

KarthikRIyer: Also without being able to use classes and therefore not being able to extend to them, would it be possible to define schemas in C?

KarthikRIyer:
@jminor @ssteinbach @meshula @reinecke could you please take a brief look at his to see if I'm going in the correct direction?

src/c-opentimelineio/opentimelineio-bindings/include/copentimelineio/typeRegistry.h
meshula: In C, the idiomatic way to deal with function pointers is to typedef them.

e.g. ```typedef int (*FuncPtr)(int, int); FuncPtr ptr = &ActualFunction;```

There are common approaches to schemas and classes in C, c.f. COM. The larger question is how they would be used in the C OTIO API. Perhaps a good next exercise would be to write a little snip of code that exercises the registry in a hypothetical way, rather than strictly wrapping the C++ interfaces. That would give us a discussion point.

meshula:
Hi there, we're on holiday for a few days :) I have a few notes for you. When there are comments left inline, there's a button down below that says "Resolve Conversation". If you address the comment, would you mind leaving a note like "applied change" or "I have more questions". If you've applied the change, go ahead and hit the "Resolve" button. If we want to reopen it again, we can, so don't worry about messing up the workflow.

I would like to suggest that instead of merging into ```PixarAnimationStudios:master```, we merge into a feature branch to help organize the work and revisions until it's complete. In other words, let's merge to a new branch, ```PixarAnimationStudios:c-otio```.

src/c-opentimelineio/opentimelineio-bindings/include/copentimelineio/typeRegistry.h
KarthikRIyer: > schemas and classes in C, c.f. COM

What is COM? :sweat_smile: 

I still don't have a proper idea of how the type registry works and how exactly the schemas are defined. Are there tests/examples of using the registry with the C++ API. The docs only focuses on python.

src/c-opentimelineio/opentime-bindings/src/timeRange.cpp
KarthikRIyer: This is for another day, so resolving the conversation.

src/c-opentimelineio/opentime-bindings/include/copentime/errorStatus.h
KarthikRIyer: Applied change

KarthikRIyer:
@meshula 
Apart from the function pointer discussion, other conversations have been resolved. Maybe we can land this in `PixarAnimationStudios:c-otio` and continue that discussion in the GitHub issue and then in a new PR.

KarthikRIyer:
I'll probably finish wrapping all the headers soon. Hopefully by tomorrow. Can we land this PR after that? We can fine tune stuff if required and add tests in subsequent PRs.

meshula:
I created a new branch ```c-otio``` to receive your PR. Please rebase against that branch so we can proceed with landing as a feature branch for iteration.

meshula:
Thanks for the rebase :) Let's circle back with the team tomorrow then push on to the next steps!

KarthikRIyer:
The wrapping is done. I think whats left is the `safely_typed_any` header, type registry doubt that I had, adding tests.

Can someone take a look if this seems fine and if anything else is left?

I think we can land this in the feature branch now, and add stuff in new PRs.

src/c-opentimelineio/opentime-bindings/include/copentime/rationalTime.h
```diff
@@ -0,0 +1,60 @@
...
+#endif
+    struct RationalTime;
+    typedef struct RationalTime RationalTime;
+
+    typedef enum
+    {
+        InferFromRate = -1,
+        ForceNo       = 0,
+        ForceYes      = 1,
+    } IsDropFrameRate_;
```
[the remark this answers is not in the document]
meshula: feature branch created.

PR #697, opened 2020-05-09
Add support for predicates from Allen's Interval Algebra for TimeRange

### LINKED ISSUE #659: Support the predicates from Allen's Interval Algebra for `TimeRange`
In https://cse.unl.edu/~choueiry/Documents/Allen-CACM1983.pdf
wikipedia summary: https://en.wikipedia.org/wiki/Allen%27s_interval_algebra

Allen defines a set of logical predicates for time ranges that we should implement in OTIO:

- [ ] before 
- [ ] meets
- [x] overlaps
- [ ] begins
- [x] during/contains
- [ ] finishes
- [x] equals

Note that in our case we have to consider the end point clusivity of the `TimeRanges`.

KarthikRIyer:
Closes #659 

Operators implemented:
- before
- meets
- begins
- finishes

@jminor @ssteinbach @reinecke @meshula please take a look.


src/opentime/timeRange.h
```diff
@@ -82,6 +82,34 @@ class TimeRange {
         return _start_time < other.end_time_exclusive() && other._start_time < end_time_exclusive();
     }
 
+    bool before(TimeRange other) const {
```
meshula, line 85: This is a question for the team at large. Is there a reason TimeRange uses pass-by-value instead of const references throughout this file?

src/opentime/timeRange.h
```diff
@@ -82,6 +82,34 @@ class TimeRange {
...
+        return _start_time < other._start_time && end_time_exclusive() <= other._start_time;
```
meshula, line 86: This is a question for the team at large. Is there any reason why a time range could ever have a start time greater than then end time? If no, through this file, there are comparisons such as this one where the _start_time < other._start_time is redundant with the subsequent check and can be eliminated. If the start IS allowed be greater than the end, then this entire file needs to be scrubbed so that all permutations of unordered TimeRanges are accommodated. If it is disallowed, then we also need code that enforces end >= start.

src/opentime/timeRange.h
```diff
@@ -82,6 +82,34 @@ class TimeRange {
...
+    bool begins(RationalTime other) const {
+        return _start_time == other;
+    }
+
+    bool finishes(TimeRange other) const {
+        return end_time_exclusive() == other.end_time_exclusive();
+    }
+
+    bool finishes(RationalTime other) const {
+        return end_time_exclusive() == other;
```
meshula, line 110: We are asking for end time exclusive in all cases for these comparisons. We should state, in the documentation what the extended rules for Allen's Algebra are for clusivity. For example, if both ranges are inclusive at both ends, and the operation is "meets" we have to decide if that is "meet" or an "overlap". I would request that a prerequisite for this part of the project is to write up a table with all the operators, versus end point clusivity combinations. Also, the associated unit test should exercise the complete table to validate that the operations conform to the clusivity described in the table.

src/opentime/timeRange.h
```diff
@@ -82,6 +82,34 @@ class TimeRange {
...
+    bool before(RationalTime other) const {
+        return end_time_exclusive() < other;
+    }
+
+    bool meets(TimeRange other) const {
+        return end_time_exclusive() == other._start_time || _start_time == other.end_time_exclusive();
+    }
+
+    bool begins(TimeRange other) const {
+        return _start_time == other._start_time;
```
meshula, line 98: This is another general comment: floating point equality comparisons are problematic at best. to be genuinely useful, these routines should all have a tolerance value, with a default tolerance value of double 0.f to force exact equivalence. A question for all is whether tolerance is straight up tolerance, or if there needs to be some sort of clamping behavior in addition to the tolerance. In other words, do we need something like

```
bool begins(TimeRange other, RationalTime tolerance);
```
where tolerance.value is the tolerance value, tolerance.rate is the rate at which to test that tolerance. A typical value might be RationalTime(1./48., 1/24) - half a frame tolerance, with an expected rate of 24 fps. 

Any thoughts on this?

meshula:
Hi Karthik! On this project, let's proceed in this manner:

1) Make a markdown document laying out Allen's Interval algebra, and extend it to describe how it behaves in light of TimeRanges that are (ex, ex) (in, ex) (ex, in), and (in, in).

For example, "before":

```
+--------+-- ex/ex ----------+--in/ex-----------+--ex/in-----------+--in/in--------------------+
| before |  a.end <= b.start | a.end <= b.start | a.end <= b.start | a.end + epsilon < b.start |   
```

In this case, a.end <= b.start is almost always okay, except for when a.end and b.start are inclusive, in which case an epsilon check is necessary.

2) Let's use Test Driven Development to proceed. Once the table is complete, a test for each comparator, and the four clusivity combinations, should be filled out. It would be good to test each combination for a) positive check, false check, and corner cases, such as start equals end on both sides of the comparison. As this set of checks could be combinatorically large, it would perhaps be expedient to be able to generate the tests procedurally.

3) once the test suite exists, fill in the functions one by one

KarthikRIyer:
@meshula do you mean to have 4 implementations of each function, or just implement the comparisons in the test? What is our motive for creating this table?

src/opentime/timeRange.h
KarthikRIyer: Problems generally occur due to rounding errors during arithmetic operations, right? Would it make sense to put in effort and handle it like they do in geometry intersection routines like here: http://www.pbr-book.org/3ed-2018/Shapes/Managing_Rounding_Error.html or is half a frame of tolerance better? And if we're usinf half a frame of tolerance, would we still need to pass in the tolerance value in the function?
Couldn't we just use `RationalTime(0.5/rate(), 1.0/rate())` directly?

src/opentime/timeRange.h
KarthikRIyer: This isn't something I've tested. How is reversing of clips handled in different NLEs n OTIO? Afair in premiere we can change the speed to -100%. So I guess that would be exported as a time warp in OTIO? In this case we won't need to allow greater start times. But this might be different for different software packages.
Other than reversing a clip, I can't think of any other use case for this.

KarthikRIyer:
There's [this java library](https://github.com/Breinify/brein-time-utilities) that implements some part of interval algebra. They ask if the user needs an `openStart` or an `openEnd` in the constructor. Could we consider such an approach and have some good defaults? This library uses closed intervals as a default and uses the same time values for comparison.
Link to relevant part of code [here](https://github.com/Breinify/brein-time-utilities/blob/bb5f91767f0adff22f775a0a72e7bd47b95bc66d/src/com/brein/time/timeintervals/intervals/NumberInterval.java#L38).

meshula:
Thanks for your questions, I should have read the code more closely in constructing my comments, instead of relying on my notions of what was actually there :)

I like the way you are thinking about this problem and driving the task forward, thank you for being proactive!

You're right about the clusivity. I had was thinking too far ahead to a revised version of the API where clusivity is explicitly represented on the object. Right now, the time range class doesn't specifiy the clusivity, and you can make inclusive and exclusive queries on it. I think this change, which you've also found implemented in that java library, is an important part of a revised version of this API.

The table is only necessary if clusivity is specified on the object, as it provides both information to the user about how to use the library, and to the implementor, for a ground truth reference. Since clusivity is not yet explicitly implemented, there's no need yet to create this table.

For the purposes of this PR, it is probably preferable to implement the remaining algebraic operators in the same manner as the existing ones. If we make the larger change, we will need to inspect all of the upstream code, such as file format adaptors, for regressions. So let's hold off on that in favor of getting the much simpler job done.

Sadly, there is no documentation in the RationalTime.h header. Could you incorporate the following notes as comments into the header in the appropriate places?

Please double check that the existing implementations of overlaps and contains match this description.

```

These relations implement James F. Allen's thirteen basic time interval relations.
Detailed background can be found here: https://www.ics.uci.edu/~alspaugh/cls/shr/allen.html

In the relations that follow, epsilon indicates the tolerance,
in the sense that if abs(a-b) < epsilon, we consider a and b to be equal.

equals(a, b, epsilon)
The start of a strictly equals the start of b. The end of b strictly equals the end of b.

precedes(a, b, epsilon)
The end of a strictly precedes the start of b by a value >= epsilon
The converse is precedes(b, a, epsilon)

meets(a, b, epsilon)
The end of a strictly precedes the start of b by a value <= to epsilon
The converse is meets(b, a, epsilon)

overlaps(a, b, epsilon)
The end of a strictly precedes the end of b, the start a of strictly precedes the start of b
The converse is overlaps(b, a, epsilon)

starts(a, b, epsilon)
The start of a equals the start of b. The end of a strictly precedes the end of b
The converse is starts(b, a, epsilon)

finishes(a, b, epsilon)
The start of a strictly antecedes the start of ab. The end of a equals the end of b
The converse is finishes(b, a, epsilon)

contains(a, b, epsilon)
The start of a strictly precedes the start of b. The end of a strictly antecdes the end of b.
The converse is contains(b, a, epsilon)

```

A further "negative" aspect of our TimeRange object, which is highlighted by the your observations on the java library, is that our TimeRange allows construction with a negative duration. However, the logic internally is written as if duration is positive. For the purposes of this PR, let's ignore that a negative duration is possible, and write the code as if a negative duration had been disallowed. Please add a comment such as

```
It is possible to construct TimeRange object with a negative duration. However, the logical predicates are written as if duration is positive, and have undefined behavior for negative durations.
```

Finally, please remember that the unit tests need to exist for these operators. There are tests in place for contains and overlaps for reference.

src/opentime/timeRange.h
meshula: In my response below, I've added an epsilon parameter to the predicates to address this. Thanks!

KarthikRIyer:
@meshula I'll add the comments and change whatever needs to be changed assuming that only positive durations exist.

You have the epsilon parameter in the comments but that isn't implemented currently. Should I add the epsilon to RationalTime? If so, should we go ahead with a margin of half a frame?

KarthikRIyer:
If we say `a meets b` should that mean `b meets a`? In a way doesn't it? So should we return true for both cases.
I say this because that is what happens in `overlaps`

meshula:
I think I can restate your question about epsilon as: "Should epsilon be a value stored on TimeRange, or should it be a parameter on the predicate functions, including the already written ones?"

My answer would be that currently, epsilon is not a quality of the TimeRange, but a quality of the question you are asking of it. I would prefer that we add epsilon as a parameter on the functions. A margin of half a frame as a default is unfortunately poorly defined. For the moment, a default epsilon of 1/1001 would allow us to capture tolerances that are compatible with NTSC frame rates which are multiples of that factor. It is also a number slightly less than one millisecond. We can leave the discussion of a better default for further discussion. For the moment, perhaps specify it as a constexpr, and we can revise it later.

If the end a abuts the start of b:

a meets b if the end of a is within epsilon of b but (a + epsilon) is not greater than or equal to b

meets by definition is this relation:

```
[   a   ][   b  ]
```

The converse relation, is "a is met by b"
```
[  b  ][  a  ]
```

so they are distinct.

Overlaps means
```
   [    a     ]
        [    b    ]
```

Thus -

```
overlaps(a, b, epsilon)
The end of a strictly precedes the end of b, the start a of strictly precedes the start of b
The converse is overlaps(b, a, epsilon)
```

The existing implementations of overlaps and contains are not as strict as the Allen Interval Algebra, unfortunately, which means that they don't function correctly in composed expressions of overlapping and containment.

As part of this task, we'll need to rewrite the existing overlaps and contains to conform to the algebra.

trackAlgorithm.cpp, track_widgets.py, and track_algo.py have one or two uses of contains() and overlaps() that should be spot checked to see if they need to be modified to conform to the revised specification. A quick glance suggests the algorithms might be fine as is, but it would be good to double check.

KarthikRIyer:
I understand. I'll make the changes by tomorrow morning. Meanwhile could you please take a look at the comment formatting and see if it's fine?

src/opentime/timeRange.h
meshula: You're right, a timewarp is the right answer for reversing. TimeRanges should never run backwards on their own.

meshula:
NTSC 23.976 is defined as 24000/1001, NTSC 29.97 is 30000/1001 and so on. That's why I'm proposing 1/1001 since it has fine enough resolution to capture those rather strange rates.

I have a question about the comment style you've implemented, I'm not familiar with it. Is that a format compatible with an automated documentation system such as doxygen?

KarthikRIyer:
I haven't used doxygen before, so I am not aware of its comment style. I used this comment style because I thought it would be easier to read. Does OTIO have a specific comment style, something compatible with doxygen?

meshula:
So far OTIO has a lack of header based documentation. As a future effort such a documentation effort should be undertaken, but for the moment, there's no precedent. What you've done is fine, I was just curious about the styling; the part I hadn't seen commonly was having two leading asterisks on each line. I'm not suggesting anything needs to change, I was just curious about the style choice.

meshula:
BTW, I will give the implementation a closer read after your changes of tomorrow morning. It's looking promising so far!

KarthikRIyer:
For some reason I'm unable to specify the epsilon as a RationalTime. The build is successful but when I import otio in python, it gives me an error:
```
ImportError: /home/karthik/venv3/lib/python3.6/site-packages/opentimelineio/_opentime.cpython-36m-x86_64-linux-gnu.so: undefined symbol: _ZN8opentime4v1_09TimeRange14defaultEpsilonE
````

So I'm using a `double epsilon` and writing the comparisons like `start.value/start.rate - end.value/end.rate >= epsilon`.

Also, using your definition of overlaps the tests fail.
```
[    a     ]
        [    b    ]
```

According to line 888 in `test_opentime.py` ,  `b.overlaps(a)` is true. So I'm keeping the rules same for the moment.

**Comment by KarthikRIyer** [comment]
I've not used epsilon in `contains`. There is a specfic test that check if a timeRange is contained in itself and it expects true. In case I change that some additional opentimelineio tests fail.

**Comment by KarthikRIyer** [comment]
Hmm...the tests pass on my local machine but not on travis. Why is that?

**Comment by KarthikRIyer** [comment]
For before:
if
```
a = TimeRange(start = 10, dur = 2)
b = TimeRange(start = 12, dur = 3)
```
shouldn't `a.before(b)` be true? I know we're using epsilon for rounding errors but in case there's a case like this without a rounding error this will turn out to be false.

**Comment by meshula** [comment]
Yes, epsilon should be a double, not a RationalTime.

As I mentioned earlier, the existing versions of overlaps and contains do not match the definitions in the interval algebra, it's expected that they would fail. Let's open a second issue to correct their operation, and not solve it on this PR. https://github.com/PixarAnimationStudios/OpenTimelineIO/issues/700

```
a = TimeRange(start = 10, dur = 2)
b = TimeRange(start = 12, dur = 3)
```

For the purposes of this PR, let's decide that dur implies an exclusive end, so we can treat the intervals as "inclusive start, exclusive end" for all of the predicates. Under that definition a is indeed before b, so in the case of the "meets" predicate, meets(a, b, eps) is true.

Let's also add a line of documentation at the top of the class:

```
The duration on a TimeRange indicates a time range that is inclusive of the start time, and
exclusive of the end time. All of the predicates are computed accordingly.
```

**Comment by KarthikRIyer** [comment]
>  I was looking at the clusivity functions in timeRange. In end_time_inclusive we're taking the floor or excluding the last frame. Shouldn't that be the case in end_time_exclusive?

This is something I wanted to ask. I might be wrong but I think the exclusive and inclusive parts are interchanged right now.

```
RationalTime end_time_inclusive() const {
                RationalTime et = end_time_exclusive();

                if ((et - _start_time.rescaled_to(_duration))._value > 1) {
                    return _duration._value != floor(_duration._value) ? et._floor() :
                           et - RationalTime(1, _duration._rate);
                } else {
                    return _start_time;
                }
            }

RationalTime end_time_exclusive() const {
            return _duration + _start_time.rescaled_to(_duration);
}
```

So if we want the `a.before(b)` thing to be true, the last frame should be excluded.

@meshula could you please take a closer look at my changes and point out if I'm wrong anywhere?

**Comment by meshula** [comment]
You are right ~ they are swapped!

**Comment by meshula** [comment]
We should swap them, and grep the codebase for usages. Since the unit tests pass currently, it is probably safe to say that all the existing usages need to be swapped.

File: src/opentime/timeRange.h [hunk 5]
```diff
@@ -5,99 +5,246 @@
...
+
+/**It is possible to construct TimeRange object with a negative duration.
+ * However, the logical predicates are written as if duration is positive,
+ * and have undefined behavior for negative durations.
+ *
+ * The duration on a TimeRange indicates a time range that is inclusive of the start time,
+ * and exclusive of the end time. All of the predicates are computed accordingly.
+ */
+
+        class TimeRange {
```
meshula, line 19: minor point, but let's not change the indentation of the file in this PR. If we are changing the indentation, and the code, it is really hard to tell what is new and what is old.

KarthikRIyer:
:+1:  I'll do this first.

src/opentime/timeRange.h
```diff
@@ -5,99 +5,246 @@
...
+             * we consider a and b to be equal
+             */
+
+            /**contains(other)
+             * params: RationalTime other
+             * The start of self precedes other.
+             * other precedes the end of self.
+             */
+            bool contains(RationalTime other) const {
+                return _start_time <= other && other < end_time_exclusive();
```
meshula, line 94: ✔️

src/opentime/timeRange.h
```diff
@@ -5,99 +5,246 @@
...
+            }
+
+            /**contains(other)
+             * params: TimeRange other
+             * The start of self precedes start of other.
+             * The end of self antecedes end of other.
+             * The converse would be other.contains(self)
+             */
+            bool contains(TimeRange other) const {
+                return _start_time <= other._start_time && end_time_exclusive() >= other.end_time_exclusive();
```
meshula, line 104: self contains other. ✔️

src/opentime/timeRange.h
```diff
@@ -5,99 +5,246 @@
...
+            }
+
+            /**overlaps(other)
+             * params: RationalTime other
+             * self contains other
+             */
+            bool overlaps(RationalTime other) const {
+                return contains(other);
```
meshula, line 112: self overlaps other ✔️

src/opentime/timeRange.h
```diff
@@ -5,99 +5,246 @@
...
+            bool overlaps(TimeRange other, double epsilon = defaultEpsilon) const {
+                double selfStart = _start_time.value() / _start_time.rate();
+                double selfEnd = end_time_exclusive().value() / end_time_exclusive().rate();
+                double otherStart = other._start_time.value() / other._start_time.rate();
+                double otherEnd = other.end_time_exclusive().value() / other.end_time_exclusive().rate();
+//                return (otherEnd - selfEnd >= epsilon) &&
+//                       (otherStart - selfStart >= epsilon) &&
+//                        (selfEnd - otherStart >= epsilon);
+//                return _start_time < other.end_time_exclusive() && other._start_time < end_time_exclusive();
+                return (otherEnd - selfStart >= epsilon) &&
```
meshula, line 130: self overlaps other ✔️

src/opentime/timeRange.h
```diff
@@ -5,99 +5,246 @@
...
+
+            /**before(other, epsilon)
+             * params: TimeRange other
+             * The end of self strictly precedes the start of other by a value >= epsilon.
+             * The converse would be other.before(self)
+             */
+            bool before(TimeRange other, double epsilon = defaultEpsilon) const {
+                double selfEnd = end_time_exclusive().value() / end_time_exclusive().rate();
+                double otherStart = other._start_time.value() / other._start_time.rate();
+                return otherStart - selfEnd >= epsilon;
```
meshula, line 142: self before other ✔️

src/opentime/timeRange.h
```diff
@@ -5,99 +5,246 @@
...
+
+            /**before(other, epsilon)
+             * params: RationalTime other
+             * The end of self strictly precedes other by a value >= epsilon.
+             */
+            bool before(RationalTime other, double epsilon = defaultEpsilon) const {
+                double selfEnd = end_time_exclusive().value() / end_time_exclusive().rate();
+                double otherTime = other.value() / other.rate();
+                return otherTime - selfEnd >= epsilon;
+            }
```
meshula, line 153: self before other ✔️

src/opentime/timeRange.h
```diff
@@ -5,99 +5,246 @@
...
+
+            /**meets(other, epsilon)
+             * params: TimeRange other
+             * The end of self strictly equals the start of other and
+             * the start of self strictly equals the end of other.
```
meshula, line 158: delete this line (start of self does not equal end of other, this line is wrong and unneccesary)

src/opentime/timeRange.h
```diff
@@ -5,99 +5,246 @@
...
+             * The converse would be other.meets(self)
+             */
+            bool meets(TimeRange other, double epsilon) const {
+                double selfEnd = end_time_exclusive().value() / end_time_exclusive().rate();
+                double otherStart = other._start_time.value() / other._start_time.rate();
+                return otherStart - selfEnd <= epsilon && otherStart - selfEnd >= 0;
```
meshula, line 164: self meets other ✔️

src/opentime/timeRange.h
```diff
@@ -5,99 +5,246 @@
...
+             * The start of self strictly equals the start of other.
+             * The end of self strictly precedes the end of other by a value >= epsilon.
+             * The converse would be other.begins(self)
+             */
+            bool begins(TimeRange other, double epsilon = defaultEpsilon) const {
+                double selfStart = _start_time.value() / _start_time.rate();
+                double selfEnd = end_time_exclusive().value() / end_time_exclusive().rate();
+                double otherStart = other._start_time.value() / other._start_time.rate();
+                double otherEnd = other.end_time_exclusive().value() / other.end_time_exclusive().rate();
+                return fabs(otherStart - selfStart) <= epsilon && otherEnd - selfEnd >= epsilon;
```
meshula, line 178: self begins other ✔️

src/opentime/timeRange.h
```diff
@@ -5,99 +5,246 @@
...
+            }
+
+            /**begins(other, epsilon)
+             * params: RationalTime other
+             * The start of self strictly equals other.
+             */
+            bool begins(RationalTime other, double epsilon = defaultEpsilon) const {
+                double selfStart = _start_time.value() / _start_time.rate();
+                double otherStart = other.value() / other.rate();
+                return fabs(otherStart - selfStart) <= epsilon;
```
meshula, line 188: self begins other ✔️

src/opentime/timeRange.h
```diff
@@ -5,99 +5,246 @@
...
+             * The start of self strictly antecedes the start of other by a value >= epsilon.
+             * The end of self strictly equals the end of other.
+             * The converse would be other.finishes(self)
+             */
+            bool finishes(TimeRange other, double epsilon = defaultEpsilon) const {
+                double selfStart = _start_time.value() / _start_time.rate();
+                double selfEnd = end_time_exclusive().value() / end_time_exclusive().rate();
+                double otherStart = other._start_time.value() / other._start_time.rate();
+                double otherEnd = other.end_time_exclusive().value() / other.end_time_exclusive().rate();
+                return fabs(selfEnd - otherEnd) <= epsilon && selfStart - otherStart >= epsilon;
```
meshula, line 202: self finishes other ✔️

src/opentime/timeRange.h
```diff
@@ -5,99 +5,246 @@
...
+            }
+
+            /**finishes(other)
+             * params: RationalTime other
+             * The end of self strictly equals other.
+             */
+            bool finishes(RationalTime other, double epsilon = defaultEpsilon) const {
+                double selfEnd = end_time_exclusive().value() / end_time_exclusive().rate();
+                double otherEnd = other.value() / other.rate();
+                return fabs(selfEnd - otherEnd) <= epsilon;
```
meshula, line 212: self finishes other ✔️

tests/test_opentime.py
```diff
@@ -923,6 +923,126 @@ def test_overlaps_timerange(self):
 
         self.assertFalse(tr.overlaps(tr_t))
 
+    def test_before_timerange(self):
```
meshula, line 926: awesome :)

src/opentime/timeRange.h
KarthikRIyer: Will do. Thanks for the review!
Do we use clang_format or something? These indentation changes were done by the autoformat thing in CLion.

meshula: AFAIK we haven't got a clang-format file, but that would be a good thing to have ultimately. Perhaps CLion is using a standard format if none is found in the project?

src/opentime/timeRange.h
```diff
@@ -5,99 +5,246 @@
...
+            }
+
+            static TimeRange range_from_start_end_time(RationalTime start_time, RationalTime end_time_exclusive) {
+                return TimeRange{start_time,
+                                 RationalTime::duration_from_start_end_time(start_time, end_time_exclusive)};
+            }
+
+        private:
+            RationalTime _start_time, _duration;
+            static constexpr double defaultEpsilon = 1.0 / 1001.0;
```
ssteinbach, line 245: Couple notes:

1. I think if we're going to add a constant like this, we should specify the units, both by a comment and probably in the name of the variable itself, so lets use `defaultEpsilon_s`, to indicate that the epsilon is in unit seconds.
2. 1001.0 is probably not small enough to be useful for audio frame rates.  In order to make it useful in those contexts, where frame rates go considerably higher, after some discussion, @meshula and I concluded the value `1/(2.0*192000)` would be better, in order to cover audio rates up to 192khz.

meshula:
@ssteinbach Tells me that we are reading the API backwards with regards to how the inclusive and exclusive methods are to be read. There is clear documentation in the original python implementation, and we should (1) copy that documentation into the .h file, and (2) we therefore don't have to modify the code. I'll close out the other issue. @ssteinbach could you drop the link here in this thread?

ssteinbach:
I put the link to the python documentation of the methods from opentime in an issue that proposes bringing that documentation forward to the new C++ api:
https://github.com/PixarAnimationStudios/OpenTimelineIO/issues/704

To make simple, here is the links to the original implementation w/ docstrings:
https://github.com/PixarAnimationStudios/OpenTimelineIO/blob/last_pure_python/opentimelineio/opentime.py

Sorry about the confusion!

KarthikRIyer:
Oh I see. So the way I've implemented the operators should be alright? I'll make the changes suggested.

Although this is documented in the pure python version, I thought it was swapped when going through the code without the doc. The current choice seems a little unintuitive. Is there a reason for the choice?

ssteinbach:
Naming things is always difficult.  In the original implementation of `TimeRange` it had a single `.end_time()` method, which would compute the exclusive end point of the time range (or the first coordinate that was outside of the range).  We found this confused people - even in our own code base the `end_time()` method was treated as inclusive and sometimes it was treated as exclusive, and there were use cases for both interpretations.  

We split the method up into `end_time_inclusive()`, representing the start of the last region of time of size rate inside the `TimeRange`, and `end_time_exclusive()`, representing the first time that is outside the `TimeRange`.  The benefit being that the clusivity of the point was explicit in the method call.  We tried to document this in the docstrings, but those weren't brought over when the codebase was rebuilt in C++ w/ Python bindings (@meshula added an issue for this).

This split was done in this PR: https://github.com/PixarAnimationStudios/OpenTimelineIO/pull/79

How were you interpreting those names?  Is there a better name you would propose?

KarthikRIyer:
I thought `end_time_exclusive` was `end_time_inclusive` and vice versa. I compared them to exclusive & inclusive class intervals.
https://www.quora.com/What-is-the-difference-between-exclusive-and-inclusive-class-interval

I think it would be easier if the implementations are swapped, but maybe they shouldn't be changed as people must have gotten used to this definition and it must be used at multiple places already.

jminor:
To elaborate on what @ssteinbach said about our thought process in naming those methods: We were thinking of the end time as included inside the clip, or the end time outside the clip.

For example, if you have a clip that runs from frame 10-90, then you often want the end time to be 90 (e.g. the last frame) and other times to be 91, the frame where the next clip should start. 

To see this visually, consider Figure 1 here: https://opentimelineio.readthedocs.io/en/stable/tutorials/otio-timeline-structure.html  Clip-001's source_range covers frames 3, 4, and 5 of Media-001. `end_time_inclusive()` returns 5, which is included in the clip, while `end_time_exclusive()` returns 6 which is outside the clip.

By introducing the two methods `end_time_inclusive()` and `end_time_exclusive()` we hoped to eliminate the various +1 or -1 calculations that we saw in many frame-based code paths. For example, folks using EDLs or image sequences do this all the time. You'll notice that the implementation of `end_time_inclusive()` uses `et - RationalTime(1, _duration._rate)` to go back "one frame" based on the rate.

Now that we're in the mindset of interval algebra, the intuitive sense of "inclusive" and "exclusive" are swapped. The specific concept is similar, but I would expect that the interval algebra's meaning of these two terms concerns the infinitesimal endpoint, not a whole frame at the end.

We could consider renaming these methods using different words (inside, outside?) if we can make the meaning clear. This is also an area where mixing continuous time with discrete frame-based time is causing confusion.

src/opentime/timeRange.h
meshula: added an issue for having a clang format file

meshula:
maybe .... **end_time_inclusive**() could be **range_end_boundary**(), and **end_time_exclusive**() could be **range_end_final_frame**(double rate)   ?

ssteinbach:
end_time_exclusive is not the "final frame" of the range, its the start of the first "frame" (really the first continuous time coordinate) that is outside of the range, and TimeRange itself has some utility beyond "Frames".  I think we should hold off on renaming it, if we rename it at all, until finishing up the discrete/continuous math design that we're working on.  We can fold this into that.

meshula:
Agreed on deferring the discussion. To be clear, I'm not proposing equivalence between the inclusive/exclusive methods and my proposed names. I'm proposing that the inclusive/exclusive naming is sufficiently vague that perhaps the way forward is to make new functions that are explicit in what they return, whether or not that happens to be an inclusive/exclusive value.

KarthikRIyer:
Would it be alright to add the clang-format file and reformat in this PR or would a separate PR be better?

meshula:
It's a good practice to keep PR's functionally distinct. So the way I think of it is that the formatting file is unrelated to the notion of a TimeRange, therefore it should be a separate PR.

Why don't we continue the discussion of the clang-format in the clang-format issue? https://github.com/PixarAnimationStudios/OpenTimelineIO/issues/706 If you have one you want to propose, could you post it there? We could all grab it and experiment a little.

meshula:
@ssteinbach When you have a moment, could you check the resolution of your requested changes? I think we are close to being able to land this PR!

src/opentime/timeRange.h
```diff
@@ -45,57 +54,199 @@ class TimeRange {
...
     bool contains(RationalTime other) const {
         return _start_time <= other && other < end_time_exclusive();
     }
 
+    /**contains(other)
+     * params: TimeRange other
+     * The start of self precedes start of other.
+     * The end of self antecedes end of other.
+     * The converse would be other.contains(self)
+     */
```
ssteinbach, line 102: I think this style is fine, but we should move to a doxygen style comment.  I'll file an issue for that and describe that as a project there.

src/opentime/timeRange.h
```diff
@@ -5,37 +5,46 @@
...
-    TimeRange& operator= (TimeRange const&) = default;
+    TimeRange &operator=(TimeRange const &) = default;
 
-    RationalTime const& start_time() const {
+    RationalTime const &start_time() const {
         return _start_time;
     }
 
-    RationalTime const& duration() const {
+    RationalTime const &duration() const {
```
ssteinbach, line 37: Did these ampersands move because of the clang format?  Could we not include reformatting with this PR?  It makes it harder to parse the contents specific to this PR vs reformatting the file in general.  Sorry about that, let me know if you need help with git commands to untangle this.

src/opentime/timeRange.h
```diff
@@ -45,57 +54,199 @@ class TimeRange {
...
-        RationalTime end {std::min(r.end_time_exclusive(), end_time_exclusive())};
-        return TimeRange {r._start_time, end - r._start_time};
+        TimeRange r{std::max(other._start_time, _start_time), other._duration};
+        RationalTime end{std::min(r.end_time_exclusive(), end_time_exclusive())};
+        return TimeRange{r._start_time, end - r._start_time};
     }
 
+    /**
+     * These relations implement James F. Allen's thirteen basic time interval relations.
+     * Detailed background can be found here: https://www.ics.uci.edu/~alspaugh/cls/shr/allen.html
```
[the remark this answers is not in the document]
ssteinbach: And rather than link to this article, the link should be to the ACM page, since its the actual publication location:  https://dl.acm.org/doi/10.1145/182.358434

src/opentime/timeRange.h
```diff
@@ -45,57 +54,199 @@ class TimeRange {
...
+                         RationalTime::duration_from_start_end_time(start_time, end_time_exclusive)};
+    }
+
+    static constexpr double getDefaultEpsilon(){
+        return defaultEpsilon_s;
     }
 
 private:
     RationalTime _start_time, _duration;
+    static constexpr double defaultEpsilon_s = 0.5 / 192000.0;
```
ssteinbach, line 249: Two things about this:

1)
```suggestion
    static constexpr double defaultEpsilon_s = 1 / (2*192000.0);
```
This is more clear about the value being the inverse of double 192k.

2) Lets add a comment before it:
```This default epsilon value is used in comparison between floating numbers.  It is computed to be twice 192khz, the fastest commonly used audio rate.  It can be changed in the future if necessary due to higher sampling rates or some other kind of numeric tolerance detected in the library.```

src/opentime/timeRange.h
KarthikRIyer: Yes, this must be because of clang format. Some help with git here would be great! I've been wondering how to make such fixes.

KarthikRIyer:
@ssteinbach 

Changes made:
- Add title and link for Interval Algebra paper
- Change pythonic reference self to this
- Remove dead code
- Change defaultEpsilon_s to DEFAULT_EPSILON_s and change it's scope to namespace
- Document DEFAULT_EPSILON_s
- Use doxygen style comments

Now, only the `&` formatting revert is left. That is included in a commit with other non formatting changes. Is it possible to revert that using git?

src/opentime/timeRange.h
ssteinbach: Don't worry about it, should be fine to include this.

meshula:
Exciting to see this land! Thanks Karthik!

KarthikRIyer:
Currently going through otio-core, working on c-bindings. Will come back to this later!

meshula:
@ssteinbach It looks like you merged this work in https://github.com/PixarAnimationStudios/OpenTimelineIO/commit/c61d309c9b00cedd13ed256db8e9c2b659128bca however, this PR is still open. Are there new commits, and that's why it's open, or do we need to manually close the PR?

PR #702, opened 2020-05-14
Addressing a handful of lint issues.

reinecke:
Somehow lint started failing for people on a lot of their branches due to issues in master, here is one such example: https://travis-ci.com/github/PixarAnimationStudios/OpenTimelineIO/jobs/333466701

Worth noting, there was one function in the kdenlive adapter I didn't know the correct fix for so I just muted the warning. This is likely broken and will require someone with more experience in that realm to address:

```
def write_keyframes(kfdict):
    """Build a MLT keyframe string"""
    # TODO: The following line is causing a lint error:
    #       F524 '...'.format(...) is missing argument(s) for placeholder(s): 1
    # This will likely cause an exception at runtime and should be addressed.
    return ';'.join('{}={}'.format(str(int(t.value))).format(v)  # noqa: F524
                    for t, v in kfdict.items())
```

### DIFF
--- a/contrib/opentimelineio_contrib/adapters/kdenlive.py
+++ b/contrib/opentimelineio_contrib/adapters/kdenlive.py
@@ -164,7 +164,10 @@ def clock(time):
 
 def write_keyframes(kfdict):
     """Build a MLT keyframe string"""
-    return ';'.join('{}={}'.format(str(int(t.value))).format(v)
+    # TODO: The following line is causing a lint error:
+    #       F524 '...'.format(...) is missing argument(s) for placeholder(s): 1
+    # This will likely cause an exception at runtime and should be addressed.
+    return ';'.join('{}={}'.format(str(int(t.value))).format(v)  # noqa: F524
                     for t, v in kfdict.items())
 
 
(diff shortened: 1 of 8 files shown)


contrib/opentimelineio_contrib/adapters/kdenlive.py
```diff
@@ -164,7 +164,10 @@ def clock(time):
 
 def write_keyframes(kfdict):
     """Build a MLT keyframe string"""
-    return ';'.join('{}={}'.format(str(int(t.value))).format(v)
+    # TODO: The following line is causing a lint error:
```
meshula, line 167: Is there any reason not to unpack this line a little? There needs to be a one to one mapping of braces to format arguments, and lint is correct. The join on semicolon is also suspicious, since there is only one string involved. Is the line intending:

```py
return '{}={}'.format(str(int(t.value)), v) + ';'
```
?

ssteinbach: There is more than one string, there is an iteration over the `.items()` of the `kfdict`, and I think this is fine.  (contrib area)

reinecke: Yeah, I agree with you @meshula that this should ideally have a real code fix.
Since this requires attention from the adapter maintainer and is in the contrib area, I figured it's best to just silence the lint error for now to get the code building again. I've filed #703 so this issue doesn't totally disappear.

reinecke:
After discussion, going ahead and merging despite the unaddressed issue in the kdenlive adapter, filed issue as #703 instead.

PR #707, opened 2020-05-16
Add .clang-format file

### LINKED ISSUE #706: add a clang format file
A clang format file would be nice, perhaps we can use the one from OpenEXR as a starting point.

KarthikRIyer:
Closes #706 

I've added a clang format file. This is the one from openEXR with a tiny change. The space before parentheses has been removed.

I'd also like to propose to disable brace wrapping after these declarations (these wrappinngs aren't present in the code currently so disabling them should be comfortable):
- class
- control-statement
- enum
- function
- namespace
- struct
- union
- extern block

Any thoughts or additional changes you'd like to propose?
If everyone agrees then I'll reformat the files and update the PR.

@jminor @meshula @ssteinbach @reinecke

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,120 @@
+---
+Language:        Cpp
+# BasedOnStyle:  LLVM
+AccessModifierOffset: -4
+AlignAfterOpenBracket: AlwaysBreak
+AlignConsecutiveAssignments: true
+AlignConsecutiveDeclarations: true
+AlignEscapedNewlines: Right
+AlignOperands:   true
+AlignTrailingComments: true
+# AllowAllArgumentsOnNextLine: true
+AllowAllParametersOfDeclarationOnNextLine: true
+# AllowAllConstructorInitializersOnNextLine: true
+AllowShortBlocksOnASingleLine: true
+AllowShortCaseLabelsOnASingleLine: true
+AllowShortFunctionsOnASingleLine: InlineOnly
+AllowShortIfStatementsOnASingleLine: true
+AllowShortLoopsOnASingleLine: false
+# AllowShortLambdasOnASingleLine: true
+AlwaysBreakAfterDefinitionReturnType: TopLevel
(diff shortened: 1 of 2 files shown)


meshula:
My preference is to disable bracewrapping. Any other opinions?

ssteinbach:
My personal style preference is to match the EXR style of brace wrapping.  Thats an aesthetic argument, though.

The non-aesthetic argument might be that sticking closer to EXR's style will help make ASWF projects more consistent.

If we're going to add a clang-format file, we may want to also add a unit test (similar to the python lint test) to make sure that new code is formatted with the style sheet.  What do folks think about that?

ssteinbach:
Something like this: https://stackoverflow.com/questions/22866609/can-clang-format-tell-me-if-formatting-changes-are-necessary

KarthikRIyer:
> If we're going to add a clang-format file, we may want to also add a unit test (similar to the python lint test) to make sure that new code is formatted with the style sheet. What do folks think about that?

Yes, we should definitely have a test for correct formatting.

I was also thinking of the open PRs that might be based on an older commit. Rebasing them might give large and ugly conflicts. So what might be a way of handling such a situation.

There's [this blog](https://engineering.mongodb.com/post/succeeding-with-clangformat-part-1-pitfalls-and-planning/) that shows how MongoDB handled the situation. But I haven't gone throught their algorithm to handle it properly yet.

ssteinbach:
Open PRs will need to be rebased, but thats ok.  Before we land this change, we can also make sure there aren't any large PRs left open.

ssteinbach:
To fix the build, you need to add this file to the `MANIFEST.in` file.  That should fix the build.  Sorry about that.  I think you can add it as `exclude`, since the .clang-format doesn't need to be installed by pip.

meshula (approved):
For review at TSC. I propose we land this.

ssteinbach (approved):

PR #710, opened 2020-05-19
Contrib application test

apetrynet:
Hi!

Here is an initial set of tests for the RV reader package.
If accepted, this can be used as a foundation for further testing with the image sequence additions mentioned in the #637 thread.
Found it easier to do it in this order than adding to an existing the PR

The tests are quite simple as most of the OTIO->RV logic is a copy from and tested in the rv_session adapter.


contrib/opentimelineio_contrib/application_plugins/tests/test_rv_reader.py
```diff
@@ -0,0 +1,246 @@
...
+    'rv',
+    'example_otio_reader'
+)
+
+# Need this for connecting to RV
+RV_PYTHON_DIR = os.path.join(RV_ROOT_DIR, 'plugins', 'Python')
+if RV_PYTHON_DIR not in sys.path:
+    sys.path.append(RV_PYTHON_DIR)
+
+import rv_tests_network     # noqa: E402
```
avrata, line 60: Cool! However I was curious about this import as I don't see it in my rv/plugins/Python dir. Is it coming from somewhere else I might not be aware of?

apetrynet: Oh, I'll check if it's in an older release of RV as well. I think the same class I used is available under src in another module

apetrynet: @avrata, there's an `RvComminnicator` class in `src/python/network/rvNetwork.py` but the network folder doesn't contain an `__init__.py`. I think I'll do what others seem to do and add the class to our test files in stead. 
Agree?

avrata: Seems fine with me, however I can't speak for any of the license issues or whatnot with including this code here. Someone closer to the project might need to chime in.

apetrynet: I believe it's what "everyone" does, but would be great to have a thumbs up from someone at RV HQ. The file itself doesn't contain any license.
Should I push the change and revert if we get the thumbs down?

mikepkes: I'm checking with the RV leadership on if this is an acceptable redistribution.

Would you prefer that RV just ships with an importable package that you can point at?  Seems like a low bar here, but if you need to support already distributed RV versions, I can understand.

contrib/opentimelineio_contrib/application_plugins/tests/test_rv_reader.py
```diff
@@ -0,0 +1,246 @@
...
+sample_timeline.tracks.append(track)
+
+
+@unittest.skipIf(
+    "OTIO_RV_ROOT_DIR" not in os.environ,
+    "OTIO_RV_ROOT_DIR not set."
+)
+@unittest.skipIf(
+    (sys.version_info > (3, 0)),
+    "RV OTIO reader plugin does not work in python 3."
```
mikepkes, line 96: Who should we add to our list to try the RV beta for when this needs to support RV + Python3?

contrib/opentimelineio_contrib/application_plugins/tests/test_rv_reader.py
```diff
@@ -0,0 +1,246 @@
...
+                  '{sample_file}' \
+                  .format(
+                      root=RV_BIN_DIR,
+                      port=9876,
+                      sample_file=sample_file.name
+                  )
+        proc = Popen(shlex.split(run_cmd), env=env)
+
+        # Dirty way to wait for RV to launch
+        time.sleep(4)
```
mikepkes, line 184: How could we make this more robust?  I'd like to believe that all RVs would run in less than 4 seconds, but would some sort of semaphore be more reliable?

contrib/opentimelineio_contrib/application_plugins/tests/test_rv_reader.py
apetrynet: For future situations like this it would be great to be able to import directly from RV so we don't get out of sync with any changes. 
It would be nice however to be able to support older versions, so let's see what legal says and deal with it accordingly.
Another option could be to use the imp module on he current rvNetwork file I guess.

contrib/opentimelineio_contrib/application_plugins/tests/test_rv_reader.py
apetrynet: Good point. I'll look at some options. Thanks!

contrib/opentimelineio_contrib/application_plugins/tests/test_rv_reader.py
ssteinbach: @mikepkes did you get this question answered?  If you ping me on the slack I can route you to someone.

PR #714, opened 2020-05-20
Fix #703 mlt keyframe string decoding/encoding

### LINKED ISSUE #703: Kdenlive adapter: write_keyframes function needs attention
The `write_keyframes` function started failing builds due to a lint issue (see #702). The expression `str(int(t.value))).format(v)` seems like it's unlikely to be correct. An integer converted to a string would never be a useful format template so it's unclear what the code is attempting to do.
Additionally the outer format on for the string `"{}={}"` is being passed a single value (a generator) rather than two values.

Code below:
```
def write_keyframes(kfdict):
    """Build a MLT keyframe string"""
    return ';'.join('{}={}'.format(str(int(t.value))).format(v)  # noqa: F524
                    for t, v in kfdict.items())
```

vpinon:
Hello,
Sorry this effects conversion was a forgotten draft (does any effects mapping exist?), it was not tested in the sample project.
I thought of removing the function entirely, but I think keeping it in place can be interesting for future extension...
Thanks for raising the issue!

### DIFF
--- a/contrib/opentimelineio_contrib/adapters/kdenlive.py
+++ b/contrib/opentimelineio_contrib/adapters/kdenlive.py
@@ -55,7 +55,7 @@ def read_keyframes(kfstring, rate):
     which are in a semicolon (;) separated list of time/value pair
     separated by = (linear interp) or ~= (spline) or |= (step)
     becomes a dict with RationalTime keys"""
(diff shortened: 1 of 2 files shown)


reinecke:
Thanks for the Fix!

PR #719, opened 2020-05-27
Openedit Design

mccartnm:
This is the relevant PR for [Issue 771](https://github.com/PixarAnimationStudios/OpenTimelineIO/issues/711)

This lays out the initial design specs for the edit platform within OTIO.


docs/design/editorial_design.md
```diff
@@ -0,0 +1,354 @@
...
+        `- openedit/
+            `- ...
+```
+
+The benefit for the extra library is the edit suite will inevitably grow and will begin crowding the core model, which ultimately should be able to run without it.
+
+## Dependencies / Namespaces
+The `openedit` library will have to communicate with both `opentime` and `openedit`.
+
+This will warrant either a new namespace `namespace openedit { ... }` or using `opentimelineio` again. The latter makes some of the bellow comments less challenging but again, might pollute an otherwise clean namespace.
```
meshula, line 35: typo 'below'

docs/design/editorial_design.md
```diff
@@ -0,0 +1,354 @@
...
+...
+```
+
+# Editing Commands
+Let's go through some of the most common commands and what the python API might look like:
+
+## Overwrite
+
+![Overwrite](../_static/edit/01_overwrite.png)
+- Anything overlapping is destroyed on contact. This may split an item.
```
meshula, line 65: It would be nice to specify things more pedantically using the nomenclature we've established in the recent Allen's Interval Algebra commit. eg.

```
for all items i
   if e contains i, or i begins e, or e ends i : delete i
   if i contains e : split i at e.start, yielding {i0, i1}. Split i1 at en.end, yielding {i2, i3}, delete i, i1 and i2, return i0 and i3.
  if i overlaps e...
  else skip 
```

meshula: You've specified the rest of the API more formally. I felt, given our new algebraic nomenclature, overlapping wasn't specific enough, also destroyed on contact was open to a bit of interpretation..

docs/design/editorial_design.md
```diff
@@ -0,0 +1,354 @@
...
+        for (auto it = completed_events.rbegin(); /*...*/)
+            (*it).revert();
+        break;
+    }
+    completed_events.push_back(event);
+}
+```
+
+## Overall
+Many of the commands mentioned have common code paths and math that is required. We can streamline many of the placement commands into a single call with different options. We can then expose that as a raw edit platform while giving users the common algorithms for ease-of-use.
```
meshula, line 353: I personally lean away from "uber" functions, and more to concise, does-one-thing functions. Do you have something specific in mind here, or a mock-up example?

docs/design/editorial_design.md
```diff
@@ -0,0 +1,354 @@
...
+    track_time: RationalTime,
+    fill_template: Item = None, # Default to Gap
+)
+```
+- In the image, Clip `C` is `item` and `track` is the what `A` and `B` are on.
+
+----
+
+## Insert
+![Insert](../_static/edit/02_insert.png)
```
meshula, line 90: The pngs are very nice, but I'm wondering if we could get some inline ASCII art as well :)

docs/design/editorial_design.md
```diff
@@ -0,0 +1,354 @@
...
+    Retainer<Composition> parent;
+    Retainer<Item> composable;
+
+    // ... Additional fields to execute the above as required
+
+    bool run(ErrorStatus* error_status);
+    bool revert();
+};
+```
+An event is atomic in nature and does "one thing" to an `Item`. Each edit maneuver (e.g. `otio.algorithms.overwrite(...)`) would generate a vector of these events that can be played forward and, possibly, backward. The result of them collectively is the commands result.
```
ssteinbach, line 338: Lets factor this out into a second PR.  It is an orthogonal concept and also relevant to other non-edit api operations in OTIO, so we should design that as its own piece.

docs/design/editorial_design.md
```diff
@@ -0,0 +1,354 @@
...
+- Overwrite
+- Trim
+- Cut
+- Slip
+- Slide
+- Ripple
+- Roll
+- Fill (3/4 Point Edit)
+
+OpenTimelineIO, while not seeking to become a fully realized NLE, could benefit from having a simple toolkit to help automate building/modifying timelines based on time rather than index alone.
```
ssteinbach, line 16: I think this could be rephrased to be closer to OpenTimelineIO's scope... maybe something like:

> OpenTimelineIO includes a suite of tools for operating on objects as raw data structures (append, etc), but is missing a library for performing higher level operations familiar to people from the editing world (Slip, 3/4 point edit, etc).  This proposal presents an API for working with OpenTimelineIO objects in those ways.

I don't think they need to be enumerated here, since you're going to be covering them in more detail later.

docs/design/editorial_design.md
```diff
@@ -0,0 +1,354 @@
...
+
+## Structure
+The underlying implementation for these edit commands will be in C++, possibly under a new library called `openedit`. This has the benefit of connecting with other languages more simply (Python, Swift, C, etc.) and, hopefully, we only need expose the functionality to the bindings.
+
```
ssteinbach, line 20: Do you mean that you'd want a separate c++ DSO?  To sit next to OpenTime.so and OpenTimelineIO.so?  I kind of think it just belongs in the regular OpenTimelineIO library, but with its own headers.  Or are you saying you'd _just_ wrap this subset of the library and not need bindings to OpenTimelineIO?

I would probably just put them under OpenTimelineIO, possibly combined with other operations into an 'algorithms' folder:

```
- opentimelineio/
         edit/
or 
         algo/
```

Things like flatten are likely peers of this and I'd like to avoid too many namespaces for people to hunt through.

In python, I was thinking a similar arrangement:
`opentimelineio.algorithms.slip()`
etc.

Curious to hear your thoughts.

docs/design/editorial_design.md
```diff
@@ -0,0 +1,354 @@
...
+        Options:
+            - Source   : Don't modify the source, overwrite if required
+            - Sequence : Don't modify the sequence, trim item if required
+            - Fit      : Apply Time Effect
+"""
+otio.algorithms.fill(
+    item: Item,
+    track: Track,
+    replace: Item = None, # Alternatively, we could have a parent-coord
```
reinecke, line 312: This feels like this should maybe be a `TimeRange` within the track. If there is just a single item to operate on the caller can use `range_in_parent` to get the value they want to use.

docs/design/editorial_design.md
```diff
@@ -0,0 +1,354 @@
...
+
+} }
+```
+
+So perhaps establishing some common aliases early might help?
+```
+using RetainedItem = SerializedObject::Retainer<Item>;
+using RetainedComposition = SerializedObject::Retainer<Composition>;
+...
+```
```
ssteinbach, line 57: `RetainedItem` seems like a reasonable alias, however, reading this feels like it should all be in the same namespace.  The edit library is going to require both OpenTimelineIO and opentime to function, and you'll need constructors, classes, etc from those to setup objects and metadata-- it seems like this might be more analogous to the split in imath between for example ImathBox.h and ImathBoxAlgo.h:
https://github.com/AcademySoftwareFoundation/openexr/blob/master/IlmBase/Imath/ImathBox.h
https://github.com/AcademySoftwareFoundation/openexr/blob/master/IlmBase/Imath/ImathBoxAlgo.h

docs/design/editorial_design.md
```diff
@@ -0,0 +1,354 @@
...
+
+---
+
+## Trim
+![Trim](../_static/edit/03_trim.png)
+- Adjust a single item's start time or duration.
+- Do not affect other clips.
+- Fill now-empty time with gap or template.
+- Clamps source_range to non-gap boundary (can only overwrite gaps)
+
```
ssteinbach, line 116: if source range isn't set, does the available range become the source range before modifying the values?

This is assuming the item is in track, is it an error condition if the item has no parent?

If the item is directly in a stack, is a track inserted wrapping the item so that gap can be put before or after the item?

docs/design/editorial_design.md
```diff
@@ -0,0 +1,354 @@
...
+            - Sequence : Don't modify the sequence, trim item if required
+            - Fit      : Apply Time Effect
+"""
+otio.algorithms.fill(
+    item: Item,
+    track: Track,
+    replace: Item = None, # Alternatively, we could have a parent-coord
+                          # RationalTime to find the item for us
+    reference_point: otio.algorithms.ReferencePoint.Source
+)
```
reinecke, line 315: This one's a bit tricky!
My instinct is that this may be trying to do too much and could yield unexpected results for users. The `reference_point` being set to `Source` is effectively the same as just using overwrite, right?
Also there is some clarification needed about whether it's most important to keep the in point or out point of the inserted item.
I wonder if maybe this operation can be broken down into a few more explicit variants instead (e.x. overwrite at time range trimming item, overwrite at time range retiming item, etc.)

docs/design/editorial_design.md
```diff
@@ -0,0 +1,354 @@
...
+    delta: RationalTime to adjust the range by.
+"""
+otio.algorithms.slip(
+    item: Item,
+    delta: RationalTime,
+)
+```
+- In the example image, `C` is `item` and `delta` is the arrow's vector.
+
+> Effectively this is already quite simple in the core API but would be nice to have here.
```
ssteinbach, line 207: I suspect the value of some of these functions might be that they hide the underlying complexity of whether or not source_range is set.

Maybe you split these functions into a "maybe" and "Definitely" pile?  that will help prioritize them and see if people come out with concerns or use cases.

This might be a controversial position, but I think I'd also prefer to not have things named too close even if they do mean a specific thing (slip/slide for example)... unless that verb means the same thing in all major editing packages.

docs/design/editorial_design.md
mccartnm: Agreed - I've got ASCII art and more explicit instructions per-command on the way (taking in the additional notes where possible). I'll add the audience information to the top along with the replaced "elevator pitch" that @ssteinbach recommended which sums it up beautifully 👍

docs/design/editorial_design.md
mccartnm: I completely agree with this. I didn't want to make this my first move without larger consensus but this will work super well and clean up a lot of clunky code I was sketching out. As @reinecke suggested, it'll be worth pointing out that no new schemas should be allowed under `algo/`

docs/design/editorial_design.md
mccartnm: I hope that the common commands are as simple as possible - with explicitly defined implementation or course. Part of that would be, possibly, having it assume the source_range is the available_range if not set before. Implementing it as an error would totally work too but do we gain anything with that?

docs/design/editorial_design.md
mccartnm: I totally see it being used more widely. I was hoping to build the common commands on top of this system. A command like `overwrite` will potentially do quite a few events - something like "sub-edits"
```
e.g.
- set A.duration = F
- set B.duration = B
- insert C into T' at index 3
```
If the general feeling is they shouldn't use it, then I'll write them as direct procedures.

docs/design/editorial_design.md
```diff
@@ -0,0 +1,354 @@
...
+    item: Item to apply trim to
+    delta_in: RationalTime that the item's source_range().start_time()
+        will be adjusted by
+    delta_out: RationalTime that the item's
+        source_range().end_time_exclusive() will be adjusted by
+"""
+otio.algorithms.trim(
+    item: Item,
+    delta_in: RationalTime = None,   #< Duration of change
+    delta_out: RationalTime = None,  #< Duration of change
```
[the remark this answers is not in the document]
mccartnm: Yes - Error cases are something I definitely need to focus on more. I'll try to get more of the edge cases documented for my next push

docs/design/editorial_design.md
mccartnm: It's for the case of actions like `overwrite` and `insert` which have the same functionality if placing an item passed the tracks current duration. I'll try to insert additional examples in the next push 👍

docs/design/editorial_design.md
```diff
@@ -62,7 +40,10 @@ Let's go through some of the most common commands and what the python API might
 ## Overwrite
 
 ![Overwrite](../_static/edit/01_overwrite.png)
-- Anything overlapping is destroyed on contact. This may split an item.
+- Augment the source range of overlapping items
```
meshula, line 43: To be honest, I still feel the wording is not as precise as it could be.  Elsewhere in OTIO we are adopting Allen's nomenclature, here, because the intent of the words is unambiguous: https://en.wikipedia.org/wiki/Allen's_interval_algebra

docs/design/editorial_design.md
```diff
@@ -73,7 +54,57 @@ Args:
     track_time: RationalTime of our track to put this item at
     fill_template: Optional[Item] that will be cloned where required
         to fill in the event that this overwrite extends the end of the
-        composition's limit
+        track's current duration
+
+Example:
```
meshula, line 59: This is great, thanks!

docs/design/editorial_design.md
meshula: Okay, I'll look forward to that :)

PR #720, opened 2020-05-27
C bindings - wrap safely typed any

KarthikRIyer:
Wrapped `safely_typed_any.h`

Next, I'll push a few return fixes, and will move on to tests.

### DIFF
--- a/src/c-opentimelineio/opentimelineio-bindings/CMakeLists.txt
+++ b/src/c-opentimelineio/opentimelineio-bindings/CMakeLists.txt
@@ -41,6 +41,7 @@ set(OPENTIMELINEIO_HEADER_FILES
         include/copentimelineio/timeline.h
         include/copentimelineio/stackAlgorithm.h
         include/copentimelineio/trackAlgorithm.h
+        include/copentimelineio/safely_typed_any.h
         )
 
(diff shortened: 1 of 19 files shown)


KarthikRIyer:
There are a few functions where I forgot to return values. GCC compiled fine without warnings. Whereas clang gave me warnings.

Also, clang gives a warning for the CPP RationalTime implementation. Does this need to be fixed?

```console
[  2%] Building CXX object build/src/opentime/CMakeFiles/opentime.dir/rationalTime.cpp.o
/home/karthik/OpenTimelineIO/src/opentime/rationalTime.cpp:293:25: warning: using integer absolute value function 'abs' when argument is of floating point type [-Wabsolute-value]
        total_seconds = abs(total_seconds);
                        ^
/home/karthik/OpenTimelineIO/src/opentime/rationalTime.cpp:293:25: note: use function 'std::abs' instead
        total_seconds = abs(total_seconds);
                        ^~~
                        std::abs

```

meshula:
That abs is harmless in context, but it is incorrect. If anyone wrote new code against total_seconds, for example, attempting to extract a frame number, that could would break. Please go ahead and add an std:: in there as the message suggests.

meshula:
Looks good!

PR #722, opened 2020-05-28
Add ImageSequenceReference schema to OpenTimelineIO

### LINKED ISSUE #69: Image sequence support

reinecke:
This resolves #69

The image sequence is expressed with:
- `target_url_base` - everything leading up to the file name in the `target_url`
- `name_prefix` - everything in the file name leading up to the frame number
- `name_suffix` - everything after the frame number in the file name
- `start_frame` - first frame number used in file names
- `frame_step` - step between frame numbers in file names (every other frame is a step of 2)
- `rate` - double frame rate if every frame in the sequence were played back (ignoring skip frames)
- `frame_zero_padding` - Number of digits to pad zeros out to (e.x. frame 10 with a pad of 4 would be `0010`)
- `missing_frame_policy` - enum `ImageSequenceReference.MissingFramePolicy` {`error`, `hold`, `black`} allows hinting about how a consuming app should behave if an image for which a url is returned should be handled when missing from disk

An example for 24fps media with a sample provided each frame numbered 1-1000 with a path `/show/sequence/shot/sample_image_sequence.%04d.exr` might be:

```
{
  "available_range": {
    "start_time": {
      "value": 0,
      "rate": 24
    },
    "duration": {
      "value": 1000,
      "rate": 24
    }
  },
  "start_frame": 1,
  "frame_step": 1,
  "rate": 24,
  "target_url_base": "file:///show/sequence/shot/",
  "name_prefix": "sample_image_sequence.",
  "name_suffix": ".exr"
  "frame_zero_padding": 4,
}
```

The same duration sequence but with only every 2nd frame available in the sequence would be:

```
{
  "available_range": {
    "start_time": {
      "value": 0,
      "rate": 24
    },
    "duration": {
      "value": 1000,
      "rate": 24
    }
  },
  "start_frame": 1,
  "frame_step": 2,
  "rate": 24,
  "target_url_base": "file:///show/sequence/shot/",
  "name_prefix": "sample_image_sequence.",
  "name_suffix": ".exr"
  "frame_zero_padding": 4,
}
```

A target url is generated using the equivalent of the following python format string:
`f"{target_url_prefix}{(start_frame + (sample_number * frame_step)):0{value_zero_padding}}{target_url_postfix}"`

Negative `start_frame` is also handled. The above example with a `start_frame` of `-1` would yield the first three target urls as:

- `file:///show/sequence/shot/sample_image_sequence.-0001.exr`
- `file:///show/sequence/shot/sample_image_sequence.-0000.exr`
- `file:///show/sequence/shot/sample_image_sequence.0001.exr`

Benefits of this approach include:

- The ability to derive the times each image map to
- Avoids having to support and potentially implement a rich format string system in otio

Downsides:

- A bit heavier than using format strings
- Not as flexible/extensible as format strings

References:

- [RV negative frame number support](http://www.tweaksoftware.com/static/documentation/rv/current/html/rv_manual.html#Command_Line_Options_Image_Sequence_Notation_Negative_Frame_Numbers)

Related PRs:

- #536 Original `ImageSequenceReference` proposal
- #602 Creation of `imagesequence` branch, C++ implementation and discussion
- #637 Support of `ImageSequenceReference` in RV
- #710 Addressing testing of RV plugin

Co-authored by: @apetrynet


docs/tutorials/otio-serialized-schema.md
```diff
@@ -354,6 +354,29 @@ parameters:
...
+
+*full module path*: `opentimelineio.schema.ImageSequenceReference`
+
+*documentation*:
+
+```
+None
+```
+
+parameters:
```
ssteinbach, line 367: can we get some of the documentation to flow through the python bindings?

src/opentimelineio/imageSequenceReference.cpp
```diff
@@ -0,0 +1,174 @@
...
+                }
+                else {
+                    // Unrecognized value
+                    ErrorStatus error_status = ErrorStatus(ErrorStatus::JSON_PARSE_ERROR,
+                               "Unknown missing_frame_policy: " + missing_frame_policy_value);
+                    reader.error(error_status);
+                    return false;
+                }
+
+                return result && Parent::read_from(reader);
```
ssteinbach, line 146: looks like this formatting got kicked to the right accidentally

docs/tutorials/otio-serialized-schema.md
reinecke: Added docstrings in the bindings. I'd love to have someone give feedback on wether or not it all makes sense. The bulk of the content is forklifted from this PR.

reinecke:
I think the build failure is a transient error, the python3.6 build just needs to be re-run.

src/opentimelineio/imageSequenceReference.cpp
```diff
@@ -0,0 +1,174 @@
...
+                      std::string const& name_prefix,
+                      std::string const& name_suffix,
+                      int start_frame,
+                      int frame_step,
+                      double const rate,
+                      int frame_zero_padding,
+                      MissingFramePolicy const missing_frame_policy,
+                      optional<TimeRange> const& available_range,
+                      AnyDictionary const& metadata)
+    : Parent(std::string(), available_range, metadata),
```
palemieux, line 15: Why can't an instance of `ImageSequenceReference` be initialized with a `name`?

reinecke: Great point. I'll see if anyone has a case for not adding that to the constructor.

reinecke: Ahh, interesting. I think I matched the constructor for `ExternalReference`. I'm guessing the main reason the api is structured that way is that `name` isn't required to be specified to have the Reference behave properly, but many of these other args are. Perhaps in python `name` should be a defaulted keyword argument toward the end of the args. Perhaps we should consider it a keyword-only argument in python 3+, but that could add to the breakage pain as studios move from python 2 to 3.

jminor: In practice, Clips usually have names, but MediaReferences often do not. When MediaReferences do have names, it is often the same as the parent Clip, or the basename of the target_url. Therefore, it makes sense to keep it optional, and less prominent.

reinecke: I think maybe if we decide to add the name arg, we should do it in a separate PR and add it to all `MediaReference` subclasses in a similar way. `name` is still in the schema, so it's an API issue, not a schema issue.

@ssteinbach @jminor What do you think?

reinecke: My previous reply was made on an un-refreshed browser tab and I missed your last message.
I'm going to resolve this and stay consistent with `ExternalReference`. If we feel we want to address it we can examine the two together and make a decision.

PR #723, opened 2020-05-28
Add tests for C-bindings

KarthikRIyer:
I'll be adding tests and any other necessary fixes in this PR.

### DIFF
--- a/src/c-opentimelineio/tests/OTIOCompositionTests.cpp
+++ b/src/c-opentimelineio/tests/OTIOCompositionTests.cpp
@@ -10,6 +10,8 @@
 #include <copentimelineio/errorStatus.h>
 #include <copentimelineio/gap.h>
 #include <copentimelineio/item.h>
+#include <copentimelineio/mediaReference.h>
+#include <copentimelineio/missingReference.h>
 #include <copentimelineio/safely_typed_any.h>
 #include <copentimelineio/serializableObject.h>
 #include <copentimelineio/serializableObjectWithMetadata.h>
@@ -43,10 +45,16 @@ protected:
     void        SetUp() override { sample_data_dir = xstr(SAMPLE_DATA_DIR); }
     void        TearDown() override {}
     const char* sample_data_dir;
-    const char* transition_example_path;
 };
 
 class OTIOEdgeCases : public ::testing::Test
+{
+protected:
+    void SetUp() override {}
+    void TearDown() override {}
+};
+
+class OTIONestingTest : public ::testing::Test
 {
 protected:
     void SetUp() override {}
@@ -306,7 +314,7 @@ TEST_F(OTIOStackTests, SerializeTest)
     Any* stack_any =
         create_safely_typed_any_serializable_object((SerializableObject*) st);
     const char* encoded = serialize_json_to_string(stack_any, errorStatus, 4);
-    Any*        decoded = /* allocate memory for destinantion */
+    Any*        decoded = /** allocate memory for destinantion */
         create_safely_typed_any_serializable_object((SerializableObject*) st);
     bool decoded_successfully =
         deserialize_json_from_string(encoded, decoded, errorStatus);
@@ -462,7 +470,7 @@ TEST_F(OTIOStackTests, RangeOfChildTest)
     Composition_insert_child(
         (Composition*) st, 2, (Composable*) clip3, errorStatus);
 
-    /* stack should be as long as longest child */
+    /** stack should be as long as longest child */
     RationalTime* length = RationalTime_create(50, 24);
     RationalTime* st_duration =
         Composable_duration((Composable*) st, errorStatus);
@@ -471,7 +479,7 @@ TEST_F(OTIOStackTests, RangeOfChildTest)
     RationalTime_destroy(st_duration);
 
     RationalTime* zero_time = RationalTime_create(0, 24);
-    /* stacked items should all start at time zero */
+    /** stacked items should all start at time zero */
     TimeRange* range_at_0 = Stack_range_of_child_at_index(st, 0, errorStatus);
     TimeRange* range_at_1 = Stack_range_of_child_at_index(st, 1, errorStatus);
     TimeRange* range_at_2 = Stack_range_of_child_at_index(st, 2, errorStatus);
@@ -562,7 +570,7 @@ TEST_F(OTIOStackTests, RangeOfChildWithDurationTest)
     RationalTime_destroy(start_time);
     RationalTime_destroy(duration);
 
-    /* range always returns the pre-trimmed range.  To get the post-trim
+    /** range always returns the pre-trimmed range.  To get the post-trim
      * range, call .trimmed_range()
      */
     ComposableRetainerVector* composableRetainerVector =
@@ -598,7 +606,7 @@ TEST_F(OTIOStackTests, RangeOfChildWithDurationTest)
     RationalTime_destroy(rt);
     RationalTime_destroy(rt2);
 
-    /* trimmed_ functions take into account the source_range */
+    /** trimmed_ functions take into account the source_range */
     TimeRange* st_trimmed_range_child_0 =
         Stack_trimmed_range_of_child_at_index(st, 0, errorStatus);
     st_sourcerange = Item_source_range((Item*) st);
@@ -617,7 +625,7 @@ TEST_F(OTIOStackTests, RangeOfChildWithDurationTest)
     RationalTime_destroy(duration);
     TimeRange_destroy(time_range);
 
-    /* get the trimmed range in the parent */
+    /** get the trimmed range in the parent */
     TimeRange* st_0_trimmed_range_in_parent =
         Item_trimmed_range_in_parent((Item*) st_0, errorStatus);
     EXPECT_TRUE(TimeRange_equal(
@@ -725,7 +733,7 @@ TEST_F(OTIOStackTests, TransformedTimeTest)
     RationalTime_destroy(test_time2);
     RationalTime_destroy(st_transformed_time);
 
-    /* ensure that transformed_time does not edit in place */
+    /** ensure that transformed_time does not edit in place */
     RationalTime* verify_test_time = RationalTime_create(0, 24);
     EXPECT_TRUE(RationalTime_equal(test_time, verify_test_time));
     RationalTime_destroy(verify_test_time);
@@ -843,7 +851,7 @@ TEST_F(OTIOTrackTests, SerializeTest)
     Any*             sq_any =
         create_safely_typed_any_serializable_object((SerializableObject*) sq);
     const char* encoded = serialize_json_to_string(sq_any, errorStatus, 4);
-    Any*        decoded = /* allocate memory for destinantion */
+    Any*        decoded = /** allocate memory for destinantion */
         create_safely_typed_any_serializable_object((SerializableObject*) sq);
     bool decoded_successfully =
         deserialize_json_from_string(encoded, decoded, errorStatus);
@@ -877,7 +885,7 @@ TEST_F(OTIOTrackTests, InstancingTest)
         Track_range_of_child_at_index(sq, 0, errorStatus);
     EXPECT_TRUE(TimeRange_equal(sq_range_of_child_0, tr));
 
-    /* Can't put item on a composition if it's already in one */
+    /** Can't put item on a composition if it's already in one */
     Track* test_track = Track_create(NULL, NULL, NULL, NULL);
     insertOK          = Composition_insert_child(
         (Composition*) test_track, 0, (Composable*) it, errorStatus);
@@ -885,7 +893,7 @@ TEST_F(OTIOTrackTests, InstancingTest)
     SerializableObject_possibly_delete((SerializableObject*) test_track);
     test_track = NULL;
 
-    /* Instancing is not allowed */
+    /** Instancing is not allowed */
     ComposableVector* composableVector = ComposableVector_create();
     ComposableVector_push_back(composableVector, (Composable*) it);
     ComposableVector_push_back(composableVector, (Composable*) it);
@@ -899,7 +907,7 @@ TEST_F(OTIOTrackTests, InstancingTest)
     ComposableVector_destroy(composableVector);
     composableVector = NULL;
 
-    /*inserting duplicates should raise error and have no side effects*/
+    /**inserting duplicates should raise error and have no side effects*/
     ComposableRetainerVector* composableRetainerVector =
         Composition_children((Composition*) sq);
     EXPECT_EQ(ComposableRetainerVector_size(composableRetainerVector), 1);
@@ -930,7 +938,7 @@ TEST_F(OTIOTrackTests, InstancingTest)
 
 TEST_F(OTIOTrackTests, DeleteParentContainerTest)
 {
-    /* deleting the parent container should null out the parent pointer */
+    /** deleting the parent container should null out the parent pointer */
     Item*            it          = Item_create(NULL, NULL, NULL, NULL, NULL);
     Track*           sq          = Track_create(NULL, NULL, NULL, NULL);
     OTIOErrorStatus* errorStatus = OTIOErrorStatus_create();
@@ -1083,7 +1091,7 @@ TEST_F(OTIOTrackTests, RangeTest)
     RationalTime_destroy(sq_duration);
     RationalTime_destroy(duration_compare);
 
-    /* add a transition to either side */
+    /** add a transition to either side */
     TimeRange* range_of_child_3 =
         Track_range_of_child_at_index(sq, 3, errorStatus);
     RationalTime* in_offset  = RationalTime_create(10, 24);
@@ -1107,7 +1115,7 @@ TEST_F(OTIOTrackTests, RangeTest)
     ASSERT_TRUE(insertOK);
     TimeRange_destroy(range_of_item);
 
-    /* range of Transition */
+    /** range of Transition */
     start_time = RationalTime_create(230, 24);
     duration   = RationalTime_create(22, 24);
     tr = TimeRange_create_with_start_time_and_duration(start_time, duration);
@@ -1134,7 +1142,7 @@ TEST_F(OTIOTrackTests, RangeTest)
     TimeRange_destroy(range_of_child_3);
 
     sq_duration = Item_duration((Item*) sq, errorStatus);
-    /* duration_compare = length x 4 + in_offset + out_offset */
+    /** duration_compare = length x 4 + in_offset + out_offset */
     duration_compare = RationalTime_create(20 + 22.0 / 24.0, 1);
     EXPECT_TRUE(RationalTime_equal(sq_duration, duration_compare));
     RationalTime_destroy(sq_duration);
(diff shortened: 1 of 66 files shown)


src/c-opentimelineio/tests/OTIOComposableTests.cpp
```diff
@@ -86,4 +57,21 @@ TEST_F(OpenTimelineIOTests, ComposableTest)
...
+{
+    Any*                   value    = create_safely_typed_any_string("bar");
+    AnyDictionary*         metadata = AnyDictionary_create();
+    AnyDictionaryIterator* it = AnyDictionary_insert(metadata, "foo", value);
+    Composable*            seqi =
+        Composable_create_with_name_and_metadata("test", metadata);
+    Any* seqi_any =
+        create_safely_typed_any_serializable_object((SerializableObject*) seqi);
+    OTIOErrorStatus* errorStatus = OTIOErrorStatus_create();
+    const char* encoded = serialize_json_to_string(seqi_any, errorStatus, 4);
```
KarthikRIyer, line 72: I don't know why but serialization doesn't work now. The casting to Any* doesn't give any errors but the error status I get here is: `type mismatch while decoding`. I'll try looking into this again tomorrow. Meanwhile, does anyone have any idea why this might be happening?

src/c-opentimelineio/tests/OTIOCompositionTests.cpp
```diff
@@ -98,4 +91,163 @@ TEST_F(OTIOCompositionTests, ConstructorTest)
...
+    ComposableVector_push_back(composableVector, (Composable*) b2);
+    ComposableVector_push_back(composableVector, (Composable*) c2);
+    resultOK = Composition_set_children(co3, composableVector, errorStatus);
+    ASSERT_TRUE(resultOK);
+    ComposableVector_destroy(composableVector);
+    composableVector = NULL;
+
+    EXPECT_FALSE(SerializableObject_is_equivalent_to(
+        (SerializableObject*) co1, (SerializableObject*) co2));
+    /*EXPECT_TRUE(SerializableObject_is_equivalent_to( //TODO Fix Segfault
```
KarthikRIyer, line 147: There are some more segfaults like this that need to be fixed, but I have no idea at the moment why they're happening.

src/c-opentimelineio/tests/OTIOCompositionTests.cpp
```diff
@@ -929,4 +929,63 @@ TEST_F(OTIOTrackTests, DeleteParentContainerTest)
...
+    Item*            item1       = Item_create(NULL, NULL, NULL, NULL, NULL);
+    Item*            item2       = Item_create(NULL, NULL, NULL, NULL, NULL);
+    Item*            item3       = Item_create(NULL, NULL, NULL, NULL, NULL);
+
+    SerializableObject* itemClone1 =
+        SerializableObject_clone((SerializableObject*) item, errorStatus);
+
+    SerializableObject* itemClone2 =
+        SerializableObject_clone((SerializableObject*) item, errorStatus);
+    // TODO segfault here
```
KarthikRIyer, line 949: Again getting a segfault here.
Tried to debug using gdb (don't have a lot of experience with this). 

It seems to be occuring here: https://github.com/PixarAnimationStudios/OpenTimelineIO/blob/eaccea3b61455705286ede0de239dd84406352bb/src/opentimelineio/serialization.cpp#L570

Any ideas why this might be happening and how I can try going about it?

I don't get a segfault when I clone `itemClone1` instead, and for the third clone I use `item` again.
The error seems to be when I clone item twice consecutively.

KarthikRIyer: @jminor @meshula @ssteinbach @reinecke

meshula: _schema_name_for_reference() is return a reference to a string, so as long as value is in scope, schema_type_name should be live. If value itself has been deallocated you'd get the segfault. If you could write a tiny cpp program to demonstrate the bug, or tell me how to reproduce it in an existing test, I'd be happy to take a look at it.

KarthikRIyer: You should be able to reproduce it using COTIOCompositionTests.

The procedure would be:

- clone the repo
- `cd OpenTimelineIO/src/c-opentimelineio`
- `mkdir build`
- `cd build`
- `cmake -DCMAKE_INSTALL_PREFIX=<install dir> ..`
- `make`

Then to run all tests:
- `make test`

But I think here it would be better to just run the composition tests:

Inside the build directory:

- `./tests/COTIOCompositionTests`

With the current code you should get a segfault at line 947 in `COTIOCompositionTests.cpp`.

I put breakpoints before and after both the clone operations which led me to the name for reference function.

If you comment out the second clone operation, you shouldn't get the segfault.

KarthikRIyer: @meshula did you get a chance to take a look?

src/c-opentimelineio/tests/OTIOCompositionTests.cpp
```diff
@@ -1639,4 +1639,66 @@ TEST_F(OTIOTrackTests, TransformedTimeTest)
...
+    //        serialize_json_to_string((Any*) neighbors, errorStatus, 4);
+    //    printf("%d\n", OTIOErrorStatus_get_outcome(errorStatus));
+    retainerComposable      = RetainerPairComposable_first(neighbors);
+    retainerComposableValue = RetainerComposable_take_value(retainerComposable);
+    EXPECT_TRUE(SerializableObject_is_equivalent_to(
+        (SerializableObject*) retainerComposableValue,
+        (SerializableObject*) fill));
+    retainerComposable      = RetainerPairComposable_second(neighbors);
+    retainerComposableValue = RetainerComposable_take_value(retainerComposable);
+    //    EXPECT_TRUE(SerializableObject_is_equivalent_to( //TODO fix segfault
```
KarthikRIyer, line 1698: Similar segfault again.

gdb again points to `schema_name_for_reference`.

src/c-opentimelineio/tests/OTIOCompositionTests.cpp
```diff
@@ -59,6 +59,43 @@ class OTIONestingTest : public ::testing::Test
...
+            (Composition*) wrapper, (Composable*) clip, errorStatus);
+
+        /* swap out the item for the wrapper */
+        bool setOK = Composition_set_child(
+            (Composition*) parent, index, (Composable*) wrapper, errorStatus);
+
+        clipWrapperPair.clip    = clip;
+        clipWrapperPair.wrapper = wrapper;
+        return clipWrapperPair;
+    }
```
KarthikRIyer, line 98: @meshula @jminor @ssteinbach @reinecke 

I'm trying to translate this:

https://github.com/PixarAnimationStudios/OpenTimelineIO/blob/eaccea3b61455705286ede0de239dd84406352bb/tests/test_composition.py#L1611

The comment below is according to the flow in the python definition of the function.

The problem I'm facing here is how to swap the `item` for the `wrapper`. I have just the pointer to the clip in the track.
Once I set the child at `index 0` to `wrapper` the `clip` has no references and will get destroyed. So the next thing would be to try to add the `clip` to the `wrapper` before setting the child. But the clip is a child of the track already so it cannot be added to come other composition.

So my query here would be how to get ownership of the clip?

For now I have implemented a workaround that I don't really like. I've cloned the `clip`, put that into the `wrapper` and then set the child in the parent. Then I've returned a struct with the `wrapper` and new `clip` pointer for further nesting.

KarthikRIyer:
Also, why is this comparison necessary?

https://github.com/PixarAnimationStudios/OpenTimelineIO/blob/eaccea3b61455705286ede0de239dd84406352bb/tests/test_composition.py#L1666

src/c-opentimelineio/tests/OTIOMediaReferenceTests.cpp
```diff
@@ -0,0 +1,159 @@
...
+    EXPECT_FALSE(SerializableObject_is_equivalent_to(
+        (SerializableObject*) filepath, (SerializableObject*) filepath2));
+}
+
+TEST_F(OTIOMediaReferenceTests, IsMissingTest)
+{
+    ExternalReference* filepath =
+        ExternalReference_create("var/tmp/foo.mov", NULL, NULL);
+    EXPECT_FALSE(
+        MediaReference_is_missing_reference((MediaReference*) filepath));
```
KarthikRIyer, line 150: @jminor @meshula @ssteinbach @reinecke 

This is an example of the inheritance stuff I was talking about. Instead of defining a method `ExternalReference_is_missing_reference` I've simply cast the pointer.

After giving it a thought I think it will be convenient for the end user if I wrap the inherited methods too. But I'd like to hear your thoughts.

meshula: Yes, wrapping inherited methods, in an unambiguous way, is the way to go.

KarthikRIyer:
@jminor @meshula @reinecke @ssteinbach 

I apologise for the decreased pace of work. 

All the inherited functions have now been wrapped and I can resume writing tests. I'm hoping to wrap up the C-bindings (apart from some unresolved bugs/segfaults) by month end and begin with the Java Bindings next month.

KarthikRIyer:
@meshula @jminor @ssteinbach @reinecke 

Writing tests is almost done! Just two of the track algo tests are remaining. After that the segfaults related to `SerializableObject` needs fixing and then this portion of the project should be done.

I'll finish the remaining tests tomorrow and get started on the Java bindings.

KarthikRIyer:
@meshula @jminor @ssteinbach @reinecke

Track algo tests done.

 Can someone review the PR?

Segfaults need to be fixed. Would it be possible for someone to take a look at the tests? I've marked them as `// TODO: fix segfault`

I'll start with the Java bindings.

meshula:
Hi @KarthikRIyer, I'm pulling now!

KarthikRIyer:
@meshula I was going through the multithreading tests. I don't understand what exactly is happening there...

Take this test for example:

```python
    def test1(self):
        self.sc = otio.schema.SerializableCollection()
        child = otio.core.SerializableObject()
        child.extra = 17
        wc = weakref.ref(child)
        self.assertEqual(wc() is not None, True)
        self.sc.append(child)
        del child

        threads = []
        for i in range(5):
            t = threading.Thread(target=self.bash_retainers1)
            t.daemon = True
            t.start()
            threads.append(t)

        for t in threads:
            t.join()

        self.assertEqual(self.sc[0].extra, 17)
        self.sc.pop()
        self.assertEqual(wc() is None, True)

    def bash_retainers1(self):
        otio._otio._testing.bash_retainers1(self.sc)
```

```cpp
static int test_bash_retainers1(SerializableCollection* sc) {
    py::gil_scoped_release release;
    SerializableObject* so = sc->children()[0];

    int total = 0;
    for (size_t i = 0; i < 1024 * 10; i++) {
        SerializableObject::Retainer<> r(so);
        if (r.value) {
            total++;
        }
    }
    
    return total;
}
```

What is `extra` ?
From what I can understand we're creating retainers in each thread, which increases the reference count by one and when the retainer is destroyed it decreases the reference count by one. So by the time the 5 threads are done with their jobs we should have one reference left, that's in the `SerializableCollection`. And once we pop that off, the reference becomes NULL.

Is that correct?

What could go wrong? What exactly are we trying to test here?

I'll go through the other multithreading tests and ask if I don't understand anything.

davidbaraff:
> On Jul 8, 2020, at 10:02 AM, Karthik Ramesh Iyer <notifications@github.com> wrote:
> 
> 
> @meshula <https://github.com/meshula> I was going through the multithreading tests. I don't understand what exactly is happening there...
> 
> Take this test for example:
> 
> What is extra ?

Every C++ OTIO object O given out to Python requires creation of a Python object P.  By setting extra = 17, I can check at the end that when I give O back out, someone didn’t create a *new* Python object and give it to me.  If the Python object I get back from that C++ object O has 17 set, then I know the bookkeeping worked.

It is common to stick extra data in Python on stuff.  If it falls off, because the system blows your Python objects away and then creates new ones when the C++ objects come back, that’s a problem.


extra is a new, dynamic attribute created on child.  C++ doesn’t know about it.  If all works correctly, at the end, we want to ensure that not only is the original C++ object still in existence, but that the Python object that corresponds with it is also exactly the same.  


> What could go wrong? What exactly are we trying to test here?
> 
> 

Oh, everything could go wrong.  In this particular case, there are multiple locks in play.  C++ OTIO has some mutex’s, and Python has the GIL (global interpreter lock).  Anything with more than one lock that involves multithreading has the potential for deadlock.  If the test doesn’t lock itself up, that’s proven something important.

The point is that the threads are executing at the same time, so there’s a lot that can mess up.

KarthikRIyer:
Thanks @davidbaraff. That helps

KarthikRIyer:
@davidbaraff Could you take a look at the last commit for the multithreaded test? I'm not very familiar with multithreading.

davidbaraff:
I will when I get a chance, but provided that the Python test is still being run, i’m not sure why you’re adding new multithreading tests for the C-bindings.  I don’t think the C-bindings have any new code that has anything to do with multi-threading, so I wouldn’t think you’d need to explicitly test the bindings under multithreading conditions.

Certainly, your C-bindings aren’t going to exercise anything new in terms of the relationship between Python and C++, if the tests are written in C.  They also aren’t going to exercise anything new on the C++ end.  The only reason to test multithreading in C is if your C bindings have introduced any kind of new “state” (or locks) that didn’t already exist in C++ or Python.

Caveat: i’ve not at all been following the work on C-bindings.

KarthikRIyer:
@davidbaraff thanks for the explanation. I am not adding anything new in terms of threading, so I'll remove the test.

KarthikRIyer:
@meshula did you get a chance to take a look at the segfaults issue? I'm facing the same problem with the Java bindings too.

meshula:
I'm working on reproducing the problem this very minute.

meshula:
Hi Karthik, it took a little while to get here. The seg_faults in the test suite are occurring because the object getting possibly_delete called on it has already been deleted, and the vtable is invalid; therefore possibly_delete cannot be invoked at all.

meshula:
basically, Track_create is creating a new Track, but nothing retains it in the test suite.

for example track_trimmed_to_range returns an unretained object.  Then, Track_is_equivalent_to is called, which creates a retainer object on the track. When that retainer object is destructed, there's no outstanding references, so the track gets destroyed. Then, when possibly_delete is invoked, it crashes.

meshula:
basically the C api doesn’t have any magic lifespan management. It seems to me that we might have to do a scrub, and have every call that returns a new object, such as track_trimmed_to_range, actually have an incremented retain count.

There are functions with the following suffixes:

```
_destroy
_managed_destroy
_possibly_delete
```

Perhaps we should gather the Paladins of OpenTimelineIO to discuss what the proper behavior should be for functions that return un-counted objects, since the fact that calling comparison functions, can as a side-effect, destroy an object, is very surprising behavior.

davidbaraff:
I would go back and reread the section of the docs about C++ lifetime management and responsibilities when dealing with raw pointers and make sure you fully understand the rules as explained in that section.

KarthikRIyer:
@meshula yes that seems to be the problem. I should have been handing out Retainer<T> instead of T.

I had the same issue with the java bindings. Yesterday I implemented the keepalive monitor for SerializableObject and when I called `isEquivalentTo` on it and tried to call other functions later, it worked fine.

@davidbaraff would it be possible for you to cross-check the SerializableObject C++ part of the last commit in the java-bindings?

davidbaraff:
I will try to find some time to look at the (C++) part of the code. ( I don't speak Java at all.)

meshula:
As suggested by David, I re-read the docs. To frame the problem, following the patterns there:

This is an expected pattern:

- create a new Track object, returned as a pointer
- call possibly delete on the Track object (refcount is zero, object destroyed)

The test program expects to do this:

- create a new Track object, returned as a pointer
- compare the Track object to another Track object
- test program calls possibly delete on the Track object (refcount is zero, object destroyed)

What actually happens is

- create a new Track object, returned as a pointer
- compare the Track object to another Track object
- the comparison routine creates a temporary retainer
- the comparison routine does the comparison
- the comparison routine deletes the retainer (refcount is zero, object is destroyed)
- test program calls possibly delete on the Track object (object already destroyed, segfault)

So the unexpected behavior is that a comparison, which should be a const, non-mutating operation, has a hidden side effect that it can destroy objects. This does not seem intentional or desirable. The bug therefore would seem to be that the comparison routine itself should not have a destructive side effect.

davidbaraff:
If you’re going to hang onto a pointer, you need to create a Retainer for it.

Basically,
 SerializableObject* s = someFunction();

At this point, if you want to immediately return s, great.  However, as soon as you start doing ANY manipulation on it, i.e. passing it off to any other function, you have to create a Retainer for it, if you expect to keep using it.

So this code
	 SerializableObject* s = someFunction();
	  someExaminationFunction(s);
	  someOtherExaminationFunction(s);

is wrong.  because someExaminationFunction(s) might have created a retainer for s, and then destroyed that Retainer.  And now s is deleted.

So the instant you create a raw pointer, either make a retainer, or return it.

don’t do ANYTHING else.

i hope that helps a bit.

meshula:
Ok. So I'll adjust my mental model to "lifespan manage via Retainer, only". Thanks for bearing with me as I work through this.

davidbaraff:
Man, I really figured the lifetime management issues would be a piece of cake for anyone whoever lived through release/retain/autorelease...

meshula:
Comparatively speaking, yes. I haven't yet fully internalized how the C++ implementation of OTIO maps onto that paradigm, but now that you mention that analogy, and if I think about how ARC works versus an NSObject pointer, it's falling into place. I should probably take a stab at the documentation, to insert a section or tweak the description such that I would have made the mental leap immediately. I mean, the information is all there, but obviously I've been getting to an understanding the long way around.

meshula:
Did a massive scrub per @davidbaraff's guidance. Fixed the unit tests to retain/release properly, all are passing now, including the previously elided tests. I'm going to close this PR in favor of a new branch, because a merge or rebase would be insanely messy.

https://github.com/PixarAnimationStudios/OpenTimelineIO/tree/c-otio-refcounting

KarthikRIyer:
Thanks a lot for the help with this @meshula!

KarthikRIyer:
In python the retainer is not exposed. What if the same was done with C?  Anyone using the bindings wouldn't get confused and wouldn't need to know about these memory management rules.

davidbaraff:
The bridge language (C, Python, Swift) should never require the end user to understand the C++ rules.  You should only expose a retainer if using it is part of the required rules in C, for example, the way pre-ARC objective-c requires users to do the retain/release dance and understand it well.

The retainer is not exposed in python either.

davidbaraff:
Let me also note that both Python and Swift do reference counting of their instances for the user.  In fact, I would assume that a C API for OTIO, because it can't build in ref counting the way C++ can, would have no choice BUT to expose an explicit retain/release model, like NSObject in objective-c.

davidbaraff:
I think C++ is in a strange spot compared with other languages.  The C++ API/implementation was forced into certain design choices because we knew that other languages were going to build on top, and we needed to provide flexibility.  Language bindings built on-top of the C++ API don't have this issue.  It would be a big mistake to try to do anything just because C++ did it that way.  Do whatever makes most sense for the language.

Pretty clearly a language like C has to have a set of memory management rules that the user needs to understand, but design what makes sense, as if C++ didn't exist.

meshula:
Initially there was some thought to wrapping the C++ library with C to make it useful as a building block for wrapping in languages that have strong support for C based ffi. The Java binding effort proceeded without taking advantage of that, and indeed a wrapper on a wrapper is getting a little contrived.

Following your note that the Retainer system was designed analogously to retain/release, the c-otio-refcounting branch I pushed has followed that notion, in not exposing the mechanisms of refcounting in the interface, in other words, not exposing the possibly_delete mechanism, but instead, adding an OTIO_RETAIN and an OTIO_RELEASE macro. This change made reasoning about application code infinitely easier, and allowed me to resolve all the test suite failures without trouble. The ease of debugging with this change alone seems like vindication of the approach.

I am still evaluating what from the API should ultimately take, for example, determination on whether a macro us the right approach, but I strongly agree based on working in that branch that retain/release for C is on the right track.

KarthikRIyer:
@meshula now I understand. My bad, I should have taken a closer look at what exactly was going on in the tests. Using the macro seems very simple.

meshula:
No problem, the C bindings were a large amount of code, I'm pleased to have a solid base in place.

PR #725, opened 2020-05-29
Remove redundant comparison in clip tests

KarthikRIyer:
I think instead of the line I removed, it might be `self.assertIsNot(cl.source_range, tr)`.


meshula:
good catch

PR #730, opened 2020-06-10
Implementation of post adapter write hook

apetrynet:
Added support for "post_adapter_write" hooks.
Post hooks are run _after_ the file is written to disk by an adapter and passes the file path to the hooks through `_filepath` key in the `argument_map`.

Updated unit tests and added example in the hooks docs.

### DIFF
--- a/docs/tutorials/write-a-hookscript.md
+++ b/docs/tutorials/write-a-hookscript.md
@@ -97,7 +97,7 @@ def hook_function(in_timeline,argument_map=None):
                 in_clip.media_reference.target_url = in_clip.media_reference.target_url.replace(r'/linux/media/path','S:')
 ```
 
-### Add an incremental copy of otio file in backup folder
+### Add an incremental copy of otio file to backup folder
 
 Example of a post adapter write hook
 
--- a/docs/tutorials/write-a-hookscript.md
+++ b/docs/tutorials/write-a-hookscript.md
@@ -99,7 +99,7 @@ def hook_function(in_timeline,argument_map=None):
 
 ### Add an incremental copy of otio file to backup folder
 
-Example of a post adapter write hook
+Example of a post adapter write hook that creates a timestamped copy of newly written file in a hidden "incremental" folder
 
 ```python
 import os
@@ -123,3 +123,6 @@ def hook_function(in_timeline, argument_map=None):
 
     return in_timeline    
 ```
+
+Please note that if your "post adapter write hook" changes `in_timeline` in any way, it's up to the user to persist those changes 
+as the hook is run _after_ the file is written to disk.
--- a/docs/tutorials/write-a-hookscript.md
+++ b/docs/tutorials/write-a-hookscript.md
@@ -124,5 +124,4 @@ def hook_function(in_timeline, argument_map=None):
     return in_timeline    
 ```
 
-Please note that if your "post adapter write hook" changes `in_timeline` in any way, it's up to the user to persist those changes 
-as the hook is run _after_ the file is written to disk.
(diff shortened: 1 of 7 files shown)


docs/tutorials/write-a-hookscript.md
```diff
@@ -89,9 +90,39 @@ An example use-case would be to create a pre-write adapter hook that checks the
 ```python
 def hook_function(in_timeline,argument_map=None):
     adapter_args = argument_map.get('adapter_arguments')
-    if if adapter_args and adapter_args.get('style') == 'nucoda':
+    if adapter_args and adapter_args.get('style') == 'nucoda':
```
ssteinbach, line 93: nice catch

docs/tutorials/write-a-hookscript.md
```diff
@@ -89,9 +90,39 @@ An example use-case would be to create a pre-write adapter hook that checks the
...
+        '.incremental',
+        backup_name
+    )   
+    shutil.copyfile(filepath, incrpath)
+
+    return in_timeline    
+```
+
+Please note that if your "post adapter write hook" changes `in_timeline` in any way, it's up to the user to persist those changes 
+as the hook is run _after_ the file is written to disk.
```
ssteinbach, line 128:
```suggestion
Please note that if a "post adapter write hook" changes `in_timeline` in any way, the api will not automatically update the already serialized file.  The changes will only exist in the in-memory object, because the hook runs _after_ the file is serialized to disk.
```

src/py-opentimelineio/opentimelineio/adapters/adapter.py
```diff
@@ -175,21 +175,30 @@ def write_to_file(
         input_otio = hooks.run("pre_adapter_write", input_otio,
                                extra_args=hook_function_argument_map)
 
+        # Store filepath for use in post_adapter_write hooks
+        hook_function_argument_map['_filepath'] = filepath
```
ssteinbach, line 179: you know, this is probably something we should pass to the `pre_adapter_write` function as well.  that's a good call. that way you can write a hook that is specific to a file extension or path or something.

apetrynet: Moved the assignment up a few lines to make this happen.

apetrynet:
I was a bit quick on the draw, so the first test didn't finish before I pushed the next one.

src/py-opentimelineio/opentimelineio/adapters/adapter.py
apetrynet: @ssteinbach 
It just struck me. Is the `_filepath` key name a good one? Will it make sense if/when persisted in a  "pre adapter write" hook's metadata?

Any suggestions for a better key name?

apetrynet: Any thoughts on this?

ssteinbach: We could deep copy the `hook_function_argument_map` on entry to the function.  I'm not sure what you mean by in_timeline metadata" though, sorry.  Can you clarify?

apetrynet: Sorry, meant `input_otio`'s metadata. 
Aaand I think I might have had a little brain fart here.. For some reason I was under the impression that all hooks modify the `input_otio`'s metadata based on the `hook_function_argument_map`. I think a deep copy will indeed do the trick and if a users hook script modifies the metadata including all of the `hook_function_argument_map` it'll be up to them to deal with it I guess

apetrynet:
@ssteinbach 
Seems like something went wrong in the py3.7 build in Travis. Could you resubmit?

PR #734, opened 2020-06-16
Added support for comma separated CDL SOP values

reinecke:
Encountered a CDL variant where SOP values looked like:

`(1.0801, 1.0747, 1.0730)`

as opposed to:

`(1.0801 1.0747 1.0730)`

Updated CDL value regex to accept this.

### DIFF
--- a/src/py-opentimelineio/opentimelineio/adapters/cmx_3600.py
+++ b/src/py-opentimelineio/opentimelineio/adapters/cmx_3600.py
@@ -421,7 +421,7 @@ class ClipHandler(object):
             sat = 1.0
 
             if asc_sop:
-                triple = r'([-+]?[\d.]+) ([-+]?[\d.]+) ([-+]?[\d.]+)'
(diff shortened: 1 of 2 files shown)


reinecke:
Yes, I just added another test case for this variant form of CDL data. I figure it's best to capture tests with each variation we encounter.

reinecke:
I *think* it may have come from Premiere Pro, but I don't have strict provenance on the EDL to say for sure. I do agree that we should try and note that in general though.

Jameclarke:
Attached is an EDL from Premiere Pro using the CDL master effect on a clip.  You are correct they use commas, however that's the first time I have ever seen that.
[Premiere Pro CDL Effect.edl.zip](https://github.com/PixarAnimationStudios/OpenTimelineIO/files/4796322/Premiere.Pro.CDL.Effect.edl.zip)

reinecke:
Thanks for the validation @Jameclarke. Would you mind if I used the EDL you just sent as the test case? The one I have is a bit hand-manufactured and I'd rather the tests included real-world output from Premiere.

Jameclarke:
Yes no problem!

PR #739, opened 2020-06-25
AAF Reader Metadata

stefanschulze:


contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py
```diff
@@ -283,22 +292,40 @@ def _transcribe(item, parents, editRate, masterMobs):
...
+        # is a MasterMob.
+        # There are some special cases where the masterMob could be:
+        # 1) For SourceClips in the CompositionMob, the mob the SourceClip is
+        #    referencing can be our MasterMob.
+        # 2) If the source clip is referencing another CompositionMob,
+        #    drill down to see if the composition holds the MasterMob
+        # 3) For everything else, it is a previously encountered parent. Find the
+        #    MasterMob in our chain, and then extract the information from that.
+
+        # Additionally, regards metadata (UserComments), they are usually stored
```
jminor, line 304: usually stored where? This looks like the comment got truncated?

contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py
```diff
@@ -283,22 +292,40 @@ def _transcribe(item, parents, editRate, masterMobs):
...
+        else:
+            transcribe_log("[found no mastermob]", indent)
+
         mastermob = child_mastermob or parent_mastermob or None
 
         if mastermob:
-            # get target path
-            mastermob_child = masterMobs.get(str(mastermob.mob_id))
+            # Get target path
+            mastermob_child = _transcribe(mastermob, list(), editRate, indent)
```
jminor, line 327: Can a MasterMob be referenced multiple times? If so, is there an opportunity to cache the result of this _transcribe? If that seems tricky, we could just leave a TODO or file an issue for later.

contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py
```diff
@@ -333,11 +360,24 @@ def _transcribe(item, parents, editRate, masterMobs):
...
+
+            # If the composition was holding UserComments and the current masterMob has
+            # no UserComments, use the ones from the CompositionMob. But if the
+            # masterMob has any, prefer them over the compositionMob, since the
+            # masterMob is the ultimate authority for a source clip.
+            if compositionUserMetadata:
+                if "UserComments" not in clipMetadata.keys():
+                    clipMetadata['UserComments'] = compositionUserMetadata
+
+            media.metadata["AAF"] = clipMetadata
```
jminor, line 375: I'm slightly confused by the variable name `clipMetadata` since it is going into the media reference. Did it come from the clip?

contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py
```diff
@@ -333,11 +360,24 @@ def _transcribe(item, parents, editRate, masterMobs):
                 otio.opentime.RationalTime(media_start, editRate),
                 otio.opentime.RationalTime(media_length, editRate)
             )
-            # copy the metadata from the master into the media_reference
-            media.metadata["AAF"] = mastermob_child.metadata.get("AAF", {})
+
+            # Copy the metadata from the master into the media_reference
+            clipMetadata = mastermob_child.metadata.get("AAF", {})
```
jminor, line 365: Since the code is modifying the dictionary later, it should be deep copied here. That way the modification doesn't accidentally get included if/when this same mastermob_child is used elsewhere.

contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py
```diff
@@ -603,6 +651,90 @@ def _find_timecode_track_start(track):
...
+                source_clips.append(slot.segment)
+
+    # No source clips, no master mob. But we still want to return
+    # the composition metadata. If there is another mastermob found 'upstream',
+    # but it does not have any UserComments metadata, we still want to use
+    # the CompositionMob's metadata.
+    if not source_clips:
+        return None, compositionMetadata
+
+    # sschulze 06/22/20 - Only expect one source clip for this case.
```
jminor, line 706: We tend to avoid leaving people's names/logins in the open source code base.

contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py
```diff
@@ -805,6 +937,9 @@ def _fix_transitions(thing):
 
 
 def _simplify(thing):
+    if not thing:
+        return thing
```
jminor, line 941: Is this meant for `None` or empty lists? Maybe a brief comment explaining the intent could be helpful.

contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py
```diff
@@ -49,7 +49,14 @@
 
 
 debug = False
-__names = set()
+
+# If enabled, output recursive traversal info of _transcribe() method.
+transcribe_debug = False
+
+
+def transcribe_log(s, indent=0, alwaysPrint=False):
```
jminor, line 57: This should have an `_` prefix since it is for internal use only.

contrib/opentimelineio_contrib/adapters/tests/test_aaf_adapter.py
```diff
@@ -884,6 +901,74 @@ def test_external_reference_paths(self):
...
+        user_comments = aaf_metadata['UserComments']
+
+        user_comment_keys = user_comments.keys()
+
+        # Check a few of the expected metadata keys of the test-AAF
+        self.assertTrue("Director" in user_comment_keys)
+        self.assertTrue("Line" in user_comment_keys)
+        self.assertTrue("Talent" in user_comment_keys)
+        self.assertTrue("Logger" in user_comment_keys)
+        self.assertTrue("Character" in user_comment_keys)
```
jminor, line 970: It would be good to verify the values also.

contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py
stefanschulze: Hm, the metadata will be assigned to the clip as well as the media reference. From my perspective it represents the metadata for a given clip.

contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py
stefanschulze: Yes, I can change that. I still think the mastermob_child will not be used anywhere else, but better safe than sorry, it should not be expensive. Just pointing out: I kept the original behavior, just added an intermediate step storing the result in a variable.

contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py
stefanschulze: sorry, that was from my tools-tree change. I'll remove that.

contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py
stefanschulze: Both. If empty or None, there is nothing to simplify.

contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py
stefanschulze: ok.

contrib/opentimelineio_contrib/adapters/tests/test_aaf_adapter.py
stefanschulze: ok, will add value-validation

contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py
stefanschulze: Usually there is one mastermob associated with a SourceClip as far as my experience goes. I don't think caching the master-mob will gain us much. But I can just test it with some of our large sequences and see if I get the same master-mob-id several times.

contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py
stefanschulze: That line is from the old comment, the updated comment right above covers everything.

contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py
stefanschulze: (But I followed the debug flag just right above, which is also internal as far as I can see. That should be _debug then as well?)

jminor: We've tried to follow that convention for functions. (e.g. `_transcode`) Maybe we should do that for module-level variables as well? @ssteinbach do you have an opinion?

You could argue that the toggles to turn this on/off are useful from outside the module - e.g. when debugging an application that uses this module.

ssteinbach: Yeah.  @meshula we should add these notes to the in progress style guide.  Some stuff:

- I would vote to remove the `transcribe_debug` variable entirely (and any if statements that were only switched by that)
- if its going to exist it should be all caps (module level constant) + prefixed by a `_` (not intended to pollute the namespace)
- I think josh was commenting on the `transcribe` function, which if its for internal usage should also be prefixed by a `_`, `_transcribe()` to not pollute the namespace and signal its intent.
- finally, we're using snake case a la pep8, so the argument should be `always_print` rather than `alwaysPrint`.

see also: https://www.python.org/dev/peps/pep-0008/#function-and-variable-names

> Function and Variable Names
Function names should be lowercase, with words separated by underscores as necessary to improve readability.
>
> Variable names follow the same convention as function names.
>
> mixedCase is allowed only in contexts where that's already the prevailing style (e.g. threading.py), to retain backwards compatibility.


Also:
>Constants
>Constants are usually defined on a module level and written in all capital letters with underscores separating words. Examples include MAX_OVERFLOW and TOTAL

ssteinbach: Just to clarify, I prefer to remove debug code - it decreases test coverage, usually requires special knowledge to turn on, and for python at least is rarely as useful as using a dedicated debugger like `pdb`.  Unless its exposed as top level interface functionality to users.  Open to counter point on that one though; what do the others think?

contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py
```diff
@@ -283,22 +292,39 @@ def _transcribe(item, parents, editRate, masterMobs):
...
+        # is a MasterMob.
+        # There are some special cases where the masterMob could be:
+        # 1) For SourceClips in the CompositionMob, the mob the SourceClip is
+        #    referencing can be our MasterMob.
+        # 2) If the source clip is referencing another CompositionMob,
+        #    drill down to see if the composition holds the MasterMob
+        # 3) For everything else, it is a previously encountered parent. Find the
+        #    MasterMob in our chain, and then extract the information from that.
+
+        child_mastermob, compositionUserMetadata = _find_mastermob_for_sourceclip(item)
```
ssteinbach, line 304: I think its probably a good idea to make sure this adapter also adheres more closely to pep8:

```suggestion
        child_mastermob, composition_user_metadata = _find_mastermob_for_sourceclip(item)
```

contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py
```diff
@@ -283,22 +292,39 @@ def _transcribe(item, parents, editRate, masterMobs):
...
+            metadata['UserComments'] = compositionUserMetadata
+
         parent_mastermobs = [
             parent for parent in parents
             if isinstance(parent, aaf2.mobs.MasterMob)
         ]
         parent_mastermob = parent_mastermobs[0] if len(parent_mastermobs) > 1 else None
+
+        if child_mastermob:
+            transcribe_log("[found child_mastermob]", indent)
```
ssteinbach, line 316: do we need these transcribe log calls?

contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py
```diff
@@ -333,11 +359,24 @@ def _transcribe(item, parents, editRate, masterMobs):
                 otio.opentime.RationalTime(media_start, editRate),
                 otio.opentime.RationalTime(media_length, editRate)
             )
-            # copy the metadata from the master into the media_reference
-            media.metadata["AAF"] = mastermob_child.metadata.get("AAF", {})
+
+            # Copy the metadata from the master into the media_reference
+            clipMetadata = copy.deepcopy(mastermob_child.metadata.get("AAF", {}))
```
ssteinbach, line 364:
```suggestion
            clip_metadata = copy.deepcopy(mastermob_child.metadata.get("AAF", {}))
```

contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py
```diff
@@ -333,11 +359,24 @@ def _transcribe(item, parents, editRate, masterMobs):
...
+
+            # If the composition was holding UserComments and the current masterMob has
+            # no UserComments, use the ones from the CompositionMob. But if the
+            # masterMob has any, prefer them over the compositionMob, since the
+            # masterMob is the ultimate authority for a source clip.
+            if compositionUserMetadata:
+                if "UserComments" not in clipMetadata.keys():
```
ssteinbach, line 371: `not in` is a check against `keys()`, don't need the extra call

```suggestion
                if "UserComments" not in clipMetadata:
```

contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py
ssteinbach: does it make sense to apply it to the media reference then?  Or is the media reference a reference to the AAF file?

contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py
```diff
@@ -186,7 +199,7 @@ def _add_child(parent, child, source):
         parent.append(child)
 
 
-def _transcribe(item, parents, editRate, masterMobs):
+def _transcribe(item, parents, editRate, indent=0):
```
ssteinbach, line 202: if we're in here:
```suggestion
def _transcribe(item, parents, edit_rate, indent=0):
```

contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py
```diff
@@ -603,6 +650,90 @@ def _find_timecode_track_start(track):
...
 
 
+def _find_mastermob_for_sourceclip(aaf_sourceclip):
+    """
+    For a given soure clip, find the related masterMob
+    """
+
+    # If the mobId of the sourceclip is a mastermob, just return that, we are done.
+    if isinstance(aaf_sourceclip.mob, aaf2.mobs.MasterMob):
+        return aaf_sourceclip.mob, None
```
ssteinbach, line 660: What are the two return values of the function?  should probably note that in the function docstring

contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py
```diff
@@ -364,6 +403,7 @@ def _transcribe(item, parents, editRate, masterMobs):
         result.out_offset = otio.opentime.RationalTime(out_offset, editRate)
 
     elif isinstance(item, aaf2.components.Filler):
+        transcribe_log("Creating Gap for {}".format(_encoded_name(item)), indent)
```
ssteinbach, line 406: If we're going to include these `trancribe_log` calls for each type, I'd rather we use a consistent format, so:

`transcribe_log("Gap", item, indent)`

...and then inside `transcribe_log()` it unrolls the name into the message `"Creating {} for {}".format(typestr, _encoded_name(item)), indent)`

...but the broader problem with these kinds of logs from adapters is that we don't have a consistent framework for it.  @reinecke just flagging this since I know its something you've thought about in the past.
and then have transcribe_log format the item name and string consistently (unless there is a lot of helpful variation in these strings)

contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py
stefanschulze: I am sorry, but I disagree in this point strongly. The AAF-object-structure is hard to understand and you need some debug tools in place to manage the complexity of _transcribe(). I actually use this flag in conjunction with pdb. I activate the flag and then do a set_trace() in the part of the traversal-tree I am interested in. So If I set it in SourceClip, I can see right away what traversal happened until this point, giving me some important context of how to introspect my current state in relation to the traversal.

contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py
stefanschulze: yes, I'll change that.

contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py
stefanschulze: will update

contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py
stefanschulze: Also just to clarify, I agree with your point in most (almost all) cases, but there are some exceptions where a complex, 'deep' structure requires some debug-bandaid-code put in place, so whoever wants to do a deep dive into that code has already something in place and does not require to scaffold the code from ground up. IMHO such situations are exceptions but should outweigh the desire of a perfect code-coverage.

jminor: I like the idea of turning it on via an adapter argument. We will be improving this adapter for a long time to come, and this sort of diagnostic feature is akin to the simplify=False option that is already an adapter argument.

jminor: Oh, and I guess I wasn't clear. It was mostly the function `transcribe_log()` that I was suggesting should be `_transcribe_log()`

stefanschulze: yes, I like the idea to have it be an adapter argument. And yes, I totally did not realize you meant the method. I'll fix that.

contrib/opentimelineio_contrib/adapters/tests/test_aaf_adapter.py
ssteinbach: Also, I think there is `assertIn`: https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertIn

ssteinbach: `self.assertIn(`

contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py
stefanschulze: yeah, good point. It started as a more generic log, but since it's now specific to the _transcribe() method, this is a good idea.

contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py
jminor: Ah, okay - this makes sense to be now.

contrib/opentimelineio_contrib/adapters/tests/test_aaf_adapter.py
jminor: It looks like `assertIn()` and `assertNotIn()` are only available in Python 3.1 and above, so we can't use them until we drop support for Python 2.7.

contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py
jminor: Yes, documenting the 2nd return value would be helpful. As it is, you have to drill down to the end of the next function to see what that is meant to be.

contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py
jminor: The original intent of the mastermob list was meant to avoid re-transcribing the same mastermobs again and again. I was under the (somewhat naive) impression that this would be common. 

In your testing so far, is the adapter faster, slower, or about the same as before this change? If it is the same or faster with a few real world AAFs then, we can leave this until another time when we do some more extensive performance testing. If it is slower, then we should look deeper since folks are already unhappy with the performance.

ssteinbach:
`assertIn` is in python 2.7: https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertIn

jminor:
Oh, you're right. I was going by this documentation which says it was introduced in 3.1: https://docs.python.org/3/library/unittest.html

contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py
stefanschulze: I would not have removed the first pass if it would have made the adapter slower. The reason (besides simplifying the code) is a 30-40% speed gain (also listed in the original commit-description), tested on three large sequences from three different shows of ours.

jminor: Faster is great! Thanks for checking.

contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py
jminor: This is an adapter argument now - does that satisfy your concern @ssteinbach ?

Issue #700, opened 2020-05-14
TimeRange contains and overlap do not match definitions


meshula:
The current implementations of contains and overlaps are sensible, but they don't match the requirements of Allen's Interval Algebra. They should be modified to match the specification as noted on this PR: https://github.com/PixarAnimationStudios/OpenTimelineIO/pull/697

ssteinbach:
Fixed by PR: #760

jminor:
Done!

PR #760, opened 2020-07-14
Enforce stricter interval algebra rules for overlaps and contains

KarthikRIyer:
#659 
Currently I have just updated the `overlaps` definition. I'll update `contains` tomorrow.


KarthikRIyer:
@meshula I've updated both the definitions.

src/opentimelineio/trackAlgorithm.cpp
```diff
@@ -27,7 +27,12 @@ Track* track_trimmed_to_range(Track* in_track, TimeRange trim_range, ErrorStatus
         }
         
         auto child_range = child_range_it->second;
-        if (!trim_range.overlaps(child_range)) {
+      if ((trim_range.duration().value() == 0) ||
+            (!trim_range.overlaps(child_range) && !child_range.overlaps(trim_range) &&
+            !trim_range.begins(child_range) && !child_range.begins(trim_range) &&
+            !trim_range.finishes(child_range) && !child_range.finishes(trim_range) &&
+            !trim_range.contains(child_range) && !child_range.contains(trim_range) &&
+            !(trim_range == child_range))) {
```
ssteinbach, line 35: - 2 space -> 4 space indentation
- this condition suggests that in addition to the foundational predicates provided by the algebra, we should provide a few high level convenience predicates.  In this case, this block is asking the _very_ useful and common question "do these two time ranges share any time points".  I think that including an inclusive `intersects` function that implements this test would be beneficial.   The intersect test should be implemented as simply and directly as possible rather than using the existing predicates exclusively.

src/py-opentimelineio/opentimelineio/algorithms/track_algo.py
```diff
@@ -47,7 +47,16 @@ def track_trimmed_to_range(in_track, trim_range):
...
+        if (trim_range.duration.value == 0) \
+                or (not trim_range.overlaps(child_range) and
+                    not child_range.overlaps(trim_range) and
+                    not trim_range.begins(child_range) and
+                    not child_range.begins(trim_range) and
+                    not trim_range.finishes(child_range) and
+                    not child_range.finishes(trim_range) and
+                    not trim_range.contains(child_range) and
+                    not child_range.contains(trim_range) and
+                    not trim_range == child_range):
```
ssteinbach, line 59: - same thing here - which means that the `intersects` test proposed previously should be exposed in python as well as C++

src/py-opentimelineio/opentime-bindings/opentime_timeRange.cpp
```diff
@@ -43,7 +43,7 @@ void opentime_timeRange_bindings(py::module m) {
         .def("clamped", (RationalTime (TimeRange::*)(RationalTime) const) &TimeRange::clamped, "other"_a)
         .def("clamped", (TimeRange (TimeRange::*)(TimeRange) const) &TimeRange::clamped, "other"_a)
         .def("contains", (bool (TimeRange::*)(RationalTime) const) &TimeRange::contains, "other"_a)
-        .def("contains", (bool (TimeRange::*)(TimeRange) const) &TimeRange::contains, "other"_a)
+        .def("contains", (bool (TimeRange::*)(TimeRange, double) const) &TimeRange::contains, "other"_a, "epsilon"_a=opentime::DEFAULT_EPSILON_s)
         .def("overlaps", (bool (TimeRange::*)(RationalTime) const) &TimeRange::overlaps, "other"_a)
         .def("overlaps", (bool (TimeRange::*)(TimeRange, double) const) &TimeRange::overlaps, "other"_a, "epsilon"_a=opentime::DEFAULT_EPSILON_s)
         .def("before", (bool (TimeRange::*)(TimeRange, double) const) &TimeRange::before, "other"_a, "epsilon"_a=opentime::DEFAULT_EPSILON_s)
```
ssteinbach, line 49: these `epsilon` arguments should be exposed as `epsilon_s` as well

tests/test_opentime.py
```diff
@@ -917,25 +917,25 @@ def test_overlaps_timerange(self):
...
 
-        self.assertTrue(tr.overlaps(tr_t))
+        self.assertFalse(tr.overlaps(tr_t))
 
         tstart = otio.opentime.RationalTime(13, 25)
         tdur = otio.opentime.RationalTime(1, 25)
         tr_t = otio.opentime.TimeRange(tstart, tdur)
 
-        self.assertTrue(tr.overlaps(tr_t))
+        self.assertFalse(tr.overlaps(tr_t))
```
ssteinbach, line 926: this is great.  we can use these old tests to test the `intersects` test and show how it differs from `overlaps`

KarthikRIyer:
@ssteinbach PR updated.

`inline constexpr bool greater_than_exclusive(lh, rh, epsilon)`
I thought we could drop exclusive here because actually we're just comparing doubles and we could pass in exclusive or inclusive times. Does that make sense?

ssteinbach:
Fixes #700

meshula:
Looks like we can't use constexpr without more changes. Why don't we make greater_than and less_than inline for now?

```
 /tmp/pip-req-build-9j_a1bx9/src/opentime/timeRange.h:28:7: note: ‘opentime::v1_0::TimeRange’ is not literal because:
453   class TimeRange {
454         ^
455  /tmp/pip-req-build-9j_a1bx9/src/opentime/timeRange.h:28:7: note:   ‘opentime::v1_0::TimeRange’ is not an aggregate, does not have a trivial default constructor, and has no constexpr constructor that is not a copy or move constructor
456  /tmp/pip-req-build-9j_a1bx9/src/opentime/timeRange.h:319:27: error: enclosing class of constexpr non-static member function ‘bool opentime::v1_0::TimeRange::lesser_than(double, double, double) const’ is not a literal type
457       inline constexpr bool lesser_than(double lhs, double rhs, double epsilon) const{
```

PR #768, opened 2020-08-03
Proposal for issue templates

meshula:
Starting with the OpenCue templates as a starting point.

https://github.com/AcademySoftwareFoundation/OpenCue/tree/master/.github


reinecke:
This is a great start!

Since users may be unaware of C++ compiler info (like when using pre-built wheels), or if they're working with C++ only python version may not be relevant, maybe we should add an "(if applicable)" to those in the bug template just to let people know we're not too stressed about it.

Also, how much do we worry about classifying feature vs. enhancement? I feel like sometimes to users the differentiation between these may be a little tough to work out - we often identify these in triage (like adding descriptive test to Markers where some research and discussion was warranted).

meshula:
I was cribbing off OpenCue, which itself is cribbed from somewhere else :) 

Worth considering starting with just Bug and Feature, since those are the two that github supports by default (https://docs.github.com/en/github/building-a-strong-community/configuring-issue-templates-for-your-repository#creating-issue-templates)

With respect to asking users to supply compiler information ~ what actually happens if you try to run pip install on a machine that does not have a compiler toolchain installed by default? Windows/MSVC or Mac/clang are obvious examples. If pip install can somehow wrangle a toolchain, then I don't think we have to ask, but if it is a prerequisite, then we'd probably end up asking the question as a follow up.

Since I suspect the latter, why don't we modify the template to say something like:

```on Mac, type clang -v, and paste the results here
on Linux, type gcc -v, and paste the results here
on Windows, type cl, and paste the results here.
If you are unable to determine your compiler via these commands, please indicate that here.
```

meshula:
> Since users may be unaware of C++ compiler info (like when using pre-built wheels)

That suggests more; the templates could be broken up by scenario

- pip and a wheel
- C++ only
- building locally for Python

PR #774, opened 2020-08-18
 delete the warning in the compilation by casting in (int)  warning: …

zebulon75018:
delete the warning in the compilation by casting in (int)  warning: …
…comparison between signed and unsigned integer expressions

just for delete a compilation message.

please just tell me if it;s not interesting. 
My two cents....
Cheer

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,7 +83,9 @@ import opentimelineio as otio
 
 timeline = otio.adapters.read_from_file("foo.aaf")
 for clip in timeline.each_clip():
-  print clip.name, clip.duration()
+  print clip.name, clip.duration() # python2
+
+  print (clip.name, clip.duration()) # python3
 ```
 
 There are more code examples here: https://github.com/PixarAnimationStudios/OpenTimelineIO/tree/master/examples
--- a/src/opentimelineio/imageSequenceReference.cpp
+++ b/src/opentimelineio/imageSequenceReference.cpp
@@ -79,7 +79,7 @@ ImageSequenceReference::ImageSequenceReference(std::string const& target_url_bas
         std::string image_num_string = std::to_string(abs(file_image_num));
 
         std::string zero_pad = std::string();
-        if (image_num_string.length() <  _frame_zero_padding) {
+        if ((int)image_num_string.length() <  _frame_zero_padding) {
             zero_pad = std::string(_frame_zero_padding - image_num_string.length(), '0');
         }
 


reinecke:
Oh nice, I shouldn't have let a compile warning get through.
I wonder if we should actually just switch `frame_zero_padding` to be an `unsigned int`, I don't think it has any valid semantic meaning if set negative.

@apetrynet What do you think?

Either way, this fix makes things better so we'd still be improving if we decided to merge as-is.

meshula:
@zebulon75018 Thanks for the catch! 

@reinecke All of the APIs in this cohort have an int as the frame number. If frame numbers must be >= 0, it makes sense to "repair" the API, and consequently the variables in the methods. If frame numbers can be negative, then a static_cast would be idiomatic in this diff, rather than a C style cast. I'm guessing, >= 0 is the answer, so what do you think about taking a quick pass over the API?

apetrynet:
@reinecke I think `frame_zero_padding` quite safely could be set to `unsigned int`. A negative frame padding would imply not having a frame number at all which makes little sense in an image sequence :)

@meshula Negative frame numbers are used in some cases where the preroll of a simulation needs to go below frame zero for instance. So, even if not that common, frame numbers should support negative numbers.

meshula:
If negative frame numbers are allowable, then the zero padding algorithm should do this ~

```
bool isNegative = frame < 0
frame = abs(frame)
zeroPad(amount - isNegative?1:0)
if (isNegative)
   prepend('-')
```
~ shouldn't it?

apetrynet:
@meshula I'm sorry, but don't think I understand what your suggesting here.

meshula:
The existing code, if zero padded to 4, will do this: `1` becomes `0001`.  `-1` becomes `00-1`. I'm proposing that `1` should become `-001`.

apetrynet:
Hmm. I'm unable to replicate that. I might of course be doing something strange here. @reinecke do you have any input?

meshula:
I'll try to repro after ASWF OpenSource Days wraps later today

meshula:
I was quite mistaken, sorry! When I was reviewing the code, every time I looked, I didn't spot that the image_num_string conversion has an abs in it. The logic does make sense for negative numbers :)

apetrynet:
Great! Thanks for double-checking!

reinecke:
Yep, negative frame numbers are allowed and (I think) have good test coverage for the examples discussed: https://github.com/PixarAnimationStudios/OpenTimelineIO/blob/887575ae1f9febe605bb9bd7889aa6e700112193/tests/test_image_sequence_reference.py#L552-L613

@meshula would you agree switching `frame_zero_padding` to be an `unsigned int` would be a good idea?

This would probably imply fixing up `static_cast` in other places. Also some additional error handling on deserialization (since al JSON ints are signed) here: https://github.com/PixarAnimationStudios/OpenTimelineIO/blob/887575ae1f9febe605bb9bd7889aa6e700112193/src/opentimelineio/imageSequenceReference.cpp#L129

Just below that point in the code is a good example of how to bubble up deserialization errors: https://github.com/PixarAnimationStudios/OpenTimelineIO/blob/887575ae1f9febe605bb9bd7889aa6e700112193/src/opentimelineio/imageSequenceReference.cpp#L148-L151

I think this would be, overall, a good update to the API and would prevent some potentially odd behavior.

@zebulon75018 Sorry to blow out the scope on what is supposed to be a simple compile warning fix, please let us know if doing all that is more than you'd want to take on.

zebulon75018:
Waou , I just want to give to the project my 2 cents ... :) , that's produced an interesting discution :).
I never imagine, but I hope that help you a little bit.

reinecke:
@zebulon75018 Thank you for pointing this one out, you managed to find a little indicator of a bigger oversight I made.

PR #792, opened 2020-09-02
xges: Add support for image sequences

thiblahute:

**Summarize your change.**

ImageSequence somehow recently landed but no support was added for the xges format. This implements that support. 

- `tests_xges_adapter.AdaptersXGESTest.test_image_sequence_example`


contrib/opentimelineio_contrib/adapters/xges.py
```diff
@@ -24,8 +24,20 @@
...
 """OpenTimelineIO GStreamer Editing Services XML Adapter."""
 import re
+import os
 import warnings
 import numbers
+try:
+    from urllib.parse import quote
+    from urllib.parse import unquote
+    from urllib.parse import urlparse
+    from urllib.parse import parse_qs
```
ssteinbach, line 34: In the OTIO codebase we usually use parentheses to group imports from the same parent package:

```python
from urllib.parse import (
    quote,
    unquote,
    urlparse,
   prase_qs,
)
```

Also we typically discourage lofting specific functions out of modules so that when you're looking at the code its more obvious that particular function comes from an imported module rather than being local to the module. I see here though that part of the issue is that in python2 the functions are split up in a couple modules, so that makes sense.

Regardless, thats 100% pure style comment, and its better to be consistent with the local module!

thiblahute: I somehow didn't now that was pep8 compliant style, will make sure to use that next time. But yes part of the issue was related to the difference between py2 and py3...

Also fwiw we do not support py2 anymore in GStreamer.

ssteinbach: good to know.  OTIO in general needs to support Python2.7 because we're following the VFX Platform, and VFX2016-2019 still has python2.7 in it: https://vfxplatform.com

Gstreamer is outside of the VFX platform but if they're not aware of it (and targeting post/animation/vfx companies), that community might also be interested in what is going on there.

thiblahute: I dropped 2.7 support for the next GStreamer release  as python2.7 is deprecated as a whole, do you happen to know what is the plan for the vfxplatform concerning py2?

PR #799, opened 2020-09-15
MLT xml adapter

### LINKED ISSUE #267: MLT XML adapter
Hi!

Just want to inform people that I've been working on an MLT (Media Lovin' Toolkit) XML adapter so we don't overlap. 
MLT is the backbone of Kdenlive, Shotcut, melt (its own command line NLE) and a whole lot of other applications. My main focus will be to support the basic elements that are supported by melt and perhaps expand to the support some other applications further down the road.

At the moment I have basic writing of the mlt format in place. 
More info on MLT [here](https://www.mltframework.org/)

apetrynet:

This PR adds an MLT xml adapter to the contrib adapters and closes #267 if/when merged 

**Summarize your change.**

The MLT adapter is a write-only adapter that creates simplified mlt files aimed at usage with the [melt](https://www.mltframework.org/docs/melt/) command line player/editor. 
Melt can both render a video file based on the mlt file and play it directly.

Features:
* Multiple tracks
* Nested tracks
* Audio/video
* LinearTimeWarp
* FreezeFrame
* ImageSequenceReference (though melt itself seems to be a bit picky on supported file formats, png works)
* Dissolve
* Fade In/Out (from/to black)

You may pass either a Timeline, Track or Clip directly to the adapter to produce an mlt file.
Arguments for the profile of the video may be passed as adapter keyword arguments when calling `otio.adapters.write_to_string()`. Please refer to the available profile data [here](https://www.mltframework.org/docs/profiles/)

Limitations:
* Audio handling/control is a bit limited. Audio clips that have the same source in video track will be ignored as MLT will include audio from the video track by default.
* Effects directly on Track and Stack is currently not implemented. I have to investigate if it's even supported in MLT.

A unittest file is added and attempts to cover all features.


apetrynet:
Closing PR
The adapter is released on [pypi](https://pypi.org/project/otio-mlt-adapter/) as discussed in TSC meetings.

PR #802, opened 2020-09-23
Fcp 7 xml better generator support

reinecke:
This updates the FCP 7 XML adapter to treat non-file source clips as generators. This means the vast majority of generators represented in Premiere Pro exports now read as `GeneratorReference` rather than `MissingReference`.

This now also parses `effectid` from `generatoritem` effects as `GeneratorReference.generator_kind`, treating that as a first-class field.


ssteinbach (approved):
LGTM

PR #803, opened 2020-09-23
CMake: Set "ARCHIVE DESTINATION" in install section

itsmattkc:

N/A

**Summarize your change.**

While `ARCHIVE DESTINATION` is usually used to define the destination of static library files, it also appears to be used for Windows/MSVC `*.lib` files in general. When it isn't defined, the `*.lib` files never get installed (only the DLLs do) and MSVC fails to link OTIO.

Change is pretty minor, simply adds an extra line in the install section of `opentimelineio` and `opentime` to ensure Windows libs (and I guess static libs if there ever were any) get installed to the `lib` folder.

`ARCHIVE DESTINATION` is described further here: https://cmake.org/cmake/help/v3.18/command/install.html

No errors or issues appear to arise from this PR.


meshula (approved):
makes sense to me.

ssteinbach (approved):

PR #804, opened 2020-09-23
Add ImageSequenceReference support to EDL adapter

### LINKED ISSUE #801: Add support for Image Sequences in the EDL adapter
## Feature Request

- [ ] New Feature
- [x] Change Existing Behavior

## Description

Currently, the EDL adapter only takes media references of type `ExternalReference` into account when storing or restoring `* FROM CLIP` style comments. 
I suggest adding basic support for `ImageSequenceReference` and landing a naming convention that can serialize and deserialize properly.

Suggested naming convention for sequences in an EDL context: `* FROM CLIP NAME: /path/filename.[1001-1200].ext`

## Context

EDL:
```
003  sq0010_sh0030_comp_v001.[1001-1092] V     C        00:00:00:00 00:00:03:20 00:00:40:02 00:00:43:22 
* FROM CLIP NAME: /mnt/storage/sq0010_sh0030_comp_v001.[1001-1092].exr

```

With the source TC and the frame range in brackets, we should have enough elements to (re)create an `ImageSequenceReference`
* `target_url_base="/mnt/storage/"` 
* `name_prefix="sq0010_sh0030_comp_v001."`
* `name_suffix=".exr"`
* `frame_zero_padding=4` # based in first frame in bracket
* `start_frame=1001`
* `available_range=[...]` # from source TC in edl

It's my experience that a lot of studios use the bracket style representation of a sequence in submission forms etc. so it should be quite familiar and readable.  
I'm a bit uncertain if the frame range should be in the reel name, but this example is a representation of what the current implementation would have done.

A few things we might need to consider:
* How do we represent a single frame? [1001-1001].ext or [1001].ext or .1001.ext
* Should the reel name include frame bracket? 
* If at all, how do applications like Resolve, Flame etc deal with edl's formed like this?
    * I think Resolve relies heavily on the source TC and not so much reel names

I believe these are good places to start implementing `ImageSequenceReference` in the adapter:

When Serializing:
https://github.com/PixarAnimationStudios/OpenTimelineIO/blob/22e2468b32f75e8e8ec8d542accba92435dcbd16/src/py-opentimelineio/opentimelineio/adapters/cmx_3600.py#L1260
and:
https://github.com/PixarAnimationStudios/OpenTimelineIO/blob/22e2468b32f75e8e8ec8d542accba92435dcbd16/src/py-opentimelineio/opentimelineio/adapters/cmx_3600.py#L579

When deserializing:
https://github.com/PixarAnimationStudios/OpenTimelineIO/blob/22e2468b32f75e8e8ec8d542accba92435dcbd16/src/py-opentimelineio/opentimelineio/adapters/cmx_3600.py#L378

 Reel Name here:
https://github.com/PixarAnimationStudios/OpenTimelineIO/blob/22e2468b32f75e8e8ec8d542accba92435dcbd16/src/py-opentimelineio/opentimelineio/adapters/cmx_3600.py#L1390

apetrynet:

closes #801

**Summarize your change.**

I was contacted by a developer who asked me if the EDL adapter would be able to store a sequence based file path in the `FROM CLIP` comment in this format: `filename.[1001-1010].ext`. As it is today, the adapter will ignore clips with `ImageSequenceReference` and treat them like `MissingReference` since  `ImageSequenceReference` lacks `target_url`. After exploring a couple of different solutions I realized it might be better to just implement native support for the ImageSequenceReference.

Changes:
* Added writing of comment (`FROM CLIP`) using the above mentioned naming convention
* Added regex parsing of the same comment on incoming EDL and create an `ImageSequenceReference` when the pattern matches.
* Made sure the new `ImageSequenceReference` get's treated like a "regular" `ExternalReference` when used as reel name etc.
* Updated feature matrix

Something to think about:
* I think the naming convention used for the sequences works, but I know this could be a point of discussion. 
    * I only accept a path as image sequence when frame range follow this pattern: `[<start_frame>-<end_frame>]`
    * Path's like: `filename.1001.ext`, `filename.[1001].ext` get treated like regular external references on "import".

new tests added to `test_cmx_3600_adapter.py`:
* `test_imagesequence_read()`
* `test_imagesequence_write()`

Please let me know if more tests are required.


reinecke:
Quick initial question: Is this format an exiting convention in EDL? Do some existing applications already support import/export of EDLs that specify image sequences in this method?

apetrynet:
@reinecke Quick answer is no, not to my knowledge. I've tried "googling" for applications that have any notion of Image Sequences in EDL's but haven't found any yet..
I'm going to check with my contact where he's using it and also do some testing of EDL's with resolve. I've seen paths with the good old "%04d" in the filename but that won't relate to the actual frames on disk.

In any case I don't think the naming convention will cause any harm as applications should just look at the `FROM CLIP` comment as a regular path. But we need to make sure of that of course..

apetrynet:
As always, if this turns out to be another case of overengineering, I'm not married to this idea and we can close the PR. Thinking about it some more, if there are no applications that can produce or read the naming convention except OTIO then that sort of defeats the purpose. Might as well go otio all the way. 
Anyway, let's hear what people have to say and I'll look for applications in the meantime.

reinecke:
I agree with your sentiment. It would be good to have the adapter able to signal something about image sequences to other tools/workflows, seems a shame to be dropping all the potentially useful `target_url` info. Would be very curious what other packages that support image sequences and EDL export do in this case - that is likely a pretty short list though.

Aanerud:
Our animation studio is built on Arnold Render, and when we are exporting EXR files from Arnold, we comp them in Nuke. 
In our Nuke Pipeline, we do as little as possible regarding metadata. On a side note, we probably should have added more metadata here.

However, this specific format of EDL files occurs when we are exporting confirmed material from Baselight, Flame, Davinchi Resolve, and more. When those applications import the EXR Sequences as video clips they are given a reel name based on the EXR Framecount and frame number that is given in the EXR name, like `q000.3_s004.0_comp_v001.[04001-04127]`

When exporting a EDL from those applications, the EDL looks like this:

```
004  q000.3_s004.0_comp_v001.[04001-04127] V     C        00:00:00:00 00:00:05:07 00:00:43:22 00:00:49:05  
* FROM CLIP NAME: q000.3_s004.0_comp_v001.[04001-04127].exr
```

What I have seen before is EDL´s printing this `q000.3_s004.0_comp_v001.%05d.exr`  
This does not confirm on Baselight or Resolve, however, Nuke Studio confirms it well.

It would be awsome if OTIO can address this, so we can make the EDL´s that confirms the correct Imagesequenses.

apetrynet:
Thanks for the clarification @Aanerud This helps a lot! I believe this PR will address you're needs and if all those apps can read edl's like these we're on good shape.

Aanerud:
I checked several EDL´s now, and it looks like the most used is this one, wher REEL NAME is just AX :
```
001  AX       V     C        00:00:00:00 00:00:01:10 01:00:00:00 01:00:01:10  
* FROM CLIP NAME: q000.8_s002.0_comp_v001.[02001-02049].exr
```

Or the one previously attached, with the actual CLIP NAME as REEL NAME, but without extencion. 

Baselight, had to have the REEL NAME in the EDL as  this 
```
001  q000.8_s002.0_comp_v001.[02001-02049]       V     C        00:00:00:00 00:00:01:10 01:00:00:00 01:00:01:10  
* FROM CLIP NAME: q000.8_s002.0_comp_v001.[02001-02049].exr
```

Resolve only needed * FROM CLIP NAME:

apetrynet:
Hey!
Thanks for checking with all these apps! I have a couple of followup questions.

Are you naming your clips in the timeline the same name including the frame range in brackets or is that something that's added at export time?
If you name them that way, what do the EDL's look like if you remove the `.[1001-1200].exr` part of the clip name in the timeline?

Also, If you remove " NAME" from the comment so it only says: `* FROM CLIP: q000.8_s002.0_comp_v001.[02001-02049].exr`
Does it still work?

Thanks!

Aanerud:
Removed `NAME` then it looks like this:
```
004  q000.3_s004.0_comp_v001.[04001-04127] V     C        00:00:00:00 00:00:05:07 00:00:43:22 00:00:49:05  
* FROM CLIP: q000.3_s004.0_comp_v001.[04001-04127].exr
```
It imported well in Resolve. 

```
007  AX       V     C        00:00:00:00 00:00:05:08 00:01:01:02 00:01:06:10
* FROM CLIP: q000.3_s007.0_comp_v001
```
This did not import at all.

apetrynet:
Awesome! Then I'm quite confident this patch will work.
Thanks!

apetrynet:
@reinecke, have you had a chance to look at this? 
I got a related question from another studio so this seems to

tin2tin:
The reel name is normally a fixed length of 8 characters. Resolve is exporting the full filename instead, which makes those edl files fail in most edl importers.

Often the file file-embedded reel name and time code is used over the file name and a time code starting at 00:00:00:00.

apetrynet:
Hi @tin2tin! You can still trunc the reel name like before. It's the default behavior of the adapter. To get the full name you have to ask for it by setting reelnamelen to None in the adapter arguments.

apetrynet:
Has anyone had a chance to look at this some more?

jminor:
Hi @apetrynet - sorry we left this unreviewed for so long. It looks like this works well and would improve compatibility with the applications mentioned in the thread. Could you rebase this onto the latest, then we could accept it.

PR #805, opened 2020-09-24
Implement subtitles schema [WIP]

### LINKED ISSUE #62: Subtitles & captions
We should support subtitles and captions.

KarthikRIyer:

Closes #62 

**Summarize your change.**

This PR adds basic support for Subtitles. I've based it on what I understood from TTML2 and from my experience with Premiere Pro. I've summarized the changes in the issue (#62). Here's an image summarizing the changes (Note change: `SerializableObject` instead of `SerializableObjectWithMetadata`):

image

Some concerns:

- I think storing a `Style` object with each `TimedText` is wasteful, but decided to have this approach because we don't have id system in otio yet. In TTML2 there's a section to specify all styles and then refer then in each TimedText. With the current approach conversion from otio to say TTML2 won't be so direct because we don't have a single set of styles. One other thing I thought of was to have a list of styles inside `Subtitle` and then have a string id in each `TimedText`. But then each `TimedText` object will have no style meaning independently/outside the Subtitle. Thoughts?

- What are `Markers` used for? Each marker has a markedRange. When we trim an `Item` or change the sourceRange shouldn't the markedRange of the markers at the start or end of the `Item` also change? I couldn't find anything that does something like this. Or is this handled in another way? Why I'm asking is if we trim the `Subtitle` or change its sourceRange, the ranges of `TimedTexts` at the start or end might change. 

Will be added later.

### DIFF
--- a/src/opentimelineio/CMakeLists.txt
+++ b/src/opentimelineio/CMakeLists.txt
@@ -27,6 +27,9 @@ set(OPENTIMELINEIO_HEADER_FILES
     stack.h
     stackAlgorithm.h
     stringUtils.h
+    subtitles.h
+    timedText.h
+    timedTextStyle.h
     timeEffect.h
     timeline.h
     track.h
@@ -62,6 +65,9 @@ add_library(opentimelineio SHARED
             stack.cpp
             stackAlgorithm.cpp
             stringUtils.cpp
+            subtitles.cpp
+            timedText.cpp
+            timedTextStyle.cpp
             timeEffect.cpp
             timeline.cpp
             track.cpp
--- a/src/opentimelineio/subtitles.cpp
+++ b/src/opentimelineio/subtitles.cpp
@@ -1,0 +1,97 @@
+#include "opentimelineio/subtitles.h"
+
+namespace opentimelineio {
+    namespace OPENTIMELINEIO_VERSION {
+
+        Subtitles::Subtitles(float extent_x,
+                             float extent_y,
+                             float padding_x,
+                             float padding_y,
+                             std::string background_color,
+                             float background_opacity,
+                             DisplayAlignment display_alignment,
+                             std::vector<TimedText *> timed_texts)
+                : Parent(),
+                  _extent_x(extent_x),
+                  _extent_y(extent_y),
+                  _padding_x(padding_x),
+                  _padding_y(padding_y),
+                  _background_color(background_color),
+                  _background_opacity(background_opacity),
+                  _display_alignment(display_alignment),
+                  _timed_texts(timed_texts.begin(), timed_texts.end()) {
+        }
+
+        Subtitles::~Subtitles() noexcept {}
+
+        TimeRange Subtitles::available_range(ErrorStatus *) const {
+            TimeRange timeRange;
+            for (auto timed_text : _timed_texts) {
+                timeRange = timeRange.extended_by(timed_text.value->marked_range());
+            }
+            return timeRange;
+        }
+
+        bool Subtitles::read_from(Reader &reader) {
+            auto result = reader.read("extent_x", &_extent_x) &&
+                          reader.read("extent_y", &_extent_y) &&
+                          reader.read("padding_x", &_padding_x) &&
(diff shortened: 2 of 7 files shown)


reinecke:
Similar to `ImageSequenceReference`, I wonder if implementing this in an adapter to try out the ergonomics might be a good idea? Also, it might be useful to see some examples of use cases and the code it takes to achieve them using the schema. In `ImageSequenceReference`, @apetrynet did a lot of iterations trying it out to make sure it worked like he expected and that really helped us make sure we had a usable schema.

apetrynet:
A usage example could be to write an adapter that can use the Subtitle Schema and write out a sub format that ffmpeg supports. _I don't think ffmpeg supports TTML2_ so converting to srt or something with an adapter might be a good test.

KarthikRIyer:
Yes, I was thinking of an srt adapter. Was waiting to hear others' thoughts on the parameters I've included and if we need to implement any other functionality.

I'll write a simple srt adapter and get back to you.

KarthikRIyer:
@apetrynet @reinecke I've implemented an srt adapter.

The tests pass for python3, but there are some extra carriage returns added to the strings with python2. Any idea why this is happening?

contrib/opentimelineio_contrib/adapters/tests/test_srt_adapter.py
```diff
@@ -0,0 +1,83 @@
...
+    "sample_data"
+)
+SRT_EXAMPLE_PATH = os.path.join(OTIO_SAMPLE_DATA_DIR, "srt_example.srt")
+SRT_OTIO_EXAMPLE_PATH = os.path.join(OTIO_SAMPLE_DATA_DIR,
+                                     "srt_example.otio")
+
+
+class SRTTest(unittest.TestCase):
+    def setUp(self):
+        fd, self.tmp_path = tempfile.mkstemp(suffix=".srt", text=True)
```
apetrynet, line 44: Hi @KarthikRIyer!
It might be a long shot, but I recognized an error when comparing multi-lined strings. Try adding `maxDiff = None` as a class variablle in `SRTTest`
```suggestion
class SRTTest(unittest.TestCase):
    maxDiff = None
    
    def setUp(self):
        fd, self.tmp_path = tempfile.mkstemp(suffix=".srt", text=True)
```

KarthikRIyer: This did't work.

maxDiff is just the max length of Diff for the compare function right?

apetrynet: Yes. I've had to use it in other tests to make them pass. The error looked similar.

KarthikRIyer:
I changed os.open() to io.open() and it worked. But I have no idea why :/

contrib/opentimelineio_contrib/adapters/tests/test_srt_adapter.py
```diff
@@ -0,0 +1,83 @@
...
+            test_data = f.read().strip()
+
+        with open(SRT_EXAMPLE_PATH) as f:
+            baseline_data = f.read().strip()
+
+        self.maxDiff = None
+        self.assertMultiLineEqual(baseline_data, test_data)
+
+    def test_otio(self):
+        st = otio.adapters.read_from_file(SRT_EXAMPLE_PATH)
```
utsab, line 70: Shouldn't this be reading SRT_OTIO_EXAMPLE_PATH instead?

KarthikRIyer: Oh, yes. I must've copied the SRT to OTIO test and forgotten to change it to test OTIO to SRT. Thanks for pointing it out. I'll fix it.

KarthikRIyer:
```
Timeline:
    Track1: [----Subtitles----]
```
If I have something like this and I do:
`Timeline.children_if(descended_from=otio.schema.Subtitles)`
This returns `[Track1, Subtitles]`. Is this the intended behaviour? Shouldn't it return just `Subtitles`?

@darbyjohnston @meshula

PR #808, opened 2020-10-01
Fix a link in documentation (to serialized schema tutorial)

Simran-B:
Discovered on Slack:
https://academysoftwarefdn.slack.com/archives/CMQ9J4BQC/p1601522521049200

**Summarize your change.**

- Fix file extension of link
- Remove `target="_blank"` as this is an internal link


jminor:
@Simran-B could we get a Contributor License Agreement from you? I know it seems like a lot for such a small change, but we need to make sure all contributions are covered by this.

See here for details:
https://opentimelineio.readthedocs.io/en/latest/tutorials/contributing.html

Simran-B:
@jminor No problem. Same for my company, although I thought that it would be less of an issue in the US than in Germany. Sent the signed CLA via email.

jminor:
CLA received. Thanks 👍

PR #820, opened 2020-10-13
Python 3 deprecation fixes (including Python 3.9 support)

reinecke:

Addressed deprecation warning:
```
opentimelineio/core/_core_utils.py:220: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.9 it will stop working
  if not isinstance(item, collections.Iterable):
```
We do this by importing `abc` from collections where available (Python 3.3+), otherwise we alias `collections` as `abc`.

This fixes OTIO in Python 3.9.

Also addressed:
- Add `__next__` (favored instead of `next` in Python 3.0+) - [PEP 3114](https://www.python.org/dev/peps/pep-3114/)
- Where available (Python 3.0+) use `inspect.getfullargspec` rather than `inspect.getargspec`

Existing tests covered the use cases already, under Python 3.9 they were failing. I successfully ran the test suite in Python 3.9 locally and verified the deprecation warnings are no longer dropping to the shell in other versions.

### DIFF
--- a/src/py-opentimelineio/opentimelineio/adapters/adapter.py
+++ b/src/py-opentimelineio/opentimelineio/adapters/adapter.py
@@ -40,6 +40,13 @@ from .. import (
 )
 
 
+try:
+    # Python 3.0+
+    getargspec = inspect.getfullargspec
+except AttributeError:
+    getargspec = inspect.getargspec
+
+
 @core.register_type
 class Adapter(plugins.PythonPlugin):
     """Adapters convert between OTIO and other formats.
@@ -307,7 +314,7 @@ class Adapter(plugins.PythonPlugin):
                 for fn_name in _FEATURE_MAP[feature]:
                     if hasattr(self.module(), fn_name):
                         fn = getattr(self.module(), fn_name)
-                        args = inspect.getargspec(fn)
+                        args = getargspec(fn)
                         docs = inspect.getdoc(fn)
                         break
 
--- a/contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py
+++ b/contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py
@@ -32,7 +32,11 @@ import os
 import sys
 import numbers
 import copy
-import collections
+try:
+    # Python 3.3+
+    import collections.abc as abc
+except ImportError:
+    import collections as abc
 import fractions
 import opentimelineio as otio
 
@@ -552,7 +556,7 @@ def _transcribe(item, parents, edit_rate, indent=0):
     #     elif isinstance(item, pyaaf.AxProperty):
     #         self.properties['Value'] = str(item.GetValue())
 
-    elif isinstance(item, collections.Iterable):
+    elif isinstance(item, abc.Iterable):
         msg = "Creating SerializableCollection for Iterable for {}".format(
             _encoded_name(item))
         _transcribe_log(msg, indent)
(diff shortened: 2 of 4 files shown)


meshula (commented):
Out of curiosity is the name "abc" a reasonable acronym for something, or is it some legacy hackiness? If hackiness, I wonder if there's some nicer name to import as. If I am reading "collections.Iterable" it's readable, but reading "abc.Iterable" is kind of out of left field for someone like me who isn't deep into Python arcana :)

apetrynet:
abc -> abstract base classes

meshula:
Ah, that does sound overly generic to be honest since that's a basic computer science term that doesn't tell us anything about what's being abstracted beyond a "class", which isn't what actually going on here. My personal preference would be that we use a descriptive name.

reinecke:
I agree about the `abc` thing feeling a bit like arcana. Ultimately, I selected this approach because:
1. It avoids inventing a new name for the module
2. The main impact is in a core module that's implementing stuff mostly focussed on leveraging these `collections.abc` classes so hopefully the readers are more inclined to be familiar with them.

Looking at the PEP that introduced this module (https://www.python.org/dev/peps/pep-3119/), it turns out there is just a root level `abc`, so my approach could cause confusion i.e. `import abc` vs. `from collections import abc`.

Looking at how [six](https://six.readthedocs.io/#module-six.moves) handles this, it uses the rename `collections_abc`. Given that all this code is doing is trying to achieve what six does without adding a dependency, I think I'll update mine to match it's behavior.

Side question for the community: How much of a problem would adding a dependency on six be? It's considered the standard for many of the Python 2/3 interoperability problems and would also flag the places in code that can be scrubbed when we drop Python 2 support.

meshula (approved):
Thanks for the change!

meshula:
The six library is a lightweight and robust dependency, in my experience ~ I'm not a Python guru, but six has never caused me a maintenance issue, so with that caveat, it gets a thumbs up from me.

reinecke:
I'm going to drop the six library thing into an issue to allow that discussion to have it's own place.

jminor (approved):

src/py-opentimelineio/opentimelineio/adapters/adapter.py
```diff
@@ -40,6 +40,13 @@
 )
 
 
+try:
+    # Python 3.0+
+    getargspec = inspect.getfullargspec
+except AttributeError:
+    getargspec = inspect.getargspec
```
ssteinbach, line 47: do we have a consistent way of labelling hacks like this?  Just something to make it easier to find these later when we eventually drop python 2 support.

Also, I wonder if it should follow the newer convention rather than older one.  In other words:

`inspect_getfullargspec = ...` rather than `getargspec = ...`

ssteinbach (commented):
One question in here, and I haven't followed the discussion fully, but this fixes the code but doesn't add the CI support - is that a mistake?  If we're going to fix the code, should we also just add CI support?  It seems like we're committing to supporting Python3.9 at that point?  That would also imply updating the badge.

src/py-opentimelineio/opentimelineio/adapters/adapter.py
reinecke: I can take a quick peek at turning on Python 3.9 for our CI.
I agree with your note about using the new `getfullargspec`, will update for that.
In terms of consistent ways of labeling the hack, I think implementing #821 would help for a lot of these, though this particular one doesn't seem to be in the cases it addresses.

Issue #333, opened 2018-09-25
Enable macOS and Windows CI on GitHub Actions


jminor:
Travis supports macOS testing too. We should enable that to ensure that we don't have platform-specific bugs/problems.

Ideally we would have Windows too, but Travis doesn't support that (yet).

jminor:
It looks like this should be very easy to turn on: https://docs.travis-ci.com/user/multi-os/

jminor:
Whoa, that sounds like a mess. I had no idea it would be such a pain. Thanks for digging into this. Let’s leave it as-is for now. 

In the long run we can look for other CI services (which might happen for other reasons), or wait until Travis improves this.

jminor:
Now that we've switched to GitHub Actions, maybe we can give this another try? It would be nice to know that everything works as expected on Mac, Windows, and Linux.

KarthikRIyer:
If tox works without changes on Mac and windows then it should be easy.

reinecke:
Given that we are looking to phase out tox, this may be a good opportunity to try building wheels using `cibuildwheel`: https://cibuildwheel.readthedocs.io/en/stable/setup/#github-actions

(see #666 also for more info about wheel building)

ssteinbach:
closed by #873

PR #831, opened 2020-11-05
Ubuntu CI build wheels [WIP]

KarthikRIyer:

#333 

**Summarize your change.**

This PR introdces cibuildwheel instead of tox. The issue linked above talks about ci for windows and MacOS too. That will be added once it works as expected for Ubuntu.

It doesn't test the adapters nor does it run and upload coverage at the moment. Only the core otio tests are run.

PS: I apologise for the number of commits. It took a long time to get this working.


ssteinbach:
I was looking at an example linked from the cibuildwheel docs, and looks like they have a specific action to build the wheels/upload them, triggered on release.  Maybe that is something we want to do as well?  I can say from experience that the manual upload-to-pypi thing we do right now is a painful part of the release process.

PR #842, opened 2020-12-10
#841 Track range_of_all_children applies TimeEffect effect to duratio…

### LINKED ISSUE #841: otioview doesn't seem to alter clip length due to a TimeWarp effect
adding this is a bug rather than a feature, as if I have a clip on a track which is e.g. slowed to half speed, then the rest of the track in otioview will be out of sync.

splidje:
```Fixes #841```

Track range_of_all_children applies TimeEffect effect to duration of clip.

### DIFF
--- a/src/opentime/rationalTime.h
+++ b/src/opentime/rationalTime.h
@@ -135,6 +135,14 @@ public:
         return RationalTime {-lhs._value, lhs._rate};
     }
 
+    friend RationalTime operator* (RationalTime lhs, double rhs) {
+        return RationalTime {lhs._value * rhs, lhs._rate};
+    }
+
+    friend RationalTime operator/ (RationalTime lhs, double rhs) {
+        return RationalTime {lhs._value / rhs, lhs._rate};
+    }
+
     friend bool operator> (RationalTime lhs, RationalTime rhs) {
          return (lhs._value / lhs._rate) > (rhs._value / rhs._rate);
     }
--- a/src/opentimelineio/linearTimeWarp.h
+++ b/src/opentimelineio/linearTimeWarp.h
@@ -2,6 +2,7 @@
 
 #include "opentimelineio/version.h"
 #include "opentimelineio/timeEffect.h"
+#include "opentime/timeRange.h"
 
 namespace opentimelineio { namespace OPENTIMELINEIO_VERSION  {
     
@@ -25,6 +26,10 @@ public:
 
     void set_time_scalar(double time_scalar) {
         _time_scalar = time_scalar;
+    }
+
+    virtual TimeRange output_range(TimeRange input_range,  ErrorStatus* error_status) const {
+        return TimeRange(input_range.start_time(), input_range.duration() / _time_scalar);
     }
 
 protected:
--- a/src/opentimelineio/track.cpp
+++ b/src/opentimelineio/track.cpp
@@ -2,6 +2,7 @@
 #include "opentimelineio/transition.h"
 #include "opentimelineio/gap.h"
 #include "opentimelineio/vectorIndexing.h"
+#include "opentimelineio/timeEffect.h"
 
 namespace opentimelineio { namespace OPENTIMELINEIO_VERSION  {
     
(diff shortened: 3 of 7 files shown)

### CODE REVIEW

File: src/opentimelineio/linearTimeWarp.h [hunk 1]
```diff
@@ -27,6 +28,10 @@ class LinearTimeWarp : public TimeEffect {
         _time_scalar = time_scalar;
     }
 
+    virtual TimeRange output_range(TimeRange input_range,  ErrorStatus* error_status) const {
+        return TimeRange(input_range.start_time(), input_range.duration() / _time_scalar);
```
reinecke, line 32: Here you could construct a`TimeTransform` and then use `.applied_to(input_range.duration()` to compute the scaled duration. 

Also, another thing to consider is what happens when `_time_scalar` is zero (like for [FreezeFrame](https://github.com/PixarAnimationStudios/OpenTimelineIO/blob/master/src/opentimelineio/freezeFrame.h)). Perhaps, if it doesn't exist, a unittest for the `FreezeFame` case should be added.

src/opentime/rationalTime.h
```diff
@@ -135,6 +135,14 @@ class RationalTime {
...
     }
 
+    friend RationalTime operator* (RationalTime lhs, double rhs) {
+        return RationalTime {lhs._value * rhs, lhs._rate};
+    }
+
+    friend RationalTime operator/ (RationalTime lhs, double rhs) {
+        return RationalTime {lhs._value / rhs, lhs._rate};
+    }
+
```
reinecke, line 145: We originally omitted overloading `/` and `*` operators for `RationalTIme` because we felt that the meaning of division and multiplication in the context of RationalTime could potentially be ambiguous. Instead we provided `TimeTransform` to make sure it was explicit how these operations work.

src/opentimelineio/track.cpp
```diff
@@ -196,7 +197,13 @@ std::map<Composable*, TimeRange> Track::range_of_all_children(ErrorStatus* error
                                       transition->out_offset() + transition->in_offset());
         }
         else if (auto item = dynamic_cast<Item*>(child.value)) {
-            auto last_range = TimeRange(last_end_time, item->trimmed_range(error_status).duration());
+            auto output_range = TimeRange(RationalTime(0), item->trimmed_range(error_status).duration());
```
reinecke, line 200: Since considerable work is potentially being done in the following `for` loop, should `error_status` be right after instead of later below?

reinecke (changes requested):
Hi @splidje,
Thanks for jumping in!

I think there is some misleading terminology in OTIO when `TimeWarp` is applied to a clip. Here is the code in python I use if I have a clip in a timeline that has a `LinearTimeWarp` applied and I want to calculate the original source frame range:
```
time_warps = [
    effect for effect in clip.effects
    if isinstance(effect, otio.schema.LinearTimeWarp)
]
time_scalar = functools.reduce(
    lambda x, y: (x * y.time_scalar if y.time_scalar != 0 else 0),
    time_warps,
    1,
 )
effect_transform = otio.opentime.TimeTransform(scale=time_scalar)

 computed_duration = effect_transform.applied_to(clip.source_range.duration)
```

This works well with EDLs I've been using. The way I think of it is that the retimed clip is acting like it's own new piece of media and if you want to evaluate the range in the original media you have to do so backwards through the time warps.

@ssteinbach @meshula @jminor Would you mind checking my mental model on this stuff, I've always been a little fuzzy about the mental model for TimeWarps.

Depending on the answer to that, if we decide to move forward we'd want to have some unittesting in place before accepting the PR.

splidje:
Thank you for the review. I wasn't aware of TimeTransform. I'll make those changes.

It makes sense in my head that a clip has `source_range` referring to the section of the source media to use, and then any number of time effects, each mapping output time to source time. So you can concatenate them all to form an effective mapping from output time to source time. Also part of that transform can helpfully tell you what the domain of the mapping is with a given range.

if you had the mapping (output -> source):
1 -> 5
2 -> 0
3 -> 3
and the source range was 0-5, then the output range would be 1-3. Then when actually trying to produce a frame, we know that the clip effectively has the "source range" 1-3, we ask it for the first frame: 1,  the time effect maps that to 5 in order to address the source media.

splidje:
I'm probably being naive though. As you say, freeze frame is an anomaly. Should it be separate from a time effect as it has special behaviour, or should time effect be designed to encompass freeze frame? Not sure.

A freeze frame needs a duration. Actually, it's just occurred to me that it does fit the model I described above. A time effect provides two things:
1) mapping from output time to source time
2) effective duration

so a freeze frame would just return its duration property for duration, and the mapping always returns the same source frame.

PR #848, opened 2020-12-14
detect cycles

davidbaraff:
```Fixes cycles in SerializableObjects cause crashes #847```

Modified the logic near #ifdef INSTANCING_SUPPORT to make use of the same table for checking for already encountered objects; however, because we only want to detect cycle and let mere instancing pass through (i.e. we don't demand things be a tree, we just demand they be acyclic) we remove an object from the table as we leave the function where we first encountered it.

In this way, we should be back to not being able to crash the system by writing bad Python code.

New tests (in tests.test_serializable_object.SerializableObjTest):
    test_instancing_without_instancing_support
    test_cycle_detection

### DIFF
--- a/tests/test_serializable_object.py
+++ b/tests/test_serializable_object.py
@@ -173,6 +173,29 @@ class SerializableObjTest(unittest.TestCase, otio_test_utils.OTIOAssertions):
         o = otio.core.SerializableObject()
         self.assertTrue(o)
 
+    def test_instancing_without_instancing_support(self):
+        o = otio.core.SerializableObjectWithMetadata()
+        c = otio.core.SerializableObjectWithMetadata()
+        o.metadata["child1"] = c
+        o.metadata["child2"] = c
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Review by meshula — APPROVED** [review]
I like this solution, thank you!

File: tests/test_serializable_object.py [hunk 1]
```diff
@@ -173,6 +173,29 @@ def test_truthiness(self):
...
+
+    def test_cycle_detection(self):
+        o = otio.core.SerializableObjectWithMetadata()
+        o.metadata["myself"] = o
+
+        # Note: If we ever enable INSTANCING_SUPPORT in the C++ code,
+        # then modify the code below to be:
+        #   oCopy = o.clone()
+        #   self.assertTrue(oCopy is oCopy.metadata["myself"])
+        with self.assertRaises(ValueError):
```
ssteinbach, line 196: ah great, thanks for putting these comments in, very helpful!

ssteinbach (approved):
LGTM

PR #849, opened 2020-12-15
Added _current_object_nest to SerializedObject

### LINKED ISSUE #847: cycles in SerializableObjects cause crashes
## Bug Report

If C++ OTIO is compiled with INSTANCING_SUPPORT then cyclic references are handled fine.  Without it, however, the following code causes a crash:

```
import opentimelineio as otio
so = otio.core.SerializableObjectWithMetadata()
so.metadata["child"] = so
so.clone()

```

However, we need to distinguish the above with the following, which needs to remain legal:
```
import opentimelineio as otio
so = otio.core.SerializableObjectWithMetadata()
soA = otio.core.SerializableObjectWithMetadata()
so.metadata["child1"] = soA
so.metadata["child2"] = soA
so.clone()
```

That is, during serialization, the object `soA` will be encountered more than once, yet this should not yield an error (because it doesn't now).  In other words, the only behavior change we're trying to make is throwing an exception (or returning an error in C++) in cases where we would certainly crash to infinite recursion.

splidje:
```Fixes #847```

Added `_current_object_nest` to SerializedObject for maintaining the 'stack' of instances currently being written during recursive calls to write methods


davidbaraff:
Note: I don't believe we want to  accept both this PR and the one I wrote, about cycle detection.

splidje:
ah yes, I saw that PR. Much better approach. I think I was avoiding stepping on the toes of the existing logic, in case I didn't understand what it was doing fully.

PR #850, opened 2020-12-16
Kdenlive improvements

splidje:
General improvements to kdenlive adapter made while trying to convert a .prproj using my otio-premiereproject adapter to .kdenlive and then opening in Kdenlive.


ssteinbach:
Did this add any new functionality?  If so, could you also add some unit tests for those cases?  the tests are in `tests/test_kdenlive_adapter.py`.  Didn't see any modifications to the tests, so just wanted to check.

splidje:
Yes, I'm sure there are some tests I can write to check on the issues I've fixed. One straightforward one would be related to my most recent fixes to converting a RationalTime to/from Kdenlive time string.

I'll get on that when I can.

ssteinbach:
Was this superceded by #869 ?  Just checking, thanks!

PR #854, opened 2021-01-01
Expose OTIO version numbers as integers

### LINKED ISSUE #844: OTIO Version numbers as integers
## Feature Request

OTIO major, minor, bugfix version numbers exposed as integers on the `otio` module

## Description

Hi!
I'm writing some functionality to "translate" a Hiero sequence into a OpenTimelineIO Timeline, based also on this PR:  https://github.com/PixarAnimationStudios/OpenTimelineIO/pull/284 (which I was thinking of contributing to at some point)

Currently in our studio we're on `0.12.1`, which lacks the `ImageSequenceReference.1` class.
To make our code as much backward compatible as possible, I'm branching off based on the `otio.__version__` (which is  string, so in our case is`"0.12.0"`).

Right now I'm just splitting the string in 3 to get the major, minor, bugfix version, but I was wondering if you guys were thinking of having separate constants for that, since it's common practice and it would make the code look a bit cleanier since we wouldn't need to have our own "split the string" one line in difference places.

## Context

Instead of this:
```
OTIO_MAJOR_VERSION, OTIO_MINOR_VERSION, OTIO_BUGFIX_VERSION =  [int(v) for v in otio.__version__.split(".")]
if OTIO_MINOR_VERSION > 13:
    # use ImageSequenceReference.1
else:
    # use ExternalReference.1
```

it would be

```
if otio.MINOR_VERSION > 13:
    # use ImageSequenceReference.1
else:
    # use ExternalReference.1
```

This is a very minor request, but I was just wondering if there was a specific reason why you went with the string approach.
I don't know how you're generating this metadata, but I'd be happy to put together a quick PR in case I can help with that.

Thanks!

Valerio

vvzen:
### Issue
Fixes #844 

### Description
Adds the major, minor and patch versions of OTIO as integers
and strings metadata so that they can be accessed on the `otio`
module, like this:

```
>>> otio.__maj_version__
0
>>> otio.__min_version__
14
>>> otio.__patch_version__
'0.dev1'
```

### Tests
I haven't added any tests since I wasn't sure if it was worth?
It's my understanding that right now the PROJECT_METADATA["version"] is set manually, so maybe it could be useful if we could avoid somebody potentially using a wrong formatting and messing up the `setup.py` ?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -288,6 +288,9 @@ PROJECT_METADATA = {
 
 METADATA_TEMPLATE = """
 __version__ = "{version}"
+__min_version__ = {major_version}
+__maj_version__ = {minor_version}
+__patch_version__ = "{patch_version}"
 __author__ = "{author}"
 __author_email__ = "{author_email}"
 __license__ = "{license}"
@@ -296,6 +299,16 @@ __license__ = "{license}"
 
 def _append_version_info_to_init_scripts(build_lib):
     """Stamp PROJECT_METADATA into __init__ files."""
+
+    # expose the maj, min, patch versions programmatically
+    version_tokens = PROJECT_METADATA["version"].split(".")
+    major_ver, minor_ver = [
+        int(v) for v in version_tokens[:2]
+    ]
+    patch_version = ".".join(version_tokens[2:])
+    PROJECT_METADATA["major_version"] = major_ver
+    PROJECT_METADATA["minor_version"] = minor_ver
+    PROJECT_METADATA["patch_version"] = patch_version
 
     for module, parentdir in [
             ("opentimelineio", "src/py-opentimelineio"),


ssteinbach:
@vvzen per the conversation on slack, what do you think about using the `packaging` module as discussed on this wiki page: https://wiki.aswf.io/display/HOME/Python+Package+Versioning ?

I think I'd like to keep this as a string for now, since that seems to be the adopted pattern in python.

vvzen:
Hi @ssteinbach !
Unfortunately, in my studio we're still mostly on 2.7 ( you wouldn't believe it, but there are even some places where we are 2.6!) and we don't have a software environment where we can simply pip install, everything is quite curated/manual and external dependencies are kept to a minimum (things are slowly changing, though).

That being said, this was really a minor thing. I'm happy to close it, and just do the splitting when needed on my side.
Thanks for the discussion on the slack and the confluence page with the link, it was quite informative! 😀

PR #861, opened 2021-01-08
Document Metadata structure

jminor:
This PR adds documentation about how metadata should be nested inside a namespace.

### DIFF
--- a/docs/tutorials/otio-file-format-specification.md
+++ b/docs/tutorials/otio-file-format-specification.md
@@ -38,8 +38,9 @@ Member fields of each data type are encoded as key/value pairs in the containing
 or dictionary. If the value is a dictionary, then it will often be an OTIO data type. In some cases (specifically metadata) it can be a regular JSON dictionary.
 
 OTIO JSON files are typically formatted with indentation to make them easier to read. This makes the files slightly larger, but dramatically improves human
-readability which makes debugging much easier. Since human readablility and ease of use are explicit goals of the OpenTimelineIO project, we recommend that you
-do not minify OTIO JSON unless absolutely necessary. If file size is really important, you should probably gzip them instead.
+readability which makes debugging much easier. Furthermore, the OTIO library will write the keys of each object in a predictable order to help with change tracking, comparisons, etc.
+
+Since human readablility and ease of use are explicit goals of the OpenTimelineIO project, we recommend that you do not minify OTIO JSON unless absolutely necessary. If file size is really important, you should probably gzip them instead.
 
 ## Nesting 
 
--- a/docs/tutorials/otio-file-format-specification.md
+++ b/docs/tutorials/otio-file-format-specification.md
@@ -25,7 +25,7 @@ OTIO does not support instancing, so you cannot reference the same object multip
 
 The top level object in an OTIO file can be any OTIO data type, but is typically a Timeline. This means that most use cases will assume that the top level
 object is a Timeline, but in specific workflows, you can read and write otio files that contain
-just a Clip, Sequence, RationalTime, or any other OTIO data type. Due to the nature of JSON, you could also read/write an array of objects, but we
+just a Clip, Track, RationalTime, or any other OTIO data type. Due to the nature of JSON, you could also read/write an array of objects, but we
 recommend that you use the OTIO SerializableCollection data type in this case so that you can attach metadata to the container itself. Code that reads an OTIO
 file should guard against unexpected top level types and fail gracefully.
 Note also, that this is the reason that there is no top level file format version in OTIO. Each data type has a version instead to allow for more granular versioning.
@@ -43,14 +43,14 @@ do not minify OTIO JSON unless absolutely necessary. If file size is really impo
 
 ## Nesting 
 
-A Timeline has one child, called "tracks" which is a Stack. Each of that Stack's children is a Sequence. From there on down each child can be any of these types: Clip, Filler, Stack, Sequence.
+A Timeline has one child, called "tracks" which is a Stack. Each of that Stack's children is a Track. From there on down each child can be any of these types: Clip, Filler, Stack, Track.
 
 In a simple case with one track of 3 clips, you have:
 
 ```
 Timeline "my timeline"
   Stack "tracks"
-    Sequence "video track"
+    Track "video track"
       Clip "intro"
       Clip "main"
       Clip "credits"
@@ -73,7 +73,7 @@ TODO: Explain how metadata works and why we do it that way.
         "OTIO_SCHEMA": "Stack.1", 
         "children": [
             {
-                "OTIO_SCHEMA": "Sequence.1", 
+                "OTIO_SCHEMA": "Track.1",
                 "children": [
                     {
                         "OTIO_SCHEMA": "Transition.1",
@@ -198,7 +198,7 @@ TODO: Explain how metadata works and why we do it that way.
                 "kind": "Video", 
                 "markers": [], 
                 "metadata": {}, 
-                "name": "Sequence1", 
+                "name": "Track1",
                 "source_range": null
             }
         ], 
--- a/docs/tutorials/otio-file-format-specification.md
+++ b/docs/tutorials/otio-file-format-specification.md
@@ -5,14 +5,10 @@
 This DRAFT describes the OpenTimelineIO JSON File Format as of OTIO Beta 13.
 
 ## Note
-We strongly recommend that you use the OpenTimelineIO library to read and write OTIO files instead of implementing your own parser or writer. However, there will
-undoubtedly be cases where this is not practical, so we document the format here for clarity.
+It is strongly recommended that everyone use the OpenTimelineIO library to read and write OTIO files instead of implementing a separate parser or writer.
 
 ## Naming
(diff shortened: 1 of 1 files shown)


docs/tutorials/otio-file-format-specification.md
```diff
@@ -2,9 +2,7 @@
 
 ## Version
 
-    This may be out of date. 
-
-This DRAFT describes the OpenTimelineIO JSON File Format as of OTIO Alpha 5.
+This DRAFT describes the OpenTimelineIO JSON File Format as of OTIO Beta 13.
```
ssteinbach, line 5: Is there a way to embed this into the automatically generated document?  Otherwise we should add to our release process updating this document.

docs/tutorials/otio-file-format-specification.md
```diff
@@ -40,19 +38,20 @@ Member fields of each data type are encoded as key/value pairs in the containing
 or dictionary. If the value is a dictionary, then it will often be an OTIO data type. In some cases (specifically metadata) it can be a regular JSON dictionary.
 
 OTIO JSON files are typically formatted with indentation to make them easier to read. This makes the files slightly larger, but dramatically improves human
-readability which makes debugging much easier. Since human readablility and ease of use are explicit goals of the OpenTimelineIO project, we recommend that you
-do not minify OTIO JSON unless absolutely necessary. If file size is really important, you should probably gzip them instead.
+readability which makes debugging much easier. Furthermore, the OTIO library will write the keys of each object in a predictable order to help with change tracking, comparisons, etc.
+
+Since human readablility and ease of use are explicit goals of the OpenTimelineIO project, we recommend that you do not minify OTIO JSON unless absolutely necessary. If file size is really important, you should probably gzip them instead.
```
ssteinbach, line 43:
```suggestion
Since human readablility and ease of use are explicit goals of the OpenTimelineIO project, it is recommended that OTIO JSON not be minified unless absolutely necessary. If a minimum file size is desired, the recommendation is to use gzip rather than minifying.
```

docs/tutorials/otio-file-format-specification.md
```diff
@@ -62,20 +61,40 @@ In order to make the tree structure easy to traverse, we use the name "children"
 
 ## Metadata
 
-TODO: Explain how metadata works and why we do it that way.
+Timeline, Stack, Track, Clip, MediaReferece, and most other OTIO objects all have a `metadata` property.
```
reinecke, line 64: It may be worth explicitly calling out here that the `metadata` property is an arbitrarily deep JSON dictionary like structure with no enforced schema.

docs/tutorials/otio-file-format-specification.md
```diff
@@ -62,20 +61,40 @@ In order to make the tree structure easy to traverse, we use the name "children"
...
+This is intended to be a place for you to put information that does not fit into the standard properties.
```
ssteinbach, line 65:
```suggestion
This is intended to be a place to put information that does not fit into the schema defined properties.
```

docs/tutorials/otio-file-format-specification.md
```diff
@@ -62,20 +61,40 @@ In order to make the tree structure easy to traverse, we use the name "children"
...
+The core of OTIO doesn't do anything with metadata, it only carries it along so that adapters, scripts,
+applications, or other workflows can use that metadata however needed. For example, you will find that
```
ssteinbach, line 67:
```suggestion
applications, or other workflows can use that metadata however needed. For example, 
```

docs/tutorials/otio-file-format-specification.md
```diff
@@ -62,20 +61,40 @@ In order to make the tree structure easy to traverse, we use the name "children"
...
+several of the adapters shipped with OTIO use metadata to store information that doesn't (yet) fit into
+the core OTIO schema.
+
+Due to the fact that many different workflows can and will use metadata, it is important to group your
+metadata inside a namespace so that all of those can coexist without stomping on each other. In the example below, you can see that there is metadata on the Timeline and on several Clips for both a hypothetical `my_playback_tool` and `my_production_tracking_system` that could coexist with anything else you add under a similar namespace.
+
+Metadata can also be useful when prototyping new OTIO schemas. An existing object can be extended with metadata which can later be upgraded into a new schema version, or a schemadef plugin.
```
reinecke, line 74: Perhaps "migrated into top-level attributes on a new schema version" would be clearer than "upgraded" to convey that the data should move out of the `metadata` dictionary and on to "top-level" properties.

docs/tutorials/otio-file-format-specification.md
```diff
@@ -62,20 +61,40 @@ In order to make the tree structure easy to traverse, we use the name "children"
...
-TODO: Explain how metadata works and why we do it that way.
+Timeline, Stack, Track, Clip, MediaReferece, and most other OTIO objects all have a `metadata` property.
+This is intended to be a place for you to put information that does not fit into the standard properties.
+The core of OTIO doesn't do anything with metadata, it only carries it along so that adapters, scripts,
+applications, or other workflows can use that metadata however needed. For example, you will find that
+several of the adapters shipped with OTIO use metadata to store information that doesn't (yet) fit into
+the core OTIO schema.
+
+Due to the fact that many different workflows can and will use metadata, it is important to group your
+metadata inside a namespace so that all of those can coexist without stomping on each other. In the example below, you can see that there is metadata on the Timeline and on several Clips for both a hypothetical `my_playback_tool` and `my_production_tracking_system` that could coexist with anything else you add under a similar namespace.
```
ssteinbach, line 72:
```suggestion
metadata inside a namespace so that all of those can coexist without encountering a name collision. In the example below, there is metadata on the Timeline and on several Clips for both a hypothetical `my_playback_tool` and `my_production_tracking_system` that could coexist with anything else added under a similar namespace.
```

reinecke (commented):
This is a good add! Just a couple small notes about clarification.

docs/tutorials/otio-file-format-specification.md
```diff
@@ -27,7 +25,7 @@ OTIO does not support instancing, so you cannot reference the same object multip
 
 The top level object in an OTIO file can be any OTIO data type, but is typically a Timeline. This means that most use cases will assume that the top level
 object is a Timeline, but in specific workflows, you can read and write otio files that contain
```
ssteinbach, line 27:
```suggestion
object is a Timeline, but in specific workflows, otio files can be read or written that contain
```

docs/tutorials/otio-file-format-specification.md
```diff
@@ -27,7 +25,7 @@ OTIO does not support instancing, so you cannot reference the same object multip
...
-just a Clip, Sequence, RationalTime, or any other OTIO data type. Due to the nature of JSON, you could also read/write an array of objects, but we
+just a Clip, Track, RationalTime, or any other OTIO data type. Due to the nature of JSON, you could also read/write an array of objects, but we
```
ssteinbach, line 28:
```suggestion
just a Clip, Track, RationalTime, or any other OTIO data type. Due to the nature of JSON, arrays of objects can also be read/written, but we
```

docs/tutorials/otio-file-format-specification.md
```diff
@@ -27,7 +25,7 @@ OTIO does not support instancing, so you cannot reference the same object multip
...
 recommend that you use the OTIO SerializableCollection data type in this case so that you can attach metadata to the container itself. Code that reads an OTIO
```
ssteinbach, line 29:
```suggestion
recommend using the OTIO SerializableCollection data type in this case so that metadata can be attached to the container itself. Code that reads an OTIO
```

ssteinbach (changes requested):
- remove "you"
- one question regarding connecting up with the autogeneration system
- some wording tweaks

let me know what you think

docs/tutorials/otio-file-format-specification.md
jminor: Given that we didn't actually update the contents of this since Alpha 5, I think it was correct. We should add a manual process to any schema enhancements to include documentation.

ssteinbach: Sorry, to include updating the documentation?  Or to include documentation in general?  

Either way, we should keep a list of which pages to check then.  We can add a todo to our TSC documentation pass list.

ssteinbach (approved):

docs/tutorials/otio-file-format-specification.md
jminor: I think both. Specifically any PR that changes schema should update this document (in addition to the autogenerated one) - and when we do a release, we should have a reminder "check that the file format spec documentation reflects any schema changes" in case we missed anything important.

jminor:
I removed all other instances of "you" and "we" from the document. Phrasing suggestions welcome :)

docs/tutorials/otio-file-format-specification.md
ssteinbach: Makes sense.  Thanks!

ssteinbach:
lgtm, thanks @jminor!

reinecke (approved):
Looks good!

PR #867, opened 2021-01-16
Check rate and value are not invalid

### LINKED ISSUE #573: RationalTime::is_valid method
Small function to make sure that neither rate nor value are invalid.

hisergiorojas:

Fixes #573 

**Summarize your change.**

To check that rate and value are not invalid.


ssteinbach (approved):
Great, thanks.

PR #869, opened 2021-01-17
Kdenlive linear timewarp

splidje:
When writing Kdenlive, LinearTimewarp effects will be included. If a clip has multiple, their effects are concatenated into a resulting speed.


ssteinbach (approved):
Looks great, thanks.

PR #880, opened 2021-02-09
add a --version flag to otiopluginfo + otioconvert

ssteinbach:
Adds a `--version` flag to `otiopluginfo` and `otioconvert`.  This:

- prints out the `__version__` string 
- searches for `pkg_resource` installed plugins, and prints out their dist info (which includes their version numbers)

Question for the community:  After seeing this I kind of wish there was a way to also burn the git commit hash into the repo some how.  Not sure if thats useful, but it something we could potentially do when we burn the version numbers in.

Output looks like this:

```
> otioconvert --version
OpenTimelineIO version: 0.14.0.dev1
No pkg_resource plugins installed.
```

### DIFF
--- a/src/py-opentimelineio/opentimelineio/console/otiopluginfo.py
+++ b/src/py-opentimelineio/opentimelineio/console/otiopluginfo.py
@@ -30,6 +30,11 @@ import fnmatch
 import textwrap
 import opentimelineio as otio
 
+# on some python interpreters, pkg_resources is not available
+try:
+    import pkg_resources
+except ImportError:
+    pkg_resources = None
 
 OTIO_PLUGIN_TYPES = ['all'] + otio.plugins.manifest.OTIO_PLUGIN_TYPES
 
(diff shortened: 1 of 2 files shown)


src/py-opentimelineio/opentimelineio/console/otiopluginfo.py
```diff
@@ -185,6 +199,21 @@ def main():
...
+        if pkg_resources:
+            pkg_resource_plugins = list(
+                pkg_resources.iter_entry_points("opentimelineio.plugins")
+            )
+            if pkg_resource_plugins:
+                print("Plugins from pkg_resources:")
+                for plugin in pkg_resource_plugins:
+                    print("   {}".format(plugin.dist))
+            else:
+                print("No pkg_resource plugins installed.")
```
reinecke, line 215: Is the intent to not exit after printing version info and continue printing the usual info?

reinecke (commented):
Just the one behavior question to make sure there isn't a missing exit, otherwise looks good!

src/py-opentimelineio/opentimelineio/console/otiopluginfo.py
ssteinbach: Correct.  The thought was that this program is just dumping information about the installation anyway, so `--version` is an additive option and doesn't prevent printing everything else.  Happy to change that if you think its too much of a break from expected behavior though.

Issue #772, opened 2020-08-18
Remove Tox


ssteinbach:
Tox was useful initially, but now that we are doing our build  matrices on CI, we don't need it anymore.

Items:
- [x] move coverage into makefile
- [x] make the CI call the makefile coverage
- [x] build docs through makefile as well
- [x] `check_manifest` in `make test`

ssteinbach:
closed by #882

PR #882, opened 2021-02-09
Remove tox dependency

ssteinbach:
We've used `tox` for some time, and its served us well.  With the recent move to GitHub actions, however, we've found that it is no longer needed.  The reasons for removing it:

- one less build dependency
- the cmake/setup.py combination is already complicated, and tox is a further complication of that system
- Our makefile wasn't running `tox`, so this ensures that developers who use our developer makefile are matching _exactly_ what the ci does
- We'd like to do further things like run ci on more platforms and push wheels directly from ci into pypi/conda/etc -- this makes that more straightforward

What this PR does:

- moves the commands that `tox` was running into a makefile target, `ci-build-test`
- removes the setup.py and ci dependencies on `tox`
- switches the ci to use the new makefile target
- Removes references to and configuration for `tox` from various places (mostly the `setup.cfg`).

Todo:

- [x] move `lcov` execution to `Makefile`
- [x] remove `tox` section of `setup.cfg`
- [x] check block in `setup.py` to see if we can remove extra logic whose comment references tox


meshula:
A big positive is that we can now see the build logs, so we can find where the build is failing :)

ssteinbach:
Making sure that the RTD build isn't busted.  Lets wait for that to finish before landing this.

ssteinbach:
Those look good too, if this looks good to folks, I think this is ready to land:
https://opentimelineio-ssteinbach.readthedocs.io/en/remove_tox/

setup.py
```diff
@@ -156,30 +163,28 @@ def __init__(self, name):
...
+    # if platform.system() in "Linux":
+    user_options = setuptools.command.build_ext.build_ext.user_options + [
+        (
+            'cxx_coverage',
+            None,
+            'Enable code coverage for C++ code.  NOTE: you will likely want to'
+            ' also set the build_tmp directory to something that does not get '
+            'cleaned up.',
+        )
+    ]
```
ssteinbach, line 177: @meshula whoops, I think we meant to delete this code, right?

ssteinbach:
Closes #772

setup.py
meshula: yes, our decision was to leave it to logic in cmake

setup.py
```diff
@@ -156,30 +163,28 @@ def __init__(self, name):
...
         setuptools.command.build_ext.build_ext.initialize_options(self)
 
     def run(self):
-        # because tox passes all commandline arguments to _all_ things being
-        # installed by setup.py (including dependencies), environment variables
         _ctx.cxx_coverage = (
             self.cxx_coverage is not False
             or bool(os.environ.get("OTIO_CXX_COVERAGE_BUILD"))
-        ) and platform.system() == "Linux"
+        )
```
meshula, line 187: for example, here, we deleted this line in that expectation

meshula (commented):
There's one chunk to remove, as you asked. The CMake side of things already has logic to disable coverage if lcov is not available on the platform (e.g. msvc)

meshula (approved):

README.md
```diff
@@ -109,22 +109,22 @@ You can install development dependencies with `pip install .[dev]`
...
+request, they:
 
 - run `make test` -- to ensure that none of the unit tests were broken
 - run `make lint` -- to ensure that coding conventions conform to pep8
 - run `make coverage` -- to detect code which isn't covered
 
 PEP8: https://www.python.org/dev/peps/pep-0008/
 
-Additionaly, if you have CI failures regarding the file manifest, you can run:
+Additionaly, if to reproduce CI failures regarding the file manifest, run:
```
meshula, line 121: grammar, delete "if"

meshula (approved):

PR #886, opened 2021-02-16
Support --prefix correctly in new cmake system

### LINKED ISSUE #885: setup.py/pip --prefix argument not handled properly
### Build Problem

The `setup.py`/`cmake` system incorrectly handles the `pip` / `setup.py` argument `--prefix`.

## To Reproduce

Either run `setup.py install` with `--prefix` or `python -m pip opentimelineio --prefix`.  For example:

`rm -rf /var/tmp/test_otio/* ; python -m pip install path/to/opentimelineio --ignore-installed -vvv`

The C++ shared libraries and headers will be missing:
```sh
find /var/tmp/test_otio -name "_opentime*" | wc -l
       0
```

which should read:
```sh
find /var/tmp/test_otio -name "_opentime*" | wc -l
       1
```

Note that this is a regression -- the current release on pypi appears to have the correct behavior.

ssteinbach:
Closes #885 

Refactors how the setup.py communicates target locations to the cmake so that `--prefix` and `--user` arguments are correctly handled.

*POTENTIALLY BREAKING CHANGE*: Now if installing via `pip`/`setup.py`, the C++ libraries are installed under the python install in a `cxx-sdk` instead of `cxx-libs` directory.

I'd like to note that the way of finding the _value_ of the `--prefix` argument is pretty gross. If there is a better/cleaner way, please let me know!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -80,7 +80,7 @@ if(OTIO_PYTHON_INSTALL)
             # In order to not require setting $PYTHONPATH to point at the .so,
             # the shared libraries are installed into the python library 
             # location.
-            set(OTIO_RESOLVED_CXX_DYLIB_INSTALL_DIR "${OTIO_PYTHON_INSTALL_DIR}/opentimelineio")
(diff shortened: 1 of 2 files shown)


meshula (approved):
The thing we should watch for is that the CMake.config files point within cxx-libs on our various platforms. This is a good step forward in any case to handling pip's --prefix complication

PR #891, opened 2021-02-20
ImageSequenceReference: Added handling for empty target_url_base

### LINKED ISSUE #890: ImageSequenceReference: If target_url_base is empty string, target_url_for_image_number raises IndexError
### Incorrect Functionality

It is sometimes useful to generate a base file name for an image without the full URL path leading up to the file name. One would expect to be able to set `target_url_base` to an empty string (the default value) to achieve this. However it does not work as expected:

```python
import opentimelineio as otio

ref = otio.schema.ImageSequenceReference(
    name_prefix="myfilename.",
    name_suffix=".exr",
    start_frame=101,
    rate=24,
    frame_zero_padding=4,
    available_range=otio.opentime.TimeRange(
        otio.opentime.from_timecode("01:25:30:04", rate=24),
        duration=otio.opentime.from_frames(48, 24)
    ),
)

print(ref.target_url_for_image_number(0))
```

Results in:
```
IndexError: basic_string
```

If we set `target_url_base` to an string of any length greater than zero, the exception goes away:

```python
>>> ref.target_url_base = " "
>>> ref.target_url_for_image_number(0)
' /myfilename.0101.exr'
```

## Expected Behavior

I would expect to be able to set target_url_base to an empty string and have it generate a bare filename:

```python
>>> ref.target_url_base = ""
>>> ref.target_url_for_image_number(0)
'myfilename.0101.exr'
```

This has usefulness in cases where your application contains better information about the containing context for the file and wants to generate just the filename itself without a full path.

reinecke:
Fixes #890 

**Summarize your change.**

There was a check to see if `target_url_base` ended in a `/` to avoid having double-/ in constructed urls. This assumed `target_url_was of at lease length 1. Added a check to not attempt the check if the length is zero.

Added test checking generation of bare filenames using `target_url_for_image_number`.

### DIFF
--- a/tests/test_image_sequence_reference.py
+++ b/tests/test_image_sequence_reference.py
@@ -675,4 +675,6 @@ class ImageSequenceReferenceTests(
             ),
         )
 
-        assert ref.target_url_for_image_number(0) == 'myfilename.0101.exr'
+        self.assertEqual(
+            ref.target_url_for_image_number(0), "myfilename.0101.exr"
+        )
--- a/tests/test_image_sequence_reference.py
+++ b/tests/test_image_sequence_reference.py
@@ -661,3 +661,18 @@ class ImageSequenceReferenceTests(
             str(exception_manager.exception),
             "Zero rate sequence has no frames.",
         )
+
+    def test_target_url_for_image_number_with_blank_target_url_base(self):
+        ref = otio.schema.ImageSequenceReference(
+            name_prefix="myfilename.",
+            name_suffix=".exr",
+            start_frame=101,
+            rate=24,
+            frame_zero_padding=4,
+            available_range=otio.opentime.TimeRange(
+                otio.opentime.from_timecode("01:25:30:04", rate=24),
+                duration=otio.opentime.from_frames(48, 24)
+            ),
+        )
+
+        assert ref.target_url_for_image_number(0) == 'myfilename.0101.exr'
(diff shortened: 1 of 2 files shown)


tests/test_image_sequence_reference.py
```diff
@@ -661,3 +661,18 @@ def test_target_url_for_image_number_with_missing_timing_info(self):
...
+            start_frame=101,
+            rate=24,
+            frame_zero_padding=4,
+            available_range=otio.opentime.TimeRange(
+                otio.opentime.from_timecode("01:25:30:04", rate=24),
+                duration=otio.opentime.from_frames(48, 24)
+            ),
+        )
+
+        assert ref.target_url_for_image_number(0) == 'myfilename.0101.exr'
```
ssteinbach, line 678:
```suggestion
        self.assertEquals(ref.target_url_for_image_number(0), 'myfilename.0101.exr')
```

ssteinbach (changes requested):
I think we're using `unittest` style assertions, does this change make sense?

tests/test_image_sequence_reference.py
reinecke: 🤦 pytest muscle memory. Fixed.

ssteinbach (approved):
Looks great!  Thanks!

PR #892, opened 2021-02-20
Address big integers bug

### LINKED ISSUE #884: Large positive integers in OTIO JSON file misinterpreted as negative
### Incorrect Functionality

A positive integer value greater than 2**31 - 1 in an OTIO JSON file will be read incorrectly by `opentimelineio.adapters.read_from_file` and may be misrepresented as a negative value due to 32-bit integer overflow.

## To Reproduce

Run this python script:

```
import opentimelineio as otio
timeline = otio.adapters.read_from_file("test_input.otio")
otio.adapters.write_to_file(timeline, "test_output.otio")
```
using the input JSON OTIO file `test_input.otio` which is as follows:

```
{
    "OTIO_SCHEMA": "Timeline.1",
    "metadata": {},
    "name": "testdata",
    "global_start_time": null,
    "tracks": {
        "OTIO_SCHEMA": "Stack.1",
        "metadata": {},
        "name": "tracks",
        "source_range": null,
        "effects": [],
        "markers": [],
        "children": [
            {
                "OTIO_SCHEMA": "Track.1",
                "metadata": {},
                "name": "",
                "source_range": null,
                "effects": [],
                "markers": [],
                "children": [
                    {
                        "OTIO_SCHEMA": "Clip.1",
                        "metadata": {
                            "foobar": {
                                 "maxint32": 2147483647,
                                 "toobig": 2147483648,
                                 "verybig": 3450100000
                            }
                        },
                        "name": "black",
                        "source_range": {
                            "OTIO_SCHEMA": "TimeRange.1",
                            "duration": {
                                "rate": 24.0,
                                "value": 24.0
                            },
                            "start_time": {
                                "rate": 24.0,
                                "value": 1.0
                            }
                        },
                        "effects": [],
                        "markers": [],
                        "media_reference": {
                            "OTIO_SCHEMA": "ExternalReference.1",
                            "metadata": {},
                            "name": "",
                            "available_range": {
                                "OTIO_SCHEMA": "TimeRange.1",
                                "duration": {
                                    "rate": 24.0,
                                    "value": 240.0
                                },
                                "start_time": {
                                    "rate": 24.0,
                                    "value": 1.0
                                }
                            },
                            "target_url": "black.mov"
                        }
                    }
                ],
                "kind": "Video"
            }
        ]
    }
}
```
Then `diff test_input.otio test_output.otio` and you will see something like this:
```
<                                  "maxint32": 2147483647,
<                                  "toobig": 2147483648,
<                                  "verybig": 3450100000
---
>                                 "maxint32": 2147483647,
>                                 "toobig": -2147483648,
>                                 "verybig": -844867296
```
Note how the integer value of `toobig` turns from positive to negative, and the value of `verybig` is changed to a negative value that is completely different due to misrepresentation as a 32-bit integer.

Tested on:

* Linux ubuntu 20.04 with Python 3.8.5 (also reproduced on RHEL 7 with Python 2.7)
* OTIO version: commit 546b333fe6d4fd28b40c243957231a0f085c4ae4
* gcc version 9.3.0 (Ubuntu 9.3.0-17ubuntu1~20.04)

## Expected Behavior

The expected behavior is that the contents of the `test_output.otio` file will be equivalent to the `test_input.otio`, that is, identical except for insignificant differences such as whitespace changes and ordering of keys in objects.

## Additional Context

I think this is a result of line 41 of `src/opentimelineio/deserialization.cpp`:

```
    bool Uint(unsigned u) {  return store(any(int(u))); }
```
I have tested the following alternative code which seems to fix the bug:
```
    bool Uint(unsigned u) {
        if (u > INT32_MAX) {
            return store(any(int64_t(u)));
        }
        return store(any(int(u)));
    }
```

ssteinbach:
Closes #884 

- Asserts that the only integer data type that OTIO serializes/deserializes is `int64_t`
- ...which means that OTIO serializes into JSON:
  - `bool`
  - `int64_t`
  - string
  - IEEE754 `double` (including `NaN`, `Inf`, `-Infinity` support)
  - list
  - dictionary
  - things that derive from `SerializableObject`
  - `RationalTime`, `TimeTransform` and `TimeRange`
- An exception is raised if an invalid type or value is inserted into PyAny:
  - `TypeError: A value of type '<type 'set'>' is incompatible with OpenTimelineIO. OpenTimelineIO only supports the following value types in AnyDictionary containers (like the .metadata dictionary): ('int', 'float', 'str', 'bool', 'list', 'dictionary', 'opentime.RationalTime', 'opentime.TimeRange', 'opentime.TimeTransform').`
  - `ValueError: A value of 36893488147419103232 is outside of the range of integers that OpenTimelineIO supports, [-9223372036854775808, 9223372036854775807], which is the range of C++ int64_t.`
- Adds documentation about this behavior and a unit test that hopefully makes this behavior explicit
- Adds an explicit test for the C++ SDK to the python unit test suite (just one function, but its a start)

TODO:
- [x] make the exception that is raised when an invalid type is inserted more readable, and have it explain what the invalid types are
- [x] update the documentation to reflect where we settled on for this behavior


src/opentimelineio/imageSequenceReference.cpp
```diff
@@ -120,13 +120,22 @@ ImageSequenceReference::ImageSequenceReference(std::string const& target_url_bas
...
             reader.read("name_suffix", &_name_suffix) &&
-            reader.read("start_frame", &_start_frame) &&
-            reader.read("frame_step", &_frame_step) &&
+            reader.read("start_frame", &start_frame_value) &&
+            reader.read("frame_step", &frame_step_value) &&
             reader.read("rate", &_rate) &&
-            reader.read("frame_zero_padding", &_frame_zero_padding);
+            reader.read("frame_zero_padding", &frame_zero_padding_value);
+
+        _start_frame = static_cast<int>(start_frame_value);
```
reinecke, line 136: I suggest we just change the `start_frame` ivar to be `int64_t`. I think it was a mistake I didn't do that in the first place.

ssteinbach:
I'm hoping to not do that in this PR... I think that entails a bunch of other small changes to true the whole class up to `int64_t`, and can be done in a later PR

reinecke:
Ahh, gotcha. Sounds good.

src/opentimelineio/deserialization.cpp
```diff
@@ -37,10 +37,13 @@ class JSONDecoder : public OTIO_rapidjson::BaseReaderHandler<OTIO_rapidjson::UTF
...
-    bool Int(int i) {  return store(any(i)); }
-    bool Uint(unsigned u) {  return store(any(int(u))); }
-    bool Int64(int64_t i) { return store(any(i)); }
-    bool Uint64(uint64_t u) { return store(any(int64_t(u))); }
+
+    // coerce all integer types to int64_t
+    bool Int(int i) {  return store(any(static_cast<int64_t>(i))); }
+    bool Uint(unsigned u) {  return store(any(static_cast<int64_t>(u))); }
+    bool Int64(int64_t i) { return store(any(static_cast<int64_t>(i))); }
+    bool Uint64(uint64_t u) { return store(any(static_cast<int64_t>(u))); }
```
meshula, line 45: if we change this to 

```cpp
bool Uint64(uint64_t u) { return store(any(static_cast<int64_t>(u & 0x7FFFFFFFFFFFFFFF))); }
```
we will avoid undefined behavior for values greater than 2^63, by defining the behavior to be that we don't allow UB overflow states.. This doesn't matter now in a practical sense, but it will show up when we start fuzz testing.

tests/test_timeline.py
```diff
@@ -47,6 +51,12 @@ def test_metadata(self):
         self.assertIsOTIOEquivalentTo(tl, decoded)
         self.assertEqual(tl.metadata, decoded.metadata)
 
+    def test_big_integers(self):
+        result = otio.adapters.read_from_file(BIG_INT_TEST)
+
+        self.assertTrue(result.tracks[0][0].metadata["foobar"]["toobig"] > 0)
+        self.assertTrue(result.tracks[0][0].metadata["foobar"]["verybig"] > 0)
+
```
meshula, line 59: here, we could add a test for setting an uint64_t to (0x7FFFFFFFFFFFFFFF + N) and verifying that the value is (N - 1)

meshula (changes requested):

tests/test_timeline.py
jminor: It seems important to check the exact value that we get, not just `> 0`.
Also, can we check that large negative numbers come through as expected?

src/opentimelineio/safely_typed_any.cpp
```diff
@@ -59,6 +63,10 @@ int64_t safely_cast_int64_any(any const& a) {
     return any_cast<int64_t>(a);
 }
 
+uint64_t safely_cast_uint64_any(any const& a) {
+    return any_cast<uint64_t>(a);
+}
+
```
ssteinbach, line 69: @meshula I left the uint64_t types in the Any code.  Do you think it would be better to remove them entirely?  (It was missing this specific cast)

meshula: I think it's harmless to have them as this won't affect things at the serialization/deserialization level.

meshula (approved):

reinecke (approved):
This is a great clarification!

PR #894, opened 2021-02-23
Enable xges unit tests on all platform/python versions

thiblahute:
After https://github.com/PixarAnimationStudios/OpenTimelineIO/pull/873 was merged, this enables xges adapter tests on the same platforms/python version as other adapters.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -57,8 +57,11 @@ INSTALL_REQUIRES = [
 ]
 # python2 dependencies
 if sys.version_info[0] < 3:
-    INSTALL_REQUIRES.append(
-        "backports.tempfile",
+    INSTALL_REQUIRES.extend(
+        [
+            "backports.tempfile",
+            'future',  # enables the builtins module
+        ]
     )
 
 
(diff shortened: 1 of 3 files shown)


thiblahute:
cc @ssteinbach

setup.py
```diff
@@ -54,6 +54,7 @@ class _Ctx(object):
 
 INSTALL_REQUIRES = [
     'pyaaf2==1.4.0',
+    'future',
```
ssteinbach, line 57: Is this to enable the `builtins` module only in python2?  If so note that on line 59 there are some python2 specific dependencies, this could go there.

ssteinbach (commented):
Nice, just one question about the import. Thanks for the quick turnaround time!

setup.py
thiblahute: I think it is needed for both python2 and python3.

thiblahute: And sorry to not answer last week, I was away. Also I think it was simpler to do it the way we did it in the end :-)

ssteinbach: Looks like its in python3: https://docs.python.org/3/library/builtins.html

ssteinbach: Could you add a comment saying "# enables the builtins module on python2"

ssteinbach (changes requested):
My one request is that this gets a comment in the `setup.py` noting that this is for the builtins module specifically used by the XGES adapter.  Otherwise looks great, thanks for the quick turnaround time!

thiblahute:
Thanks for the review!

Issue #895, opened 2021-02-25
Transition duration incorrect in CMX 3600


JoshBurnell:
Hey, guys,

The [cmx_3600.py](https://github.com/PixarAnimationStudios/OpenTimelineIO/blob/master/src/py-opentimelineio/opentimelineio/adapters/cmx_3600.py) adapter appears to use a duration for Transitions that doesn't conform to the CMX 3600 spec.

According to [this spec](https://xmil.biz/EDL-X/CMX3600.pdf) (Page 30), the duration is stored in a "RATE" field.

[This guide](http://www.edlmax.com/EdlMaxHelp/Edl/maxguide.html) makes it a little easier to read:
```
75 frame dissolve from reel 123 to reel 123B on video and A1:

Edit # Reel Name  Channel Trans      Dur      Source IN          Source OUT          Record IN         Record OUT
105 123    B     C        03:05:57:17 03:05:57:17 01:00:21:20 01:00:21:20
105 123B   B     D    075 03:15:33:09 03:15:35:24 01:00:21:20 01:00:24:05
```

However, the `cmx_3600.py` adapter appears to use the length of the previous clip:
```
 668             transition_duration = clip.duration()
```
https://github.com/PixarAnimationStudios/OpenTimelineIO/blob/84c8245c070b12d7e627c145435a5db9add0d2de/src/py-opentimelineio/opentimelineio/adapters/cmx_3600.py#L668

Test EDL:
[cmx_3600_transition.edl.zip](https://github.com/PixarAnimationStudios/OpenTimelineIO/files/6039542/cmx_3600_transition.edl.zip)

Recreation Steps:
```
In [13]: cat cmx_3600_transition.edl
TITLE:   0800_Cut_v3_ToLayout_021121_EDL 
FCM: NON-DROP FRAME
001  ABC0100. V     C        01:00:12:15 01:00:17:01 01:00:48:20 01:00:53:06 
002  ABC0200. V     C        01:00:11:07 01:00:14:09 01:00:53:06 01:00:56:08 
003  ABC0200. V     C        01:00:14:09 01:00:14:09 01:00:56:08 01:00:56:08 
003  ABC0300. V     D    026 01:00:10:08 01:00:13:09 01:00:56:08 01:00:59:09 
* BLEND, DISSOLVE 
004  ABC0400. V     C        01:00:11:09 01:00:17:20 01:00:59:09 01:01:05:20

In [14]: import opentimelineio as otio

[...]

PR #896, opened 2021-02-25
Change cmx_3600 adapter to use transition duration frames (Issue #895)

JoshBurnell:
The [cmx_3600.py](https://github.com/PixarAnimationStudios/OpenTimelineIO/blob/master/src/py-opentimelineio/opentimelineio/adapters/cmx_3600.py) adapter appears to use a duration for Transitions that doesn't conform to the CMX 3600 spec.

According to [this spec](https://xmil.biz/EDL-X/CMX3600.pdf) (Page 30), the duration is stored in a "RATE" field.

[This guide](http://www.edlmax.com/EdlMaxHelp/Edl/maxguide.html) makes it a little easier to read:
```
75 frame dissolve from reel 123 to reel 123B on video and A1:

Edit # Reel Name  Channel Trans      Dur      Source IN          Source OUT          Record IN         Record OUT
105 123    B     C        03:05:57:17 03:05:57:17 01:00:21:20 01:00:21:20
105 123B   B     D    075 03:15:33:09 03:15:35:24 01:00:21:20 01:00:24:05
```

However, the `cmx_3600.py` adapter appears to use the length of the previous clip:
https://github.com/PixarAnimationStudios/OpenTimelineIO/blob/84c8245c070b12d7e627c145435a5db9add0d2de/src/py-opentimelineio/opentimelineio/adapters/cmx_3600.py#L668

Test:
```
➜  OpenTimelineIO-JoshBurnell git:(issue-895) python3 tests/test_cmx_3600_adapter.py    
..............................
Ran 30 tests in 0.285s

OK
➜  OpenTimelineIO-JoshBurnell git:(issue-895) python3 tests/test_cmx_3600_adapter.py EDLAdapterTest.test_transition_duration
.
Ran 1 test in 0.092s

OK
```


ssteinbach:
First of all, kudos for citing a document from 1988!  Is there some way to get that as a citation/reference with a comment above the block where you set the transition duration?

JoshBurnell:
lol If you look closely, I bet you can see the perforated edges of the dot matrix paper that that document was printed on. :-P

Great idea. I couldn't find a good reference link directly to something that said "Duration works like so..." but I'll add something.

ssteinbach:
Do you know what the legal status of that document is?  Is that something that we or ASWF could host?  @jmertic do you know if there is a place for documents like that?

jmertic:
Good question! Do you know the org that owns this spec? Or who used to :-)?

mikekoetter:
**TIL** the CMX from cmx 3600 comes from CMX Editing Systems [_(a member of the Chyron group)_](https://en.wikipedia.org/wiki/ChyronHego)
and the 3600 was a model of a "Highly popular online videotape editor" :)

jminor:
SMPTE ST 258:2004 is the most official spec I know of for EDL
https://ieeexplore.ieee.org/document/7291839

The proper citation format is something like: "ST 258:2004 - SMPTE Standard - For Television — Transfer of Edit Decision Lists," in ST 258:2004 , vol., no., pp.1-37, 6 April 2004, doi: 10.5594/SMPTE.ST258.2004.

jminor:
and there are mentions/links to two of the docs mentioned above here: https://opentimelineio.readthedocs.io/en/latest/tutorials/adapters.html

ssteinbach:
Ok great, so in that case SMPTE is the organization that hosts the document and we should link to it with the citation that @jminor notes.  Thanks everyone!

jmertic:
Citation makes complete sense here :-)

One other thing to think about is if within the project there is an implementation of the spec; if so you likely would want to review the copyright/IP policy of the standards owner to ensure your use is allowed in that context. You probably are fine, but just mentioning it :-)

JoshBurnell:
I updated the Citation per Josh's suggestion.

tests/test_cmx_3600_adapter.py
```diff
@@ -1038,6 +1039,14 @@ def test_speed_effects(self):
             )
         )
 
+    def test_transition_duration(self):
+        tl = otio.adapters.read_from_file(TRANSITION_DURATION_TEST)
+        self.assertEqual(len(tl.tracks[0]), 5)
+
+        self.assertTrue(isinstance(tl.tracks[0][2], otio.schema.Transition))
```
ssteinbach, line 1046:
```suggestion
        self.assertIsInstance(tl.tracks[0][2], otio.schema.Transition)
```

ssteinbach (commented):
Looks great -- one more last thing, if you accept the change, we can use `assertIsInstance` instead of `assertTrue(isinstance(...))`

PR #911, opened 2021-03-11
Avoid exporting build tree libraries

### LINKED ISSUE #910: Unable to link with downstream project libraries that are not exported
### Build Problem

If trying to link with a downstream project that does not export its libraries (Imath in this example) this will occur when building:

```
-- OTIO Defaulting C++ install to 
-- Building shared libs
-- Installing C++ bindings to: /opentimelineio/cxx-sdk
-- Installing C++ dynamic libraries to: /opentimelineio
--   Installing 'any' and 'nonstd' for C++ (OTIO_DEPENDENCIES_INSTALL=ON)
-- Installing Python bindings to: 
-- Building C++ with Coverage: OFF
-- Checking git repo is available:
true
-- root: Updating git submodules to make sure they are up to date
-- pybind11 v2.3.dev0
-- Configure Imath Version: 3.0.0 Lib API: 26.0.0
-- Imath is configuring as a cmake sub project
-- Configuring done
CMake Error in src/opentimelineio/CMakeLists.txt:
  export called with target "opentimelineio" which requires target "Imath"
  that is not in any export set.
```

## Additional Context

This occurs because opentime and opentimelineio call export in their CMakeFiles like this:
```
    export(TARGETS opentimelineio
           NAMESPACE OTIO::
           FILE "${CMAKE_CURRENT_BINARY_DIR}/OpenTimelineIOConfig.cmake")
```

When exporting, all the dependencies must also be exported.  Meaning that opentimelineio and opentimeio will not be able to link with any downstream projects that do not also export.

From the CMake docs:
> Creates a file <filename> that may be included by outside projects to import targets from the current project's build tree. This is useful during cross-compiling to build utility executables that can run on the host platform in one project and then import them into another project being compiled for the target platform. If the NAMESPACE option is given the <namespace> string will be prepended to all target names written to the file.

Target installations are associated with the export <export-name> using the EXPORT option of the install(TARGETS) command.

It's not likely that any upstream project would ever need to reference the libraries directly from the build tree.  In fact, its likely undesirable since this is most often a temporary location.  Upstream projects would instead want to access the installed location created by the `install( EXPORT )` command, which is also present in the CMakeFiles already.

This seems to have been added to support a local pip install, so any solution should make sure that it is still possible to call `pip install .`.

rogernelson:
```Fixes #910 ```

**Summarize your change.**

This branch removes the CMake `export` of the opentime and opentimelineio libraries.  In CMake, exporting generates a CMakeFile that makes the target available from the build tree to upstream projects.  An example use case is when there's a tool facilitating the build process that might also be needed by the upstream project.  However, this is not desirable from a library perspective, it would allow upstream projects to reference the library in the build folder (which is temporary) instead of the installed location (which is permanent).  See Issue for more details.

This is needed to properly link with libraries from other downstream projects, since an export will fail if its dependencies are not also exported.  Specifically, we would like to link with the Imath library for an upcoming PR, but this fails because Imath does not export its libraries.

From the git history this was originally done to facilitate a local pip install.  I tried doing this via `pip install .` and it works, so I assume its since been fixed by other CMake changes.

A clean build on all platforms should be sufficient.

### DIFF
--- a/src/opentime/CMakeLists.txt
+++ b/src/opentime/CMakeLists.txt
@@ -47,10 +47,6 @@ if(OTIO_CXX_INSTALL)
             LIBRARY DESTINATION "${OTIO_RESOLVED_CXX_DYLIB_INSTALL_DIR}"
             RUNTIME DESTINATION "${OTIO_RESOLVED_CXX_DYLIB_INSTALL_DIR}")
 
-    export(TARGETS opentime
-           NAMESPACE OTIO::
-           FILE "${CMAKE_CURRENT_BINARY_DIR}/OpenTimeConfig.cmake")
-
     install(EXPORT OpenTimeConfig
             DESTINATION "${OTIO_RESOLVED_CXX_INSTALL_DIR}"
             NAMESPACE OTIO:: )
--- a/src/opentimelineio/CMakeLists.txt
+++ b/src/opentimelineio/CMakeLists.txt
@@ -113,10 +113,6 @@ if(OTIO_CXX_INSTALL)
            LIBRARY DESTINATION "${OTIO_RESOLVED_CXX_DYLIB_INSTALL_DIR}"
            RUNTIME DESTINATION "${OTIO_RESOLVED_CXX_DYLIB_INSTALL_DIR}")
 
-    export(TARGETS opentimelineio
-           NAMESPACE OTIO::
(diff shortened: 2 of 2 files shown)


meshula:
It passed on all platforms, thanks for the investigation!

meshula (approved):

ssteinbach (approved):

PR #918, opened 2021-03-13
Cxx examples

### LINKED ISSUE #917: Add C++ examples
It would be nice to have C++ examples demonstrating use of the OTIO API similar to the Python examples. Ideally the example code would be as similar as possible between C++ and Python so that new developers could see how the API is used in each language.

This could also address these issues:
https://github.com/PixarAnimationStudios/OpenTimelineIO/issues/796
https://github.com/PixarAnimationStudios/OpenTimelineIO/issues/171

I've done some initial work on this and converted the example "summarize_timing.py" to C++; I'll open a PR for the changes, though it is still a WIP...

darbyjohnston:
```Fixes #917```

This is WIP code to convert the Python examples to C++. I've successfully converted "summarize_timing.py" and started converting "flatten_video_tracks.py". With these changes, and the CMake variable OTIO_CXX_EXAMPLE set to "ON", you can run:

> ./examples/summarize_timing ../tests/sample_data/premiere_example.xml
> ./examples/flatten_video_tracks ../tests/sample_data/premiere_example.xml tmp.xml

Converting the examples brought up a couple of questions:
* It looks like the Python API has diverged a bit from C++? For example what are the C++ equivalents of each_clip() and deepcopy()?
* How should a user of the C++ API handle memory management? The "C++ Implementation Details" documentation mentions the use of possibly_delete() and Retainers?
* Can a user of the C++ API utilize the Python adapters for file I/O? I've implemented a utility class that provides this, and it seems to work OK initially, but is it the right approach? If so maybe the code should be moved into the OTIO library for general usage?

Note this code currently only works for Linux/macOS, there is some additional code that needs to be added for Windows.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -34,6 +34,7 @@ set(OTIO_PYTHON_INSTALL_DIR "" CACHE STRING "Python installation dir (such as th
 # Build options
 option(OTIO_SHARED_LIBS          "Build shared if ON, static if OFF" ON)
 option(OTIO_CXX_COVERAGE         "Invoke code coverage if lcov/gcov is available" OFF)
+option(OTIO_CXX_EXAMPLES         "Build CXX examples (also requires OTIO_PYTHON_INSTALL=ON)" OFF)
 option(OTIO_AUTOMATIC_SUBMODULES "Fetch submodules automatically" ON)
 
 #------------------------------------------------------------------------------
@@ -208,7 +209,7 @@ if(OTIO_PYTHON_INSTALL)
     add_subdirectory(src/py-opentimelineio)
 endif()
 
-if(OTIO_BUILD_CXX_EXAMPLES AND OTIO_PYTHON_INSTALL)
+if(OTIO_CXX_EXAMPLES AND OTIO_PYTHON_INSTALL)
     add_subdirectory(examples)
 endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -210,7 +210,7 @@ if(OTIO_PYTHON_INSTALL)
     add_subdirectory(src/py-opentimelineio)
 endif()
 
-if(OTIO_CXX_EXAMPLES AND OTIO_PYTHON_INSTALL)
+if(OTIO_CXX_EXAMPLES)
     add_subdirectory(examples)
 endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -130,7 +130,6 @@ endif()
 #------------------------------------------------------------------------------
 # Global language settings
 
-
 if (NOT CMAKE_CXX_STANDARD)
     set(CMAKE_CXX_STANDARD 11)
 endif()
@@ -153,6 +152,8 @@ if(WIN32)
     # to the Microsoft debug CRT.
     set(OTIO_DEBUG_POSTFIX "d")
 endif()
+
+set_property(GLOBAL PROPERTY USE_FOLDERS ON)
 
 #------------------------------------------------------------------------------
 # Fetch or refresh submodules if requested
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -207,3 +207,8 @@ add_subdirectory(src/opentimelineio)
 if(OTIO_PYTHON_INSTALL)
     add_subdirectory(src/py-opentimelineio)
 endif()
+
+if(OTIO_BUILD_CXX_EXAMPLES AND OTIO_PYTHON_INSTALL)
+    add_subdirectory(examples)
+endif()
+
--- a/examples/flatten_video_tracks.cpp
+++ b/examples/flatten_video_tracks.cpp
@@ -17,9 +17,8 @@ int main(int argc, char** argv)
     }
     
     // Read the file
-    PythonAdapters adapters;
     otio::ErrorStatus error_status;
-    auto timeline = adapters.read_from_file(argv[1], &error_status);
+    otio::SerializableObject::Retainer<otio::Timeline> timeline(dynamic_cast<otio::Timeline*>(otio::Timeline::from_json_file(argv[1], &error_status)));
     if (!timeline)
     {
         print_error(error_status);
@@ -63,7 +62,7 @@ int main(int argc, char** argv)
     // ...and save it to disk.
     std::cout << "Saving " << newtimeline.value->video_tracks().size() << " video tracks and " <<
         newtimeline.value->audio_tracks().size() << " audio tracks." << std::endl;
-    if (!adapters.write_to_file(newtimeline, argv[2], &error_status))
+    if (!timeline.value->to_json_file(argv[2], &error_status))
     {
         print_error(error_status);
         return 1;            
@@ -71,4 +70,3 @@ int main(int argc, char** argv)
 
     return 0;
 }
-
--- a/examples/flatten_video_tracks.cpp
+++ b/examples/flatten_video_tracks.cpp
@@ -1,0 +1,74 @@
+#include "util.h"
+
+#include <opentimelineio/stackAlgorithm.h>
+#include <opentimelineio/timeline.h>
+
+#include <iostream>
+#include <sstream>
+
+namespace otio = opentimelineio::OPENTIMELINEIO_VERSION;
+
+int main(int argc, char** argv)
+{
+    if (argc != 3)
+    {
+        std::cout << "Usage: flatten_video_tracks (inputpath) (outputpath)" << std::endl;
+        return 1;
+    }
+    
+    // Read the file
+    PythonAdapters adapters;
+    otio::ErrorStatus error_status;
+    auto timeline = adapters.read_from_file(argv[1], &error_status);
+    if (!timeline)
+    {
+        print_error(error_status);
+        return 1;
+    }
+    auto video_tracks = timeline->video_tracks();
+    auto audio_tracks = timeline->audio_tracks();
+    
+    std::cout << "Read " << video_tracks.size() << " video tracks and " <<
+        audio_tracks.size() << " audio tracks." << std::endl;
+
+    // Take just the video tracks - and flatten them into one.
+    // This will trim away any overlapping segments, collapsing everything
+    // into a single track.
+    std::cout << "Flattening " << video_tracks.size() << " video tracks into one..." << std::endl;
+    auto onetrack = otio::flatten_stack(video_tracks, &error_status);
+    if (!onetrack)
+    {
+        print_error(error_status);
+        return 1;
+    }
+
+    // Now make a new empty Timeline and put that one Track into it
+    std::string name;
+    std::stringstream ss(name);
+    ss << timeline->name() << " Flattened";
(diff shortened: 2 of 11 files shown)

### CODE REVIEW

**Comment by darbyjohnston** [comment]
I've updated the code to use Retainers and added the missing code for Windows.

File: examples/summarize_timing.cpp [hunk 1]
```diff
@@ -0,0 +1,131 @@
...
+#include <iostream>
+
+namespace otio = opentimelineio::OPENTIMELINEIO_VERSION;
+namespace otime = opentime::OPENTIME_VERSION;
+
+void _summarize_effects(otio::SerializableObject::Retainer<otio::Item> const& item)
+{
+    for (auto effect : item.value->effects())
+    {
+        if (auto freezeFrame = dynamic_cast<otio::FreezeFrame*>(effect.value))
```
meshula, line 24: per your notes elsewhere, this definitely begs a dynamic_retained_cast.

examples/flatten_video_tracks.cpp
```diff
@@ -0,0 +1,74 @@
...
+int main(int argc, char** argv)
+{
+    if (argc != 3)
+    {
+        std::cout << "Usage: flatten_video_tracks (inputpath) (outputpath)" << std::endl;
+        return 1;
+    }
+    
+    // Read the file
+    PythonAdapters adapters;
```
meshula, line 20: This is a cool thing, but I'd suggest breaking out a separated "embedding a python interpreter" example, rather than using it as a utility in cpp in general. I can imagine people starting to proliferate python embeddings via copy-pasta, not understanding whether that's a requirement of OTIO or not. Reading and writing a JSON file via C++ is supported out of the box, so in terms of being instructive, the latter is the approach I'd recommend.

examples/summarize_timing.cpp
```diff
@@ -0,0 +1,131 @@
...
+                "(" << typeid(effect.value).name() << ")" << std::endl;
+        }
+    }
+}
+
+void _summarize_range(std::string const& label, otio::TimeRange const& time_range, otio::ErrorStatus const& errorStatus)
+{
+    if (errorStatus.outcome != otio::ErrorStatus::Outcome::OK)
+    {
+        std::cout << '\t' << label << ": None" << std::endl;
```
meshula, line 46: std::cerr would be appropriate here (and wherever errors are reported)

examples/summarize_timing.cpp
```diff
@@ -0,0 +1,131 @@
...
+                    _summarize_effects(item);
+                }
+            }
+        }
+    }
+}
+
+int main(int argc, char** argv)
+{
+    PythonAdapters adapters;
```
meshula, line 116: same comment, on using native facilities here

meshula (commented):
This looks very helpful, I've dropped a few notes inline.

meshula:
Might be an idea to mark this PR as "WIP" until it's ready to land? That's the convention we're using in this repo, in general, when things are being iterated and thought through.

ssteinbach:
Rather than the `wip` tag, marking it as a draft PR is even better.  Then we get a nice notification when its no longer draft.

examples/flatten_video_tracks.cpp
jminor: In the past we've encouraged people to execute `otioconvert` or `otiocat` as sub-processes to avoid the subject of embedding Python in their application. Example programs that show how to accomplish either or both of these would be very helpful. Then folks can pick whichever suits their use case.

jminor: Also, using `otiocat` (or equivalent code via embedded Python) would avoid the need for a temp file when reading. You could probably avoid the temp file when writing also.

examples/summarize_timing.cpp
darbyjohnston: Maybe something like:
```
template <class T, class U>
otio::SerializableObject::Retainer<T> dynamic_retainer_cast(otio::SerializableObject::Retainer<U> const& retainer)
{
    return dynamic_cast<T*>(retainer.value);
}
```

examples/flatten_video_tracks.cpp
darbyjohnston: Sure thing, I'll make a dedicated example like you suggest and the other examples will just work with JSON files.

darbyjohnston: Thanks Josh, I'll checkout otioconvert/otiocat, I wasn't thrilled with the temp file approach.

darbyjohnston:
Any thoughts about the Python functions each_clip() and deepcopy() for C++?

examples/summarize_timing.cpp
meshula: yes, exactly that ~ I think I'm leaning to sending this over to OTIO as a PR. It's in some sense trivial, but it would be nice to do it the same everywhere

meshula:
each_clip() and deepcopy() both seem handy for c++ use. Maybe a quick sketch and discussion if they should be a topic for a PR for OTIO? (Since deepcopy() is a Pythonism, I'd suggest deep_copy for C++.)

examples/summarize_timing.cpp
darbyjohnston: It might also be nice to add the "->" operator to Retainer so things like this:
`timeline.value->tracks()`

Could be simplified to:
`timeline->tracks()`

meshula: I like it

darbyjohnston:
I've broken out the C++ code for using Python adapters into two separate examples, "python_adapters_child_process.cpp" and "python_adapters_embed.cpp", for using the adapters via child processes and an embedded Python interpreter respectively.

I think I prefer the embedding approach, that way there is less environment setup to worry about (making sure otioconvert is in the search path), and there is less platform specific code (child process creation and pipes seem more complicated on Windows).

I still need to take a look and see if I can get rid of the temporary files, and the flatten_video_tracks.cpp example is missing a C++ version of deepcopy().

@meshula Have you used CreateProcess() on Windows before? The MSDN documentation says to pass "cmd.exe" in to execute bat files, but I got the error "The system cannot find the file specified.". If I use "C:\windows\system32\cmd.exe" it works OK, but it seems like a bad idea to hard code the path...

examples/summarize_timing.cpp
darbyjohnston: @meshula Just double-checking, did you want me to open a PR for the dynamic_retainer_cast()? If so do you think the other variants like static_retainer_cast(), etc., are useful as well?

meshula:
The embedding approach is valid, but it's a whole other can of worms. Not everyone will want to bind Python into their application, because there's so much other environment to set up to make that fully functional. Embedded IS interesting, but so is calling out. 

CreateProcess with a full path to system32\cmd.exe is actually fine; cmd has been there forever, and is unlikely to move in the foreseeable future. It's like the assumption that we make in expecting to be able to find sh at /bin/sh until the end of time.

darbyjohnston:
OK, I've changed the example "python_adapters_embed.cpp" to use Python directly without needing temporary files. I did this by converting the timeline into JSON and passing that to/from Python where the adapters are used for file IO. It's a bit more code than before, but it's not too bad and doesn't require any platform specific code.

I also looked into deepcopy() and apparently it's just a wrapper around SerializableObject::clone(), so maybe it's not needed for C++. I updated the example "flatten_video_tracks.cpp" to use clone() and it should be complete now.

I think that addresses all of the feedback, should I "un-draft" this PR? Or should this also include C++ versions of the rest of the Python examples?

CMakeLists.txt
```diff
@@ -153,6 +153,8 @@ if(WIN32)
     set(OTIO_DEBUG_POSTFIX "d")
 endif()
 
+set_property(GLOBAL PROPERTY USE_FOLDERS ON)
```
ssteinbach, line 156: just googled what this does, neat.

CMakeLists.txt
```diff
@@ -207,3 +209,8 @@ add_subdirectory(src/opentimelineio)
 if(OTIO_PYTHON_INSTALL)
     add_subdirectory(src/py-opentimelineio)
 endif()
+
+if(OTIO_CXX_EXAMPLES AND OTIO_PYTHON_INSTALL)
+    add_subdirectory(examples)
+endif()
```
ssteinbach, line 215: Should this write an error if `OTIO_CXX_EXAMPLES` is on but `OTIO_PYTHON_INSTALL` is off?  Or a message of some kind?

examples/summarize_timing.cpp
ssteinbach: Just so I understand, that would make this code:
```c++
if (auto freezeFrame = otio::dynamic_retainer_cast<otio::FreezeFrame*>(effect)) {
```

ssteinbach (commented):
Thanks for putting this together, @darbyjohnston.  Both `deepcopy` and `each_clip` are based on pythonisms, and yeah -- deep copy() is just exposing the `clone` functionality.  I wonder if `each_clip` would become a C++11 style iterator?  Or are those just not good?

meshula:
The interface for a C++ style iterator is not hard to satisfy; here's a nice overview. https://internalpointers.com/post/writing-custom-iterators-modern-cpp

That gives you compatibility with lots of STL features like std::copy, so iterators would be a definite nice-to-have.

examples/summarize_timing.cpp
darbyjohnston: Actually even a little bit simpler since the "*" isn't necessary:
```
if (auto freezeFrame = otio::dynamic_retainer_cast<otio::FreezeFrame>(effect)) {
```

CMakeLists.txt
darbyjohnston: It definitely helps keep things neat in IDEs like Visual Studio, though one downside is you need to manually specify the folders for each target:
`set_target_properties(${example} PROPERTIES FOLDER examples)`

CMakeLists.txt
darbyjohnston: Actually now that the C++/Python code is in separate examples, maybe this should be changed to just:
```
if(OTIO_CXX_EXAMPLES)
    add_subdirectory(examples)
endif()
```
And then the examples/CMakeLists.txt could optionally add the C++/Python code when Python is enabled:
```
if(OTIO_PYTHON_INSTALL)
    list(APPEND examples python_adapters_child_process)
    list(APPEND examples python_adapters_embed)
endif()
```
That way the pure C++ examples can be built without Python.

ssteinbach: Ah yeah -- lets do that.  That sounds simpler and clearer.  Thanks!

examples/summarize_timing.cpp
darbyjohnston: I just created a draft PR that add this:
https://github.com/PixarAnimationStudios/OpenTimelineIO/pull/923

darbyjohnston:
Un-drafting this since it seems like most of the feedback has been resolved and the other work for Retainers and each_clip() should probably be separate PRs...

ssteinbach (approved):
Looks good to me!  Thanks @darbyjohnston!

meshula (approved):

PR #923, opened 2021-03-23
Cxx conveniences

### LINKED ISSUE #922: Add C++ conveniences
It would be nice to add a couple of convenience functions to streamline the C++ code that works with the SerializableObject::Retainer type, specifically functions for casting and operator overloads that makes the type feel more like a regular pointer.

Old:
```
if (auto freezeFrame = dynamic_cast<otio::FreezeFrame*>(effect.value)) {
```

New:
```
if (auto freezeFrame = otio::dynamic_retainer_cast<otio::FreezeFrame>(effect)) {
```

Old:
```
if (!_media_reference.value->available_range()) {
```

New:
```
if (!_media_reference->available_range()) {
```

darbyjohnston:
```Fixes #922```

Add convenience functions and operator overloads for working with SerializableObject::Retainer types.

### CODE REVIEW

**Review by ssteinbach — COMMENTED** [review]
Looks good, just had a couple questions

File: src/opentimelineio/clip.cpp [hunk 1]
```diff
@@ -41,13 +41,13 @@ TimeRange Clip::available_range(ErrorStatus* error_status) const {
...
     
-    if (!_media_reference.value->available_range()) {
+    if (!_media_reference->available_range()) {
         *error_status = ErrorStatus(ErrorStatus::CANNOT_COMPUTE_AVAILABLE_RANGE,
                                     "No available_range set on media reference on clip", this);
         return TimeRange();
     }
     
-    return *_media_reference.value->available_range();
+    return *_media_reference->available_range();
```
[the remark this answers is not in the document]
darbyjohnston: Another option would be to use "value()" instead of the "*" operator here to get the value from the optional, maybe even more readable?
```
_media_reference->available_range().value()
```

src/opentimelineio/composition.cpp
```diff
@@ -161,7 +161,7 @@ Composition::handles_of_child(Composable const* /* child */, ErrorStatus* /* err
 
 int Composition::_index_of_child(Composable const* child, ErrorStatus* error_status) const {
     for (size_t i = 0; i < _children.size(); i++) {
-        if (_children[i].value == child) {
+        if (_children[i] == child) {
```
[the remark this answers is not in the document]
darbyjohnston: In this case the compiler is using the cast operator on the Retainer to automatically convert it to the pointer type, this change actually could have been made without the operator I added. For example you can also write:
```
Composable* c = _children[i];
```

src/opentimelineio/timeline.cpp
```diff
@@ -27,8 +27,8 @@ void Timeline::write_to(Writer& writer) const {
 
 std::vector<Track*> Timeline::video_tracks() const {
     std::vector<Track*> result;
-    for (auto c: _tracks.value->children()) {
-        if (Track* t = dynamic_cast<Track*>(c.value)) {
+    for (auto c: _tracks->children()) {
+        if (auto t = dynamic_retainer_cast<Track>(c)) {
```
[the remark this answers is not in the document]
darbyjohnston: This would be a `Retainer<Track>`, a `Retainer` class templated with the `Track` type.

I'm a big fan of the `auto` keyword but it can be a bit mysterious sometimes, especially with references. I think I've read some places have coding policies against using it, but it is so convenient, especially with the STL classes. For example this:
```
auto c = _tracks->children().begin();
```
Vs:
```
std::vector<SerializableObject::Retainer<Composable>>::const_iterator c = _tracks->children().begin();
```
And if you make a typo in the above, this is the handy error message you get :)
```
Severity	Code	Description	Project	File	Line	Suppression State
Error (active)	E0312	no suitable user-defined conversion from "std::_Vector_const_iterator<std::_Vector_val<std::conditional_t<true, std::_Simple_types<opentimelineio::v1_0::SerializableObject::Retainer<opentimelineio::v1_0::Composable>>, std::_Vec_iter_types<opentimelineio::v1_0::SerializableObject::Retainer<opentimelineio::v1_0::Composable>, size_t, ptrdiff_t, opentimelineio::v1_0::SerializableObject::Retainer<opentimelineio::v1_0::Composable> *, const opentimelineio::v1_0::SerializableObject::Retainer<opentimelineio::v1_0::Composable> *, opentimelineio::v1_0::SerializableObject::Retainer<opentimelineio::v1_0::Composable> &, const opentimelineio::v1_0::SerializableObject::Retainer<opentimelineio::v1_0::Composable> &>>>>" to "std::_Vector_iterator<std::_Vector_val<std::conditional_t<true, std::_Simple_types<opentimelineio::v1_0::SerializableObject::Retainer<opentimelineio::v1_0::Composable>>, std::_Vec_iter_types<opentimelineio::v1_0::SerializableObject::Retainer<opentimelineio::v1_0::Composable>, size_t, ptrdiff_t, opentimelineio::v1_0::SerializableObject::Retainer<opentimelineio::v1_0::Composable> *, const opentimelineio::v1_0::SerializableObject::Retainer<opentimelineio::v1_0::Composable> *, opentimelineio::v1_0::SerializableObject::Retainer<opentimelineio::v1_0::Composable> &, const opentimelineio::v1_0::SerializableObject::Retainer<opentimelineio::v1_0::Composable> &>>>>" exists	opentimelineio	C:\dev\otio\OpenTimelineIO\src\opentimelineio\timeline.cpp	30	
```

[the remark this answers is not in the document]
meshula: My general rule is for auto is to apply DRY.

good auto:

```cpp
auto foo = std::make_shared<Some::Inconvenient<type::NameFromHell>>()
```
auto saves us from typing/reading the same strangled types twice

bad auto

```cpp
auto arentIMysterious = someWeirdoFunction();
```

src/opentimelineio/serializableObject.h
```diff
@@ -543,5 +547,11 @@ class SerializableObject {
     AnyDictionary _dynamic_fields;
     friend class TypeRegistry;
 };
+
+template <class T, class U>
+SerializableObject::Retainer<T> dynamic_retainer_cast(SerializableObject::Retainer<U> const& retainer)
```
meshula, line 552: I wonder if we need a different name for this? given class Foo, and class Bar::Foo, std:::dynamic_ptr_cast<Bar*>(myFoo) will return a shared_ptr<Bar>. dynamic_retainer_cast actually does a value unboxing, so think that perhaps there is some other C++ idiom we need here. https://en.cppreference.com/w/cpp/types/remove_reference ~ might remove_retainer make more sense?

meshula: dynamic_value_cast maybe? Since remove_reference doesn't also do a cast?

darbyjohnston: The dynamic_retainer_cast returns a Retainer by automatically constructing it from the unboxed value, isn't that similar to std:::dynamic_ptr_cast?

meshula: Ah right. Gremlins munching on my cerebrum, I focused on the return T, not the implicit conversion. I am now recalling we talked this through aeons ago ;)

darbyjohnston: No worries! I forgot, did we decide if we wanted the other casts as well, i.e. static?

PR #926, opened 2021-03-25
C++ each_child()/each_clip()

### LINKED ISSUE #925: Add C++ versions of each_child()/each_clip()
It would be useful to have the each_child()/each_clip() functionality available in C++ for iterating over objects in a timeline. Currently these functions are Python only, located in timeline.py and composition.py.

darbyjohnston:
```Fixes #925```

This is a draft PR for adding the each_child()/each_clip() functionality to C++, currently the functions are only available from Python.

I took a look at the custom iterators suggested in the other PR (https://internalpointers.com/post/writing-custom-iterators-modern-cpp), but it seems like that's only applicable if you are using custom containers? Since OTIO uses the std containers I tried implementing each_child()/each_clip() to return a vector which seems to work nicely:
```
for (const auto& clip : timeline->each_clip(error_status))
{
    ...
}
```

I tried to keep the C++ code as similar as possible to the Python code (located in composition.py), except for returning a vector vs. the Python generator syntax. I'm not that familiar with Python generators so I'm not sure if that code should stay or be replaced with the C++ code?

The only testing I've done so far is to port the Python conform.py example to C++ and try it on some of the sample data; as far as I can tell it seems to work OK... I've also only tested it on Windows, there's still some missing utility code for POSIX system that I need to add.

### CODE REVIEW

File: src/opentimelineio/composition.h [hunk 1]
```diff
@@ -72,12 +78,170 @@ class Composition : public Item {
...
+            out.push_back(valid_child);
+
+        // if not a shallow_search, for children that are compositions,
+        // recurse into their children
+        if (!shallow_search)
+            if (auto composition = dynamic_cast<Composition*>(child.value))
+            {
+                if (search_range)
+                {
+                    search_range = transformed_time_range(*search_range, composition, error_status);
```
meshula, line 155: In some of these checks, the loop is aborted, and an empty return results if error_status is true. error_status reports things such as "no media reference set on clip."

1 - when is an error not an error? For example, if I have just created a clip, and not yet set a media reference, is that clip in an erroneous state? The existing implementation says yes, but the default construction is an an erroneous object. Is it sensible for OTIO to construct by default, erroneous things? This question's resolution is outside the scope of this PR, but it leads to

2- would I legitimately want to use these iterators to collect clips and test for being in an erroneous state, as part of a sanity checking tool?

3- even if there is an erroneous clip, would I still want all the non-erroneous clips?

4- should there be another argument to specify whether erroneous clips result in an empty return? In other words, might I want versions of this routine that are (a) all or nothing, (b) all, or (c) only non-erroneous? TBH, I'm only seeing the utility of (b). If an error_status resulted in the (a) case, I need to find some other way to iterate to discover the bad results, yet here we are in an otherwise lovely iterator...

5- is the above the behavior of the python equivalent, and ... same set of questions :)

src/opentimelineio/composition.h
```diff
@@ -72,12 +78,170 @@ class Composition : public Item {
...
+template<typename T>
+inline std::vector<SerializableObject::Retainer<T>> Composition::each_child(
+    ErrorStatus* error_status,
+    optional<TimeRange> search_range,
+    bool shallow_search) const
+{
+    std::vector<Retainer<Composable>> children;
+    if (search_range)
+    {
+        auto range_map = range_of_all_children(error_status);
```
meshula, line 110: I'd recommend busting this out to a separate non-templated "filter" function. It's a generally useful function, likely to be used more than once. Busting out would reduce the amount of per-class code instantiated per <T>.

src/opentimelineio/errorStatus.h
```diff
@@ -8,7 +8,7 @@ namespace opentimelineio { namespace OPENTIMELINEIO_VERSION  {
 class SerializableObject;
 
 struct ErrorStatus {
-    operator bool () {
+    operator bool () const {
```
meshula, line 11: could you bust out this general improvement with util.cpp/h please?

examples/python_adapters_child_process.cpp
```diff
@@ -19,8 +19,6 @@
 #define WIN32_LEAN_AND_MEAN
```
meshula, line 19: general improvement

examples/python_adapters_embed.cpp
```diff
@@ -129,7 +129,7 @@ bool PythonAdapters::write_to_file(
 
         // Convert the C++ timeline to a string and pass that to Python.
         const auto string = timeline.value->to_json_string(error_status);
-        if (error_status->outcome != otio::ErrorStatus::Outcome::OK)
+        if (!error_status->outcome)
```
meshula, line 132: general improvement

examples/summarize_timing.cpp
```diff
@@ -39,9 +39,9 @@ void summarize_effects(otio::SerializableObject::Retainer<otio::Item> const& ite
     }
```
meshula, line 39: general improvement

examples/util.cpp
```diff
@@ -4,13 +4,11 @@
 
 #include <Python.h>
```
meshula, line 6: general improvement

examples/util.h
```diff
@@ -2,12 +2,42 @@
 
 #include <opentimelineio/errorStatus.h>
```
meshula, line 4: general improvement

meshula (changes requested):
This is great, thank you for the work!

Could I ask you to separate out the "general improvement" prerequisites to another PR please? Those can get landed quickly. They add a lot of bulk to the PR, and I'd like to focus on the each* specifically, rather than combining a bunch of things in one PR.

darbyjohnston:
Sure, sorry about the general changes, I'll split those into a separate PR. Should I also split out the example code, conform.cpp? I included it for testing the each_clip() functionality but it does bring in a bunch of extra code.

Also any thoughts on whether this should replace the Python implementations? If so it should be covered by the current tests, otherwise it would need it's own C++ tests?

src/opentimelineio/composition.h
darbyjohnston: Good points! I should have looked closer at the error codes, I was assuming they were all fatal and should stop any further processing.

Taking a look at the Python code it looks like they throw exceptions (otio_errorStatusHandler.cpp)? The OTIO docs mention that exceptions shouldn't be used in C++ code so maybe that's another complication...

So the (b) option would be to not return immediately on an error, and continue processing? That does seem useful for uses cases like the sanity checking tool, though if there are multiple errors the user would only know about the last one.

src/opentimelineio/composition.h
darbyjohnston: In composition.py there is also a function called child_at_time() which is sort of similar, but for a specific time instead of a time range.

src/opentimelineio/composition.h
meshula: Throwing python exceptions is the responsibility of Python code :) Indeed the C++ code shouldn't throw, but that doesn't constrain the Python implementation.

I'd suggest, continue processing. If an error was encountered, return a new error status, perhaps "one or more invalid children encountered".

src/opentimelineio/composition.h
meshula: child_at_time() and children_in_range() perhaps?

meshula:
In the spirit of Don't Repeat Yourself, and also in fully consistent behavior, I would personally advocate that yes, the Python implementations should expose these new C++ implementations (imposing the pythonic exception behavior at the binding level). Benefits: DRY, consistency, performance, availability to all our other language bindings.

darbyjohnston:
OK, I think I have addressed all the feedback so far, though the Python bindings still need work. I managed to replace the "child_at_time()" Python implementation with the C++ version; I'm not that familiar with pybind, but the unit tests pass, so that's good. :) I haven't tried replacing the Python "each_child()/each_clip()" implementations yet, I need to look into the pyside docs for how to deal with C++ templates.

In the "child_at_time()" pybind wrappings, is it correct to call "take_value()" on the C++ Retainers?

meshula:
take_value steals the value out of the retainer and nulls the value in the retainer, so it's not meant for transferring from retained container to another. I personally find retain count reasoning a little mind bending, and prefer to write out each line involving Retainers, even documenting on each line what I expect the count to do. In the test suite for the C bindings it was very helpful to add tests for expected reference counts for patterns we puzzled over in development, as a sort of document and reminder of troubleshooting we did.  

Here's the docs on take_value:

.. Note::
    We do not expect the following scenario to arise, but it
    is certainly possible to write a function which returns a raw pointer
    back to the user *and* also gives them the responsibility for possibly
    deleting it: ::
```cpp
        Item* remove_and_return_named_item(Composition* c, std::string const& name) {
            auto& children = c->children();
            for (size_t i = 0; i < children.size(); ++i) {
                if (children[i].value->name() == name) {
                    SerializableObject::Retainer<Item> r_item(children[i]);
                    c->remove_child(i);
                    return r_item.take_value();
                }
            }
            return nullptr;
        }
```
    The raw pointer in a ``Retainer`` object is accessed via the ``value`` member.
    The call to ``take_value()`` decrements 
    the reference count of the pointed to object but does not delete the instance
    if the count drops to zero.  The pointer is returned to the caller, and
    the ``Retainer`` instance sets its internal pointer to null.
    Effectively, this delivers a raw
    pointer back to the caller, while also giving them the responsibility to try to delete
    the object if they were the only remaining owner of the object.

darbyjohnston:
Thanks, transferring ownership is definitely not intended, I tried removing the "retainer.take_value()" calls and replaced them with "retainer.value".

I just checked in a change to replace the Python "each_child()" functions with the C++ version, unfortunately the confidence tests are seg-faulting in "test_filter_algorithms.py" so I need to look more into that.

The way I Python wrapped the C++ "each_child()" function is maybe not ideal, but I couldn't find anything about how to deal with passing Python types into a pybind wrapped function. So I tried specializing each function call, for example "each_child_clip()", "each_child_track()", etc.

darbyjohnston:
I figured out a way to use the Python type for ```each_child()``` so it now works the same way it did before. The trick was to use the pybind11 py::type class like:
```descended_from_type.is(py::type::handle_of<T>())```

Unfortunately "test_filter_algorithms.py" is still seg-faulting so there are still some issues with the memory management...

**Comment by darbyjohnston** [comment]
Trying to track down the seg-fault, I reduced one of the tests in "test_filter_algorithms.py" to the bare minimum amount of code that triggers the crash:
```
import opentimelineio as otio
import copy

if __name__ == "__main__":

    # Create a track and child clip
    tr = otio.schema.Track(name='foo')
    tr.append(otio.schema.Clip(name='cl1'))

    # Deep copy the track
    tr_copy = copy.deepcopy(tr)
    #tr_copy = tr

    seg_fault = True
    if seg_fault:
        # Call each_child()
        children = tr_copy.each_child()
        # Remove the child and add it back
        del tr_copy[0]
        # BOOM!
        tr_copy.append(children[0])
    else:
        # Remove the child and add it back
        child = tr_copy[0]
        del tr_copy[0]
        tr_copy.append(child)

    # Success?    
    print("tr_copy:", tr_copy)
```

**Comment by meshula** [comment]
looks like something didn't add a retain on the clip, that should have added a retain.

**Comment by darbyjohnston** [comment]
Yes, though I'm not sure what the correct way is to add the retain? The other wrappers return the pointer to the C++ object like I am doing, and seem to work OK? (Except for Track "neighbors_of", which returns ```py::cast(result.first.take_value())``` which doesn't seem right...) Is there something different I need to do because I'm returning a list?

**Comment by meshula** [comment]
Sorry not to have a better suggestion, but the way I've debugged this issue in the past (debugging counts in the C bindings) was to take a snip of code like you've demonstrated and then reason about the counts, and add assertions on all the counts of objects between every operation to help fix things. It exposed issues,,, or hidden assumptions.... on the retains and releases at both levels of the bindings.

**Comment by darbyjohnston** [comment]
OK, I think I have a solution; instead of using py::list in the wrapper methods I tried using std::vector and including "pybind11/stl.h" for the automatic conversion to Python. It seems to be working now and the tests are all passing.

**Comment by meshula** [comment]
that's a nice solution!

**Comment by darbyjohnston** [comment]
Thanks! Since everything seems to be working OK I will "undraft" this PR for any final review. If this lands OK I also have a C++ version of the "conform.py" example that demonstrates the use of "each_clip()" from C++.

File: src/py-opentimelineio/opentimelineio-bindings/otio_serializableObjects.cpp [hunk 9]
```diff
@@ -51,6 +52,47 @@ namespace {
...
+        return l;
+    }
+    
+    template<typename T>
+    std::vector<SerializableObject*> each_clip(T* t, optional<TimeRange> const& search_range, bool shallow_search = false) {
+        std::vector<SerializableObject*> l;
+        for (const auto& child : t->each_clip(ErrorStatusHandler(), search_range, shallow_search)) {
+            l.push_back(child.value);
+        }
+        return l;
```
ssteinbach, line 94: The subtle distinction here from the original implementation is that it returns a static list rather than an iterator?  Meaning, it has to build the entire list in memory and then return that, rather than using the `yield` style iterator?

darbyjohnston: Yes, AFAIK there's not an equivalent to ```yield``` in C++. Since the vector is just holding pointers to the objects it shouldn't be too expensive, unless you anticipate large sizes like over 10K. The speed of the function might also be faster in C++ though I haven't done any benchmarks comparing the two implementations.

If that changes the style of the Python usage too much we could bring back the Python implementation, though it seemed nice to share the code. Another possibility would be sort of a hybrid approach where instead of Python code calling the C++ wrapper function directly, there was a Python version of ```each_child()/each_clip()``` that used the C++ functions but exposed a more Python like interface.

ssteinbach: @reinecke what do you think?  Is it ok to give up the iterator-interface of `each_clip()`?  I kind of feel like the name is misleading if it isn't an iterator.  Something like `all_clips()` is more intuitive to me if it builds a list.

darbyjohnston: Something like this might work as a hybrid approach?
```
def each_clip(self, search_range=None, descended_from_type=_otio.Composable):
    # _each_clip() is the C++ implementation
    for child in _each_clip(search_range, descended_from_type):
        yield child
```

meshula: The c++ equivalent of yield is any of the container accessors that return an iterator, since an iterator is somewhat equivalent to a generator. The big difference is that a generator is going to hold a reference on a container so the container won't vanish under foot, but a C++ iterator does not hold a reference. Due to the complexity of the OTIO data structures, an iterator would be complex to implement and fragile in the sense that the iterator has to be very cognizant of implementation details. Returning a vector makes more sense for C++ makes a lot more sense to avoid the complexity and fragility.

It seems to me that you don't have to give up the public facing part of the implementation, as a python generator could be wrap the list from the C++ implementation. The generator would be responsible for cleaning up the vector when the vector is exhausted, since the magic thing about generators is you can only iterate them forward, once.

So I think we can have the (assumed) efficiency of the C++ implementation, but still return a generator for iteration. @ssteinbach since you're saying the `yield` style iterator, I have read that as `generator`. Hopefully I haven't read too much into that.

rogernelson: I'm not sure if you looked into using pybind's make_iterator but it might be helpful to you.  There are some examples here:
https://github.com/pybind/pybind11/blob/master/tests/test_sequences_and_iterators.cpp

If you really wanted to add yield-like functionality on the c++ side as well, you'd likely have to add your own iterator class: 
https://www.cplusplus.com/reference/iterator/iterator/ 

That would probably be a fair amount of work.  A less involved solution might be to use make_iterator on the returned vector.  That way at least the generation of the Python instances from the c++ ones would be generated.

darbyjohnston: Hi Roger, thanks for the links! I did not know about the pybind11 iterators, I will check them out. Do you think that's a better approach than the small Python code snippet I posted above?

rogernelson: No problem, I hope they help!

I think one thing that is nice about the make_iterator approach is that you can add this at the end (as the examples do):
```
py::keep_alive<0, 1>()
```
So you effectively bind the lifetime of the generator and the vector (in your case) together and ensuring the vector only needs to be build once.  By calling the c++ method from python each time, I think it would generate the vector on each call, even though the python objects are generated, right?  

I haven't tried this approach myself, so I might be wrong about it.  It might not be worth the effort if you try it and it starts crashing everywhere because of the vector's lifetime.

darbyjohnston: I tried adding ```py::make_iterator``` to the returned vector but unfortunately got seg-faults in the tests again, which I assume is because the vector is being created on the stack and getting destroyed prematurely. It seems like you would need a custom container class to get it to work right.

I also tried the hybrid Python approach and it seemed to work OK, and didn't re-generate the vector on each call. Which makes sense since otherwise how would it be able to iterate? Python must keep the state of the function around for the lifespan of the iterator and "yields" control flow back to the calling function. However one downside of this approach is that each Python class (Timeline, Stack, Track, etc.) needs the function definitions which is a fair bit of boiler-plate code.

So, to sum up it seems like there are three options:

1. Leave changes as-is. Pros: probably the least amount of code of the three options. Cons: maybe not "Pythonic" without the iterators?
2. Use the "hybrid" approach. Pros: fairly simple and keeps the "Pythonic" iterators. Cons: a bit of Python boiler-plate code as I mentioned above.
3. Use a custom container class with "py::make_iterator". Pros: probably the most "correct" way of the three options. Cons: probably adds the most code of the three options...

ssteinbach: The fourth option is to keep the implementation of `each_child` and `each_clip` for the python implementation _in_ python, duplicating the code in each place.

darbyjohnston: True, but that duplicates quite a bit of code, which isn't that complex but isn't trivial either. I think that also implies adding C++ confidence tests to verify that the C++ versions of the functions work the same way as the Python versions.

I think the "hybrid" approach is functionally the same, it leaves the existing Python API as-is but replaces the implementations with the C++ functions. Maybe it was wrong to call it boiler-plate code in that respect.

darbyjohnston: My personal preference (C++ biased) would be for option 1, since it is the least amount of code and anything in Python inheriting from ```Composition``` will automatically get support for ```each_child()```. Though if the Python iterators are important to keep I like option 2, since it is also a relatively small amount of code and just needs some additional Python to decorate each class to support ```each_child()```.

meshula:
I propose we take up this discussion in the next TSC.

ssteinbach:
Summary of the discussion from the TSC meeting:

- each_* is for a function that returns an iterator
- introduce a new function, I would propose `children_if` or `all_children_if` to get the conversation started, which returns a list rather than an iterator
- swap the implementation of each_child such that it returns the iterator for the list generated by the c++ implementation of the above new function
- note in the docstring for `each_child` and `each_clip` that they are deprecated

darbyjohnston:
I renamed the C++ ```each_child()``` to ```children_if()```, and ```each_clip()``` to ```clip_if()```. I also put back the original Python functions and replaced their contents with calls to the C++ functions. Not all of the Python functions had docstrings, so I added them where they were missing with the deprecation note.

I think that addresses everything, unless I should convert the tests to use the new function names?

darbyjohnston:
Sorry for the delay; at the last TSC meeting I had suggested updating the tests to use the new function names, but on second thought maybe that should be done as a separate pass to keep the code changes in this PR limited. I'm happy to do a follow up PR with the name changes, otherwise I think this PR is complete?

AliRAMADA (approved):

src/opentimelineio/composition.cpp
```diff
@@ -335,4 +335,150 @@ bool Composition::has_child(Composable* child) const {
...
 }
 
+SerializableObject::Retainer<Composable> Composition::child_at_time(
+    RationalTime const& search_time,
+    ErrorStatus* error_status,
+    bool shallow_search) const
+{
+    auto range_map = range_of_all_children(error_status);
+    if (!error_status)
+        *error_status = ErrorStatus(ErrorStatus::INTERNAL_ERROR, "one or more invalid children encountered");
```
ssteinbach, line 345:
```suggestion
    if (!error_status) {
        *error_status = ErrorStatus(ErrorStatus::INTERNAL_ERROR, "one or more invalid children encountered");
    }
```

src/opentimelineio/composition.cpp
```diff
@@ -335,4 +335,150 @@ bool Composition::has_child(Composable* child) const {
...
+
+    // find the first item whose end_time_exclusive is after the
+    const auto first_inside_range = _bisect_left(
+        search_time,
+        [&range_map](Composable* child) { return range_map[child].end_time_exclusive(); },
+        error_status);
+    if (!error_status)
+        *error_status = ErrorStatus(ErrorStatus::INTERNAL_ERROR, "one or more invalid children encountered");
```
ssteinbach, line 353:
```suggestion
    if (!error_status) {
        *error_status = ErrorStatus(ErrorStatus::INTERNAL_ERROR, "one or more invalid children encountered");
    }
```

src/opentimelineio/composition.cpp
```diff
@@ -335,4 +335,150 @@ bool Composition::has_child(Composable* child) const {
...
+
+    // find the last item whose start_time is before the
+    const auto last_in_range = _bisect_right(
+        search_time,
+        [&range_map](Composable* child) { return range_map[child].start_time(); },
+        error_status,
+        first_inside_range);
+    if (!error_status)
+        *error_status = ErrorStatus(ErrorStatus::INTERNAL_ERROR, "one or more invalid children encountered");
```
ssteinbach, line 362:
```suggestion
    if (!error_status) {
        *error_status = ErrorStatus(ErrorStatus::INTERNAL_ERROR, "one or more invalid children encountered");
    }
```

src/opentimelineio/composition.cpp
```diff
@@ -335,4 +335,150 @@ bool Composition::has_child(Composable* child) const {
...
+
+    // limit the search to children who are in the search_range
+    std::vector<Retainer<Composable>> possible_matches;
+    for (auto child = _children.begin() + first_inside_range; child < _children.begin() + last_in_range; ++child)
+        possible_matches.push_back(child->value);
+
```
ssteinbach, line 368:
```suggestion
    for (auto child = _children.begin() + first_inside_range; child < _children.begin() + last_in_range; ++child) {
        possible_matches.push_back(child->value);
    }
```

src/opentimelineio/composition.cpp
```diff
@@ -335,4 +335,150 @@ bool Composition::has_child(Composable* child) const {
...
+    for (auto child = _children.begin() + first_inside_range; child < _children.begin() + last_in_range; ++child)
+        possible_matches.push_back(child->value);
+    
+    Retainer<Composable> result;
+    for (const auto& thing : possible_matches)
+        if (range_map[thing].overlaps(search_time))
+        {
+            result = thing;
+            break;
+        }
```
ssteinbach, line 375:
```suggestion
    for (const auto& thing : possible_matches) {
        if (range_map[thing].overlaps(search_time))
        {
            result = thing;
            break;
        }
    }
```

src/opentimelineio/composition.cpp
```diff
@@ -335,4 +335,150 @@ bool Composition::has_child(Composable* child) const {
...
+
+    // if the search cannot or should not continue
+    auto composition = Retainer<Composition>(dynamic_cast<Composition*>(result.value));
+    if (!result || shallow_search || !composition)
+        return result;
```
ssteinbach, line 380:
```suggestion
    if (!result || shallow_search || !composition) {
        return result;
    }
```

src/opentimelineio/composition.cpp
```diff
@@ -335,4 +335,150 @@ bool Composition::has_child(Composable* child) const {
...
+
+    // before you recurse, you have to transform the time into the
+    // space of the child
+    const auto child_search_time = transformed_time(search_time, composition.value, error_status);
+    if (!error_status)
+        *error_status = ErrorStatus(ErrorStatus::INTERNAL_ERROR, "one or more invalid children encountered");
```
ssteinbach, line 386:
```suggestion
    if (!error_status) {
        *error_status = ErrorStatus(ErrorStatus::INTERNAL_ERROR, "one or more invalid children encountered");
    }
```

src/opentimelineio/composition.cpp
```diff
@@ -335,4 +335,150 @@ bool Composition::has_child(Composable* child) const {
...
+
+    result = composition.value->child_at_time(child_search_time, error_status, shallow_search);
+    if (!error_status)
+        *error_status = ErrorStatus(ErrorStatus::INTERNAL_ERROR, "one or more invalid children encountered");
```
ssteinbach, line 390:
```suggestion
    if (!error_status) {
        *error_status = ErrorStatus(ErrorStatus::INTERNAL_ERROR, "one or more invalid children encountered");
    }
```

src/opentimelineio/composition.cpp
```diff
@@ -335,4 +335,150 @@ bool Composition::has_child(Composable* child) const {
...
+    return result;
+}
+
+std::vector<SerializableObject::Retainer<Composable>> Composition::children_in_range(
+    TimeRange const& search_range,
+    ErrorStatus* error_status) const
+{
+    auto range_map = range_of_all_children(error_status);
+    if (!error_status)
+        *error_status = ErrorStatus(ErrorStatus::INTERNAL_ERROR, "one or more invalid children encountered");
```
ssteinbach, line 400:
```suggestion
    if (!error_status) {
        *error_status = ErrorStatus(ErrorStatus::INTERNAL_ERROR, "one or more invalid children encountered");
    }
```

src/opentimelineio/composition.cpp
```diff
@@ -335,4 +335,150 @@ bool Composition::has_child(Composable* child) const {
...
+
+    // find the first item whose end_time_inclusive is after the
+    // start_time of the search range
+    const auto first_inside_range = _bisect_left(
+        search_range.start_time(),
+        [&range_map](Composable* child) { return range_map[child].end_time_inclusive(); },
+        error_status);
+    if (!error_status)
+        *error_status = ErrorStatus(ErrorStatus::INTERNAL_ERROR, "one or more invalid children encountered");
```
ssteinbach, line 409:
```suggestion
    if (!error_status) {
        *error_status = ErrorStatus(ErrorStatus::INTERNAL_ERROR, "one or more invalid children encountered");
    }
```

src/opentimelineio/composition.cpp
```diff
@@ -335,4 +335,150 @@ bool Composition::has_child(Composable* child) const {
...
+
+    // find the last item whose start_time is before the
+    // end_time_inclusive of the search_range
+    const auto last_in_range = _bisect_right(
+        search_range.end_time_inclusive(),
+        [&range_map](Composable* child) { return range_map[child].start_time(); },
+        error_status,
+        first_inside_range);
+    if (!error_status)
+        *error_status = ErrorStatus(ErrorStatus::INTERNAL_ERROR, "one or more invalid children encountered");
```
ssteinbach, line 419:
```suggestion
    if (!error_status) {
        *error_status = ErrorStatus(ErrorStatus::INTERNAL_ERROR, "one or more invalid children encountered");
    }
```

src/opentimelineio/composition.cpp
```diff
@@ -335,4 +335,150 @@ bool Composition::has_child(Composable* child) const {
...
+
+    // limit the search to children who are in the search_range
+    std::vector<Retainer<Composable>> children;
+    for (auto child = _children.begin() + first_inside_range; child < _children.begin() + last_in_range; ++child)
+        children.push_back(child->value);
```
ssteinbach, line 424:
```suggestion
    for (auto child = _children.begin() + first_inside_range; child < _children.begin() + last_in_range; ++child) {
        children.push_back(child->value);
    }
```

src/opentimelineio/composition.cpp
```diff
@@ -335,4 +335,150 @@ bool Composition::has_child(Composable* child) const {
...
+    optional<int64_t> lower_search_bound,
+    optional<int64_t> upper_search_bound) const
+{
+    if (*lower_search_bound < 0)
+    {
+        *error_status = ErrorStatus(ErrorStatus::Outcome::INTERNAL_ERROR, "lower_search_bound must be non-negative");
+        return 0;
+    }
+    if (!upper_search_bound)
+        upper_search_bound = optional<int64_t>(_children.size());
```
ssteinbach, line 441:
```suggestion
    if (!upper_search_bound) {
        upper_search_bound = optional<int64_t>(_children.size());
    }
```

src/opentimelineio/composition.cpp
```diff
@@ -335,4 +335,150 @@ bool Composition::has_child(Composable* child) const {
...
+    int64_t midpoint_index = 0;
+    while (*lower_search_bound < *upper_search_bound)
+    {
+        midpoint_index = static_cast<int64_t>(std::floor((*lower_search_bound + *upper_search_bound) / 2.0));
+
+        if (tgt < key_func(_children[midpoint_index]))
+            upper_search_bound = midpoint_index;
+        else
+            lower_search_bound = midpoint_index + 1;
```
ssteinbach, line 450:
```suggestion
        if (tgt < key_func(_children[midpoint_index])) {
            upper_search_bound = midpoint_index;
        } 
        else {
            lower_search_bound = midpoint_index + 1;
        }
```

src/opentimelineio/composition.cpp
```diff
@@ -335,4 +335,150 @@ bool Composition::has_child(Composable* child) const {
...
+    optional<int64_t> lower_search_bound,
+    optional<int64_t> upper_search_bound) const
+{
+    if (*lower_search_bound < 0)
+    {
+        *error_status = ErrorStatus(ErrorStatus::Outcome::INTERNAL_ERROR, "lower_search_bound must be non-negative");
+        return 0;
+    }
+    if (!upper_search_bound)
+        upper_search_bound = optional<int64_t>(_children.size());
```
ssteinbach, line 469:
```suggestion
    if (!upper_search_bound) {
        upper_search_bound = optional<int64_t>(_children.size());
    }
```

src/opentimelineio/composition.cpp
```diff
@@ -335,4 +335,150 @@ bool Composition::has_child(Composable* child) const {
...
+    int64_t midpoint_index = 0;
+    while (*lower_search_bound < *upper_search_bound)
+    {
+        midpoint_index = static_cast<int64_t>(std::floor((*lower_search_bound + *upper_search_bound) / 2.0));
+
+        if (key_func(_children[midpoint_index]) < tgt)
+            lower_search_bound = midpoint_index + 1;
+        else
+            upper_search_bound = midpoint_index;
+    }
```
ssteinbach, line 479:
```suggestion
        if (key_func(_children[midpoint_index]) < tgt) {
            lower_search_bound = midpoint_index + 1;
        }
        else {
            upper_search_bound = midpoint_index;
        }
    }
```

src/opentimelineio/composition.h
```diff
@@ -72,12 +96,93 @@ class Composition : public Item {
...
+    else
+    {
+        // otherwise search all the children
+        children = _children;
+    }
+    std::vector<Retainer<T>> out;
+    for (const auto& child : children)
+    {
+        if (auto valid_child = dynamic_cast<T*>(child.value))
+            out.push_back(valid_child);
```
ssteinbach, line 161:
```suggestion
        if (auto valid_child = dynamic_cast<T*>(child.value)) {
            out.push_back(valid_child);
        }
```

src/opentimelineio/composition.h
```diff
@@ -72,12 +96,93 @@ class Composition : public Item {
...
+                    search_range = transformed_time_range(*search_range, composition, error_status);
+                    if (!error_status)
+                        *error_status = ErrorStatus(ErrorStatus::INTERNAL_ERROR, "one or more invalid children encountered");
+                }
+
+                const auto valid_children = composition->children_if<T>(error_status, search_range, shallow_search);
+                if (!error_status)
+                    *error_status = ErrorStatus(ErrorStatus::INTERNAL_ERROR, "one or more invalid children encountered");
+                for (const auto& valid_child : valid_children)
+                    out.push_back(valid_child);
```
ssteinbach, line 180:
```suggestion
                if (!error_status) {
                    *error_status = ErrorStatus(ErrorStatus::INTERNAL_ERROR, "one or more invalid children encountered");
                }
                for (const auto& valid_child : valid_children) {
                    out.push_back(valid_child);
                }
```

src/opentimelineio/serializableCollection.h
```diff
@@ -46,4 +70,45 @@ class SerializableCollection : public SerializableObjectWithMetadata {
...
+        // recurse into their children
+        if (!shallow_search)
+        {
+            if (auto collection = dynamic_cast<SerializableCollection*>(child.value))
+            {
+                const auto valid_children = collection->children_if<T>(error_status, search_range);
+                if (!error_status)
+                    *error_status = ErrorStatus(ErrorStatus::INTERNAL_ERROR, "one or more invalid children encountered");
+                for (const auto& valid_child : valid_children)
+                    out.push_back(valid_child);
```
ssteinbach, line 99:
```suggestion
                if (!error_status) {
                    *error_status = ErrorStatus(ErrorStatus::INTERNAL_ERROR, "one or more invalid children encountered");
                }
                for (const auto& valid_child : valid_children) {
                    out.push_back(valid_child);
                }
```

src/opentimelineio/serializableCollection.h
```diff
@@ -46,4 +70,45 @@ class SerializableCollection : public SerializableObjectWithMetadata {
...
+            }
+            else if (auto composition = dynamic_cast<Composition*>(child.value))
+            {
+                const auto valid_children = composition->children_if<T>(error_status, search_range);
+                if (!error_status)
+                    *error_status = ErrorStatus(ErrorStatus::INTERNAL_ERROR, "one or more invalid children encountered");
+                for (const auto& valid_child : valid_children)
+                    out.push_back(valid_child);
```
ssteinbach, line 107:
```suggestion
                if (!error_status) {
                    *error_status = ErrorStatus(ErrorStatus::INTERNAL_ERROR, "one or more invalid children encountered");
                }
                for (const auto& valid_child : valid_children) {
                    out.push_back(valid_child);
                }
```

src/opentimelineio/composition.h
```diff
@@ -72,12 +96,93 @@ class Composition : public Item {
...
+    optional<TimeRange> search_range,
+    bool shallow_search) const
+{
+    std::vector<Retainer<Composable>> children;
+    if (search_range)
+    {
+        // limit the search to children who are in the search_range
+        children = children_in_range(*search_range, error_status);
+        if (!error_status)
+            *error_status = ErrorStatus(ErrorStatus::INTERNAL_ERROR, "one or more invalid children encountered");
```
ssteinbach, line 150:
```suggestion
        if (!error_status) {
            *error_status = ErrorStatus(ErrorStatus::INTERNAL_ERROR, "one or more invalid children encountered");
        }
```

src/opentimelineio/composition.h
```diff
@@ -72,12 +96,93 @@ class Composition : public Item {
...
+        // recurse into their children
+        if (!shallow_search)
+        {
+            if (auto composition = dynamic_cast<Composition*>(child.value))
+            {
+                if (search_range)
+                {
+                    search_range = transformed_time_range(*search_range, composition, error_status);
+                    if (!error_status)
+                        *error_status = ErrorStatus(ErrorStatus::INTERNAL_ERROR, "one or more invalid children encountered");
```
ssteinbach, line 173:
```suggestion
                    if (!error_status) {
                        *error_status = ErrorStatus(ErrorStatus::INTERNAL_ERROR, "one or more invalid children encountered");
                    }
```

ssteinbach (commented):
Just adding `{` and `}` for single line if and for statements to conform to the style of other .cpp files in the library. There is an outstanding issue to add some style linting for C++ code as well.

ssteinbach (approved):
Going to land after committing the style fixes.

meshula (approved):

darbyjohnston:
Thanks! I'll create a follow up PR to update the code with the new API.

src/opentimelineio/composition.cpp
```diff
@@ -335,4 +335,167 @@ bool Composition::has_child(Composable* child) const {
     return _child_set.find(child) != _child_set.end();
 }
 
+SerializableObject::Retainer<Composable> Composition::child_at_time(
+    RationalTime const& search_time,
+    ErrorStatus* error_status,
+    bool shallow_search) const
+{
+    auto range_map = range_of_all_children(error_status);
+    if (!error_status) {
```
gfgit, line 344: Hi, I didn't test the code but the if branch clearly leads to segmentation fault. The `!` should be removed otherwise a null pointer gets dereferenced. This is repeated also in following lines of code.

darbyjohnston: OK, I've created a new issue #1060, and PR #1061, to fix checking the error_status. I did not add checks for whether the error_status pointer is valid, since that doesn't happen in the surrounding code.

PR #931, opened 2021-04-05
Add dependabot

ssteinbach:
Adds a dependabot.yml to opentimelineio.

Dependabot is a GitHub service that opens Pull requests to update the version of dependencies in the library.  For more information:
https://docs.github.com/en/code-security/supply-chain-security/enabling-and-disabling-version-updates#enabling-github-dependabot-version-updates

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,7 +7,7 @@ updates:
   #   schedule:
   #     interval: "daily"
 
-  # Maintain dependencies for npm
+  # Maintain dependencies for pip
   - package-ecosystem: "pip"
     directory: "/"
     schedule:
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,7 +7,6 @@ updates:
   #   schedule:
   #     interval: "daily"
 
-  # Maintain dependencies for pip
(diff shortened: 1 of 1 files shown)


jhodges10:
Good call making it Monday's only @ssteinbach, all too often dependabot gets incredibly annoying because of incessant notifications.

.github/dependabot.yml
```diff
@@ -0,0 +1,19 @@
+version: 2
+updates:
+
+  # Maintain dependencies for GitHub Actions
+  # - package-ecosystem: "github-actions"
+  #   directory: "/"
+  #   schedule:
+  #     interval: "daily"
+
+  # Maintain dependencies for npm
```
reinecke, line 10: This should say `pip` instead of `npm`.

reinecke (commented):
Just one copy/paste issue in a comment. This is a good add though, especially for our submodules.

.github/dependabot.yml
ssteinbach, line 10:
```suggestion
  # Maintain dependencies for pip
```

reinecke (approved):

PR #939, opened 2021-04-12
Removed inadvertent unicode char

avrata:
Fixes: 
https://github.com/PixarAnimationStudios/OpenTimelineIO/issues/938

Removed non-ascii char to allow python parsing


ssteinbach (approved):
good catch, thanks!

PR #942, opened 2021-04-15
Localize test otiod/z test data to sample_data directory

avrata:
Localizes the test data for the otiod/z adapter into the sample_data dir. 

I am reticent to add these two new files so if there are other options, please let me know.

Thanks,
Robyn


ssteinbach:
Good fix, thanks.

ssteinbach (approved):

ssteinbach (approved):
This is better than reaching into the docs directory, thanks.

PR #945, opened 2021-04-16
Only load a manifest if it has not been loaded

ssteinbach:
- During plugin loading, checks to see which source_files have already been loaded before loading a new plugin.
- For pkg_resource plugins that use the function approach rather than the manifest.json approach, generates a source_files path based on the path to the module containing the function.
- Adds unit tests to check duplicate entries.
- Also updates and clarifies some documentation around how the plugin manifests are loaded.

### DIFF
--- a/docs/tutorials/otio-serialized-schema.md
+++ b/docs/tutorials/otio-serialized-schema.md
@@ -221,11 +221,13 @@ parameters:
 ```
 Defines an OTIO plugin Manifest.
 
-    This is an internal OTIO implementation detail.  A manifest tracks a
-    collection of adapters and allows finding specific adapters by suffix
+    This is considered an internal OTIO implementation detail.
 
-    For writing your own adapters, consult:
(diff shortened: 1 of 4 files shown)


avrata:
This all looks good to me. I did have two thoughts but in thinking about it more I don't know if they are really worth any changes:

* Any concern with relative vs absolute paths? This code did not try and normalize them before, but with using the file path as a key now, there could be some inadvertent missed collisions if they are intermixed. 
* Could/should this all be localized into `Manifest.extend()`? I don't see an easy way to do this currently though using `source_files`. (such that people might want to manually call this, but could get into trouble if so)

I still wish I could track down what the "duplicate schema def" issue was and if we should be super strict about never allowing duplicates in any of these internal lists or not.

ssteinbach:
@avrata I think this function could use a clean up pass and refactor how the manifests aggregate and are loaded.  But this adds a check for relative vs absolute paths.

avrata:
Looks great! Thanks

meshula (approved):

PR #947, opened 2021-04-19
Ascii doc fix

avrata:
Found an additional non-ascii character in a doc string related to a citation.


ssteinbach (approved):
these are so weird, thanks.

PR #949, opened 2021-04-21
Allow test to run in both python 2.7 and python 3.3+ wrt reloading modules

avrata:
The test_plugin_detection test was failing for me in python 2.7 due to a missing `importlib.reload()`. I believe this is only available in python 3.4+. 

This change lets the test fall back to `imp.reload()` in cases where it can't find the `importlib` one. 

There is a related question about how the 2.7 tests did not detect this which I think is related to a missing `mock` module. I'll follow up offline about that.

### DIFF
--- a/tests/test_plugin_detection.py
+++ b/tests/test_plugin_detection.py
@@ -44,6 +44,7 @@ try:
     # Python3:  use importlib.reload
     from importlib import reload as import_reload
 except ImportError:
+    # Python2:
     from imp import reload as import_reload
 
 import opentimelineio as otio
--- a/tests/test_plugin_detection.py
+++ b/tests/test_plugin_detection.py
@@ -41,7 +41,7 @@ except ImportError:
         could_import_mock = False
 
 try:
-    # Python 3.4 forwards use importlib.reload
+    # Python3:  use importlib.reload
     from importlib import reload as import_reload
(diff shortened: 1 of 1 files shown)


tests/test_plugin_detection.py
```diff
@@ -41,6 +40,11 @@
         # Mock appears to not be installed
         could_import_mock = False
 
+try:
+    # Python 3.4 forwards use importlib.reload
```
ssteinbach, line 44:
```suggestion
    # Python3:  use importlib.reload
```
very small nitpick to make it easier to find this later and remove it (when we eventually drop python2 support).

tests/test_plugin_detection.py
```diff
@@ -41,6 +40,11 @@
...
+    from importlib import reload as import_reload
+except ImportError:
+    from imp import reload as import_reload
```
ssteinbach, line 47:
```suggestion
    # Python2:
    from imp import reload as import_reload
```

ssteinbach (changes requested):

ssteinbach:
Very small request - just want to put comments in that are like the other comments in the codebase to make it easier to find this via future Grepping.  thanks!

ssteinbach (approved):

PR #955, opened 2021-04-28
Refactor setup.py for cmake updates

reinecke:
Refactors `setup.py` to properly leverage the newly cleaned up cmake builds (reworked in #837).

**Summarize your change.**

Our `setup.py` had been arranged to try and figure out the final install location for for the libraries and place products in there. The python packaging utilities actually want you to assemble your build into a temp location where it will then pick it up to either package or install to the proper place.

This refactor now tells the cmake build to install into that temp location. Because of this, we have to manually maintain much less code that supports determining the user's desired end action with the assembled package (e.x. install, install using --user, bdist, bdist_wheel).

Also removed `_Ctx` class instance that was storing build config as a global - it made it hard to follow the flow of code and would likely be prone to bugs.

PR #837, opened 2020-12-08
Cmake cleanup

meshula:
This commit allows a full cmake build.

- consolidate options at the top level
- detect a "reasonable" place to install the python components
- restrict gcov coverage to platforms that support it (those built with gcc)
- use cmake build from setup,py on all platforms (cmake will call make appropriately, so the outcome is identical to how it was)
- update submodules (any/optional/etc) BEFORE installing them
- add ability to build OTIO libraries as static
- add optional debug postfix because windows release/debug builds are not compatible
- add platform appropriate rpaths (@loader_path on Mac, $ORIGIN on Linux)
- add ability for cmake to do an install into the python environment so that a cmake build automatically results in a functional python build
- more granular build control (turn python builds or c++ builds on or off independently)
- For IDEs, rename the .sln or .xcode file to OpenTimelineIO (was OPENTIMELINEIO_ROOT)

PR #956, opened 2021-04-28
Refactor setup.py to build and stage into temp area for setup.py to install from

reinecke:
Refactors `setup.py` to properly leverage the newly cleaned up cmake builds (reworked in #837).

(Note: this replaces #955 that wouldn't reopen due to GitHub oddness)

**Summarize your change.**

Our `setup.py` had been arranged to try and figure out the final install location for for the libraries and place products in there. The python packaging utilities actually want you to assemble your build into a temp location where it will then pick it up to either package or install to the proper place.

This refactor now tells the cmake build to install into that temp location. Because of this, we have to manually maintain much less code that supports determining the user's desired end action with the assembled package (e.x. install, install using --user, bdist, bdist_wheel).

Other changes include:
- Removed `_Ctx` class instance that was storing build config as a global - it made it hard to follow the flow of code and would likely be prone to bugs.
- Removed setuptools and pip version checks [per this conversation](https://github.com/PixarAnimationStudios/OpenTimelineIO/pull/955#discussion_r622365200)
- Updated `-jx` flag generation to be based on detected number of system CPUs (better compile speeds!)


setup.py
```diff
@@ -87,142 +64,106 @@ def cmake_version_check():
...
+        )
+
+        cmake_args += [
+            '-DOTIO_PYTHON_INSTALL_DIR=' + install_dir,
+            '-DCMAKE_INSTALL_PREFIX=' + cmake_install_prefix,
+        ]
+
+        if platform.system() == "Windows":
+            if sys.maxsize > 2**32:
+                cmake_args += ['-A', 'x64']
```
reinecke, line 133: Migrated from @meshula on [previous PR](https://github.com/PixarAnimationStudios/OpenTimelineIO/pull/955#discussion_r621901333):
> per usd_build.py, we should detect VS2019 or greater, as this argument doesn't apply for VS2017 and less, where 64 bit builds are indicated by appending "Win64" to the build target.

meshula: Alternatively we could add a note to our readme that currently vs2019 or later is required.

meshula (approved):

setup.py
ssteinbach: VFX Platform seems to support minimum vs2017 across the board, FWIW

ssteinbach (approved):
One note about vs 2017 vs 2019 and a formatting note suggestion to match the rest of the file.  Thanks!

setup.py
reinecke: In this re-factor I didn't actually change any of the build behavior beyond re-targeting the locations build products land in.

On this issue, I was thinking we address this when we do the cibuildwheel change because it will exercise a lot of wacky environment combos that will test what we want to address. How does that sound?

meshula: That's what I was implying by "currently vs2019". It's not a heavy lift to add vs2017 support, but at the same time, it is also not a regression due to this change; so a note in the Readme, now that we've noted the issue, suffices IMO

setup.py
```diff
@@ -87,183 +63,109 @@ def cmake_version_check():
...
+            os.makedirs(self.build_temp_dir)
+
+        debug = (self.debug or bool(os.environ.get("OTIO_CXX_DEBUG_BUILD")))
+        self.build_config = ('Debug' if debug else 'Release')
+
+        self.cmake_generate()
+        self.cmake_install()
+
+    def cmake_generate(self):
+        # Determine the target directory for the built library
```
JeanChristopheMorinPerso, line 104: I would add a more explicit explanation to say that the built library needs to go into the build's directory so that the it's picked up correctly by the rest of setuptools build? It might make it easier to remember what was needed in order to have the libraries included in the wheels.

reinecke: Good idea, done!

JeanChristopheMorinPerso:
When some hacks are removed from a setup.py file, it makes me extremely happy 😁

Great work!

ssteinbach (approved):
This looks great.  Anything else you want to change on this?  Or is it ready to land?

reinecke:
Let's land this one and move on! I'll do it now.

PR #958, opened 2021-04-30
uncomment github actions dependabot config

ssteinbach:
- add dependabot support for versioning GitHub actions


meshula (approved):
daily makes sense to me, as updates are likely to have critical fixes against things like mining attacks.

ssteinbach:
Exactly

PR #966, opened 2021-05-11
Add support for multiple timecode objects in an AAF

stefanschulze:
The current implementation of the AAF reader is very strict if it comes to the number of discovered TimeCode objects in the AAF. If there is more than one TimeCode object found, the reader throws an exception.

However, there are situations where you can have multiple TimeCode objects, for example one created via the AuxTC24 bin-column in MediaComposer. 

This change relaxes the rule a bit by accepting multiple TimeCode objects as long as they have the same start and length, in which case they are considered 'same'.

Also added an AAF with such multiple TimeCode objects and updated test_aaf_adapter to just read that AAF to make sure it can be read without throwing an exception.


contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py
```diff
@@ -162,14 +162,40 @@ def _find_timecode_mobs(item):
...
     """Given a mob with a single timecode slot, return the timecode and length
     in that slot as a tuple
     """
     timecodes = [slot.segment for slot in mob.slots
                  if isinstance(slot.segment, aaf2.components.Timecode)]
 
-    if len(timecodes) == 1:
+    # Only use timecode if we have just one or multiple ones with same
+    # start/length.
+    if len(timecodes) == 1 or timecode_values_are_same(timecodes):
```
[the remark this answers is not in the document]
ssteinbach: right, so:
```suggestion
    if timecode_values_are_same(timecodes):
```

contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py
```diff
@@ -162,14 +162,40 @@ def _find_timecode_mobs(item):
     return mobs
 
 
+def timecode_values_are_same(timecodes):
+    """
+    We can have a valid AAF with multiple timecode objects (for example
+    an auxTC24 value got added via the Avid Bin column). But as long as they
+    have the same start and length values, we can use them, treat them as 'same'
```
ssteinbach, line 169: we've been trying to move comments away from using 'we', to be more direct:

```suggestion
    An AAF can have multiple timecode objects (for example an auxTC24 value
    that got added via the Avid Bin column). As long as they have the same
    start and length values, they can be treated as being the same.
```

contrib/opentimelineio_contrib/adapters/tests/test_aaf_adapter.py
```diff
@@ -1012,6 +1017,17 @@ def test_aaf_composition_metadata_mastermob(self):
...
+    def test_aaf_multiple_timecode_objects(self):
+        """
+        Make sure we can read AAFs with multiple timecode objects of the
+        same start value and length.
+        """
+
+        timeline = otio.adapters.read_from_file(
+            MULTIPLE_TIMECODE_OBJECTS_PATH)
+
+        self.assertTrue(timeline is not None)
```
ssteinbach, line 1029: There is a nice unit test assertion for this that is more direct:
```suggestion
        self.assertIsNotNone(timeline)
```

contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py
```diff
@@ -162,14 +162,40 @@ def _find_timecode_mobs(item):
...
     timecodes = [slot.segment for slot in mob.slots
                  if isinstance(slot.segment, aaf2.components.Timecode)]
 
-    if len(timecodes) == 1:
+    # Only use timecode if we have just one or multiple ones with same
+    # start/length.
+    if timecode_values_are_same(timecodes):
         timecode = timecodes[0]
         timecode_start = timecode.getvalue('Start')
         timecode_length = timecode.getvalue('Length')
```
jminor, line 201: In the exception below (line 205) instead of "Error: mob has more than one timecode slots" it would be better to say "Error: mob has more than one timecode slot with different values".

contrib/opentimelineio_contrib/adapters/tests/test_aaf_adapter.py
```diff
@@ -197,6 +197,11 @@
     "with_usercomments_no_mastermob_usercomments.aaf"
 )
 
+MULTIPLE_TIMECODE_OBJECTS_PATH = os.path.join(
+    SAMPLE_DATA_DIR,
+    "multiple_timecode_objects.aaf"
```
jminor, line 202: Can you triple check to make sure the contents of this AAF are clean to share publicly?

contrib/opentimelineio_contrib/adapters/tests/test_aaf_adapter.py
```diff
@@ -1012,6 +1017,17 @@ def test_aaf_composition_metadata_mastermob(self):
...
         self._verify_user_comments(aaf_metadata, expected_md)
 
+    def test_aaf_multiple_timecode_objects(self):
+        """
+        Make sure we can read AAFs with multiple timecode objects of the
+        same start value and length.
+        """
+
+        timeline = otio.adapters.read_from_file(
+            MULTIPLE_TIMECODE_OBJECTS_PATH)
```
jminor, line 1027: It would be good to check the value of the relevant timecode extracted from the file.

jminor (changes requested):
Looks great. I made a couple of small requests.

contrib/opentimelineio_contrib/adapters/tests/test_aaf_adapter.py
stefanschulze: I removed any notion of production-related info, only things left are Company ("Pixar Animation Studios") and the tool name used ("aafWrapMov"), which I did not bother to change.

contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py
stefanschulze: Makes sense, I'll update that.

contrib/opentimelineio_contrib/adapters/tests/test_aaf_adapter.py
stefanschulze: I don't think that is too relevant for this test-case, but I'll add the timecode-value-check, does not do any harm.

stefanschulze: I added some timecode check

contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py
```diff
@@ -164,9 +164,9 @@ def _find_timecode_mobs(item):
 
 def timecode_values_are_same(timecodes):
     """
-    An AAF can have multiple timecode objects (for example an auxTC24 value
-    that got added via the Avid Bin column). As long as they have the same
-    start and length values, they can be treated as being the same.
+    An SourceClip can have multiple timecode objects (for example an auxTC24
```
meshula, line 167:
```suggestion
    A SourceClip can have multiple timecode objects (for example an auxTC24
```

contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py
```diff
@@ -164,9 +164,9 @@ def _find_timecode_mobs(item):
...
+    value  that got added via the Avid Bin column). As long as they have the
```
meshula, line 168:
```suggestion
    value that got added via the Avid Bin column). As long as they have the
```

meshula (commented):
minor typos

ssteinbach (approved):
Looks good, thanks @stefanschulze

jminor (approved):

PR #993, opened 2021-06-22
add example timeline builder

ssteinbach:
Adds a simple example that builds a timeline from scratch using a list of files and times as input.

- [x] add a command line argument to take a path of where to put the resulting file
- [x] add a more helpful docstring


meshula (approved):
lgtm

PR #995, opened 2021-06-22
Cxx conform example

### LINKED ISSUE #994: C++ version of conform.py example
Now that the C++ children_if()/clip_if() functionality is available it would be nice to have a C++ version of the conform.py example.

darbyjohnston:
```Fixes #994```

This PR adds a C++ version of the conform.py example. The code is mostly identical to the Python version except for the ```argparse``` command line parsing which doesn't have an equivalent in C++.

There are also a couple minor fixes to other example files, I can break those out into a separate PR if necessary. Specifically conform.py had some syntax errors with my Python 3 setup, hopefully the changes I made are also compatible with Python 2. It might be worthwhile to add the examples to the CI at some point...

### DIFF
--- a/examples/conform.cpp
+++ b/examples/conform.cpp
@@ -117,7 +117,7 @@ int conform_timeline(
         // relink to the found path
         clip->set_media_reference(new otio::ExternalReference(
             "file://" + new_path,
-            otio::nullopt // we don't know the available range
+            otio::nullopt // the available range is unknown without opening the file
         ));
         count += 1;
     }
--- a/examples/conform.cpp
+++ b/examples/conform.cpp
@@ -85,6 +85,12 @@ std::string find_matching_media(std::string const& name, std::string const& fold
 // Look for replacement media for each clip in the given timeline.
 //
 // The clips are relinked in place if media with a matching name is found.
+//
+// Note the use of otio::SerializableObject::Retainer to wrap the timeline,
+// it provides a safe way to manage the memory of otio objects by keeping an
+// internal reference count. For more details on the usage of Retainers see
+// the C++ documentation:
+// https://opentimelineio.readthedocs.io/en/latest/cxx/cxx.html
 int conform_timeline(
     otio::SerializableObject::Retainer<otio::Timeline> const& timeline,
     std::string const& folder)
@@ -95,14 +101,14 @@ int conform_timeline(
     const auto clips = timeline->clip_if(&error_status);
     if (error_status)
     {
-        std::cout << error_status.outcome << std::endl;
-        throw error_status;
+        print_error(error_status);
+        exit(1);
     }
     
-    for (auto clip : clips)
+    for (const otio::SerializableObject::Retainer<otio::Clip>& clip : clips)
     {
         // look for a media file that matches the clip's name
-        const auto new_path = find_matching_media(clip->name(), folder);
+        const std::string new_path = find_matching_media(clip->name(), folder);
 
         // if no media is found, keep going
         if (new_path.empty())
@@ -130,33 +136,28 @@ int main(int argc, char** argv)
     const std::string folder = normalize_path(argv[2]);
     const std::string output = normalize_path(argv[3]);
     
-    try
-    {   
-        otio::ErrorStatus error_status;
-        otio::SerializableObject::Retainer<otio::Timeline> timeline(
-            dynamic_cast<otio::Timeline*>(otio::Timeline::from_json_file(input, &error_status)));
-        if (!timeline || error_status)
-        {
-            throw error_status;
-        }
-        const int count = conform_timeline(timeline, folder);
-        std::cout << "Relinked " << count << " clips to new media." << std::endl;
-        if (!timeline.value->to_json_file(output, &error_status))
-        {
-            throw error_status;
-        }
-        const auto clips = timeline->clip_if(&error_status);
-        if (error_status)
-        {
-            throw error_status;
-        }
-        std::cout << "Saved " << output << " with " << clips.size() << " clips." << std::endl;
-    }
-    catch (otio::ErrorStatus const& e)
+    otio::ErrorStatus error_status;
+    otio::SerializableObject::Retainer<otio::Timeline> timeline(
+        dynamic_cast<otio::Timeline*>(otio::Timeline::from_json_file(input, &error_status)));
+    if (!timeline || error_status)
     {
-        print_error(e);
-        return 1;
+        print_error(error_status);
+        exit(1);
     }
+    const int count = conform_timeline(timeline, folder);
+    std::cout << "Relinked " << count << " clips to new media." << std::endl;
+    if (!timeline.value->to_json_file(output, &error_status))
+    {
+        print_error(error_status);
+        exit(1);
+    }
+    const auto clips = timeline->clip_if(&error_status);
+    if (error_status)
+    {
+        print_error(error_status);
+        exit(1);
+    }
+    std::cout << "Saved " << output << " with " << clips.size() << " clips." << std::endl;
     
     return 0;
 }
--- a/examples/conform.cpp
+++ b/examples/conform.cpp
@@ -114,7 +114,7 @@ int conform_timeline(
         if (new_path.empty())
             continue;
 
-        // if we found one, then relink to the new path
+        // relink to the found path
         clip->set_media_reference(new otio::ExternalReference(
             "file://" + new_path,
             otio::nullopt // we don't know the available range
--- a/examples/conform.cpp
+++ b/examples/conform.cpp
@@ -23,7 +23,7 @@
 //
 
 // Example OTIO script that reads a timeline and then relinks clips
-// to movie files found in a given folder, based on matching names.
+// to movie files found in a given folder, based on matching clip names to filenames.
 //
 // Demo:
 //
--- a/examples/conform.cpp
+++ b/examples/conform.cpp
@@ -54,11 +54,11 @@ namespace otio = opentimelineio::OPENTIMELINEIO_VERSION;
 // Look for media with this name in this folder.
 std::string find_matching_media(std::string const& name, std::string const& folder)
 {
-    // In this case we're looking in the filesystem.
-    // In your case, you might want to look in your asset management system
-    // and you might want to use studio-specific metadata in the clip instead
-    // of just the clip name.
-    // Something like this:
+    // This function is an example which searches the file system for matching media.
+    // A real world studio implementation would likely look in an asset management system
+    // and use studio-specific metadata in the clip's metadata dictionary instead
+    // of matching the clip name.
+    // For example:
     // shot = asset_database->find_shot(
     //    otio::any_cast<std::map<std::string, std::string> >(clip->metadata()["mystudio"])["shotID"]);
     // new_media = shot->latest_render("mov");
@@ -161,4 +161,3 @@ int main(int argc, char** argv)
     
     return 0;
 }
-
--- a/examples/conform.cpp
+++ b/examples/conform.cpp
@@ -63,7 +63,7 @@ std::string find_matching_media(std::string const& name, std::string const& fold
     //    otio::any_cast<std::map<std::string, std::string> >(clip->metadata()["mystudio"])["shotID"]);
     // new_media = shot->latest_render("mov");
     
-    const auto matches = glob(folder + "/" + name + ".*");
+    const auto matches = glob(folder, name + ".*");
     
     if (matches.size() == 0)
     {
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: examples/conform.cpp [hunk 1]
```diff
@@ -0,0 +1,163 @@
...
+            name << "', using '" << matches[0] << "'";
+        return matches[0];
+    }
+}
+
+// Look for replacement media for each clip in the given timeline.
+//
+// The clips are relinked in place if media with a matching name is found.
+int conform_timeline(
+    otio::SerializableObject::Retainer<otio::Timeline> const& timeline,
```
meshula, line 89: conform_timeline doesn't do anything that will change the retain count on the Timeline, so passing a Retainer isn't required in this case. Passing Timeline const& might emphasize that point.

OTOH, the Retainer semantics are difficult to build a mental model for, so it might be reasonable in a tutorial to do the unbreakable thing.

Would you mind adding some notes around why a Retainer is used here, to help people using this code in copy-pasta later?

examples/conform.cpp
```diff
@@ -0,0 +1,163 @@
...
+    
+    otio::ErrorStatus error_status;
+    const auto clips = timeline->clip_if(&error_status);
+    if (error_status)
+    {
+        std::cout << error_status.outcome << std::endl;
+        throw error_status;
+    }
+    
+    for (auto clip : clips)
```
meshula, line 102: nice :)

`const auto clips` occults the type ~ could you add a note explaining the type, or use the type explicitly on line 95? In general, our preference is to use auto when the type can be understood by looking at the code. e.g. `auto foo = std::make_shared<Foo>()` tells us the result is a `shared_ptr<Foo>`.

examples/conform.cpp
```diff
@@ -0,0 +1,163 @@
...
+    {
+        // look for a media file that matches the clip's name
+        const auto new_path = find_matching_media(clip->name(), folder);
```
meshula, line 105: here as well, the guideline for auto isn't really met, new_path is not a hairy incomprehensible template, and find_matching_media doesn't tell us in its name if it is returning a `set` or a `std::file_system::path`, so naming the `std::string` would be preferable to auto

examples/conform.cpp
```diff
@@ -0,0 +1,163 @@
...
+    const std::string output = normalize_path(argv[3]);
+    
+    try
+    {   
+        otio::ErrorStatus error_status;
+        otio::SerializableObject::Retainer<otio::Timeline> timeline(
+            dynamic_cast<otio::Timeline*>(otio::Timeline::from_json_file(input, &error_status)));
+        if (!timeline || error_status)
+        {
+            throw error_status;
```
meshula, line 140: I think we'd prefer an exit(error_status) here, or see alternate below

examples/conform.cpp
```diff
@@ -0,0 +1,163 @@
...
+        }
+        const int count = conform_timeline(timeline, folder);
+        std::cout << "Relinked " << count << " clips to new media." << std::endl;
+        if (!timeline.value->to_json_file(output, &error_status))
+        {
+            throw error_status;
```
meshula, line 146: same

examples/conform.cpp
```diff
@@ -0,0 +1,163 @@
...
+        }
+        const auto clips = timeline->clip_if(&error_status);
+        if (error_status)
+        {
+            throw error_status;
```
meshula, line 151: same

examples/conform.cpp
```diff
@@ -0,0 +1,163 @@
...
+    }
+    catch (otio::ErrorStatus const& e)
+    {
+        print_error(e);
+        return 1;
+    }
+    
+    return 0;
+}
+
```
meshula, line 163:
```suggestion
    return 0;
}
catch (std::exception& exc)
{
    std::cerr << "Unexpected termination due to " << exc.what() << std::endl;
    return 1;
}
```

examples/conform.cpp
```diff
@@ -0,0 +1,163 @@
...
+            "file://" + new_path,
+            otio::nullopt // we don't know the available range
+        ));
+        count += 1;
+    }
+    
+    return count;
+}
+
+int main(int argc, char** argv)
```
meshula, line 122:
```suggestion
int main(int argc, char** argv) try
```
What's this try you say? See below

meshula (changes requested):
Thanks for the demo - it's going to be useful.

Some notes about exceptions and auto

examples/conform.cpp
darbyjohnston: I guess I was thinking of Retainer like std::shared_ptr, and that it would be a good example to always use it to point to OTIO objects? I you don't think that's a good idea I'm happy to change it to a regular pointer.

examples/conform.cpp
darbyjohnston: Interesting, I haven't see a try/catch block used like that with main() before. Maybe exit() would be preferable since exceptions aren't used by OTIO?

examples/conform.cpp
meshula: I'm happy to have just a line of comment there as you'd find in a tutorial, just to explain that Retainer manages the internal reference count on the retained object. (It's an intrusive pointer system, rather than a shared pointer).

examples/conform.cpp
meshula: exit(1) is fine/preferable, I'd say. Pedantic aside: This try on main pattern is semantically equivalent to having a try/catch inside main. It's more useful on object constructors, like Foo() try { throw exc(); } catch(exc&) { printf("wut"); } ~~~ weirdly the catch after constructors and destructors automatically rethrows, but it's a chance to do something special after an exception.

darbyjohnston: On another note I find some of the error handling a bit confusing with the C++ API; for example ```otio::Timeline::from_json_file()``` can return an error as either a null timeline or an error status. But could it return a non-null timeline and an error status? If so is the timeline  OK to use and the error is recoverable?

Similar also with ```otio::Timeline::to_json_file()```, it returns a bool and an error status. Could it ever return true but with a non-OK error status?

meshula (approved):

examples/conform.cpp
meshula: The way it's coded, if an error occurred, nullptr is returned. No guarantees are made about ErrorStatus. Since you passed in the original ErrorStatus, it might already have a value; it's not reset by OTIO. So the reasonable check is on the return value, with the expectation that additional new information is available in the error status if the return indicated an error. I notice there is unfortunately no documentation in the header file on these functions.

examples/conform.cpp
```diff
@@ -0,0 +1,164 @@
...
+std::string find_matching_media(std::string const& name, std::string const& folder)
+{
+    // In this case we're looking in the filesystem.
+    // In your case, you might want to look in your asset management system
+    // and you might want to use studio-specific metadata in the clip instead
+    // of just the clip name.
+    // Something like this:
+    // shot = asset_database->find_shot(
+    //    otio::any_cast<std::map<std::string, std::string> >(clip->metadata()["mystudio"])["shotID"]);
+    // new_media = shot->latest_render("mov");
```
ssteinbach, line 64: Light editing

```suggestion
    // This function is an example which searches the file system for matching media.
    // A real world studio implementation would likely look in an asset management system
    // and use studio-specific metadata in the clip's metadata dictionary instead
    // of matching the clip name.
    // For example:
    // shot = asset_database->find_shot(
    //    otio::any_cast<std::map<std::string, std::string> >(clip->metadata()["mystudio"])["shotID"]);
    // new_media = shot->latest_render("mov");
```

examples/conform.cpp
```diff
@@ -0,0 +1,164 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the Apache License with the above modification is
+// distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied. See the Apache License for the specific
+// language governing permissions and limitations under the Apache License.
+//
+
+// Example OTIO script that reads a timeline and then relinks clips
+// to movie files found in a given folder, based on matching names.
```
ssteinbach, line 26:
```suggestion
// to movie files found in a given folder, based on matching clip names to filenames.
```

examples/util.h
```diff
@@ -2,12 +2,20 @@
...
+std::string absolute(std::string const&);
+
 // Create a temporary directory.
 std::string create_temp_dir();
 
+// Get a list of files from a directory.
+std::vector<std::string> glob(std::string const& path, std::string const& pattern);
+
 // Print an error to std::cerr.
 void print_error(opentimelineio::OPENTIMELINEIO_VERSION::ErrorStatus const&);
```
ssteinbach, line 20: I think it might help for locating this code to put it into an `example` or `util` namespace-  I see you left the otio code in the `otio` namespace, which is great, and to make it super obvious to folks who are only scanning the code on GitHub where these helper functions come from, I think throwing them in a namespace as well helps.

examples/conform.cpp
```diff
@@ -0,0 +1,164 @@
...
+// the C++ documentation:
+// https://opentimelineio.readthedocs.io/en/latest/cxx/cxx.html
+int conform_timeline(
+    otio::SerializableObject::Retainer<otio::Timeline> const& timeline,
+    std::string const& folder)
+{
+    int count = 0;
+    
+    otio::ErrorStatus error_status;
+    const auto clips = timeline->clip_if(&error_status);
```
ssteinbach, line 101: This is more a comment for the other PR, but you were proposing adding `child_if` which stops at the first matching child -- I wonder if this should be `clips_if` rather than the singular form to make it clear that it won't stop at the first thing it finds but rather goes and builds an entire list?  What do you think?  @darbyjohnston @meshula

examples/conform.cpp
```diff
@@ -0,0 +1,164 @@
...
+    for (const otio::SerializableObject::Retainer<otio::Clip>& clip : clips)
+    {
+        // look for a media file that matches the clip's name
+        const std::string new_path = find_matching_media(clip->name(), folder);
+
+        // if no media is found, keep going
+        if (new_path.empty())
+            continue;
+
+        // if we found one, then relink to the new path
```
ssteinbach, line 117:
```suggestion
        // relink to the found path
```

examples/conform.cpp
```diff
@@ -0,0 +1,164 @@
...
+        clip->set_media_reference(new otio::ExternalReference(
+            "file://" + new_path,
+            otio::nullopt // we don't know the available range
```
ssteinbach, line 120:
```suggestion
            otio::nullopt // the available range is unknown without opening the file
```

examples/conform.py
```diff
@@ -53,12 +54,14 @@
 
 def parse_args():
     """ parse arguments out of sys.argv """
-    parser = argparse.ArgumentParser(description=__doc__)
+    parser = argparse.ArgumentParser(description=__doc__,
+                                     formatter_class=RawTextHelpFormatter)
```
ssteinbach, line 58: interesting, I usually use the formatter with defaults.  The raw text one preserves the formatting in the file docstring block comment?

examples/conform.py
```diff
@@ -90,7 +93,7 @@ def _find_matching_media(name, folder):
     # new_media = shot.latest_render(format='mov')
 
     matches = glob.glob("{0}/{1}.*".format(folder, name))
-    matches = map(os.path.abspath, matches)
+    matches = list(map(os.path.abspath, matches))
 
     if len(matches) == 0:
```
ssteinbach, line 98:
```suggestion
    if not matches:
```

ssteinbach (changes requested):
This is great, thanks Darby!  I edited some of the wording in the comments and asked a couple questions... since this is an example, just poking it at it while we're stirring it around to see if we can make it a little better for new folks coming into the project.  Let me know what you think and thanks for looking into this!

examples/conform.cpp
darbyjohnston: Looks good, this text was taken from the Python example so I'll make the changes there as well.

examples/util.h
darbyjohnston: That's a good idea, I like ```example``` just so it's obvious this is not part of the actual otio API.

examples/conform.cpp
darbyjohnston: I was thinking the same thing, should I do that as part of the other PR or a new one? It looks like there are about 30 occurences of ```clip_if``` in the otio code.

examples/conform.py
darbyjohnston: Yes, the default formatter was removing all the line breaks.

examples/conform.cpp
ssteinbach: new PR I'd say!

ssteinbach (approved):
lgtm!

ssteinbach:
* Adds a C++ version of the conform.py example

PR #997, opened 2021-06-22
Inline functions

### LINKED ISSUE #996: C++ headers with separate declarations/definitions
Currently a number of the C++ headers have the function declarations combined with the definitions, i.e.:
```
class Foo
{
public:
    void print_hello_world()
     {
         std::cout << "hello world!" << std::endl;
     }
};
```

As a C++ developer it would be nicer to browse the class API if the function definitions were separated, i.e.:
```
class Foo
{
public:
    void print_hello_world();
};

inline void Foo::print_hello_world()
{
    std::cout << "hello world!" << std::endl;
}
```

This example is trivial but with large classes with a lot of implementation code it can obscure the class API. OpenEXR does this fairly consistently:
https://github.com/AcademySoftwareFoundation/openexr/blob/master/src/lib/OpenEXR/ImfChannelList.h

The code changes would be cosmetic only, no functionality would be changed. I'll submit a PR to show what this looks like for a couple of header files, if the changes sound reasonable I can update the rest of them. However if this seems like too large of a change I'm also fine to close this issue.

darbyjohnston:
```Fixes #996```

Currently a number of the C++ headers have the function declarations combined with the definitions, i.e.:
```
class Foo
{
public:
    void print_hello_world()
     {
         std::cout << "hello world!" << std::endl;
     }
};
```

As a C++ developer it would be nicer to browse the class API if the function definitions were separated, i.e.:
```
class Foo
{
public:
    void print_hello_world();
};

inline void Foo::print_hello_world()
{
    std::cout << "hello world!" << std::endl;
}
```

This example is trivial but with large classes with a lot of implementation code it can obscure the class API. OpenEXR does this fairly consistently:
https://github.com/AcademySoftwareFoundation/openexr/blob/master/src/lib/OpenEXR/ImfChannelList.h

The code changes would be cosmetic only, no functionality would be changed. I'll submit a PR to show what this looks like for a couple of header files, if the changes sound reasonable I can update the rest of them. However if this seems like too large of a change I'm also fine to close this issue.

### DIFF
--- a/src/opentime/rationalTime.h
+++ b/src/opentime/rationalTime.h
@@ -16,88 +16,43 @@ enum IsDropFrameRate : int {
     
 class RationalTime {
 public:
-    explicit RationalTime(double value = 0, double rate = 1)
-    : _value {value}, _rate {rate} {}
-
+    explicit RationalTime(double value = 0, double rate = 1);
     RationalTime(RationalTime const&) = default;
     RationalTime& operator= (RationalTime const&) = default;
 
-    bool is_invalid_time() const {
-        if(std::isnan(_rate) || std::isnan(_value)) {
-            return true;
-        }
-
-        return _rate <= 0;
-    }
+    bool is_invalid_time() const;
     
-    double value() const {
-        return _value;
-    }
+    double value() const;
+    double rate() const;
 
-    double rate() const {
-        return _rate;
-    }
+    RationalTime rescaled_to(double new_rate) const;
+    RationalTime rescaled_to(RationalTime rt) const;
 
-    RationalTime rescaled_to(double new_rate) const {
-        return RationalTime {value_rescaled_to(new_rate), new_rate};
-    }
+    double value_rescaled_to(double new_rate) const;
+    double value_rescaled_to(RationalTime rt) const;
 
-    RationalTime rescaled_to(RationalTime rt) const {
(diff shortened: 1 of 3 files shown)


meshula:
This change highlights another issue related to the very notion of inlining.

I hadn't considered this point previously, but reading the diff made me think about what was being split, and what impact it might have on code generation.

Our methods are marked const, but the majority of things marked const should be marked constexpr. Also pretty much every function in these two headers should be marked noexcept. I just added a new issue. https://github.com/PixarAnimationStudios/OpenTimelineIO/issues/999

As a style point, inlines as we have them look fine, and splitting things out as we have them looks fine, so I'll defer the style question to others.

darbyjohnston:
I don't think this change would have an impact on code generation, but that's a good point about noexcept/constexpr, I'd be happy to make a pass at that in a separate PR.

darbyjohnston:
So, does this seem like a good idea? If so I would like to un-draft it, otherwise I'll close it and the corresponding issue:
https://github.com/PixarAnimationStudios/OpenTimelineIO/issues/996

meshula:
This is a stylistic change that I don't feel strongly about one way or another. My bias is that it looks easier to read as a user of the API, and harder to maintain as a developer of the API, mostly because I don't use this pattern myself because it's harder to visually spot missing implementations; not really an issue here, given the relative maturity of the interface.

Perhaps someone else has some thoughts.

PR #998, opened 2021-06-22
Spatial Bounds with Imath

### LINKED ISSUE #771: MediaReference: spatial bounding box
**Is the feature new development, or does it address existing functionality?**

The existing OTIO structure already contains well-defined bounds for temporal metadata, but there is currently no way to express spatial bounds.  When viewing a series of clips, particularly ones of different resolutions, it is useful to be able to define or determine the spatial bounds (ie viewport size) so that all the clips can be contained within it. 

The MediaReference metadata of a clip currently contains a TimeRange that expresses the temporal bounds as the combination of a start_time and a duration property, and each of these are defined as Time structure that is a combination of rate and value properties.  We would like to expand this concept to also express spatial bounds in a similar way.

Ideally, the spatial bounds would be expressed as a Box, which contains three properties: center, width and height.  The width and height are similar to Time's value property in that they are unitless values as to be agnostic of any on particular coordinate system (pixels, cm, etc).  The center is a Point structure that contains x and y properties that represent location of the center of the box in x and y coordinates.  

The combination of the four related values width, height, x and y can then be used to derive the valid coordinate system for the spatial bounds.  For example, a 16x9 box with a center at (0,0) would have x values in the range (-8, 8) and y values in the range (-4.5,4.5).  These values can then be scaled to the actual image size being viewed while remaining independent of the units used to express the size of the image.  Futhermore, since the center of the box is a user defined value, no assumptions about the location of the origin are required either.  For example for a 16x9 Box in a coordinate system where the origin is in the bottom left, the center would be defined as (8,4.5) whereas in a coordinate system where the origin is the center the Box's center would be defined as (0,0). 

**Describe relevant implementation notes**

The Box and Point primitives can be added to the existing opentime library, so they can live along side the Time and TimeRange ones and be de/serialized in a similar fashion.  In addition, the Box should be supported as a template parameter to Optional so that older serialized json files are still valid even if they do not define bounds. 

The MediaReference and its derivates can be modified to optionally take bounds argument defaulting to optional<Box> after their existing constructor arguments.  This way both existing calls to the API and existing serialized json files still remain valid without requiring any client code to change. 

The proposed json additions to the MediaReference schema looks like this:
```
"bounds": {
  "OTIO_SCHEMA": "Box.1"
  "center": {  
     "OTIO_SCHEMA": "Point.1"
     "x": 0.0,
     "y": 0.0
   }
   "width": 16.0,
   "height: 9.0
}
```

As with TimeRange, it is also useful to be able to determine the bounds of all the Clips in a container, so for this reason a bounds function can be added to the Composition and Timeline classes.  When called it should calculate the union of all Boxes contained within any Clip MediaReferences yielding the minimum sized Box required to contain all the Composition's Clips. 

**Additional context**

Slack discussion: https://academysoftwarefdn.slack.com/archives/CMQ9J4BQC/p1596741543292800

rogernelson:
```Fixes #771```
MediaReference: spatial bounding box

**Summary**

This change addresses the need to have spatial bounds expressed in the MediaReference.  These are used to define a measurement-agnostic coordinate system representing the viewing area of the clip(s), which can be used by media players to determine how to display or scale/resize/center (if necessary) the media referenced by the Clips.  

The Imath library is used to implement the spatial bounding box, specifically the Box class and its dependency Vec.  Imath was selected since its already used by OpenExr, another ASWF project, and this helps establish a common standard between the two projects.  Since the 2d template specializations (ie Box2d, a box using 2 dimensional doubles) made the most sense in this context, I only implemented support for it, and its V2d (a 2d vector) dependency.  This, however, does not close the door to implementing support for more dimensions or other vector types if needed in the future.  The Imath types are contained within a Bounds class which derives from SerializableObject.  This encapsulation allows not only the serialization of bounds, but also for the concept to be extended in the future -- for example, support for additional Imath types or for a Transformation type to allow transformations between different coordinate systems. 

For a more in-depth explanation of the idea and motivations please see the issue linked here:
https://github.com/PixarAnimationStudios/OpenTimelineIO/issues/771

I have flagged as many potential discussion points that I think might arise with **Note** below.  

**Files changed**
- .gitmodules
Add Imath as a submodule

- *.md files
Lists the additions to the schemas (automatically generated by Makefiles).

- src/deps/CMakeLists.txt
Include Imath submodule and set the option to build it statically (without overriding the option for the base OTIO project).  We want to build statically to keep the libraries self-contained so they can be made easily into Python wheels.

- src/opentime/CMakeLists.txt
Add new bounds class to compilation, link with static Imath library and remove OpenTimelineIOConfig export.  Remove export of build tree (this is the same issue seen here: https://github.com/PixarAnimationStudios/OpenTimelineIO/pull/911).

- src/opentimelineio/bounds.cpp
Implement the Bounds class.  For the time being it contains only an optional Imath::Box2d.  The box is optional to avoid selecting default values for the Box.  Since its optional we can simply have no box as the default.

- src/opentimelineio/bounds.h
Add mutators and accessors to the public interface.
**Python Bindings**: Also added

- src/opentimelineio/clip.cpp
Implement the bounds method.  Simply returns the Bounds if it has been set on the clip or an error otherwise (conforming to the standard of the available_range method).

- src/opentimelineio/clip.h
Adds the bounds method to the public interface.
**Python Bindings**: Also added

- src/opentimelineio/composable.cpp
Implements the bounds method to return an error so that any derived classes that do not explicitly override the virtual method return the error (it is overriden by Clip, Stack and Track).  

- src/opentimelineio/composable.h
Adds the bounds method to the public interface
**Python Bindings**: Also added

- src/opentimelineio/composition.cpp
Implements the has_clips method for composition, which returns true if the Composition contains at least one clip.  This function is used by bounds method of the derived Stack class to exclude any tracks that do not contain any clips when it calculates the union of the bounds.  Since the bounds default to width, height and center of {0, 0, (0, 0)} we want to prevent the default values from affecting the result and thus use the has_clips method to skip over any tracks that do not contain clips. 

- src/opentimelineio/composition.h
Adds the has_clips method to the public interface. 
**Python Bindings**: Also added
**Note**: Even though this method is only needed by Stack, I thought it could be useful elsewhere so I made it public.  If you would prefer I could make it protected so only Stack could access it.  Or if you do not want it in the interface at all, I can also implement it as a lambda function (or in the anonymous namespace) inside of Stack::bounds.

- src/opentimelineio/deserialization.cpp
Implements the de/serialization of Imath::Box2d Imath::V2d and Bounds, including optional support for Imath::Box2d.  The code for serializing the Imath types follows the existing standard provided by Time and TimeRange.

- src/opentimelineio/errorStatus.cpp
Adds a new error value for when the bounds cannot be determined.  This occurs when the user asks for bounds on a Clip that does not contain bounds or asks for bounds on a Composition that has a Clip that does not contain bounds.

- src/opentimelineio/externalReference.cpp, src/opentimelineio/generatorReference.cpp, src/opentimelineio/imageSequenceReference.cpp, src/opentimelineio/missingReference.h
Pass bounds constructor argument to MediaReference parent constructor 

- src/opentimelineio/externalReference.h, src/opentimelineio/generatorReference.h, src/opentimelineio/imageSequenceReference.h, src/opentimelineio/missingReference.h
Adds optional constructor argument for bounds.

- src/opentimelineio/mediaReference.cpp
Initializes bounds member and add it to the read/write method for serialization.  A Retainer is used to store Bounds so the memory can be managed in the Python bindings similar to the Effect class.

- src/opentimelineio/mediaReference.h
Add bounds member and constructor argument as well as a mutator and accessor method.
**Python Bindings**: Also added (includes derived classes)

- src/opentimelineio/safely_typed_any.cpp
Adds casting support for Box2d  and V2d (required by serialization code).

- src/opentimelineio/safely_typed_any.h
Adds casting methods for Box2d and V2d to public interface.

- src/opentimelineio/serializableObject.h
Adds de/serialization support for Box2d and V2d found within the Bounds schema.

- src/opentimelineio/serialization.cpp
Adds de/serialization code for Box2d and V2d.

- src/opentimelineio/stack.cpp
Implements bounds method for a stack.  We begin by finding the first bounding box, defined as the Box of the first Clip or of the first Composition that contains at least one Clip. After this is found, we then extend the box to contain each subsequent Clip or Composition containing at least one Clip.  If no Clips are found or no Boxes are found, we return nullptr.  If an error is found, we return the error (as a parameter). This follows the standard defined by the available_range method.

- src/opentimelineio/stack.h
Adds bounds method to the public interface.
**Python Bindings**: Also added

- src/opentimelineio/timeline.h
Adds bounds to the public interface and implements it by redirecting the call to it's Stack.
**Python Bindings**: Also added

- src/opentimelineio/track.cpp
Implements bounds method for a track.  Similarly to Stack, we begin by finding the first bounding box, however in this case we don't need to consider Compositions, only Clips.  Once we have the bounding Box of the first clip, we extend it with the Box of each subsequent Clip.  If no Clips or Boxes are found, we return nullptr.  If an error is found, we return the error (as a parameter) and the default Box. This follows the standard defined by the available_range method.

- src/opentimelineio/track.h
Adds bounds method to the public interface.
**Python Bindings**: Also added

- src/opentimelineio/typeRegistry.cpp
Register Bounds as a SerializableObject.

- src/py-opentimelineio/opentimelineio-bindings/CMakeLists.txt
Adds compilation of Python bindings for Imath types.

- src/py-opentimelineio/opentimelineio-bindings/otio_bindings.cpp
Initialize Imath bindings

- src/py-opentimelineio/opentimelineio-bindings/otio_bindings.h
Imath bindings function prototype 

- src/py-opentimelineio/opentimelineio-bindings/otio_imath.cpp
Create pybind bindings for Imath types.
**Note**: I did not use the existing Python bindings for Imath here because they are done using boost.  Using them would create a dependency on boost that we want to avoid.  We don't want to break the self-contained nature of the Python binding libraries by depending on a dynamic system library.  Because of this if we want to use the Imath types in the Python OTIO library with the Imath Python library, we would need to create some kind bridge code. 

- src/py-opentimelineio/opentimelineio-bindings/otio_serializableObjects.cpp
Adds all added opentimelineio public c++ interface methods to the Python bindings
**Note**: These are denoted by **Python Bindings** in descriptions above.

- src/py-opentimelineio/opentimelineio/core/mediaReference.py
Adds bounds to the __repr__ and __str__ method for MediaReference.

- src/py-opentimelineio/opentimelineio/schema/__init__.py 
Import schemas for Bounds, Box2d and V2d.

- src/py-opentimelineio/opentimelineio/schema/bounds.py
Create a Bounds schema

- src/py-opentimelineio/opentimelineio/schema/box2d.py
Create a Box schema

- src/py-opentimelineio/opentimelineio/schema/image_sequence_reference.py
Adds bounds to the __repr__ and __str__ method for ImageSequence.

- src/py-opentimelineio/opentimelineio/schema/v2d.py
Create a Vector schema

**Tests**

- tests/baselines/*.json
Adds empty bounds to the json representations
**Note**: This is necessary even if the bounds don't exist in the metadata since the serialization code will add a null bounds member in this case.  If you'd prefer I could make it exclude the bounds member entirely if it is not defined.

- tests/sample_data/clip_example.otio
Added bounds to the example clip and associated test (test_documentation.py).

- tests/test_bounds.py
Test all the public bounds methods.  This is essentially just the schemas and getting and setting the box.

- tests/test_box2d.py 
Test all public Box2d methods. 
**Note**:  I did test each method but only on a basic level to ensure it was bound to the correct underlying C++ method. Imath has its own more rigorous unit testing to make sure the methods are working correctly.

- tests/test_clip.py
Updated the str test to include the bounds.  Also adds a bounding box to a clip and tests that they were set correctly using the contains methods to find the edges of the Box.

- tests/test_composition.py
Tests the added has_clips method on empty stacks and tracks and ones containing no Clips (only Gaps).  Also tests that the bounds are correctly applied to empty Stacks, single clip Stacks, multi-clip Stacks, Stacks with gaps, and Stacks containing Tracks containing Clips. Tracks are also tested for empty Tracks, gap-only Tracks, single Clip Tracks and multi-Clip Tracks.

- tests/test_media_reference.py
Tests for new bounds in schema

- tests/test_v2d.py
Test all public V2d methods. 
**Note**:  I did test each method but only on a basic level to ensure it was bound to the correct underlying C++ method. Imath has its own more rigorous unit testing to make sure the methods are working correctly.

- tests/test_generator_reference.py, tests/test_image_sequence_reference.py, tests/test_media_reference.py
Ensures the bounds can be passed to the constructor and the same bounds are returned from the accessor.
ImageSequenceReference and MediaRefererence also test the __str__ and __repr__ methods return the expected string.

- tests/test_timeline.py
Timeline bounds are tested to ensure a single track timeline has equal bounds between the Track and the Timeline.

### DIFF
--- a/src/opentimelineio/composition.cpp
+++ b/src/opentimelineio/composition.cpp
@@ -1,5 +1,6 @@
 #include "opentimelineio/composition.h"
 #include "opentimelineio/vectorIndexing.h"
+#include "opentimelineio/clip.h"
 
 #include <assert.h>
 #include <set>
@@ -496,6 +497,20 @@ int64_t Composition::_bisect_left(
     }
 
     return *lower_search_bound;
+}
+
+bool Composition::has_clips() const {
+    for (auto child: children()) {
+        if (dynamic_cast<Clip*>(child.value)) {
+            return true;
+        }
+        else if (auto child_comp = dynamic_cast<Composition*>(child.value)) {
+            if( child_comp->has_clips() ) {
+               return true;
+            }
+        }
+    }
+    return false;
 }
 
 } }
--- a/src/opentimelineio/stack.cpp
+++ b/src/opentimelineio/stack.cpp
@@ -93,4 +93,45 @@ std::vector<SerializableObject::Retainer<Clip>> Stack::clip_if(
     return children_if<Clip>(error_status, search_range, shallow_search);
 }
 
+SerializableObject::Retainer<Bounds> 
+Stack::bounds(ErrorStatus* error_status) const {
+    optional<Imath::Box2d> box;
+    bool found_first_child = false;
+    for (auto child: children()) {
+        optional<Imath::Box2d> child_box;
+        if (auto comp = dynamic_cast<Composition*>(child.value)) {
+            if (!comp->has_clips()) {
+                continue;
+            }
+            if (auto child_bounds = comp->bounds(error_status)) {
+               
+                child_box = (*child_bounds).box();
+            }
+        }
+        else if (auto clip = dynamic_cast<Clip*>(child.value)) {
+            if (auto child_bounds = clip->bounds(error_status)) {
+                child_box = (*child_bounds).box();
+            }
+        }
+        else {
+           continue;
+        }
+
+        if (*error_status) {
+           return Retainer<Bounds>();
+        }
+
+        if ( child_box ) {
+            if (found_first_child) {
+                box->extendBy(*child_box);
+            }
+            else {
+                box = child_box;
+                found_first_child = true;
+            }          
+        }
+    }
+    return Retainer<Bounds>( box ? new Bounds( std::string(), *box ) : nullptr );
+}
+
 } }
--- a/src/opentimelineio/track.cpp
+++ b/src/opentimelineio/track.cpp
@@ -1,6 +1,7 @@
 #include "opentimelineio/track.h"
 #include "opentimelineio/clip.h"
 #include "opentimelineio/transition.h"
+#include "opentimelineio/clip.h"
 #include "opentimelineio/gap.h"
 #include "opentimelineio/vectorIndexing.h"
 
@@ -216,6 +217,32 @@ std::vector<SerializableObject::Retainer<Clip>> Track::clip_if(
     bool shallow_search) const
 {
     return children_if<Clip>(error_status, search_range, shallow_search);
+}
+
+SerializableObject::Retainer<Bounds> 
+Track::bounds(ErrorStatus* error_status) const {
+    optional<Imath::Box2d> box;
+    bool found_first_clip = false;
+    for (auto child: children()) {
+        if (auto clip = dynamic_cast<Clip*>(child.value)) {
+            if (auto clip_bounds = clip->bounds(error_status)) {
+                auto clip_box = (*clip_bounds).box();
+                if (clip_box) {
+                    if (found_first_clip) {
+                        box->extendBy(*clip_box);
+                    }
+                    else {
+                        box = clip_box;
+                        found_first_clip = true;
+                    }
+                }
+            }
+            if (*error_status) {
+                return Retainer<Bounds>();
+            }
+        }
+    }
+    return Retainer<Bounds>( box ? new Bounds( std::string(), *box ) : nullptr );
 }
 
 } }
--- a/.gitmodules
+++ b/.gitmodules
@@ -10,3 +10,6 @@
 [submodule "src/deps/any"]
 	path = src/deps/any
 	url = https://github.com/thelink2012/any.git
+[submodule "src/deps/Imath"]
+	path = src/deps/Imath
+	url = https://github.com/AcademySoftwareFoundation/Imath
--- a/docs/tutorials/otio-serialized-schema-only-fields.md
+++ b/docs/tutorials/otio-serialized-schema-only-fields.md
@@ -60,6 +60,7 @@ parameters:
 
 parameters:
 - *available_range*
+- *bounds*
 - *metadata*
 - *name*
 
@@ -128,6 +129,13 @@ parameters:
 
 ## Module: opentimelineio.schema
 
+### Bounds.1
+
+parameters:
+- *box*
+- *metadata*
+- *name*
+
 ### Clip.1
 
 parameters:
@@ -149,6 +157,7 @@ parameters:
 
 parameters:
 - *available_range*
+- *bounds*
 - *metadata*
 - *name*
 - *target_url*
@@ -174,6 +183,7 @@ parameters:
 
 parameters:
 - *available_range*
+- *bounds*
 - *generator_kind*
 - *metadata*
 - *name*
@@ -183,6 +193,7 @@ parameters:
 
 parameters:
 - *available_range*
+- *bounds*
 - *frame_step*
 - *frame_zero_padding*
 - *metadata*
@@ -214,6 +225,7 @@ parameters:
 
 parameters:
 - *available_range*
+- *bounds*
 - *metadata*
 - *name*
 
--- a/docs/tutorials/otio-serialized-schema.md
+++ b/docs/tutorials/otio-serialized-schema.md
@@ -114,6 +114,7 @@ None
 
 parameters:
 - *available_range*: 
+- *bounds*: 
 - *metadata*: 
 - *name*: 
 
@@ -257,6 +258,21 @@ parameters:
 
 ## Module: opentimelineio.schema
 
+### Bounds.1
+
+*full module path*: `opentimelineio.schema.Bounds`
+
+*documentation*:
+
+```
+None
+```
+
+parameters:
+- *box*: 
+- *metadata*: 
+- *name*: 
+
 ### Clip.1
 
 *full module path*: `opentimelineio.schema.Clip`
@@ -302,6 +318,7 @@ None
 
 parameters:
 - *available_range*: 
+- *bounds*: 
 - *metadata*: 
 - *name*: 
 - *target_url*: 
@@ -351,6 +368,7 @@ None
 
 parameters:
 - *available_range*: 
+- *bounds*: 
 - *generator_kind*: 
 - *metadata*: 
 - *name*: 
@@ -428,6 +446,7 @@ Negative ``start_frame`` is also handled. The above example with a ``start_frame
 
 parameters:
 - *available_range*: 
+- *bounds*: 
 - *frame_step*: Step between frame numbers in file names.
 - *frame_zero_padding*: Number of digits to pad zeros out to in frame numbers.
 - *metadata*: 
@@ -483,6 +502,7 @@ None
 
 parameters:
 - *available_range*: 
+- *bounds*: 
 - *metadata*: 
 - *name*: 
 
--- a/src/deps/CMakeLists.txt
+++ b/src/deps/CMakeLists.txt
@@ -1,6 +1,6 @@
 
 # detect if the submodules haven't been updated
-set(DEPS_SUBMODULES any optional-lite pybind11 rapidjson)
+set(DEPS_SUBMODULES any optional-lite pybind11 rapidjson Imath)
 foreach(submodule IN LISTS DEPS_SUBMODULES)
     file(GLOB SUBMOD_CONTENTS ${submodule})
     list(LENGTH SUBMOD_CONTENTS SUBMOD_CONTENT_LEN)
@@ -22,3 +22,15 @@ if(OTIO_CXX_INSTALL AND OTIO_DEPENDENCIES_INSTALL)
     install(FILES optional-lite/include/nonstd/optional.hpp
     	    DESTINATION "${OTIO_RESOLVED_CXX_INSTALL_DIR}/include/opentimelineio/deps/nonstd")
 endif()
+
+# preserve BUILD_SHARED_LIBS options for this project, but set it off for Imath
+option(BUILD_SHARED_LIBS "Build shared libraries" ON)
+set(BUILD_SHARED_LIBS OFF)
+
+# If we do not want Imath to install headers and CMake files use the EXCLUDE_FROM_ALL option
+if(OTIO_CXX_INSTALL AND OTIO_DEPENDENCIES_INSTALL)
+  add_subdirectory(Imath)
+else()
+  add_subdirectory(Imath EXCLUDE_FROM_ALL)
+endif()
+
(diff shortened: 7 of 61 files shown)

### CODE REVIEW

File: src/opentimelineio/composition.cpp [hunk 1]
```diff
@@ -498,4 +499,18 @@ int64_t Composition::_bisect_left(
     return *lower_search_bound;
 }
 
+bool Composition::has_clips() const {
+    for (auto child: children()) {
```
darbyjohnston, line 503: I recently added some functions for working with child objects, you could also write this as:
```
ErrorStatus error_status;
return children_if<Clip>(&error_status).size() > 0;
```
```

File: src/opentimelineio/stack.cpp [hunk 2]
```diff
@@ -93,4 +93,45 @@ std::vector<SerializableObject::Retainer<Clip>> Stack::clip_if(
     return children_if<Clip>(error_status, search_range, shallow_search);
 }
 
+SerializableObject::Retainer<Bounds> 
+Stack::bounds(ErrorStatus* error_status) const {
+    optional<Imath::Box2d> box;
+    bool found_first_child = false;
+    for (auto child: children()) {
```
darbyjohnston, line 100: You could potentially also use the ```children_if()``` function here, like:
```
for (auto clip : children_if<Clip>(error_status))
{
    optional<Imath::Box2d> child_box;
    if (auto child_bounds = clip->bounds(error_status)) {
        child_box = (*child_bounds).box();
    }
    if (*error_status) {
        return Retainer<Bounds>();
    }
    if ( child_box ) {
        if (found_first_child) {
            box->extendBy(*child_box);
        }
        else {
            box = child_box;
            found_first_child = true;
        }          
    }
}
```

src/opentimelineio/composition.cpp
rogernelson: That's nice! I noticed there were some new changes to the cpp when I rebased. I don't suppose there's a way to short-circuit the children_if call so it stops on the first match? That would be the advantage of keeping this as-is, but I admit it's not nice to have duplicate functionality.

src/opentimelineio/stack.cpp
rogernelson: I like this.  Will do.

src/opentimelineio/track.cpp
```diff
@@ -218,4 +219,30 @@ std::vector<SerializableObject::Retainer<Clip>> Track::clip_if(
     return children_if<Clip>(error_status, search_range, shallow_search);
 }
 
+SerializableObject::Retainer<Bounds> 
+Track::bounds(ErrorStatus* error_status) const {
+    optional<Imath::Box2d> box;
+    bool found_first_clip = false;
+    for (auto child: children()) {
```
darbyjohnston, line 226: Also here; ```children_if()``` has a flag to do a shallow search if you don't need it to recurse.

src/opentimelineio/composition.cpp
darbyjohnston: Good point, the short-circuit is nice to have.

rogernelson: I could perhaps add a `children_has<Clip>` to fit more with the existing pattern, that would do the short circuit.

darbyjohnston: The template seems nice, that way you could also check for the existence of any transitions, etc. I wonder if it would be useful to return the instance instead of a bool?

rogernelson: Sure.  I think it makes sense to do something like `optional<T> children_first<T>` that would return the first instance of T found.  In my case I simply need to check whether the optional is empty.

darbyjohnston: I guess I was thinking of something more like:
```
template<typename T>
Retainer<T> child_if(
    ErrorStatus* error_status,
    optional<TimeRange> search_range = nullopt) const;
```
To match the ```children_if()``` function; that way you could just write:
```
ErrorStatus error_status;
return child_if<Clip>(&error_status);
```
If this seems like too much for this PR I would also be happy to add it as a separate PR since I implemented the other functions.

rogernelson: I'm good with that too.  Because of the CMake issues on Windows, I'm going to have to put it on hold for a bit until I can fix that. If you do it in the meantime, I'll update and use it.  If not, I'll do it as part of this PR.  So essentially I'll leave it up to you :)

darbyjohnston: I created a draft PR for the child_if() functions:
https://github.com/PixarAnimationStudios/OpenTimelineIO/pull/1002

By making it a general function and supporting the various otio container types it turned out to be more code changes than I was hoping for...

meshula:
This is a thorough PR, all the details are well covered. I like the pybind11 wrapping of the Imath classes, I look forward to a future revision of Imath itself using pybind11!

As an overall first remark, it's not clear to me that the bounds should itself be SerializableObject, rather than simply a quality of an Item. I have been personally thinking about bounds as a plain-old-data quality of an Item, most analogous to TimeRange. My inclination is to simplify things slightly by treating Bounds as a subclass of Box2D, if it needs special methods, or simply a type alias if Box2D already provides needed functionality. As a further modification, like TimeRange, I think a POD Bounds would be pass by value, or const-reference, which would be nice in terms of just treating it like a value, rather than a fancy boxed and reference counted thing.

rogernelson:
> This is a thorough PR, all the details are well covered. I like the pybind11 wrapping of the Imath classes, I look forward to a future revision of Imath itself using pybind11!

That would be really nice.  It would have saved me quite a bit of time, but the best part would be that we wouldn't need any 'bridge' code to go from a Python OTIO Imath objects to Imath objects.

> As an overall first remark, it's not clear to me that the bounds should itself be SerializableObject, rather than simply a quality of an Item. I have been personally thinking about bounds as a plain-old-data quality of an Item, most analogous to TimeRange. My inclination is to simplify things slightly by treating Bounds as a subclass of Box2D, if it needs special methods, or simply a type alias if Box2D already provides needed functionality. As a further modification, like TimeRange, I think a POD Bounds would be pass by value, or const-reference, which would be nice in terms of just treating it like a value, rather than a fancy boxed and reference counted thing.

I can definitely see the appeal of using value semantics with Bounds, and I am open to changing if thats the consensus.  The reason I used a SerializableObject was to make it easily adaptable to the other box template specializations, I didn't necessarily want to lock people into the Box2d.  I thought by having the Box in the Bounds container, the box could be changed to the `any` type if necessary and then getters/setters that cast from/to the requested type of Box could be added.   The other reason I had the SerializableObject is that I thought that custom metadata might be necessary in some cases, like maybe for transformations.  

If we're okay with tying the bounds to a Box2d specifically and adding any additional metadata that is required as part of the Item containing the bounds (instead of directly in the bounds themselves) I would be happy to simplify it to a POD type.

rogernelson:
Closing since I re-submitted it to the spatial_coordinates branch.

PR #1000, opened 2021-06-23
Python build: Use CMake FindPython

### LINKED ISSUE #987: Resolve installation into Blender
This issue is to track building OTIO for installation into Blender, and also the creation of wheels that Blender can consume.

JeanChristopheMorinPerso:
Fixes #987.

**Summary**

Use CMake native `FindPython` to get the Python interpreter and the Python headers needed to build OTIO. `pybind11` calls that the "FindPython mode", see https://pybind11.readthedocs.io/en/stable/compiling.html#findpython-mode. Note that our `setup.py` will force CMake to use the Python interpreter that was used to execute the script.

The advantage over the current approach is that we now use [FindPython](https://cmake.org/cmake/help/latest/module/FindPython.html#module:FindPython) module instead if [FindPythonIterp](https://cmake.org/cmake/help/latest/module/FindPythonInterp.html)/[FindPythonLibs](https://cmake.org/cmake/help/latest/module/FindPythonLibs.html) modules that pybind11 uses by default. `FindPython` was introduced in CMake to fix `FindPythonInterp` and `FindPythonLibs` modules. This basically allow us to support a lot more corner cases and scenarios now.

Now, why not use the [new FindPython mode (`-DPYBIND11_FINDPYTHON=ON`)](https://pybind11.readthedocs.io/en/stable/cmake/index.html?highlight=PYBIND11_FINDPYTHON#new-findpython-mode) of pybind11? This mode uses the new `FindPython` module and should technically work. But... it's not working for our use case. The thing with `PYBIND11_FINDPYTHON` is that it calls `find_package(Python REQUIRED COMPONENTS Interpreter Development)` (note the `Development` component that is requested). The Development component requires the `libpython` library to be present. The problem is that it's not always present (for example it's not present in the `manylinux` docker images that are used to build the Linux wheels). Also we don't link against the python library, so there is no need to require it.

If someone wants to build with cmake directly without going through the `setup.py`, then Python interpreter can be specified using `-DPython_EXECUTABLE` or it can also be prepended to the `PATH` environment variable.

I've tested these cases:

* `manylinux` docker images. (Linux only)
* Executing `/path/to/python3.7 setup.py bdist_wheel`, make sure it's not in `PATH` and also make sure a more recent version of python (for example 3.9) is present in `PATH`. (All platforms)
* Executing `/path/to/python3.7 -m pip install . --user`, make sure it's not in `PATH` and also make sure a more recent version of python (for example 3.9) is present in `PATH`. (All platforms)
* Executing `/path/to/python3.7 -m pip wheel .`, make sure it's not in `PATH` and also make sure a more recent version of python (for example 3.9) is present in `PATH`. (All platforms)
* Execute `python setup.py bdist_wheel` from inside a virtualenv. (All platforms)
  Making the assumption that if this works, `pip install .`, `python setup.py bdist_whell`, etc will also work.
* Try with Blender on Windows (just to confirm that my previous tests still work)
* Test with RV's Python (tested by @rogernelson, see https://github.com/PixarAnimationStudios/OpenTimelineIO/issues/987#issuecomment-860692337)
* Build without using the `setup.py` file (so calling cmake manually).

I welcome everyone to test this PR in your setup using your preferred workflow just to make sure I'm not missing something obvious.

Also, I hopefully added enough documentation so that we don't have to redo as much investigation as I had to do. But let me know if you feel more is needed and I'll happily add more.


CMakeLists.txt
```diff
@@ -41,6 +37,17 @@ option(OTIO_AUTOMATIC_SUBMODULES "Fetch submodules automatically" ON)
 # Set option dependent variables
 
 if(OTIO_PYTHON_INSTALL)
+    # Find the python interpreter using FindPython introduced in CMake 3.12.
+    # The Development.Module component is used instead of the Development one
```
meshula, line 41:
```suggestion
    # The Development.Module component introduced in CMake 3.18.2 is used instead of the Development one
```
I suggest documenting this point, if I've understood it properly, since you changed the minimum version to 3.18.2, but mention 3.12 in the comment here.

CMakeLists.txt
```diff
@@ -41,6 +37,17 @@ option(OTIO_AUTOMATIC_SUBMODULES "Fetch submodules automatically" ON)
...
+    # because we don't need the libpython library since we don't link against it.
+    # pybind11 will detect that we already found a python
+    # interpreter and will use what was found previously (here).
+    # See https://pybind11.readthedocs.io/en/stable/compiling.html#findpython-mode
+    # Also, the "manylinux" docker images don't have the libpythons anyway.
```
meshula, line 46:
```suggestion
    # Also, the "manylinux" docker images don't have libpython, so requiring the libraries is an issue there.
```
This suggestion is so that in the future, people don't try to turn it back on.

meshula (commented):
This looks good to me. If we can get a few more eyes on this, I think it's mergeable.

setup.py
```diff
@@ -108,7 +108,12 @@ def cmake_generate(self):
             install_dir += os.path.sep
 
         cmake_args = [
-            '-DPYTHON_EXECUTABLE=' + sys.executable,
+            # Python_EXECUTABLE is important as it tells CMake's FindPython
+            # which Python executable to use. We absolutely want to use the
+            # interpreter that was used to execute the setup.py.
+            # See https://cmake.org/cmake/help/v3.20/module/FindPython.html#artifacts-specification
```
meshula, line 114: The linter is failing the compile checks because this line is too long. Is there an escape to allow long lines to evade the linter?

CMakeLists.txt
JeanChristopheMorinPerso: Actually, `Development.Module` was introduced in 3.18.0 but pybind11 recommends 3.18.2+. I'll fix that and document that pybind11 recommends 3.18.2+.

CMakeLists.txt
JeanChristopheMorinPerso: Perfect, will change that!

setup.py
JeanChristopheMorinPerso: Yep, `# noqa: E501` at the end of the line should do it.

meshula: Thanks!

CMakeLists.txt
JeanChristopheMorinPerso: @meshula I adapted the comment and updated the comment for the `cmake_minimum_required`, which should hopefully be enough. Let me know if it's clearer.

reinecke:
I'll carve some time out this afternoon to try it in a few of my use scenarios and report back.

meshula (approved):

CMakeLists.txt
```diff
@@ -41,6 +39,17 @@ option(OTIO_AUTOMATIC_SUBMODULES "Fetch submodules automatically" ON)
...
 
 if(OTIO_PYTHON_INSTALL)
+    # Find the python interpreter using FindPython module.
+    # The Development.Module component is used instead of the Development one
+    # because we don't need the libpython library since we don't link against it.
+    # pybind11 will detect that we already found a python
+    # interpreter and will use what was found previously (here).
+    # See https://pybind11.readthedocs.io/en/stable/compiling.html#findpython-mode
+    # Also, the "manylinux" docker images don't have libpython, so requiring the libraries is an issue there.
+    find_package(Python REQUIRED COMPONENTS Interpreter Development.Module)
```
ssteinbach, line 49: This block is in the section that runs regardless of whether the user has specified a specific python installation or not, is that correct?  I ask because the original line that was deleted was only in the branch where the user had not specified one.

ssteinbach (commented):
generally LGTM, just wanted to ask a question about the code moving.

CMakeLists.txt
JeanChristopheMorinPerso: Exactly. The new behaviour is similar to the old one.

Before:
* If `OTIO_PYTHON_INSTALL_DIR` and `CMAKE_INSTALL_PREFIX` are not set, use `find_program(Python...)`
* Else don't call it explicitly and let pybind11 do the work of finding the interpreter later in the CMake file.

Now:
* No matter what happens, always use `find_program`. This should not change the previous behaviour as the `Python_*` variables created by `find_program(Python...)` are not used in the `else` block at lines 77-96.

Does that clarify things a little bit more @ssteinbach ?

ssteinbach: Yup, thanks!

reinecke:
I tried both running the development build (`python setup.py install`) and downloaded the wheels to test in our linux runtime and my mac dev machine (both python 3.8) and everything looked great!

Nice fix!

reinecke (changes requested):
One very small note, setup.py version check and messaging just needs to be brought up to date with the new cmake version requirement: https://github.com/PixarAnimationStudios/OpenTimelineIO/blob/d33bf24ee7e7f639edcb408f0a97365254671aec/setup.py#L46-L63

JeanChristopheMorinPerso:
@reinecke Isn't it redundant to check version in both `setup.py` and `CMakeLists.txt`? I have a feeling that the `setup.py` should just make sure that `cmake` is available and then let cmake handle the minimum requirement... I'm guessing the version check in `setup.py` is to have a more user friendly error message?

meshula:
yes, we added the check in setup.py only because the issue is not immediately obvious if we rely on cmake to report the problem via setup.py (it's submerged in a log that makes it hard to spot the error)

reinecke:
@JeanChristopheMorinPerso I took a look at this and came up with a diff that would run a "Pre-flight" cmake check that outputs a human-friendly error while deferring to CMake to actually check and fail:
```diff
--- a/setup.py
+++ b/setup.py
@@ -42,27 +42,6 @@ if sys.version_info[0] < 3:
     )


-def cmake_version_check():
-    if platform.system() == "Windows":
-        required_minimum_version = '3.17.0'
-    else:
-        required_minimum_version = '3.12.0'
-
-    try:
-        out = subprocess.check_output(['cmake', '--version'])
-        cmake_version = LooseVersion(
-            re.search(r'version\s*([\d.]+)', out.decode()).group(1)
-        )
-        if cmake_version < required_minimum_version:
-            raise RuntimeError("CMake >= " + required_minimum_version +
-                               " is required to build OpenTimelineIO's runtime"
-                               " components")
-    except OSError:
-        if platform.system() == "Windows":
-            raise RuntimeError("CMake >= 3.17.0 is required")
-        raise RuntimeError("CMake >= 3.12.0 is required")
-
-
 def _debugInstance(x):
     for a in sorted(dir(x)):
         print("%s:     %s" % (a, getattr(x, a)))
@@ -97,9 +76,21 @@ class OTIO_build_ext(setuptools.command.build_ext.build_ext):
         debug = (self.debug or bool(os.environ.get("OTIO_CXX_DEBUG_BUILD")))
         self.build_config = ('Debug' if debug else 'Release')

+        self.cmake_preflight_check()
         self.cmake_generate()
         self.cmake_install()

+
+    def cmake_preflight_check(self):
+        proc = subprocess.Popen(
+            ["cmake", "--check-system-vars", SOURCE_DIR],
+            stderr=subprocess.PIPE,
+            cwd=self.build_temp_dir,
+        )
+        _, stderr = proc.communicate()
+        if proc.returncode != 0:
+            raise RuntimeError(stderr.strip())
+
     def cmake_generate(self):
         # Use the provided build dir so setuptools will be able to locate and
         # either install to the correct location or package.
@@ -249,9 +240,6 @@ def test_otio():
     return unittest.TestLoader().discover('tests')


-cmake_version_check()
-
-
 # copied from first paragraph of README.md
 LONG_DESCRIPTION = """OpenTimelineIO is an interchange format and API for
 editorial cut information. OTIO is not a container format for media, rather it
 ```
 
 Here is an example of the failure output:
 ```
 -- Configuring incomplete, errors occurred!
  File "setup.py", line 378, in <module>
    **PROJECT_METADATA
  File "/opt/python/cp27-cp27m/lib/python2.7/site-packages/setuptools/__init__.py", line 162, in setup
    return distutils.core.setup(**attrs)
  File "/opt/python/cp27-cp27m/lib/python2.7/distutils/core.py", line 151, in setup
    dist.run_commands()
  File "/opt/python/cp27-cp27m/lib/python2.7/distutils/dist.py", line 953, in run_commands
    self.run_command(cmd)
  File "/opt/python/cp27-cp27m/lib/python2.7/distutils/dist.py", line 972, in run_command
    cmd_obj.run()
  File "/opt/python/cp27-cp27m/lib/python2.7/site-packages/setuptools/command/install.py", line 67, in run
    self.do_egg_install()
  File "/opt/python/cp27-cp27m/lib/python2.7/site-packages/setuptools/command/install.py", line 109, in do_egg_install
    self.run_command('bdist_egg')
  File "/opt/python/cp27-cp27m/lib/python2.7/distutils/cmd.py", line 326, in run_command
    self.distribution.run_command(command)
  File "/opt/python/cp27-cp27m/lib/python2.7/distutils/dist.py", line 972, in run_command
    cmd_obj.run()
  File "/opt/python/cp27-cp27m/lib/python2.7/site-packages/setuptools/command/bdist_egg.py", line 172, in run
    cmd = self.call_command('install_lib', warn_dir=0)
  File "/opt/python/cp27-cp27m/lib/python2.7/site-packages/setuptools/command/bdist_egg.py", line 158, in call_command
    self.run_command(cmdname)
  File "/opt/python/cp27-cp27m/lib/python2.7/distutils/cmd.py", line 326, in run_command
    self.distribution.run_command(command)
  File "/opt/python/cp27-cp27m/lib/python2.7/distutils/dist.py", line 972, in run_command
    cmd_obj.run()
  File "/opt/python/cp27-cp27m/lib/python2.7/site-packages/setuptools/command/install_lib.py", line 11, in run
    self.build()
  File "/opt/python/cp27-cp27m/lib/python2.7/distutils/command/install_lib.py", line 111, in build
    self.run_command('build_ext')
  File "/opt/python/cp27-cp27m/lib/python2.7/distutils/cmd.py", line 326, in run_command
    self.distribution.run_command(command)
  File "/opt/python/cp27-cp27m/lib/python2.7/distutils/dist.py", line 972, in run_command
    cmd_obj.run()
  File "setup.py", line 65, in run
    self.build()
  File "setup.py", line 79, in build
    self.cmake_preflight_check()
  File "setup.py", line 92, in cmake_preflight_check
    raise RuntimeError(stderr.strip())
RuntimeError: CMake Error at CMakeLists.txt:3 (cmake_minimum_required):
  CMake 3.18.2 or higher is required.  You are running version 3.12.0
  ```

**Comment by JeanChristopheMorinPerso** [comment]
Oh nice @reinecke ! I was searching for something like that yesterday and this morning but couldn't find how to do it.

This is much cleaner than duplicating the version check.

**Comment by JeanChristopheMorinPerso** [comment]
Thanks @reinecke foir the suggestion. I implemented your change (with a little tweak to make it work with Python 2 and 3).

**Comment by JeanChristopheMorinPerso** [comment]
Ok, I fix the linting and I had to force push because I noticed most of my commits had my email address in them (author and committer). So modified to use the email I wanted and forced push.

**Comment by JeanChristopheMorinPerso** [comment]
@ssteinbach @meshula @reinecke Can one of you re-run the GitHub actions please for me please? I'm not authorized to run them. Thanks!

**Comment by reinecke** [comment]
Hi @JeanChristopheMorinPerso - re-ran the actions, looking up the build log it looks like this is the root failure message:
```
  CMake Error at /private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/pip-build-env-c9zz6x7z/overlay/lib/python3.7/site-packages/cmake/data/CMake.app/Contents/share/cmake-3.20/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
4607
    Could NOT find Python (missing: Interpreter Development.Module) (found
4608
    version "3.7.10")
4609
  Call Stack (most recent call first):
4610
    /private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/pip-build-env-c9zz6x7z/overlay/lib/python3.7/site-packages/cmake/data/CMake.app/Contents/share/cmake-3.20/Modules/FindPackageHandleStandardArgs.cmake:594 (_FPHSA_FAILURE_MESSAGE)
4611
    /private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/pip-build-env-c9zz6x7z/overlay/lib/python3.7/site-packages/cmake/data/CMake.app/Contents/share/cmake-3.20/Modules/FindPython.cmake:556 (find_package_handle_standard_args)
4612
    CMakeLists.txt:49 (find_package)
```

**Comment by JeanChristopheMorinPerso** [comment]
So the actions failed for Python 3.7 and 3.8 on OSX but worked for 2.7 (OSX). Weird.

Early investigation shows me that:
```
$PATH = /Users/runner/hostedtoolcache/Python/3.8.10/x64/bin:/Users/runner/hostedtoolcache/Python/3.8.10/x64:/usr/local/opt/pipx_bin:/Users/runner/.cargo/bin:/usr/local/lib/ruby/gems/2.7.0/bin:/usr/local/opt/ruby@2.7/bin:/usr/local/opt/curl/bin:/usr/local/bin:/usr/local/sbin:/Users/runner/bin:/Users/runner/.yarn/bin:/Users/runner/Library/Android/sdk/tools:/Users/runner/Library/Android/sdk/platform-tools:/Users/runner/Library/Android/sdk/ndk-bundle:/Library/Frameworks/Mono.framework/Versions/Current/Commands:/usr/bin:/bin:/usr/sbin:/sbin:/Users/runner/.dotnet/tools:/Users/runner/.ghcup/bin:/Users/runner/hostedtoolcache/stack/2.7.1/x64

"which pip" = /Users/runner/hostedtoolcache/Python/3.8.10/x64/bin/pip
"which python" = /Users/runner/hostedtoolcache/Python/3.8.10/x64/bin/python
```

I've also looked at how they build their Python interpreters and nothing standouts. I'll have to test on my macbook I guess. Will do that later, potentially tomorrow.

**Comment by jhodges10** [comment]
Might be worth adding `pyenv` for managing the versions at build time.

**Comment by reinecke** [comment]
@jhodges10 I think that's handled in this build step: https://github.com/PixarAnimationStudios/OpenTimelineIO/blob/master/.github/workflows/python-package.yml#L34-L37

**Comment by JeanChristopheMorinPerso** [comment]
Yeah, we rely on pythons that are compiled for the GitHub Actions environment. Most of them are already in the VM images. If we were to go the route of pyenv, it would add a considerable amount of time to the build times.

It would also solve the issue itself. It might just temporarily push it away.

The issue is not that Python doesn't exixts, but more that CMake can't find either the interpreter or the headers. Most probably the headers since it says `(found version "3.7.10")`, so it found the interpreter.

I'm trying now with 3.7 and 3.7 on OSX.

**Comment by JeanChristopheMorinPerso** [comment]
@reinecke Interestingly, the problem seems to be caused by `cmake --check-system-vars`. Since we do not provide the command with the full arguments that we normally pass to CMake, it fails to find the right python (since Python_EXECUTABLE is not specified) and the CMakeCache is basically wrong for the actual build.

There is multiple options available to us:
1. Run `cmake --check-system-vars` with the same set of arguments that we use to actually build OTIO.
2. Run `cmake --check-system-vars` in a temporary folder (other than the build temp directory) so that the cache doesn't pollute subsequent runs. This would double the CMake generation time (since it would happen both during the pre-flight check and the build).
3. We could stop using `cmake --check-system-vars` and simply parse the `CMakeFiles.txt` to get the minimum required version and then do a manual version check with `cmake --version`.
4. Probably others that I can't think of at the time of writing.

**Comment by reinecke** [comment]
@JeanChristopheMorinPerso Oh wow - good catch.

To me the best thing to do would be 1 + 2: make a temp dir (a good use case for using the [mkdtmp context manager](https://docs.python.org/3.8/library/tempfile.html#tempfile.mkdtemp)), but use the same command args so they stay as similar as possible.
The common arg generation could be factored into a shared `_cmake_args` method that just returns the `cmake_args` array in `cmake_generate`.

What do you think? If you like, I can submit a PR to your branch with the changes - I feel somewhat responsible for this issue coming up in the first place.

**Comment by reinecke** [comment]
Oh, also it may be good to rebase the branch onto #1004 before making this change - it's stepping in the same lines of code for setup.py

**Comment by JeanChristopheMorinPerso** [comment]
@reinecke 1+2 seems good to me. I had no string opinion anyway. I'll do the change and will rebase.

File: setup.py [hunk 5]
```diff
@@ -83,6 +62,7 @@ def run(self):
         # This works around the fact that we build _opentime and _otio
         # extensions as a one-shot cmake invocation. Usually we'd build each
         # separately using build_extension.
+        self.announce('running OTIO build_ext', level=2)
```
JeanChristopheMorinPerso, line 65: I thought it would be a nice touch to have a little more feedback in the logs.

meshula:
@reinecke I think that @JeanChristopheMorinPerso has addressed your requested change? If so, can you mark it as resolved? I've resolved the other conversation, and approved the change to run the CI.

reinecke (approved):
Looks great! Sorry that version check thing became bigger than expected.

JeanChristopheMorinPerso:
Don't be sorry @reinecke ! I'm the one that questioned the version check in the `setup.py`, so I kind of started it.

Thanks for merging the PR! I'm really happy about it!

Next stop, Windows wheels and Uploading wheels to PyPI.

PR #1002, opened 2021-06-24
Add child_if()

### LINKED ISSUE #1001: Add a child_if() function to complement children_if()?
In PR https://github.com/PixarAnimationStudios/OpenTimelineIO/pull/998 I suggested it might be nice to use the general purpose children_if() function to test if a composition has any child clips. @rogernelson pointed out that children_if() might be expensive since it iterates over every child, as opposed to just stopping at the first instance. So maybe it would be nice to have a child_if() function that is similar to children_if(), but only returns the first child? In addition to Roger's use case this could also be useful for other tests, like if a timeline has any transitions.

darbyjohnston:
```Fixes #1001```

In PR #998 I suggested it might be nice to use the general purpose children_if() function to test if a composition has any child clips. @rogernelson pointed out that children_if() might be expensive since it iterates over every child, as opposed to just stopping at the first instance. So maybe it would be nice to have a child_if() function that is similar to children_if(), but only returns the first child? In addition to Roger's use case this could also be useful for other tests, like if a timeline has any transitions.

The caveat here is that by making this a general function and supporting the various otio container types, this turned out to be more code changes than I was hoping for... I'm setting this as a draft PR for now, if this seems like a good change I can add test coverage.

### DIFF
--- a/src/opentimelineio/serializableCollection.h
+++ b/src/opentimelineio/serializableCollection.h
@@ -51,11 +51,22 @@ public:
 
     // Return a vector of all objects that match the given template type.
     //
-    // An optional search_time may be provided to limit the search.
+    // An optional search_range may be provided to limit the search.
     //
     // If shallow_search is false, will recurse into children.
     template<typename T = Composable>
     std::vector<Retainer<T>> children_if(
+        ErrorStatus* error_status,
+        optional<TimeRange> search_range = nullopt,
+        bool shallow_search = false) const;
+
+    // Return the first object that matches the given template type.
+    //
+    // An optional search_range may be provided to limit the search.
+    //
+    // If shallow_search is false, will recurse into children.
+    template<typename T = Composable>
+    Retainer<T> child_if(
         ErrorStatus* error_status,
         optional<TimeRange> search_range = nullopt,
         bool shallow_search = false) const;
@@ -67,6 +78,16 @@ protected:
     virtual void write_to(Writer&) const;
 
 private:
+    enum ChildrenIfOptions {
+        match_first = 1,
+        shallow_search = 2
+    };
+    template<typename T = Composable>
+    std::vector<Retainer<T>> _children_if(
+        ErrorStatus* error_status,
+        optional<TimeRange> search_range,
+        int options) const;
+
     std::vector<Retainer<SerializableObject>> _children;
 };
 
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by ssteinbach** [comment]
When looking at the review for #995 it struck me that:

1. I like this child_if function suggestion
2. I think, if we do this, then we probably want to rename `clip_if` to `clips_if` and add a `clip_if` so that the naming convention is consistent.   Thoughts?

Apologies if this is retreading ground!

**Comment by darbyjohnston** [comment]
Yes, I was thinking the same thing about the naming; should I do that as part of this PR or a separate one? It looks like there are about 30 occurrences of ```clip_if``` in the otio code base.

**Comment by meshula** [comment]
@darbyjohnston This diff is already complex enough, could I request a separate PR for clip(s)_if in order to keep the diff focussed?

**Comment by darbyjohnston** [comment]
@meshula Sure thing, this PR just has the new functions, I'll make a separate PR for renaming "clip_if()".

File: src/opentimelineio/serializableCollection.h [hunk 1]
```diff
@@ -60,13 +60,34 @@ class SerializableCollection : public SerializableObjectWithMetadata {
...
     virtual ~SerializableCollection();
 
     virtual bool read_from(Reader&);
     virtual void write_to(Writer&) const;
 
 private:
+    enum ChildrenIfOptions {
+        match_first = 1,
+        shallow_search = 2
+    };
```
rogernelson, line 84: This is pretty minor, but I'm used to seeing enums that are meant for bitwise operations expressed with bit shift operators.
```
enum ChildrenIfOptions {
   match_first = 1 << 0,
   shallow_search = 1 << 1
};
```
This may come down to personal preference, but I think it helps to express the intent that the options can be or'ed together, especially when there's only 2 (it's more obvious when the third option has a value of 4).

meshula: Same for me.

meshula: There isn't a precedent in OTIO yet on this pattern, we've got one enum declared as int, and the rest are C style, so a minimum edit with the bitshifts is sufficient IMO

darbyjohnston: Sounds good, I'll make the change to the enum.

Overall does this PR seem like a worthwhile change vs. the more specialized function in Roger's PR?

meshula: Could you drop a link to the more specialized function here so we can see the comparison in context of this thread?

darbyjohnston: https://github.com/PixarAnimationStudios/OpenTimelineIO/pull/998#discussion_r656588870

darbyjohnston: So, does this seem like a good idea? If so I would like to un-draft it, otherwise I'll close it and the corresponding issue:
https://github.com/PixarAnimationStudios/OpenTimelineIO/issues/1001

meshula:
@darbyjohnston I'm trying to follow the threads around these predicated functions, it's a bit confusing :) Could you add a note about where the conversation has picked up?

darbyjohnston:
@meshula This PR was for adding a similar function to ```children_if()```` that just returned a single child. It's an optimization that allows the function to stop searching the hierarchy when it finds the first instance.

meshula:
Ah, I thought that was a good idea. I see no one responded to your question of Aug 5.

PR #1006, opened 2021-06-30
allow commandline arguments to be passed through to cmake

ssteinbach:
Sometimes its helpful to be able to pass command line arguments through setup.py to cmake.  This change appends the CMAKE_ARGS variable to the flags passed to cmake.  It can be used like this:

`env CMAKE_ARGS="-DCMAKE_VAR_1=VALUE1 -DCMAKE_VAR_2=VALUE2" python setup.py build`
or
`env CMAKE_ARGS="-DCMAKE_VAR=VALUE -DCMAKE_VAR_2=VALUE2" pip install .`


reinecke (approved):
Nice add!
Something we might think about is finding a place to document these so they don't stay in the realm of tribal knowledge.

ssteinbach:
Yes, although part of the reason I'd prefer to break up the build systems is so that it's possible to communicate directly with cmake using standard cmake tools. 

Not having to document these weird pass throughs the setup.py would be better... They aren't discoverable and are pretty non standard as far as I've seen.

JeanChristopheMorinPerso:
@ssteinbach There is ways to expose custom arguments.

For example we could have:

```python
class OTIO_build_ext(setuptools.command.build_ext.build_ext):

    user_options = setuptools.command.build_ext.build_ext.user_options + [
        ('cmake=', None, 'JC custom extension')
    ]

    def initialize_options(self):
        setuptools.command.build_ext.build_ext.initialize_options(self)
        self.cmake = None  # Really important. If you don't do this, the option will not be available
```
Running `python setup.py build_ext --cmake="-DCMAKE_VAR_1=something -DCMAKE_VAR_2=else"` would make the variables accessible in the `self.cmake` variable.

the `--cmake` is discoverable when using `python setup.py build_ext --help`.

Not sure if that would be an acceptable solution.

The downside is that I don't think we can pass this argument to `pip` in any way. The usual `--global-option`/`--install-option`/`--build-option` don't seem to work.

meshula (approved):

ssteinbach:
Yeah... I think in this case we need to pass these options through pip since that is the way that wheels are to be built.  But that is a great suggestion... pip doesn't have functionality like that?

JeanChristopheMorinPerso:
>I think in this case we need to pass these options through pip since that is the way that wheels are to be built.

You can create a wheel like this:
```sh
python setup.py build_ext
python setup.py bdist_wheel
```
and because `build_ext` is run before `bdist_wheel`, there will only be a single compilation and `bdist_wheel` just takes the existing artifacts and creates a wheel with them.

>pip doesn't have functionality like that
Pip has some flags, like `--global-option`, `--install-option`. It's not too clear what the global flag does. As for the install flag, it passes the flag to the `setup.py install` command.

I've tried them and I'm still not too sure how everything works with these flags. There is also the fact that if you specify flags like these, they will apply to all dependencies being installed... Which means the `cmake` python package will be completely rebuilt. One thing I noticed is that if you `pip install .`, our `setup.py` will be invoked with `setup.py bdist_wheel`. But the thing is we don't have a custom class/command for `bdist_wheel`. Which means setuptools invokes/uses our custom `build_ext` class/command somehow when `bdist_wheel` is run. I'm not sure how it's done but I have a feeling that by understanding this it would probably answer if it's doable to do that with pip.

meshula (approved):

ssteinbach:
Add some documentation to the build instructions in the readthedocs/README files.

ssteinbach:
> Nice add!
> Something we might think about is finding a place to document these so they don't stay in the realm of tribal knowledge.

@reinecke I added some notes to the README and to the Quickstart.md.  Any notes on that or does that look good?

PR #1017, opened 2021-07-19
Replace description-file metadata key with description_file

JeanChristopheMorinPerso:
**Summary**

Latest `setuptools` prints a deprecation warning as follows:

```
    /tmp/pip-build-env-cje71c9_/overlay/lib/python3.9/site-packages/setuptools/dist.py:697: UserWarning: Usage of dash-separated 'description-file' will not be supported in future versions. Please use the underscore name 'description_file' instead
```
So this PR simply replaces `description-file` with `description_file`.


meshula:
I can't find documentation, other than this, which doesn't mention the deprecation reported in the warning. https://docs.openstack.org/pbr/latest/user/features.html - our tests have all passed, including making wheels, so it seems fine!

meshula (approved):

JeanChristopheMorinPerso:
See https://github.com/pypa/setuptools/pull/2588 and https://github.com/pypa/setuptools/issues/1608 for more information on this warning.

meshula:
Thanks for the links. Looks like setuptools' maintainers decided to deviate from PEP508 which states that dashes and underscores are supposed to be equivalent.

ssteinbach:
Are the other hyphenated fields ok?

ssteinbach:
(meaning: will we have to change the rest to underscores as well?)

ssteinbach:
Looking through the file, it looks like a mix of hyphens and underscores.  This seems fine, we can fix them as they come up.  Thanks!

JeanChristopheMorinPerso:
>Looks like setuptools' maintainers decided to deviate from PEP508 which states that dashes and underscores are supposed to be equivalent.

The PEP isn't clear on that, so they decided one something, which is this warning.

>Looking through the file, it looks like a mix of hyphens and underscores. This seems fine, we can fix them as they come up.

This warning should only apply to sections owned by setuptools/distutils, see https://github.com/pypa/setuptools/issues/2595#issuecomment-798923933. So the PR I linked was a little problematic, which was fixed in https://github.com/pypa/setuptools/pull/2600.

So the other sections not owned by setuptools/distutils are fine.

PR #1029, opened 2021-07-29
AAF adapter: added `clip_timecode_offset` argument

### LINKED ISSUE #1028: AAF adapter: track timecode offsets clip source range
## Feature Request

New feature

## Description

Currently the AAF adapter adds the track timecode to the source range of any clip. This allows for roundtripping of the information, but makes it cumbersome to retrieve the "real" source range for clips. This behaviour could be made optional with a new adapter argument, so that correct source range information is more easily accessible.

## Context

Related bit of code:

https://github.com/PixarAnimationStudios/OpenTimelineIO/blob/master/contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py#L303-L305

timlehr:

Fixes #1028

**Summarize your change.**

**AAF adapter: added `clip_timecode_offset` argument**

Added `clip_timecode_offset` as new optional argument. By default, the behaviour is the same as before. However, if the parameter is set to `False`, the track timecode offset will not be added to the clip source range. This allows for easy access to the actual source range used in the clip.

- updated `test_aaf_adapter.py`

### DIFF
--- a/contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py
+++ b/contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py
@@ -57,6 +57,9 @@ debug = False
 # If enabled, output recursive traversal info of _transcribe() method.
 _TRANSCRIBE_DEBUG = False
 
+# If enabled, the track timecode offset will be included in the clip source_range
+_CLIP_TIMECODE_OFFSET = True
+
 
 def _transcribe_log(s, indent=0, always_print=False):
     if always_print or _TRANSCRIBE_DEBUG:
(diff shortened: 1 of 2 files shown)


jminor:
In discussion with @timlehr @andrewmoore-nz we would like to hold on this PR until Andrew updates his older PR that touches this same code. The current implementation finds timecode for a clip only from a single mob, but should sum the offsets from a chain of mobs instead. Once that is in place, we can re-examine this PR with a better understanding of how timecode for a clip should be retrieved.

timlehr:
Playing around a bit more with #1063, we get the source range we expect internally. This one ignores the mob chain entirely based on some wrong understanding I had of the AAF parser code last year. I think this one can probably be closed. @jminor

meshula:
Closing i n favor of #1063

PR #1032, opened 2021-08-04
Added linguist override to detect .otio files for JSON highlighting

reinecke:
Adds a [linguist override](https://github.com/github/linguist/blob/master/docs/overrides.md) to syntax highlight `.otio` files as JSON.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.otio linguist-language=JSON


meshula (approved):
nifty!

meshula:
As an aside, I think codecov isn't configured properly. Most of the decrease in coverage is mapped to blank lines!!

PR #1040, opened 2021-08-07
Fix missing Python coverage in Codecov reports

### LINKED ISSUE #1039: Python coverage is missing from Codecov reports
With the update of the `codecov/codecov-action` action from v1 to v2, the coverage reports are off by a lot.

The update happened in https://github.com/PixarAnimationStudios/OpenTimelineIO/pull/1026 and the report on this PR reports:

```diff
@@            Coverage Diff             @@
##           master    #1026      +/-   ##
==========================================
- Coverage   85.64%   79.31%   -6.34%     
==========================================
  Files         192      113      -79     
  Lines       18387     6086   -12301     
  Branches     2056        0    -2056     
==========================================
- Hits        15748     4827   -10921     
+ Misses       2113     1259     -854     
+ Partials      526        0     -526     
```

So basically the covered lines went from 18387 to 6086. By looking at the reports in the UI:

Before: https://codecov.io/gh/PixarAnimationStudios/OpenTimelineIO/tree/1b08d491b7538e2bcdb32d43822b73e32e8021e4
After: https://codecov.io/gh/PixarAnimationStudios/OpenTimelineIO/tree/068225674752a226381ea43fddaef22159e410fb

We can see that Python coverage seems to be missing.

By looking at it I'm not entirely sure what's wrong. Except maybe that the new codecov upload script (or action) used by the action stopped running the `coverage xml` command and we might have to generate the XML ourselves.

JeanChristopheMorinPerso:
Fixes #1039 

**Summary**

As stated in #1039 , with the recent update of the `codecov/codecov-action` GH action, the Python coverage is missing from the Codecov reports.

This is due to the fact that the old Codecov action was using a different uploader script that was taking care of calling `coverage xml` when it found Python coverage files. The new action doesn't seem to do it.

So the fix is as simple as calling `coverage xml` from the `make coverage` target.


JeanChristopheMorinPerso:
For transparency: I've let this PR as a draft while I find time to investigate a little bit more. I'd like to get to the bottom of why the `ignore` list in `.codecov.yaml` doesn't seems to have any effect.

If anyone wants to participate/help, feel free to replay :)

I was also unable to produce C++ coverage on my Linux laptop, which annoys be quite a lot.

JeanChristopheMorinPerso:
This is now ready to go.

PR #1046, opened 2021-08-18
Update Python code to use the new functions children_if()/clip_if()

### LINKED ISSUE #1045: Update Python code to use the new functions children_if()/clip_if()
As a follow up to https://github.com/PixarAnimationStudios/OpenTimelineIO/issues/925, this issue is for updating the Python code to use the new functions children_if() and clip_if().

darbyjohnston:
Fixes #1045

As a follow up to https://github.com/PixarAnimationStudios/OpenTimelineIO/pull/926, this PR is for updating the Python code to use the new functions children_if() and clip_if().


darbyjohnston:
Whoops, looks like the adapter tests are failing with my changes; those don't get tested with the "make test" from the root OpenTimelineIO directory?

darbyjohnston:
OK, I checked in fixes for the adapters so I was going to un-draft this PR since all of the tests now pass.

meshula (approved):
lgtm

JeanChristopheMorinPerso:
I'm curious about what we should do with the `each_child` methods. Is the plan to eventually remove/deprecate them?

darbyjohnston:
I added comments that those functions are deprecated, but it would be nice to remove them at some point. Maybe after the next version/release?

jradha11:
I had a name change suggestion for `children_if` and `clip_if`. Can we rename them to `get_children` and `get_clips` respectively?

meshula:
TSC notes:

- `children_if` makes sense because it returns a zero or more children
- `clip_if` returns zero or more clips, so should be named `clips if`
- the naming `foo_if` exists because the arguments to the function are predicates. I would like a token to indicate the default predication, because `clip_if()` leaves ambiguity on the part of a reader, but `clip_if(any_exist)` directly expresses intent.

darbyjohnston:
I'm not that familiar with Python, what is a token?

reinecke:
Here's a question: We're re-litigating api naming that was [previously settled on](https://github.com/PixarAnimationStudios/OpenTimelineIO/pull/926) in a PR that is meant to just move forward with API deprecation. Given there is confusion about the API naming, the discussion is a good one -  perhaps it would be best to let that exist as a separate issue? If we rename the python API, there is a discussion to be had about whether we should rename the C++ API too.
I feel like this PR should stay focussed on advancing the proposed deprecation, if we want to do a name change perhaps that should be a separate PR?

meshula:
@darbyjohnston by "token" I simply mean some sort of argument to signal the default intent, because whenever I discuss children_if() and clip_if() with people, the immediate reaction is puzzlement because children_if() without an argument simply begs the question of "if what now, exactly?". C++ naming has latterly given me an allergy to vague naming that causes puzzlement :)

@reinecke That's a fair point, although I was recording the meeting notes for this issue in the issue ;) It suggests that if we're in agreement on the naming update proposed during the TSC, the C++ API should be amended first and this PR rebased upon it. Our convention is an issue and a PR that closes it, so as you suggest, that could be the next step prior to landing this.

darbyjohnston:
If we are going to change the names I think it's good to see that in context with both the C++ and Python, so I personally would prefer a single PR. Could we update the title of this PR to expand the scope, or close it and open a new one?

meshula:
In response to @darbyjohnston's suggestion:

Tagging @jminor as the person who flagged the naming issue.
Tagging @reinecke as the person who suggested a new issue.
Tagging @ssteinbach as a person to provide quorum.
My vote is 👍

PR #1047, opened 2021-08-22
Add constexpr/noexcept

### LINKED ISSUE #999: c++ methods to be marked constexpr and noexcept as appropriate
## Feature Request

We need to do a scrub on the C++ API to further annotate functions with constexpr and noexcept. Currently only const is in use.

For example,  in RationalTime,

```cpp
     double value() const {
         return _value;
     }
```
could be 
```cpp
     constexpr double value() noexcept {
         return _value;
     }
```

darbyjohnston:
```Fixes #999```

I tried adding constexpr/noexcept decorators to OpenTime, if this looks reasonable I can also do a pass over OpenTimelineIO. I haven't used constexpr/noexcept much so hopefully I didn't go overboard, it certainly makes the headers look more busy.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -130,7 +130,7 @@ endif()
 # Global language settings
 
 if (NOT CMAKE_CXX_STANDARD)
-    set(CMAKE_CXX_STANDARD 11)
+    set(CMAKE_CXX_STANDARD 14)
 endif()
 
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
--- a/src/opentime/rationalTime.h
+++ b/src/opentime/rationalTime.h
@@ -3,6 +3,7 @@
 #include "opentime/version.h"
 #include "opentime/errorStatus.h"
 #include <cmath>
+#include <limits>
 #include <string>
 
 namespace opentime { namespace OPENTIME_VERSION  {
@@ -17,7 +18,7 @@ constexpr double fabs(double val) noexcept
 {
     union { double f; uint64_t i; }
     bits = { val };
-    bits.i &= -1ULL / 2;
+    bits.i &= std::numeric_limits<uint64_t>::max() / 2;
     return bits.f;
 }
 
--- a/src/opentime/rationalTime.h
+++ b/src/opentime/rationalTime.h
@@ -22,7 +22,7 @@ public:
     constexpr RationalTime(RationalTime const&) noexcept = default;
     RationalTime& operator= (RationalTime const&) noexcept = default;
 
-    constexpr bool is_invalid_time() const noexcept {
+    bool is_invalid_time() const noexcept {
         return (std::isnan(_rate) || std::isnan(_value)) ? true : (_rate <= 0);
     }
     
@@ -50,7 +50,7 @@ public:
         return value_rescaled_to(rt._rate);
     }
 
-    constexpr bool almost_equal(RationalTime other, double delta = 0) const noexcept {
+    bool almost_equal(RationalTime other, double delta = 0) const noexcept {
         return fabs(value_rescaled_to(other._rate) - other._value) <= delta;
     }
 
@@ -171,7 +171,7 @@ private:
     static RationalTime _invalid_time;
     static constexpr double _invalid_rate = -1;
     
-    constexpr RationalTime _floor() const noexcept {
+    RationalTime _floor() const noexcept {
         return RationalTime {floor(_value), _rate};
     }
 
--- a/src/opentime/rationalTime.h
+++ b/src/opentime/rationalTime.h
@@ -13,7 +13,14 @@ enum IsDropFrameRate : int {
     ForceYes = 1,
 };
     
-    
+constexpr double fabs(double val) noexcept
+{
+    union { double f; uint64_t i; }
+    bits = { val };
+    bits.i &= -1ULL / 2;
+    return bits.f;
+}
+
 class RationalTime {
 public:
     explicit constexpr RationalTime(double value = 0, double rate = 1) noexcept
@@ -50,7 +57,7 @@ public:
         return value_rescaled_to(rt._rate);
     }
 
-    bool almost_equal(RationalTime other, double delta = 0) const noexcept {
+    constexpr bool almost_equal(RationalTime other, double delta = 0) const noexcept {
         return fabs(value_rescaled_to(other._rate) - other._value) <= delta;
     }
 
@@ -107,7 +114,7 @@ public:
     
     std::string to_time_string() const;
 
-    RationalTime const& operator+= (RationalTime other) noexcept {
+    constexpr RationalTime const& operator+= (RationalTime other) noexcept {
         if (_rate < other._rate) {
             _value = other._value + value_rescaled_to(other._rate);
             _rate = other._rate;
@@ -118,7 +125,7 @@ public:
         return *this;
     }
 
-    RationalTime const& operator-= (RationalTime other) noexcept {
+    constexpr RationalTime const& operator-= (RationalTime other) noexcept {
         if (_rate < other._rate) {
             _value = value_rescaled_to(other._rate) - other._value;
             _rate = other._rate;
--- a/src/opentime/rationalTime.h
+++ b/src/opentime/rationalTime.h
@@ -16,50 +16,46 @@ enum IsDropFrameRate : int {
     
 class RationalTime {
 public:
-    explicit RationalTime(double value = 0, double rate = 1)
+    explicit constexpr RationalTime(double value = 0, double rate = 1) noexcept
     : _value {value}, _rate {rate} {}
 
-    RationalTime(RationalTime const&) = default;
-    RationalTime& operator= (RationalTime const&) = default;
+    constexpr RationalTime(RationalTime const&) noexcept = default;
+    RationalTime& operator= (RationalTime const&) noexcept = default;
 
-    bool is_invalid_time() const {
-        if(std::isnan(_rate) || std::isnan(_value)) {
-            return true;
-        }
-
-        return _rate <= 0;
+    constexpr bool is_invalid_time() const noexcept {
+        return (std::isnan(_rate) || std::isnan(_value)) ? true : (_rate <= 0);
     }
     
-    double value() const {
+    constexpr double value() const noexcept {
         return _value;
     }
 
-    double rate() const {
+    constexpr double rate() const noexcept {
         return _rate;
     }
 
-    RationalTime rescaled_to(double new_rate) const {
+    constexpr RationalTime rescaled_to(double new_rate) const noexcept {
         return RationalTime {value_rescaled_to(new_rate), new_rate};
     }
 
-    RationalTime rescaled_to(RationalTime rt) const {
+    constexpr RationalTime rescaled_to(RationalTime rt) const noexcept {
         return RationalTime {value_rescaled_to(rt._rate), rt._rate};
     }
 
-    double value_rescaled_to(double new_rate) const {
+    constexpr double value_rescaled_to(double new_rate) const noexcept {
         return new_rate == _rate ? _value : (_value * new_rate) / _rate;
     }
     
-    double value_rescaled_to(RationalTime rt) const {
+    constexpr double value_rescaled_to(RationalTime rt) const noexcept {
         return value_rescaled_to(rt._rate);
     }
 
-    bool almost_equal(RationalTime other, double delta = 0) const {
+    constexpr bool almost_equal(RationalTime other, double delta = 0) const noexcept {
         return fabs(value_rescaled_to(other._rate) - other._value) <= delta;
     }
 
     static RationalTime
-    duration_from_start_end_time(RationalTime start_time, RationalTime end_time_exclusive) {
+    constexpr duration_from_start_end_time(RationalTime start_time, RationalTime end_time_exclusive) noexcept {
         return start_time._rate == end_time_exclusive._rate ?
             RationalTime {end_time_exclusive._value - start_time._value, start_time._rate} :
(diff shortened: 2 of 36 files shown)

### CODE REVIEW

**Review by meshula — COMMENTED** [review]
It's the nature of C++ code to become busy like this, it's expected. That's how you know you're Modern :p It does make sense to do it, because some compilers really do take advantage of the hinting at least partially.

What prevents the TimeRange comparisons from also becoming constexpr?

**Comment by meshula** [comment]
At the previous TSC the consensus was moving to c++14. @darbyjohnston I think we can move ahead.

**Comment by darbyjohnston** [comment]
OK, I updated CMakeLists.txt to C++14. That change along with your ```fabs()``` function allowed a number of additional functions in RationalTime/TimeRange to be marked constexpr.

I'm not sure why my changes are showing this though:
```
Submodule pybind11 updated 77 files
```
Is there a way I can revert those changes?

**Comment by ssteinbach** [comment]
We should check the docs and the README.md badges as well -- might want to split the C++14 thing into its own pr and land that first just to clarify it

**Comment by meshula** [comment]
@darbyjohnston a rebase at your end to master, and a force push to this PR should resolve the spurious commits.

Landing the CMakeLists compiler version change first makes sense, just to let the CI kick through any consequences.

**Comment by darbyjohnston** [comment]
@mehsula Thanks, that didn't seem to fix it though; it looks like those changes actually got included in one of my commits, ```5400daa51f88b71e3fac2ead9b91ed1cc792c359```, possibly from merging? I guess I could delete that commit, manually redo the changes, and then another force push?

**Comment by darbyjohnston** [comment]
I created a new PR for the C++14 change:
https://github.com/PixarAnimationStudios/OpenTimelineIO/pull/1069

The only references to C++11 I saw in the docs were about the ```any/optional``` classes, which maybe doesn't need to change. And the older C++ section:
https://opentimelineio.readthedocs.io/en/latest/cxx/older.html

**Comment by meshula** [comment]
@darbyjohnston Unfortunately the history-surgery you propose is probably the only solution.

File: CMakeLists.txt [hunk 1]
```diff
@@ -130,7 +130,7 @@ endif()
 # Global language settings
 
 if (NOT CMAKE_CXX_STANDARD)
-    set(CMAKE_CXX_STANDARD 11)
+    set(CMAKE_CXX_STANDARD 14)
```
meshula, line 133: I merged this already, so hopefully a rebase makes this part of the diff go away?

src/opentime/rationalTime.h
```diff
@@ -13,61 +13,64 @@ enum IsDropFrameRate : int {
     ForceYes = 1,
 };
     
-    
+constexpr double fabs(double val) noexcept
+{
+    union { double f; uint64_t i; }
+    bits = { val };
+    bits.i &= -1ULL / 2;
```
meshula, line 20: some compilers are going to reject a -1ULL, we should just use `(numeric_limits<uint64_t>::max() /2)` here.

src/opentimelineio/composition.h
```diff
@@ -23,7 +23,7 @@ class Composition : public Item {
 
     virtual std::string const& composition_kind() const;
```
meshula, line 24:
```suggestion
    virtual std::string composition_kind() const;
```

src/opentimelineio/composition.h
```diff
@@ -23,7 +23,7 @@ class Composition : public Item {
...
 
-    std::vector<Retainer<Composable>> const& children() const {
+    std::vector<Retainer<Composable>> const& children() const noexcept {
```
meshula, line 26: return by ref is scary here, but I don't want to mess with it just now. The vector is const, but the Composables can be mutated, not what the intent of the const here.

src/opentimelineio/effect.h
```diff
@@ -18,7 +18,7 @@ class Effect : public SerializableObjectWithMetadata {
            std::string const& effect_name = std::string(),
            AnyDictionary const& metadata = AnyDictionary());
 
-    std::string const& effect_name() const {
+    std::string const& effect_name() const noexcept {
         return _effect_name;
```
meshula, line 22:
```suggestion
    std::string effect_name() const noexcept {
        return _effect_name;
```
const ref in a return makes no guarantees that the string actually lives as long as the reference, so returning const& in general is an anti-pattern.

src/opentimelineio/externalReference.h
```diff
@@ -18,7 +18,7 @@ class ExternalReference final : public MediaReference {
                       optional<TimeRange> const& available_range = nullopt,
                       AnyDictionary const& metadata = AnyDictionary());
         
-    std::string const& target_url() const {
+    std::string const& target_url() const noexcept {
```
meshula, line 21:
```suggestion
    std::string target_url() const noexcept {
```

src/opentimelineio/generatorReference.h
```diff
@@ -20,19 +20,19 @@ class GeneratorReference final : public MediaReference {
                        AnyDictionary const& parameters = AnyDictionary(),
                        AnyDictionary const& metadata = AnyDictionary());
         
-    std::string const& generator_kind() const {
+    std::string const& generator_kind() const noexcept {
```
meshula, line 23:
```suggestion
    std::string generator_kind() const noexcept {
```

src/opentimelineio/imageSequenceReference.h
```diff
@@ -31,67 +31,67 @@ class ImageSequenceReference final : public MediaReference {
                       optional<TimeRange> const& available_range = nullopt,
                       AnyDictionary const& metadata = AnyDictionary());
         
-    std::string const& target_url_base() const {
+    std::string const& target_url_base() const noexcept {
```
meshula, line 34:
```suggestion
    std::string target_url_base() const noexcept {
```

src/opentimelineio/imageSequenceReference.h
```diff
@@ -31,67 +31,67 @@ class ImageSequenceReference final : public MediaReference {
...
         return _target_url_base;
     }
     
     void set_target_url_base(std::string const& target_url_base) {
         _target_url_base = target_url_base;
     }
         
-    std::string const& name_prefix() const {
+    std::string const& name_prefix() const noexcept {
```
meshula, line 42:
```suggestion
    std::string name_prefix() const noexcept {
```

src/opentimelineio/item.h
```diff
@@ -29,27 +29,27 @@ class Item : public Composable {
     virtual bool visible() const;
     virtual bool overlapping() const;
 
-    optional<TimeRange> const& source_range () const {
+    optional<TimeRange> const& source_range () const noexcept {
```
meshula, line 32:
```suggestion
    optional<TimeRange> source_range () const noexcept {
```

src/opentimelineio/marker.h
```diff
@@ -33,19 +33,19 @@ class Marker : public SerializableObjectWithMetadata {
            std::string const& color = Color::green,
            AnyDictionary const& metadata = AnyDictionary());
 
-    std::string const& color() const {
+    std::string const& color() const noexcept {
```
meshula, line 36:
```suggestion
    std::string color() const noexcept {
```

src/opentimelineio/marker.h
```diff
@@ -33,19 +33,19 @@ class Marker : public SerializableObjectWithMetadata {
...
         return _color;
     }
     
     void set_color(std::string const& color) {
         _color = color;
     }
 
-    TimeRange const& marked_range() const {
+    TimeRange const& marked_range() const noexcept {
```
meshula, line 44:
```suggestion
    TimeRange marked_range() const noexcept {
```

src/opentimelineio/mediaReference.h
```diff
@@ -20,7 +20,7 @@ class MediaReference : public SerializableObjectWithMetadata {
                    optional<TimeRange> const& available_range = nullopt,
                    AnyDictionary const& metadata = AnyDictionary());
 
-    optional<TimeRange> const& available_range () const {
+    optional<TimeRange> const& available_range () const noexcept {
```
meshula, line 23:
```suggestion
    optional<TimeRange> available_range () const noexcept {
```

src/opentimelineio/serializableObjectWithMetadata.h
```diff
@@ -17,19 +17,19 @@ class SerializableObjectWithMetadata : public SerializableObject {
     SerializableObjectWithMetadata(std::string const& name = std::string(),
                                    AnyDictionary const& metadata = AnyDictionary());
 
-    std::string const& name() const {
+    std::string const& name() const noexcept {
```
meshula, line 20:
```suggestion
    std::string name() const noexcept {
```

src/opentimelineio/timeline.h
```diff
@@ -34,7 +34,7 @@ class Timeline : public SerializableObjectWithMetadata {
 
     void set_tracks(Stack* stack);
     
-    optional<RationalTime> const& global_start_time() const {
+    optional<RationalTime> const& global_start_time() const noexcept {
```
meshula, line 37:
```suggestion
    optional<RationalTime> global_start_time() const noexcept {
```

src/opentimelineio/track.h
```diff
@@ -31,7 +31,7 @@ class Track : public Composition {
           std::string const& = Kind::video,
           AnyDictionary const& metadata = AnyDictionary());
 
-    std::string const& kind() const {
+    std::string const& kind() const noexcept {
```
meshula, line 34:
```suggestion
    std::string kind() const noexcept {
```

src/opentimelineio/unknownSchema.h
```diff
@@ -14,11 +14,11 @@ class UnknownSchema : public SerializableObject {
 
     UnknownSchema(std::string const& original_schema_name, int original_schema_version);
 
-    std::string const& original_schema_name() const {
+    std::string const& original_schema_name() const noexcept {
```
meshula, line 17:
```suggestion
    std::string original_schema_name() const noexcept {
```

meshula (changes requested):
Looking almost ready! Notes

- rebase for c++14
- numeric_limits instead of -1ULL
- some const& elimination on return values, where impact is low - if you haven't used the suggestion feature before, the "merge suggestion" button updates the diff to save you work

src/opentime/rationalTime.h
darbyjohnston: Are the parentheses ```()``` needed?

src/opentimelineio/composition.h
darbyjohnston: I was planning on submitting a separate PR for const cleanup changes, though I guess it could be part of this one? There's a number of these that could be changed.

src/opentime/rationalTime.h
meshula: that snip was from my crusty brain matter, not documentation :)

src/opentimelineio/composition.h
meshula: I'm inclined to rip off the bandaid and get it done as part of constexpr just because we have to address general constness with this PR. a change to C++ version is a separate concern, so it made sense to me to split it. I don't see constness as meaningfully separable though.

src/opentimelineio/composition.h
darbyjohnston: Out of curiosity why not change it if we're removing the other ```const&```?

meshula: the intent of const& here is that if we get children we cannot mutate them, so we wouldn't get the intended safety by removing the const&.

 In other words, children() should really be returning something like `std::vector<Immutable<Composable>>` but we haven't got an Immutable Retainer thingy or a ConstRetainer, or a Retainer<const T>, as far as I know. So making that change seems out of scope to me for this change, as we'd first need to figure out what the returned vector should really be, possibly write a new class for it, then patch it in everywhere appropriate.

darbyjohnston: In this particular case I don't think the ```const&``` was meant to imply immutable Composables, but rather as an optimization/pessimization to avoid copying the list. For example:
https://github.com/PixarAnimationStudios/OpenTimelineIO/blob/835fc5c12791dd29c333e5c6d7e633565b2d792f/src/opentimelineio/timeline.cpp#L35

darbyjohnston: Hmm, what about this:
https://github.com/PixarAnimationStudios/OpenTimelineIO/blob/835fc5c12791dd29c333e5c6d7e633565b2d792f/src/opentimelineio/serializableObjectWithMetadata.h#L28

Replace with ```set_metadata```?

meshula: Since all we are doing is exposing a member with full access, it seems like the member should be public, and the accessors omitted....

meshula: I agree with your point, but not only should the children list be immutable, but c in the `auto c` would itself should be const. I think a reasonable interim would be to actually return a copy of the vector, since the children are reference counted...

darbyjohnston: I personally like the accessors since it gives us the ability to add extra functionality later, and seems similar to the rest of the API. My vote would be for ```set_metadata()```.

meshula: I think I agree? I'd prefer

```cpp
void set_metadata(const AnyDictionary&);
void merge_metadata(const AnyDictionary&);
```
These aren't accessors, but the idea that well-meaning arbitrary end users can reach in and tickle the internal state gives me the shivers. I prefer a functional approach for this kind of thing.

meshula: In order to unblock and land this work, I'd like to defer this discussion about accessors to another PR or Issue.

meshula (approved):
This looks ready to land. Thanks again for the scrub!

PR #1054, opened 2021-08-30
Add ALE adapter round-trip CDL values

### LINKED ISSUE #1050: ALE adapter round-trip CDL Values
## Feature Request

New feature

## Description

The ALE adapter reads CDL values into the `cdl` namespaces of clip metadata dictionaries from the `ASC_SAT` and `ASC_SOP` fields, however it does not write them back out. It would be helpful to have the ALE adapter round-trip this data. This would also match the EDL adapter behavior.

## Context

For how the adapter reads CDL values, look [here](https://github.com/PixarAnimationStudios/OpenTimelineIO/blob/6ce63ff7bdb5b90f66aa090d5689f566ca7b5b33/contrib/opentimelineio_contrib/adapters/ale.py#L123-L157).
For an example of how the EDL adapter writes out this metadata, look [here](https://github.com/PixarAnimationStudios/OpenTimelineIO/blob/6ce63ff7bdb5b90f66aa090d5689f566ca7b5b33/src/py-opentimelineio/opentimelineio/adapters/cmx_3600.py#L1348-L1368)

Viraj-Rana008:
```Fixes #1050```

**Summary**
Appended ASC_SOP and ASC_SAT values for the round-trip of data.

To provide a similar behavior as EDL adapter.

### DIFF
--- a/contrib/opentimelineio_contrib/adapters/ale.py
+++ b/contrib/opentimelineio_contrib/adapters/ale.py
@@ -134,12 +134,29 @@ def _parse_data_line(line, columns, fps):
 
             if cdl:
                 del metadata['ASC_SOP']
+                asc_sop = cdl.get('asc_sop')
+                if asc_sop:
+                    line.append(
+                        "*ASC_SOP ({} {} {}) ({} {} {}) ({} {} {})".format(
+                            asc_sop['slope'][0],
+                            asc_sop['slope'][1],
+                            asc_sop['slope'][2],
+                            asc_sop['offset'][0],
+                            asc_sop['offset'][1],
+                            asc_sop['offset'][2],
+                            asc_sop['power'][0],
+                            asc_sop['power'][1],
+                            asc_sop['power'][2],
+                         ))
 
         if metadata.get('ASC_SAT'):
             try:
                 asc_sat_value = float(metadata['ASC_SAT'])
                 cdl.update(asc_sat=asc_sat_value)
                 del metadata['ASC_SAT']
+                asc_sat = cdl.get('asc_sat')
+                if asc_sat:
+                    line.append("*ASC_SAT {}".format(asc_sat))
             except ValueError:
                 pass
 

### CODE REVIEW

File: contrib/opentimelineio_contrib/adapters/ale.py [hunk 1]
```diff
@@ -134,12 +134,30 @@ def _parse_data_line(line, columns, fps):
...
         if metadata.get('ASC_SAT'):
             try:
                 asc_sat_value = float(metadata['ASC_SAT'])
                 cdl.update(asc_sat=asc_sat_value)
                 del metadata['ASC_SAT']
+                asc_sat = cdl.get('asc_sat')
+                if asc_sat:
+                    line += "\t*ASC_SAT {}\t".format(
+                        asc_sat
+                    )
```
reinecke, line 160: I think this update code is in the reader, not the writer. Also, the string looks like it's in a format meant for EDL?

contrib/opentimelineio_contrib/adapters/ale.py
```diff
@@ -134,12 +134,30 @@ def _parse_data_line(line, columns, fps):
...
+                        asc_sop['slope'][0],
+                        asc_sop['slope'][1],
+                        asc_sop['slope'][2],
+                        asc_sop['offset'][0],
+                        asc_sop['offset'][1],
+                        asc_sop['offset'][2],
+                        asc_sop['power'][0],
+                        asc_sop['power'][1],
+                        asc_sop['power'][2]
+                    )
```
reinecke, line 149: This looks like it's using EDL syntax to try and write to the ALE, but in the parsing code rather than the writer code?

reinecke (changes requested):
Hi @Viraj-Rana008,

Thanks for contributing! It looks like the changes might be in the read code instead of the writer code - also it looks like it's somewhat matching EDL syntax instead of ALE syntax.

It might be best to start by adding a unittest to make sure CDL data round-trips properly and that will help make clearer whether or not the behavior is correct.

reinecke (changes requested):
Looks like it's coming along! Looks like the next big thing is to add unittesting to make sure this behaves as expected.

jminor:
Waiting on tests to be added.

contrib/opentimelineio_contrib/adapters/ale.py
```diff
@@ -318,6 +318,11 @@ def write_to_string(input_otio, columns=None, fps=None, video_format=None):
             for key in fields.keys():
                 if key not in columns:
                     columns.append(key)
+            if clip.metadata.get('cdl'):
+                if 'ASC_SOP' not in columns:
+                    columns.append('ASC_SOP')
+                if 'ASC_SAT' not in columns:
+                    columns.append('ASC_SAT')
```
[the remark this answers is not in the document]
Viraj-Rana008: How about this?

`cdl = clip.metadata.get('cdl', {})`
`if cdl.get('asc_sop') and 'ASC_SOP' not in column:`
`...columns.append('ASC_SOP')`
`if cdl.get('asc_sat') and 'ASC_SAT' not in column:`
`...columns.append('ASC_SAT')`

Viraj-Rana008:
Yes, I'll work on it.

reinecke:
Hi @Viraj-Rana008,
Because the ALE adapter would do blind pass-through deserializing metadata into the `ALE` namespace, then serialize anything in that namespace back out I think it's possible that while the functionality changed as expected there could be a  regression we wouldn't know about that the round-trip would hide. My personal rule-of-thumb is that if functionality changes and the unit tests don't, then it means our tests aren't validating the change we made - the code could accidentally get reverted in a merge error and no one would know.

An ALE CDL write test should be able to cover it. You could create a `SerializableCollection` with a single clip that has both `ALE` and `CDL` metadata, then use `write_to_string` to output and check the result. The nice part is that if there are other specialized kinds of metadata that would need to be translated, that same test could be expanded to handle them.

reinecke:
This change should move to [OpenTimelineIO/otio-ale-adapter](https://github.com/OpenTimelineIO/otio-ale-adapter/).

jminor:
Now that the OTIO adapters have moved to their own repositories, this PR should be continued on this repo: https://github.com/OpenTimelineIO/otio-ale-adapter/

PR #1061, opened 2021-09-06
Error fixes

### LINKED ISSUE #1060: Error handling fixes
As pointed out by @gfgit, there is an issue with the error handling in the new C++ functions ```Composition::child_at_time()```, ```Composition::children_in_range()```, ```Composition::children_if()```, and ```SerializableCollection::children_if()```.

The code currently looks like this:
```
if (!error_status) {
...
}
```
But should look like this (similar to the surrounding code):
```
if (*error_status) {
...
}
```
We don't need to check whether ```error_status``` is null since the API assumes that it is a valid pointer.

darbyjohnston:
Fixes #1060

These changes fix the error handling in ```Composition::child_at_time()```, ```Composition::children_in_range()```, ```Composition::children_if()```, and ```SerializableCollection::children_if()``` to properly check the error_status and return from the function.

I also added an additional Python test for the various options to ```children_if()/each_child()```, though it might not be necessary if it doesn't increase test coverage. I was hoping to test the various error conditions that were fixed, though it's not obvious how to trigger them since error_status is passed down to so many different functions.

### DIFF
--- a/src/opentimelineio/composition.cpp
+++ b/src/opentimelineio/composition.cpp
@@ -340,9 +340,11 @@ SerializableObject::Retainer<Composable> Composition::child_at_time(
     ErrorStatus* error_status,
     bool shallow_search) const
 {
+    Retainer<Composable> result;
+
     auto range_map = range_of_all_children(error_status);
-    if (!error_status) {
-        *error_status = ErrorStatus(ErrorStatus::INTERNAL_ERROR, "one or more invalid children encountered");
+    if (*error_status) {
+        return result;
     }
 
     // find the first item whose end_time_exclusive is after the
@@ -350,8 +352,8 @@ SerializableObject::Retainer<Composable> Composition::child_at_time(
         search_time,
         [&range_map](Composable* child) { return range_map[child].end_time_exclusive(); },
         error_status);
-    if (!error_status) {
-        *error_status = ErrorStatus(ErrorStatus::INTERNAL_ERROR, "one or more invalid children encountered");
+    if (*error_status) {
+        return result;
     }
 
     // find the last item whose start_time is before the
@@ -360,8 +362,8 @@ SerializableObject::Retainer<Composable> Composition::child_at_time(
         [&range_map](Composable* child) { return range_map[child].start_time(); },
         error_status,
         first_inside_range);
-    if (!error_status) {
-        *error_status = ErrorStatus(ErrorStatus::INTERNAL_ERROR, "one or more invalid children encountered");
+    if (*error_status) {
+        return result;
     }
 
     // limit the search to children who are in the search_range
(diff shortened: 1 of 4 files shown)


meshula (commented):
The reason the null checks were on the error_status was because we pass in a pointer, that can be null, which causes crashes upon dereference. Clearly that's confusing as there's ambiguity as to whether the error status is supposed to be checked, or if the error status is merely a place to write.

This code is meant to write to error_status, if error_status was applied. Not, if error_status pointer was supplied return as if an error occurred.

I believe the real issue, to @gfgit's original review, is that we are using a C style idiom in a C++ interface. Having a pointer to a place to store any errors in the interface is the root source of the issue. I'm open to a new interface proposal.

src/opentimelineio/errorStatus.cpp
```diff
@@ -1,7 +1,12 @@
 #include "opentimelineio/errorStatus.h"
 
 namespace opentimelineio { namespace OPENTIMELINEIO_VERSION  {
-    
+
+bool ErrorStatus::is_ok(const ErrorStatus* es)
+{
+    return es && Outcome::OK == es->outcome;
```
meshula, line 7: So the new behavior is to default to a nullptr ErrorStatus, and to treat the default as an error. I think that's a low touch change, as a step to getting rid of the bool operator first, and upgrading the error reporting pattern wholesale in the future.

darbyjohnston: It's a fairly small change to the API though it will require a fair amount of work to make sure all the code using ```ErrorStatus``` can deal with null pointers.

I was looking at codecov to see how well the error cases were covered, and does this look right?
https://codecov.io/gh/PixarAnimationStudios/OpenTimelineIO/src/master/src/opentimelineio/composition.cpp#L165

In ```Composition::_index_of_child()``` it looks like the coverage shows the function never returns successfully in the tests, but the lines setting the error status are not recorded as hit (i.e., not green)? Actually the coverage in the rest of the files looks odd as well, the green numbers should show the number of times that line was hit, but they are all == 1?
https://docs.codecov.com/docs/viewing-source-code

meshula: I'm not a codecov person, so I'll defer to someone else.

>  it will require a fair amount of work to make sure all the code using ErrorStatus can deal with null pointers.

Yep. The checks I put in were specifically ones I hit. We really need an audit of every function that takes an ErrorStatus.

darbyjohnston: I've replaced the ErrorStatus operator with the ```is_ok``` function and checked all the uses of ```error_status``` to make sure the pointers are valid. It's a lot of changes but the tests are passing...

meshula (approved):
Huge thanks for grinding this through!

Could I request one more reviewer please? There's a lot to read here, and it would be good for someone else to also check.

Issue #523, opened 2019-06-03
AAF Reader: Incorrect Clip Source Times


andrewmoore-nz:
So, bear with me on this one... it's a little complicated; a little confusing; and I'm still trying to piece together as many of the exact details as I can. 

I'm just starting to notice a very serious issue with `available_range`'s in a clip's `media_reference` (and `source_range`) coming through from Media Composer AAFs.

There appears to be inconsistency in when the the correct `start_time` is used from the underlying AAF data. I am yet to see a specific reason/pattern as to when this is happening, so best I can do at this point is just give as much info as I can for the situation. 

----------------------

**ExampleClip_col_mat12_v025_p005**

This clip has source timecode values of In: `00:00:41:21` and Out: `00:01:08:16`, but the `available_range`/`source_range` object generated by OTIO comes through as:

```
"available_range": {
    "OTIO_SCHEMA": "TimeRange.1", 
    "duration": {
        "OTIO_SCHEMA": "RationalTime.1", 
        "rate": 23.97602397602398, 
        "value": 643
    }, 
    "start_time": {
        "OTIO_SCHEMA": "RationalTime.1", 
        "rate": 23.97602397602398, 
        "value": 4
    }
},
...
"source_range": {
    "OTIO_SCHEMA": "TimeRange.1", 
    "duration": {
        "OTIO_SCHEMA": "RationalTime.1", 
        "rate": 23.97602397602398, 
        "value": 643
    }, 
    "start_time": {
        "OTIO_SCHEMA": "RationalTime.1", 
        "rate": 23.97602397602398, 
        "value": 4
    }
}
```

Using this info, we get source timecode values of In: `00:00:00:04` and Out: `00:00:26:23`. The correct timecode values above were generated by opening the AAF in Media Composer and exporting EDLs, and the incorrect timecode values (the `available_rage` object) were generated by reading the same AAF using OTIO. 

Consulting this specific clip's info in the Media Composer bin we see:

![ExampleClip_col_mat12_v025_p005](https://user-images.githubusercontent.com/19657812/58782313-57246300-8632-11e9-961e-59b3b429c5d2.png)

From this, it appears that the `Start (TC)` and `Aux TC1` fields are being ignored, and simply the frames being used are determining the `available_range` - frames 4 (`00:00:00:04`) through 647 (`00:00:26:23`) of the total 651 available. 

----------------------

[...]

PR #1063, opened 2021-09-12
AAF Parser - Calculating Source Values

andrewmoore-nz:
Following on from this weeks AAF call, this is the main process being used to correctly calculate the source timecode values for the SourceClip component. 

@jminor @timlehr @stefanschulze 

To give a brief recap, the source timecode values are calculated by looking at a chain of Media Objects (Mobs) and summing and converting a selection of values. The chain of mobs will be a combination of CompositionMobs, MasterMobs, and SourceMobs. There can be a source clip Origin, source clip Start, and a mob slot timecode object, all of which can effect the frame count that ends up being the source start value. 

The function `_find_mob_chain_and_timecode()` condenses a few things in to one process for efficiency:

    """ This is combining several processes in to one for efficiency:
         - Traversing the SourceClip's Mob structure using the clip_id to return all
           the necessary Mobs.
         - Extracting the necessary Start/Origin/Offset values for each object, and
           converts the rate if necessary. This value can come from several different
           places.
         - Searching for Operation Group objects that may contain information on how to
           correctly calculate the starting frame value.
         - Storing the individual SourceClips so that at the end of the process the
           correct Length can be derived from the MasterMob."""

Because all the relevant media objects for this particular source clip are returned from this process, there are several efficiencies that can be gained further on in this part of the `_transcribe()` function. I haven't included any of those here though, as I just want to focus on the source timecode values for now. I can't completely separate the pull requests, as the further changes are reliant on those mobs being available in that particular way. 

This

### DIFF
--- a/contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py
+++ b/contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py
@@ -128,95 +128,146 @@ def _transcribe_property(prop):
         return str(prop)
 
 
-def _find_timecode_mobs(item):
-    mobs = [item.mob]
+def _find_source_clip(item, op_group_found=None):
+    if isinstance(item, aaf2.components.SourceClip):
+        return item, op_group_found
+    elif isinstance(item, aaf2.components.OperationGroup):
+        return _find_source_clip(item.segments[0], item)
+    elif isinstance(item, aaf2.components.Sequence):
+        comp = [c for c in item.components if not isinstance(c, aaf2.components.Filler)]
+        comp = comp[0] if len(comp) > 0 else None
+        return _find_source_clip(comp, op_group_found)
+    elif isinstance(item, aaf2.components.EssenceGroup):
+        choice = item['Choices'][0] if len(item['Choices']) > 0 else None
+        return _find_source_clip(choice, op_group_found)
+    elif isinstance(item, aaf2.components.NestedScope):
+        slot = item.slots[0] if len(item.slots) > 0 else None
+        return _find_source_clip(slot, op_group_found)
+    elif isinstance(item, aaf2.components.Selector):
+        alt = [a for a in item.alternates.value]
+        alt = alt[0] if len(alt) > 0 else None
+        if alt is not None:
+            return _find_source_clip(item.alternates.value[0], op_group_found)
+        # return _find_source_clip(item.selected.value, op_group_found)
+    elif isinstance(item, aaf2.components.Filler) or not item:
+        return None, None
+    elif isinstance(item, aaf2.components.Pulldown) or not item:
+        return None, None
+    else:
+        raise AAFAdapterError("Error: _find_source_clip() parsing {} not "
+                              "supported".format(type(item)))
 
-    for c in item.walk():
-        if isinstance(c, aaf2.components.SourceClip):
-            mob = c.mob
-            if mob:
-                mobs.append(mob)
-            else:
-                continue
+
+def _find_mobslot_timecode(mob, phys_track_num):
+    timecode = (0, 0.0, 'mob slot timecode')
+    slot = [s for s in mob.slots if s['PhysicalTrackNumber'].value == phys_track_num
+            and s.media_kind == 'Timecode']
+    tc = slot[0].segment if len(slot) > 0 else None
+    if isinstance(tc, aaf2.components.Sequence):
+        comps = [c for c in tc.components]
+        tc = comps[0] if len(comps) > 0 else None
+    if isinstance(tc, aaf2.components.Timecode):
+        timecode = (tc.start, float(slot[0].edit_rate), 'mob slot timecode')
+    return timecode
+
+
+def _absolute_offset_from_tc_chain(timecode_chain, edit_rate, mob_chain):
+    frame_count = 0
+    for start, rate, source in timecode_chain:
+        if start == 0:
+            continue
+        elif rate == edit_rate:
+            frame_count += start
         else:
-            # This could be 'EssenceGroup', 'Pulldown' or other segment
-            # subclasses
-            # For example:
-            # An EssenceGroup is a Segment that has one or more
-            # alternate choices, each of which represent different variations
-            # of one actual piece of content.
-            # According to the AAF Object Specification and Edit Protocol
-            # documents:
-            # "Typically the different representations vary in essence format,
-            # compression, or frame size. The application is responsible for
-            # choosing the appropriate implementation of the essence."
-            # It also says they should all have the same length, but
-            # there might be nested Sequences inside which we're not attempting
-            # to handle here (yet). We'll need a concrete example to ensure
-            # we're doing the right thing.
-            # TODO: Is the Timecode for an EssenceGroup correct?
-            # TODO: Try CountChoices() and ChoiceAt(i)
-            # For now, lets just skip it.
+            if rate <= 0:
+                err = "Error: an incompatible rate of {} has been found when " \
+                      "attempting to calculate the source timecode value for clip {}"
+                raise AAFAdapterError(err.format(rate, mob_chain[0].name))
+            frame_count += int(round(start / (rate / edit_rate)))
+    return frame_count
+
+
+def _find_mob_chain_and_timecode(source_clip, edit_rate):
+    """ This is combining several processes in to one for efficiency:
+         - Traversing the SourceClip's Mob structure using the clip_id to return all
+           the necessary Mobs.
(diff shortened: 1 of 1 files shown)


andrewmoore-nz:
This approach directly addresses the the situation I outlined in my issue from 2019 - #523 

An example of how/why this approach is needed... when a `TapeName` value is added to a clip, it changes the final SourceMob in the chain from a mob with an `ImportDescriptor` essence to a mob with a `TapeDescriptor` essence. This then changes where one of the required start frame values is being stored, from in the last mob in the chain, to the second to last mob in the chain - a SourceMob with a `CDCIDescriptor` essence.

timlehr:
@andrewmoore-nz are you still on this issue?

andrewmoore-nz:
@timlehr This pull request was supposed to be a jumping off point to look at addressing the AAF parsing issues present in the current version of the parser, which we have fixed in the version currently being used in Matchbox. I'm still very keen to get these changes rolled in.

timlehr:
@jminor @reinecke can we reserve some time for this in an upcoming TSC?

andrewmoore-nz:
@jminor @reinecke What needs to happen to get some traction on this? It's been well over two years since Roger Sacilotto confirmed in our call with him that this is the correct approach, so would be great to actually get this rolled in and available for people to use, and some further development can continue.

meshula:
Hi! I've tagged this for TSC review. Up until now it's been marked as a Draft/Work In Progress, so I've changed the status to Open.

I think the core fix in this PR is this:

```
An example of how/why this approach is needed... when a TapeName value is added to a clip, it changes the final SourceMob in the chain from a mob with an ImportDescriptor essence to a mob with a TapeDescriptor essence. This then changes where one of the required start frame values is being stored, from in the last mob in the chain, to the second to last mob in the chain - a SourceMob with a CDCIDescriptor essence.
```

If there was a test case added around this that fails with the existing code, but accepts with the patch, I think we'll be good to go.

andrewmoore-nz:
@meshula There are a few test clips from back in 2019 that should cover that test case pretty easily so I'll add that to the PR this weekend when I get a chance to look at it again. 

This was marked as a Draft/Work in Progress as that's what it is meant to be. We had a call back in September discussing this topic, and this Draft PR was supposed to be a jumping off point for a discussion about the problem, which never happened. If this PR specifically was going to be merged, it will need some further work to clean up other, now redundant, parts of the code.

meshula:
Gotcha, I understood your question on getting traction to mean specifically this PR, as opposed to re-starting a conversation.

contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py
```diff
@@ -128,95 +128,146 @@ def _transcribe_property(prop):
...
+
+            # If an operation group was discovered during the _find_source_clip()
+            # process, check for a time effect which may change the starting frame value
+            time_effect = None
+            if op_group_found and op_group_found.operation.name == 'Motion Control':
+                op_stack = _transcribe(op_group_found, list(), edit_rate)
+                time_effect = op_stack.effects[0] if len(op_stack.effects) > 0 else None
+
+            time_scalar = time_effect.time_scalar if hasattr(time_effect, 'time_scalar') else None
+            if time_scalar:
```
jminor, line 242: Are you checking for None or 0 or both? If it was a freeze frame, the time_scalar will be 0 which Python treats as false.

contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py
```diff
@@ -128,95 +128,146 @@ def _transcribe_property(prop):
...
+        return _find_source_clip(choice, op_group_found)
+    elif isinstance(item, aaf2.components.NestedScope):
+        slot = item.slots[0] if len(item.slots) > 0 else None
+        return _find_source_clip(slot, op_group_found)
+    elif isinstance(item, aaf2.components.Selector):
+        alt = [a for a in item.alternates.value]
+        alt = alt[0] if len(alt) > 0 else None
+        if alt is not None:
+            return _find_source_clip(item.alternates.value[0], op_group_found)
+        # return _find_source_clip(item.selected.value, op_group_found)
```
jminor, line 151: If alt was None, this should return something. Also, should the Selector's "Selected" item be used here instead of the first alternate?

jminor (commented):
At a high level this all makes sense. I can't claim that I understand all the details, but I trust that if this passes the unit tests, especially for the wider range of sample AAFs that you have, then this is a nice step forwards.
Sorry for the extremely long delay before looking at it in detail.

contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py
```diff
@@ -128,95 +128,146 @@ def _transcribe_property(prop):
...
-    mobs = [item.mob]
-
-    for c in item.walk():
-        if isinstance(c, aaf2.components.SourceClip):
-            mob = c.mob
-            if mob:
-                mobs.append(mob)
-            else:
-                continue
+def _find_source_clip(item, op_group_found=None):
```
jminor, line 131: From Zoom code review: we're wondering what `item` is passed in here?

contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py
```diff
@@ -128,95 +128,146 @@ def _transcribe_property(prop):
...
-        raise otio.exceptions.NotSupportedError(
-            "Error: mob has more than one timecode slot with different values."
-            " This is currently not supported by the AAF adapter. Found:"
-            " {} slots,  mob name is: '{}'".format(len(timecodes), mob.name)
-        )
-    else:
-        return None
+        slot = [s for s in mob.slots if s.slot_id == slot_id]
+        slot = slot[0] if len(slot) > 0 else None
+        result = _find_source_clip(slot.segment) if slot else (None, None)
```
jminor, line 225: From Zoom code review: this is where the code attempts to drill into the SourceClip to find the details of what's inside it

contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py
jminor: This was the segment of the 1st slot of the source clip from  _find_mob_chain_and_timecode, which might just be a SourceClip, but we might need to dig into the structure under it to find the SourceClip.

jminor: This if/elif chain is attempting to handle the variety of AAFs in our test suite (both in OTIO, and from Matchbox) but we are still unclear of the full breadth of structures we might find within a SourceClip.

contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py
```diff
@@ -288,14 +339,15 @@ def _transcribe(item, parents, edit_rate, indent=0):
...
-        last_mob = mobs[-1] if mobs else None
-        timecode_info = _extract_timecode_info(last_mob) if last_mob else None
+        # Get all relevant mobs down the tree and their source clips
+        # These are necessary for calculating correct starting values
+        mc_tc_info = _find_mob_chain_and_timecode(item, edit_rate)
+        mobs, timecode_info, source_start_multiplier = mc_tc_info
 
         source_start = int(metadata.get("StartTime", "0"))
+        if source_start_multiplier:
+            source_start = int(source_start * source_start_multiplier)
```
jminor, line 349: Is the multiplier just from varying frame rate clips, or is this from timewarp effects in the composition?
0 or None means it was not found, and should be ignored.

contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py
jminor: 0 would only come up via a FreezeFrame. If that was the case, we should already have the right first frame. So 0 and None should both skip this code block.
WBN to have a unit test for that, or at least verify manually to confirm this.

jminor, line 242:
```suggestion
            if time_scalar is None or time_scalar == 0:  # FreezeFrames have a time_scale of 0
```

jminor (changes requested):
If you could turn some of these informal code review notes from our video conf into comments in the code that would be great.

timlehr:
I think we can close this one in favour of https://github.com/OpenTimelineIO/otio-aaf-adapter/pull/44.

jminor:
Closing in favor of https://github.com/OpenTimelineIO/otio-aaf-adapter/pull/44

PR #1064, opened 2021-09-12
AAF Parser - Simplify no longer removes track of all muted clips

andrewmoore-nz:
A bug that recently came to my attention, so figured I'd submit a PR to match the change we've put in to our branch for use in Matchbox. 

Looks to be an oversight in how the Gap objects are actually being used, as there is absolutely something valuable in that object when it's a muted clip and contains all the information on the clip that is muted.

### DIFF
--- a/contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py
+++ b/contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py
@@ -1147,7 +1147,9 @@ def _contains_something_valuable(thing):
 
     if isinstance(thing, otio.schema.Gap):
         # TODO: Are there other valuable things we should look for on a Gap?
-        return False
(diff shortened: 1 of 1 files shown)


timlehr:
@jminor Is there a a reason this hasn't been merged yet?

jminor:
Now that #1175 has landed, we can update the AAF adapter to use the `enabled` flag instead of making muted clips into Gaps. That will be much cleaner and will make this change unnecessary.

jminor:
@andrewmoore-nz and @timlehr how about this instead? https://github.com/PixarAnimationStudios/OpenTimelineIO/pull/1212

jminor:
@andrewmoore-nz do we still need this PR now that your #1277 has landed?

andrewmoore-nz:
@jminor This is redundant now, so no longer needed. I'll close it.

Issue #828, opened 2020-10-30
Python 3.9 Support


reinecke:
There currently exists a rare crash condition that can occur with the combination of python 3.9.0 and pybind11 (for context see [this ASWF thread](https://academysoftwarefdn.slack.com/archives/C014LNJUWKH/p1603206819024200) or [this pybind11 issue](https://github.com/python/cpython/pull/22670).

The tl;dr: There could be unexpected and unpredictable crashing in Python 3.9.0, this issue is set to be addressed in 3.9.1. After 3.9.1 is released we should enable CI for 3.9 and consider adding 3.9 to the badge in the readme.

We may also consider updating the `python_requires` to exclude 3.9.0, though we'd want to make sure we can do this in a way that signals to users why this is and steps to remedy (use 3.9.1 instead).

ssteinbach:
It looks like 3.9.1 is released:
https://www.python.org/downloads/release/python-391/

I'm not sure about how the python_requires works, but I know we do some manual checks at the top of the `setup.py` to guard against ancient pip/setuptools versions.

ssteinbach:
Note that there are a bunch of build warnings we'll need to tackle in 3.9:

```
src/deps/pybind11/include/pybind11/detail/internals.h:200:9: warning: 'PyEval_InitThreads' is deprecated [-Wdeprecated-declarations]
        PyEval_InitThreads();
        ^
```

meshula:
That looks like motivation to get pybind11 updated

ssteinbach:
yeah, hoping that addresses this issue.  Note that pybind11 is also seeing some weird CI failure: #874

Tilix4:
I don't know if it's completely relevant, but I've tried both the current master and the reinecke's `feature/wheel-builds` branch with Python3.9 and I get this error when importing:
```py
>>> import opentimelineio
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Users\Felix\AppData\Local\Programs\Python\Python39\lib\site-packages\opentimelineio\__init__.py", line 35, in <module>
    from . import (
  File "C:\Users\Felix\AppData\Local\Programs\Python\Python39\lib\site-packages\opentimelineio\exceptions.py", line 26, in <module>
    from . _otio import ( # noqa
ImportError: DLL load failed while importing _otio: A dynamic link library (DLL) initialization routine failed.
```

The funny thing is, I only get this error on my personal windows computer.

[...]

PR #1070, opened 2021-09-18
Fix failing import of opentimelineio in Python 3.9 on Windows

JeanChristopheMorinPerso:
Relates to issue #828

I've been working on building wheels for Python 3.9 and I've been blocked by an issue where `opentimelineio` couldn't be imported once compiled. The issue only happened on Windows with Python 3.9 when the extension was statically linked (which is the default).

After a lot of back and forth and getting some help from @meshula and @LazyDodo, I was able to finally fix the problem.

The issue would manifest itself like this:
```
>>> import opentimelineio
  File "<stdin>", line 2, in <module>
  File "C:\Users\jeanchristophemorin1\AppData\Roaming\Python\Python39\site-packages\opentimelineio\__init__.py", line 35, in <module>
    from . import (
  File "C:\Users\jeanchristophemorin1\AppData\Roaming\Python\Python39\site-packages\opentimelineio\exceptions.py", line 26, in <module>
    from . _otio import ( # noqa
ImportError: DLL load failed while importing _otio: A dynamic link library (DLL) initialization routine failed.
```

So `_otio.pyd` was properly found but failed to be initialized/loaded. Attaching a debugger would give this call stack:

```
>	python39_d.dll!PyErr_Occurred() Line 250	C
 	_otio_d.cp39-win_amd64.pyd!pybind11::arg_v::arg_v<std::string>(pybind11::arg && base, std::string && x, const char * descr) Line 1066	C++
 	_otio_d.cp39-win_amd64.pyd!pybind11::arg_v::arg_v<std::string>(const pybind11::arg & base, std::string && x, const char * descr) Line 1080	C++
 	_otio_d.cp39-win_amd64.pyd!pybind11::arg::operator=<std::string>(std::string && value) Line 1110	C++
 	_otio_d.cp39-win_amd64.pyd!`dynamic initializer for 'name_arg''() Line 126	C++
```

Somehow, CPython isn't happy with
```c++
static py::arg_v name_arg = ("name"_a = std::string());
static py::arg_v metadata_arg = ("metadata"_a = py::none());
```
in `src/py-opentimelineio/opentimelineio-bindings/otio_serializableObjects.cpp`. As my C/C++ knowledge is limited, I can't explain nor understand why it's causing a problem, but I found how to fix it.

Simply removing the static vars from the global scope and in lining them (Not sure if it's the right term, but it feel like it to me. Let me know if it's the wrong term) fixes the problem.

Note that CI for Python 3.9 will be added in a future PR. In the meantime, you can see a build in my fork: https://github.com/JeanChristopheMorinPerso/OpenTimelineIO/runs/3637946770?check_suite_focus=true.

### DIFF
--- a/src/py-opentimelineio/opentimelineio-bindings/otio_serializableObjects.cpp
+++ b/src/py-opentimelineio/opentimelineio-bindings/otio_serializableObjects.cpp
@@ -123,9 +123,6 @@ static std::string str(AnyDictionary& value) {
 }
 */
 
-static py::arg_v name_arg = ("name"_a = std::string());
-static py::arg_v metadata_arg = ("metadata"_a = py::none());
-
 template <typename CONTAINER, typename ITEM>
 class ContainerIterator {
 public:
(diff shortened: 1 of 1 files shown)


meshula (approved):
This is wonderful, thanks so much for your perseverance on this difficult issue!

PR #1072, opened 2021-09-18
Fix Makefile format: target

### LINKED ISSUE #1071: Fix Makefile format: target
The Makefile format: target recipe is indented with spaces causing it to run when ```make test``` is run (or maybe any target?). Replacing the spaces with tabs fixes the issue.

darbyjohnston:
Fixes #1071

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -143,11 +143,11 @@ endif
 
 # format all .h and .cpp files using clang-format
 format:
-    opentimeFiles = src/opentime/*.h src/opentime/*.cpp
-    opentimelineioFiles = src/opentimelineio/*.h src/opentimelineio/*.cpp
-    pyopentimelineio-opentimeFiles = src/py-opentimelineio/opentime-bindings/*.h src/py-opentimelineio/opentime-bindings/*.cpp
-    pyopentimelineio-opentimelineioFiles = src/py-opentimelineio/opentimelineio-bindings/*.h src/py-opentimelineio/opentimelineio-bindings/*.cpp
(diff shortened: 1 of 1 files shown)


ssteinbach:
ah, good catch.  One small request- the Makefile has a pattern in it:

```
PYCODESTYLE_PROG := $(shell command -v pycodestyle 2> /dev/null)

# ...
ifndef PYCODESTYLE_PROG
	$(error $(newline)$(ccred)pycodestyle is not available on $$PATH please see:$(newline)$(ccend)\
	$(ccblue)	https://pypi.python.org/pypi/pep8#installation$(newline)$(ccend)\
	$(dev_deps_message))
endif
```

So it prints a small error message when the program is missing indicating that the target won't work.  Adding that for `clang-format` would help people out in case they don't have it on their path.  I can do this in a later PR also if you'd like to break it up.

meshula (approved):

meshula:
@darbyjohnston I approved before I saw Stephan's request, so I'll hold off merging...

ssteinbach:
In this case I don't know if we need to provide instructions on how to install clang-format, merely messaging that it wasn't found in $PATH.

darbyjohnston:
@ssteinbach I would prefer that to be a separate change, this issue/PR addresses a specific bug that is causing clang-format to be run when ```make test``` is executed (and probably the other make targets as well). I've already lost work to this when running ```make test``` to check my changes and ending up with every .h/.cpp file reformatted.

meshula:
Added #1078 to catch the clang-format catcher.

ssteinbach:
fair enough!

ssteinbach (approved):

PR #1077, opened 2021-09-22
Fix RationalTime Representation

### LINKED ISSUE #682: RationalTime repr inexact value can be hard to debug
When the value gets large enough on a RationalTime, the repr switches to scientific notation. This can make debugging output less exact than needed in common debugging situations:

For instance, in the following example the repr appears to be three frames forward in time:
```
>>> otio.opentime.from_timecode("11:58:01:13", rate=24)
otio.opentime.RationalTime(value=1.03396e+06, rate=24)
>>> otio.opentime.from_timecode("11:58:01:13", rate=24).value
1033957.0
>>> otio.opentime.RationalTime(1033957, rate=24)
otio.opentime.RationalTime(value=1.03396e+06, rate=24)
```

visajshah:
Fixes #682 RationalTime Representation

### DIFF
--- a/src/py-opentimelineio/opentime-bindings/opentime_rationalTime.cpp
+++ b/src/py-opentimelineio/opentime-bindings/opentime_rationalTime.cpp
@@ -1,6 +1,5 @@
 #include <pybind11/pybind11.h>
 #include <pybind11/operators.h>
-#include <cmath.h>
 
 #include "opentime/rationalTime.h"
 #include "opentimelineio/stringUtils.h"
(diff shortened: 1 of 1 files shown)


jminor:
Thanks for taking a look at this! Could you add some tests also to show that the intended use case (e.g. round-trip from RationalTime to string to back again) are working properly?
Also, there is some background information and suggested strategy for this here: https://stackoverflow.com/questions/16839658/printf-width-specifier-to-maintain-precision-of-floating-point-value/19897395

visajshah:
Yeah, sure, I will look into it. I will also have to look at the build tests and clear that.

meshula:
@visajshah Are you still interested in pursuing this one? The intention seemed worthwhile.

meshula:
Closing due to inactivity.

PR #1079, opened 2021-09-22
Makefile "format" target tuning

### LINKED ISSUE #1078: Makefile should detect missing clang-format
the Makefile has a pattern in it:

```
PYCODESTYLE_PROG := $(shell command -v pycodestyle 2> /dev/null)

# ...
ifndef PYCODESTYLE_PROG
	$(error $(newline)$(ccred)pycodestyle is not available on $$PATH please see:$(newline)$(ccend)\
	$(ccblue)	https://pypi.python.org/pypi/pep8#installation$(newline)$(ccend)\
	$(dev_deps_message))
endif
```

So it prints a small error message when the program is missing indicating that the target won't work. Adding that for clang-format would help people out in case they don't have it on their path. ~ @ssteinbach

ssteinbach:
Fix #1078 

Also tunes the logic in the target to use wildcards.  I'm considering tuning the style file itself, and maybe even adding in a post-formatted version of the code to this PR as well.


jminor (approved):

jminor:
In a separate PR, let's look at auto-formatting the code.

PR #1080, opened 2021-09-24
Support Python 3.9

### LINKED ISSUE #828: Python 3.9 Support

JeanChristopheMorinPerso:
Fixes #828.

Add support for Python 3.9.

I also took the occasion to set the `python_requires` to explicitly set the supported Python versions and prohibit Python 3.9.0. Additionally I removed the logic around the install requires for Python < 3 to use a more modern and static method.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -32,19 +32,6 @@ PLAT_TO_CMAKE = {
     "win-amd64": "x64",
 }
 
-INSTALL_REQUIRES = [
-    'pyaaf2~=1.4.0',
-]
-# python2 dependencies
-if sys.version_info[0] < 3:
-    INSTALL_REQUIRES.extend(
-        [
-            "backports.tempfile",
-            'future',  # enables the builtins module in the XGES adapter
-            "pathlib2"  # used in the otioz adapter to conform to unix paths
-        ]
-    )
-
 
 def _debugInstance(x):
     for a in sorted(dir(x)):
@@ -295,6 +282,7 @@ setup(
         'Programming Language :: Python :: 3',
         'Programming Language :: Python :: 3.7',
         'Programming Language :: Python :: 3.8',
+        'Programming Language :: Python :: 3.9',
         'Operating System :: OS Independent',
         'Natural Language :: English',
     ],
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -329,7 +317,17 @@ def test_otio():
         'opentimelineview': 'src/opentimelineview',
     },
 
-    install_requires=INSTALL_REQUIRES,
+    # Disallow 3.9.0 because of https://github.com/python/cpython/pull/22670
+    python_requires='>2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*, !=3.5.*, !=3.6.*, !=3.9.0',  # noqa: E501
```
meshula, line 321: this is a nice way to handle versioning!

meshula (approved):

ssteinbach (changes requested):
The README.md also needs to be updated (badges and so on).  Thanks for doing this clean up!

JeanChristopheMorinPerso:
@ssteinbach I modified the README as requested. I also tried to see if there was other references to supported Python versions and couldn't find anything outside what I have in this PR. Let me know if there is more places to update.

setup.py
JeanChristopheMorinPerso: Something like `~=2.7.0, >3.7, <3.10` would be more intuitive IMO, but unfortunately it's not how it works.
```python
>>> packaging.specifiers.SpecifierSet('~=2.7.0, >3.8, <3.10, !=3.9.0').contains('3.8.1')
False
```
:shrug:

ssteinbach:
Yup this looks good.  I think we can ask the question of whether we can drop the older VFX platform support and add 2021.

ssteinbach (approved):

ssteinbach:
Thanks @JeanChristopheMorinPerso!

JeanChristopheMorinPerso:
I was unsure about the VFX platform version as I wasn't sure if we fully support 2022. For example I don't know if the OTIO UI works with PySide 5.15.x.

PR #1087, opened 2021-09-27
bounds -> available_image_bounds and remove imath inclusives

### LINKED ISSUE #771: MediaReference: spatial bounding box

rogernelson:
Fixes #771
MediaReference: spatial bounding box

**Summary**

This is a followup on the previous spatial bounds merge [here](https://github.com/PixarAnimationStudios/OpenTimelineIO/pull/1022)

After our meeting, we decided to make a few changes which this branch introduces:
- Rename `bounds` to `available_image_bounds`.  The `image` is added to indicate the bounds are being applied to the visual/video/image data and `available` is added to indicate that this is the total available size (similar in concept to available range).  This allows us to later add constraints on the bounds (like `source_range`). 

-  Remove some problematic imath bindings.  Calculations in imath related to borders/edges are inclusive, which is not what we want, so we'll remove these until we can add exclusive methods in imath.

**Tests**

Since this is only renaming/removal the existing imath and bounds testings will cover this completely.

### DIFF
--- a/docs/tutorials/otio-serialized-schema-only-fields.md
+++ b/docs/tutorials/otio-serialized-schema-only-fields.md
@@ -59,8 +59,8 @@ parameters:
 ### MediaReference.1
 
 parameters:
+- *available_image_bounds*
 - *available_range*
-- *bounds*
 - *metadata*
 - *name*
 
@@ -149,8 +149,8 @@ parameters:
 ### ExternalReference.1
 
 parameters:
+- *available_image_bounds*
 - *available_range*
-- *bounds*
 - *metadata*
 - *name*
 - *target_url*
@@ -175,8 +175,8 @@ parameters:
 ### GeneratorReference.1
 
 parameters:
+- *available_image_bounds*
 - *available_range*
-- *bounds*
 - *generator_kind*
 - *metadata*
 - *name*
@@ -185,8 +185,8 @@ parameters:
 ### ImageSequenceReference.1
 
 parameters:
+- *available_image_bounds*
 - *available_range*
-- *bounds*
 - *frame_step*
 - *frame_zero_padding*
 - *metadata*
@@ -217,8 +217,8 @@ parameters:
 ### MissingReference.1
 
 parameters:
+- *available_image_bounds*
 - *available_range*
-- *bounds*
 - *metadata*
 - *name*
 
--- a/docs/tutorials/otio-serialized-schema.md
+++ b/docs/tutorials/otio-serialized-schema.md
@@ -113,8 +113,8 @@ None
 ```
 
 parameters:
+- *available_image_bounds*: 
 - *available_range*: 
-- *bounds*: 
 - *metadata*: 
 - *name*: 
 
@@ -302,8 +302,8 @@ None
 ```
 
 parameters:
+- *available_image_bounds*: 
 - *available_range*: 
-- *bounds*: 
 - *metadata*: 
 - *name*: 
 - *target_url*: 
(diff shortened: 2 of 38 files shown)


meshula (commented):
This looks good to me, not exposing the problematic functions through Python dodges the issue of inconsistent implementation on the Imath side.

At this point, what do we the think the answer is to this:

Pixel grid x axis:
[0 1 2 3 4 5 6 7 8 9]

Box 1: min.x 3, max.x 5
Box 2: min.x 5, max.x 8

Does pixel 5 belong to box 2? Or both box 1 and box 2?

My answer is that it belongs to box 2.

Whatever our consensus, I think we should lay it out in a line of documentation.

Does

rogernelson:
> Box 1: min.x 3, max.x 5 Box 2: min.x 5, max.x 8
> 
> Does pixel 5 belong to box 2? Or both box 1 and box 2?
> 
> My answer is that it belongs to box 2.
> 
> Whatever our consensus, I think we should lay it out in a line of documentation.

I 100% agree it belongs to box 2.  I'll tag @desruie so he can add this to the [doc](https://github.com/desruie/OpenTimelineIO/blob/autodesk_desruie_spatial_doc/docs/tutorials/spatial-coordinates.md) he started.

PR #1093, opened 2021-10-01
add build job that just builds c++

ssteinbach:
Adding a CI job that _just_ compiles the C++ from cmake, without compiling and running the python.


.github/workflows/python-package.yml
```diff
@@ -16,7 +16,25 @@ on:
     branches: [ master ]
 
 jobs:
-  build:
+  cpp-build:
```
JeanChristopheMorinPerso, line 19: Extremely minor: I would keep the naming consistent between jobs. So either use dashes underscores.

PR #1099, opened 2021-10-11
Publish wheels and sdist in PyPI

### LINKED ISSUE #988: PyPI Publish github action
## Description

To automate the release process, make it more predictable, and improve our security footprint;  PyPI publishes should be handled by a github action using artifacts generated by build actions.

## Context

All previous OTIO releases have been done via a manual checklist. A source tarball is then hand built as well as wheels built in either docker or a macOS VM on a developer machine. This is problematic because:

- It requires a human executing the the right commands each time to get proper results.
- It is time-consuming and stressful for the human running those commands.
- The results are potentially unpredictable -  the developer's machine state could impact the build in unexpected ways.
- It is a possible threat vector for malicious actors - if the developer who runs the commands has a compromised machine it could pollute the artifacts.

A PyPI publish action should be triggered by some github trigger. Ideally, by the end of a publish we should have:

- A version tag in the github repo
- A release in github
- Artifacts matching the ones associated with the github release and bearing the tag version published to PyPI

We should validate that the solution we make allows us to properly do pre-releases (I believe this is [governed completely by the version string](https://www.python.org/dev/peps/pep-0440/#pre-releases))

## References

The [otio-mlt-adapter plugin](https://github.com/apetrynet/otio-mlt-adapter/blob/main/.github/workflows/deploy.yaml) by @apetrynet appears to do exactly this already.

[OpenTimelineIO-Java-Bindings](https://github.com/OpenTimelineIO/OpenTimelineIO-Java-Bindings/blob/main/.github/workflows/publish.yml) also does something like this for releases.

## Questions
- Should we continue to drive setting the version string in the `setup.py` using hand-updates, or would we like to try to automate that in some way?
- Do we want to try enabling a way to push to the PyPI test repo at all? What should govern that?

JeanChristopheMorinPerso:
Publish wheels and sdist (source code only) in PyPI.

Fixes #988 .

The method I used is simple. I added a job in the `python-package` workflow that will only run when the GitHub Action is triggered by tag pushed in the repo.

That's the simplest option. But it has one major downside. The artifacts will be regenerated for the release. This is a problem because there could be some amount of time between the last PR is merged and the tag creation and during this time some things might have changed (transitive dependencies, GitHub Actions environment, etc). On Linux it isn't much a problem as wheels are generated in docker images, but on macOS and Windows it's another story.

A safer approach would be to have the publish job be in its own workflow. This workflow would fetch the wheels from the last successful build (or a specific build if we make the workflow that can be manually triggered with custom inputs). This is the approach the the [Java bindings](https://github.com/OpenTimelineIO/OpenTimelineIO-Java-Bindings) is using, with the difference that the publish workflow is triggered by a release creation, not a tag pushed.

I will be happy to go with later (safer) approach if TSC prefers that.

# Notes

* For this to work, you will need to add a secret named `PYPI_API_TOKEN` to the repository. The value needs to be a PyPI token created from the user interface (see https://pypi.org/help/#apitoken for instructions). You will need to specify a name when creating the token. I suggest choosing a name that clearly states that it's going to be used by GitHub actions .
* I left out GPG signing for now, but I'll be more than happy to add that if requested.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -12,6 +12,7 @@ env:
 on:
   push:
     branches: [ master ]
+    tags: [ '*' ]
   pull_request:
     branches: [ master ]
 
@@ -21,7 +22,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
-        python-version: [2.7, 3.7, 3.8, 3.9]
+        python-version: ['2.7', '3.7', '3.8', '3.9']
 
     env:
       OTIO_CXX_COVERAGE_BUILD: ON
@@ -102,3 +103,33 @@ jobs:
         with:
           name: wheels
           path: ./wheelhouse/*.whl
+
+  pypi_publish:
+    needs: package_wheels
+    runs-on: ubuntu-latest
+    if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags')
+    steps:
+      - uses: actions/checkout@v2.3.4
+        with:
+          submodules: 'recursive'
+
+      - name: Set up Python
+        uses: actions/setup-python@v2.2.2
+
+      - name: Install pypa/build
+        run: python -m pip install build --user
+
+      - name: Generate sdist
+        run: python -m build --sdist --outdir dist .
+
+      - name: Download wheels
+        uses: actions/download-artifact@v2
+        with:
+          name: wheels
+          path: dist
+
+      - name: Publish distribution 📦 to PyPI
+        uses: pypa/gh-action-pypi-publish@release/v1
+        with:
+          user: __token__
+          password: ${{ secrets.PYPI_API_TOKEN }}


ssteinbach:
It looks like it builds an sdist separately for the wheels - that is for platforms that don't have compatible wheels, correct?

JeanChristopheMorinPerso:
> I don't fully understand the risks of the less safe option. Is the risk that the released package could be broken due to drift outside the OTIO source code between the two builds?

Exactly this yes. On Windows and macOS the compiler could be updated, etc.

> Is there any security risk in there too?

I don't believe so.

> I think the 'verification' is less important here because in a way, the GitHub action itself is doing the verification -- the tests are running and passing and so on, using the wheels that will be uploaded to pypi, right?

Not yet unfortunately. We don't run the tests against the produced wheels. To run the tests, we run `pip install .[dev]` and then we run the tests. This actually creates a wheel and it installs it but this wheel it not the one we upload to PyPI.

> When we have done pypi package releases in the past, we did them to the test pypi index first, verified them, then pushed to the official one. Is there a similar step here where we can vet the final package before releasing it to the world? Perhaps that would come included in the safer approach you mentioned.

The "safer" approach is similar to what you described indeed.

JeanChristopheMorinPerso:
@ssteinbach I can certainly set the URL to the test PyPI. Just let me know once the credentials are setup in the repo and I'll do the change.

> Alternatively, it might just make sense to publish packages from commits on master to the test pypi. Is that commonly done?

As far as I know it's uncommon. Though what we could have is a manual action where you specify a build ID (https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/) and it would take care of publishing in test.pipy.org or something similar.

PR #1110, opened 2021-10-19
PyPI publish on release

### LINKED ISSUE #988: PyPI Publish github action

JeanChristopheMorinPerso:
Fixes #988 .

Take two of publishing to PyPI. This builds on #1099 .

This PR introduces a completely new workflow called `publish` (original isn't it?). This new workflow is triggered by a release. The `python-package` also has a separate job to build the `sdist`.

Compared to #1099 , the publishing is no more triggered by pushing tags.

The release process would look like this:
1. Merge all PRs that needs to go into the release.
2. Create a pull request for bumping the version to the one we want to use for the release, add changelog, etc.
3. Merge PR from step 2 and wait for the build to finish.
4. If we want to test the release, now it's time. We can use `maintainers/download_gha_artifact.py` to download the wheels locally and manually upload them in https://test.pypi.org if we want.
5. Once the artifacts are confirmed to be working as expected, we can tag the commit that resulted from the merge at step 3.
6. Push the tag.
7. Create a release in Github.
8. The new `publish` workflow will kick in. It will download the artifacts from the tag associated to the release and upload them to https://pypi.org.
9. Now bump the version to the next version and a `.dev...` at the end.

I tested all this in my OpenTimelineIO fork. You can see the commit that I used for the tag (v0.14.0.dev5) [here](https://github.com/JeanChristopheMorinPerso/OpenTimelineIO/actions/runs/1356958062) and the actual release [here](https://github.com/JeanChristopheMorinPerso/OpenTimelineIO/runs/3933341872?check_suite_focus=true). The release can be seen [here](https://test.pypi.org/project/jcmorin-OpenTimelineIO/0.14.0.dev5/)

This PR also introduces a script to download wheels from a build. Handy if you want to upload wheels in https://test.pypi.org for example. The script by default only requires to pass the commit hash (the full length one) and a GitHub Personal Access Token.

```
./maintainers/download_gha_artifact.py <commit hash> <your token>
```

Additional parameters can be used to set the workflow name from which to download the artifact., the artifact name and the directory where the artifact will be extracted to.

The output looks like:
```
Found workflow:
  Name:       OpenTimelineIO
  Branch:     pypi_upload
  Commit:     7c521b04a50df40596feae0c2802be7ff3198a16
  Committer:  {'name': 'Jean-Christophe Morin', 'email': 'jean_christophe_morin@hotmail.com'}
  Run Number: 87
  Status:     completed
  Conclusion: success
  URL:        https://github.com/JeanChristopheMorinPerso/OpenTimelineIO/actions/runs/1356958062
Getting list of artifacts
Downloading 'wheels' artifact and unzipping to 'dist'
Writing 'jcmorin_OpenTimelineIO-0.14.0.dev5-cp27-cp27m-macosx_10_9_x86_64.whl' to 'dist/jcmorin_OpenTimelineIO-0.14.0.dev5-cp27-cp27m-macosx_10_9_x86_64.whl'
... truncated for the purpose of the PR description
```

---

I hope this covers what we discussed in the last TSC meeting. If not let me know and I'll adjust.

### DIFF
--- a/maintainers/download_gha_artifact.py
+++ b/maintainers/download_gha_artifact.py
@@ -1,0 +1,116 @@
+#!/usr/bin/env python3
+"""
+This script downloads an artifact from a GitHub Action workflow run,
+unzips and and stores the files in a directory of your choice.
+
+Use cases:
+* You want to locally test or inspect the generated artifact.
+* You want to upload the artifact to a remote repository (test.pypi.org for example)
+  before a release.
+"""
+import io
+import os
+import sys
+import json
+import time
+import zipfile
+import argparse
+
+import urllib.request
+
+parser = argparse.ArgumentParser(
+    description="Download an artifact from a GitHub Action workflow run."
+)
+parser.add_argument("sha", help="Commit hash")
+parser.add_argument("token", help="GitHub Personal Access Token.")
+parser.add_argument(
+    "-w",
+    "--workflow",
+    default="OpenTimelineIO",
+    help="Name of the workflow to download artifact from.",
+)
+parser.add_argument(
+    "-a", "--artifact", default="wheels", help="Artifact name to download."
+)
+parser.add_argument(
+    "-d", "--directory", default="dist", help="Directory where to write the artifact."
+)
+
+args = parser.parse_args()
+
+headers = {
+    "Accept": "application/vnd.github.v3+json",
+    "Authorization": "token {args.token}".format(args=args),
+}
+
+if os.path.exists(args.directory) and os.listdir(args.directory):
+    sys.stderr.write(
+        "{0!r} directory contains files. It should be empty.".format(args.directory)
+    )
+    sys.exit(1)
+
+if not os.path.exists(args.directory):
+    os.makedirs(args.directory)
+
+request = urllib.request.Request(
+    "https://api.github.com/repos/PixarAnimationStudios/OpenTimelineIO/actions/runs?status=success",  # noqa: E501
+    headers=headers,
+)
+response = urllib.request.urlopen(request).read()
+workflow_runs = json.loads(response)["workflow_runs"]
+for run in workflow_runs:
+    if run["head_sha"] == args.sha and run["name"] == args.workflow:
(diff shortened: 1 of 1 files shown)


JeanChristopheMorinPerso:
> I propose we get an hour on the calendar, go through the workflow once with the TSC, and commit this, ASAP.

Makes sense. I'll let the TSC coordinate?

Also thanks @tin2tin for the time you took to test:)

PR #1143, opened 2021-11-13
Use platform specific command to find path

ThomasWilshaw:
**Summarize your change.**

Tests on Windows were failing as the test code uses the Unix command `which`. Here we change that to use a platform specific command (`where`) when running on Windows. `shutil.which()` cannot be used as it's Python3.3+ only.

I should also say I don't have a Linux machine to test on so this has only been tested on Windows.

### DIFF
--- a/tests/test_console.py
+++ b/tests/test_console.py
@@ -84,7 +84,7 @@ class ConsoleTester(otio_test_utils.OTIOAssertions):
             try:
                 subprocess.check_call(
                     [
-                        "which", sys.argv[0]
+                        "where" if os.name == "nt" else "which", sys.argv[0]
                     ],
                     stdout=subprocess.PIPE
                 )
--- a/tests/test_console.py
+++ b/tests/test_console.py
@@ -83,7 +83,6 @@ class ConsoleTester(otio_test_utils.OTIOAssertions):
     def run_test(self):
         if self.SHELL_OUT:
             # make sure its on the path
-            
             console_script = os.path.join(sysconfig.get_path('scripts'), sys.argv[0])
             if platform.system() == 'Windows':
                 console_script += '.exe'
--- a/tests/test_console.py
+++ b/tests/test_console.py
@@ -29,7 +29,6 @@ import sys
 import os
 import subprocess
 import sysconfig
-import fnmatch
 import platform
 
 try:
--- a/tests/test_console.py
+++ b/tests/test_console.py
@@ -30,6 +30,7 @@ import os
 import subprocess
 import sysconfig
 import fnmatch
+import platform
 
 try:
     # python2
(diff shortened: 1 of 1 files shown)


ssteinbach (approved):
considered doing this another way, but I think this is fine, actually.  Thanks!

tests/test_console.py
```diff
@@ -84,7 +84,7 @@ def run_test(self):
             try:
                 subprocess.check_call(
                     [
-                        "which", sys.argv[0]
+                        "where" if os.name == "nt" else "which", sys.argv[0]
```
JeanChristopheMorinPerso, line 87: I'm wondering if doing something like:
```python
import os
import sysconfig

console_script = os.path.join(sysconfig.get_path('scripts'), sys.argv[0])
if not os.path.exists(console_script):
    self.fail(
        "Could not find '{}'.  Tests that explicitly shell"
        " out can be disabled by setting the environment variable "
        "OTIO_DISABLE_SHELLOUT_TESTS.".format(console_script)
    )
```
might be a little bit better. It would at least guarantee that the console script is really the one we are looking for.

meshula: it's true that would be a more exact test

jfpanisset: Although in common use, "which" is not POSIX, and is being deprecated by Debian. "command -v" is the POSIX compliant alternative. https://lwn.net/ml/debian-devel/9e3a2df146f73437fe8cce10272ddf3913e48757.camel@debian.org/

meshula: Removing `which` is weird, it's been in common use for decades. I took a peek at the issue thread, sigh... If some yak shaving is going on in the debian bike shed, then using Jean-Christophe's suggestion is a good way to avoid having a yak in that race.

ThomasWilshaw: Unfortunately this fails on Windows as the files all have a `.exe` extension. Would this be an exceptable addition:
```python
 console_script = os.path.join(sysconfig.get_path('scripts'), sys.argv[0])
            if not os.path.exists(console_script):
                # We might be in Windows
                if not os.path.exists(console_script + ".exe"):
                    self.fail(
                    "Could not find '{}'.  Tests that explicitly shell"
                    " out can be disabled by setting the environment variable "
                    "OTIO_DISABLE_SHELLOUT_TESTS.".format(console_script)
                    )
```

meshula: This PR was previously approved, but conversation has stalled. Where do we stand? As is? Or do we want further modifications?

JeanChristopheMorinPerso: What about

```
console_script = os.path.join(sysconfig.get_path('scripts'), sys.argv[0])
If platform.platform() == 'windows':
    console_script += '.exe'

if not os.path.exists(console_script):
    self.fail(
        "Could not find '{}'.  Tests that explicitly shell"
        " out can be disabled by setting the environment variable "
        "OTIO_DISABLE_SHELLOUT_TESTS.".format(console_script)
    )
```

meshula: Ok, yes, that *almost* works for me. It needs to be platform.system() and check for Capitalized `Windows`. platform.platform() returns `Windows-10-10.0.19041-SP0` or similar depending... platform.system() returns `Windows`. On Mac, it platform.platform returns `macOS-12.0-arm64-arm-64bit`, platform.system() returns `Darwin`, so clearly platform() returns a tuple cross on any platform.

If you add your variant, with the change `platform.system() == 'Windows'` on the PR, I'm happy to land it when the tests pass.

meshula:
linux failing with 

```
./tests/test_console.py:32:1: F401 'fnmatch' imported but unused
15
./tests/test_console.py:87:1: W293 blank line contains whitespace
16
make: *** [Makefile:146: lint] Error 1
17
Error: Process completed with exit code 2.
```
any thoughts?

ThomasWilshaw:
Sorry I'd left that in from a previous test. I've removed it in the latest commit.

meshula (approved):

meshula:
Thanks!

PR #1144, opened 2021-11-13
Improve the developer quickstart documentation

JeanChristopheMorinPerso:
As note in https://academysoftwarefdn.slack.com/archives/CMQ9J4BQC/p1636760978147000, the developer quickstart documentation is out of date right now.

This PR removes obsolete instructions, adds some more instructions and also use modern methodologies (e.g. call `python -m pip` install of `pip`, don't call `python setup.py`, etc).

Hopefully it should be clear enough and the additions I made will help future contributors.

### DIFF
--- a/docs/tutorials/quickstart.md
+++ b/docs/tutorials/quickstart.md
@@ -13,7 +13,7 @@ You probably want the prebuilt binary for your platform.  PySide2 generally incl
 
 ## Install OTIO
 
-- `pip install opentimelineio`
+- `python -m pip install opentimelineio`
 
 ## Configure Environment Variables for extra adapters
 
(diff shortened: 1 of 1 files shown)


JeanChristopheMorinPerso:
@ThomasWilshaw Do you think the proposed changed would have made your experience better?

docs/tutorials/quickstart.md
```diff
@@ -55,41 +55,28 @@ The CMAKE_INSTALL_PREFIX variable must be set to a path with no spaces in it, be
 
 ## To build OTIO for Python development:
 
-### Linux
+**Note** You are highly encouraged to use virtual environements to develop OTIO.
```
meshula, line 58: typo: environments
suggest:
**Note** The use of virtual environments is highly encouraged for OTIO development.

docs/tutorials/quickstart.md
```diff
@@ -105,9 +92,9 @@ To use opentime without opentimelineio, link with -lopentime instead, and compil
 
 ### Linux / GDB / LLDB
 
-From your virtual environment, compile with debug flags:
+From your virtual environment, compile with debug flags (also works on Windows):
```
meshula, line 95: the syntax `env foo=1 python` doesn't actually work on windows

meshula (changes requested):
Thanks for the update! I've noted a couple of issues.

docs/tutorials/quickstart.md
JeanChristopheMorinPerso: environements: The merge of the french word `environnements` and the english word `environments` :upside_down_face: 

I will fix that and I will also rephrase as you suggested. Your suggestion looks better than mine.

docs/tutorials/quickstart.md
JeanChristopheMorinPerso: What do you think if I go with:

> To compile in debug mode, set the `OTIO_CXX_DEBUG_BUILD` environment variable to any value and then `python -m pip install .`

?

meshula: Sure!

ThomasWilshaw:
> 
> 
> @ThomasWilshaw Do you think the proposed changed would have made your experience better?

Yes that's definitely clearer but there's still a few things I don't follow.

1. In the section "To build OTIO for Python development", where is the given command meant to be run from? The source directory or somewhere else? Also is a note of what the `--user` option does necessary or will most people know what it's for already?

2. In the section: "To build OTIO for both C++ and Python development" it implies that you should run `setup.py` to create all the build files and compile OTIO, is that what we have just done in the previous step when we ran `python -m pip install . --user`? It also says using `setup.py` is highly discouraged, despite just saying to run it. 

Am I right in thinking that if I were to edit the code base in a way that affected both the Python and C++ sections, the best way for me to compile everything and run the tests would be to run `python -m pip install. --user` whereas if I were just modifying the C++ code I should use the standard `cmake --build...` command?

I should just clarify I'm not a very experienced dev (and I particularly don't like actually compiling software ;)) so I could just be misunderstanding things or generally no know what I'm doing :)

JeanChristopheMorinPerso:
@ThomasWilshaw I just pushed an update which hopefully should answer your questions. Let me know if it's still unclear.

meshula:
One thing we haven't gotten to is a comprehensive test suite at the C++ level. We are relying on the python bindings to expose problems. If you are developing in the C++ code, you should do a pip install and run the python tests before submitting a PR. It's up to you whether you also do a pip install while you are iterating the C++ code.

New C++ PRs are expected to have a C++ based test to go along with them. We will need to add a comprehensive test suite at the C++ level.

ThomasWilshaw:
> @ThomasWilshaw I just pushed an update which hopefully should answer your questions. Let me know if it's still unclear.

That's defintely a lot clearer thanks.

meshula (approved):

PR #1147, opened 2021-11-15
Fix RV plugin test timing

### LINKED ISSUE #1146: The RV application plugin test fails

The RV application plugin test fails when it is run.  Output:
```
FAIL: test_read_otio_file (test_rv_reader.RVSessionAdapterReadTest)
  File "/Users/nelsonr/git/public/OpenTimelineIO/contrib/opentimelineio_contrib/application_plugins/tests/test_rv_reader.py", line 219, in test_read_otio_file
    self.assertEqual(clip2, 'clip2.mov')
AssertionError: 'clip3.mov' != 'clip2.mov'

Ran 3 tests in 14.975s
```

## To Reproduce

1. Operating System: Mac OS 11.6
2. Python version: System (Python 2.7.16)
3. Example snippet that demonstrates the issue - if it's a build issue: 
4. OpenTimelineIO release version or commit hash: 1ba6d78
5. Compiler information: Apple LLVM version 10.0.0 (clang-1000.11.45.2)

## Expected Behavior
The tests pass

rogernelson:
Fixes #1146 

**Summarize your change.**

The tests for the RV plugin will always fail when they are run.  This is because the setup creates three clips with a duration of RationalTime(30,24) and the test is expecting new clips at RV global frames 1, 20, and 40.  

To fix the issue, I just lowered the duration to RationalTime(3,24) and check for clips at RV global frames 1, 4 and 7. 

Only the modified test: test_rv_reader.py

### DIFF
--- a/contrib/opentimelineio_contrib/application_plugins/tests/test_rv_reader.py
+++ b/contrib/opentimelineio_contrib/application_plugins/tests/test_rv_reader.py
@@ -84,7 +84,7 @@ for clipnum in range(1, 4):
             ),
             source_range=otio.opentime.TimeRange(
                 otio.opentime.RationalTime(11, 24),
-                otio.opentime.RationalTime(30, 24)
+                otio.opentime.RationalTime(3, 24)
             )
         )
     )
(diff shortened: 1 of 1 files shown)


meshula (approved):

jminor:
Thanks @rogernelson !

PR #1149, opened 2021-11-15
Fix RV plugin test for Mac

### LINKED ISSUE #1148: RV plugin test does not run on Mac
The RV plugin test does not run on Mac.

```
$ python -m unittest discover -s . 
.EE
ERROR: test_install_plugin (test_rv_reader.RVSessionAdapterReadTest)
  File "/Users/nelsonr/git/public/OpenTimelineIO/contrib/opentimelineio_contrib/application_plugins/tests/test_rv_reader.py", line 121, in test_install_plugin
    rc = install_package(source_package_path)
  File "/Users/nelsonr/git/public/OpenTimelineIO/contrib/opentimelineio_contrib/application_plugins/tests/test_rv_reader.py", line 244, in install_package
    rc = call(shlex.split(install_cmd))
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/subprocess.py", line 172, in call
    return Popen(*popenargs, **kwargs).wait()
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/subprocess.py", line 394, in __init__
    errread, errwrite)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/subprocess.py", line 1047, in _execute_child
    raise child_exception
OSError: [Errno 2] No such file or directory

ERROR: test_read_otio_file (test_rv_reader.RVSessionAdapterReadTest)
  File "/Users/nelsonr/git/public/OpenTimelineIO/contrib/opentimelineio_contrib/application_plugins/tests/test_rv_reader.py", line 161, in test_read_otio_file
    install_package(source_package_path)
  File "/Users/nelsonr/git/public/OpenTimelineIO/contrib/opentimelineio_contrib/application_plugins/tests/test_rv_reader.py", line 244, in install_package
    rc = call(shlex.split(install_cmd))
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/subprocess.py", line 172, in call
    return Popen(*popenargs, **kwargs).wait()
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/subprocess.py", line 394, in __init__
    errread, errwrite)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/subprocess.py", line 1047, in _execute_child
    raise child_exception
OSError: [Errno 2] No such file or directory

Ran 3 tests in 0.019s

FAILED (errors=2)
```

## To Reproduce

1. Operating System: Mac OS 11.6
2. Python version: System (2.7.16)'
3. Example snippet that demonstrates the issue - if it's a build issue:
4. OpenTimelineIO release version or commit hash: 1ba6d78
5. Compiler information: Apple LLVM version 10.0.0 (clang-1000.11.45.2)

## Expected Behavior

The test is able to run

rogernelson:
Fixes #1148

**Summarize your change.**

There are a few Mac OS-specific fixes here:
- The RV_BIN_DIR is expected to be named 'bin'.  This is not the case on Mac where RV is packaged as a Mac application and the binaries relative to the install root are in `RV.app/Contents/MacOS/`.  This is fixed by checking for Mac and changing `bin` for `MacOS` if it's the case.

- The package installation tries to verify the package was installed correctly by matching the installation path to the list of RV packages.  This fails because the test installs in Mac's tempdir (`/var/tmp`) which is actually a symlink to `/private/var/tmp`, and RV displays the path with the symlink resolved.  Fix this by using `realpath` to resolve the symlinks prior to the installation so the paths match.

- On Mac the executable name is upper-case, so add a check for executable name based on OS.

- Increase the wait time for rvNetwork to start.  On my machine at least it took more than 10 iterations through the loop, so I doubled it.

- Add a sleep of 10 seconds after the network connection succeeds.  There is a delay between the launch of the network server and the time RV is fully initialize.  We'll give it lots of time to ensure start up is complete before we start issuing commands.

Only the modified test is effected: test_rv_reader.py

### DIFF
--- a/contrib/opentimelineio_contrib/application_plugins/tests/test_rv_reader.py
+++ b/contrib/opentimelineio_contrib/application_plugins/tests/test_rv_reader.py
@@ -38,6 +38,7 @@ import shutil
 import shlex
 import time
 import imp
+import platform
 from subprocess import call, Popen, PIPE
 
 import opentimelineio as otio
@@ -46,7 +47,10 @@ RV_OTIO_READER_NAME = 'Example OTIO Reader'
 RV_OTIO_READER_VERSION = '1.0'
 
 RV_ROOT_DIR = os.getenv('OTIO_RV_ROOT_DIR', '')
-RV_BIN_DIR = os.path.join(RV_ROOT_DIR, 'bin')
+RV_BIN_DIR = os.path.join(
+    RV_ROOT_DIR,
+    'MacOS' if platform.system() == 'Darwin' else 'bin'
+)
 
 RV_OTIO_READER_DIR = os.path.join(
     '..',
@@ -146,7 +150,7 @@ class RVSessionAdapterReadTest(unittest.TestCase):
             'I L - {version} "{package_name}" {pkg_path}'.format(
                 version=RV_OTIO_READER_VERSION,
                 package_name=RV_OTIO_READER_NAME,
-                pkg_path=installed_package_path
+                pkg_path=os.path.realpath(installed_package_path)
             )
 
         self.assertIn(desired_result, stdout.split('\n'))
@@ -171,13 +175,14 @@ class RVSessionAdapterReadTest(unittest.TestCase):
             delete=False
         )
         otio.adapters.write_to_file(sample_timeline, sample_file.name)
-        run_cmd = '{root}/rv ' \
+        run_cmd = '{root}/{exe} ' \
                   '-nc ' \
                   '-network ' \
                   '-networkHost localhost ' \
                   '-networkPort {port} ' \
                   '{sample_file}' \
                   .format(
+                      exe='RV' if platform.system() == 'Darwin' else 'rv',
                       root=RV_BIN_DIR,
                       port=9876,
                       sample_file=sample_file.name
@@ -196,10 +201,15 @@ class RVSessionAdapterReadTest(unittest.TestCase):
                 if not rvc.connected:
                     time.sleep(.5)
 
-                if attempts == 10:
+                if attempts == 20:
                     raise socket.error(
                         "Unable to connect to RV!"
                     )
+
+            # some time can pass between the RV connection
+            # and the complete startup of RV
+            print("Waiting for RV startup to complete")
+            time.sleep(10)
 
             # Check clips at positions
             clip1 = rv_media_name_at_frame(rvc, 1)


contrib/opentimelineio_contrib/application_plugins/tests/test_rv_reader.py
```diff
@@ -196,11 +201,16 @@ def test_read_otio_file(self):
...
-                if attempts == 10:
+                if attempts == 20:
                     raise socket.error(
                         "Unable to connect to RV!"
                     )
 
+            # some time can pass between the RV connection
+            # and the complete startup of RV
+            print("Waiting for RV startup to complete")
+            time.sleep(10)
```
jminor, line 212: Is this sleep(10) meant to be in addition to the sleep(.5) above? So after rvc is connected it still needs to wait for RV to finish launching? I wonder if there's a way to detect that RV is fully ready within the while loop that would avoid waiting longer than necessary.
I know this test is not usually run for most people, so it might not matter too much, but the faster the tests are the more likely they will be run regularly.
If this is not an easy fix, I'm fine with this PR as-is. Someday we can have CI runners with RV installed and this can be reassessed at that time.

rogernelson: > Is this sleep(10) meant to be in addition to the sleep(.5) above?

Yes, those sleeps are waiting for rvNetwork to connect whereas this one is waiting between the time the connection succeeds and the full startup of RV. 

> So after rvc is connected it still needs to wait for RV to finish launching? 

Yes, there is a delay between the time the server loads and RV is fully initialized.  I am running a debug build of RV and on an 8 year old Mac, so those factors probably make the delay more significant on my machine.  I can try a release build if you think it's worth seeing if we can lower that value, but I thought i was worth keeping something reasonably large in case someone else runs into the same issue.

As an alternative, I could also put the sleep value in an envar that defaults to 0 and print a message if the first command fails that the envar may need to be set.

meshula:
I was reacting to the indeterminacy of it. I've worked in test suites previously that had timing-imposed accommodations, and the problems just compound over time when automation is involved. This test isn't in that kind of an environment where there is a tree of test-dependencies (e.g. another test waiting on this one), so don't consider my comment a blocker on the review.

jminor (approved):

contrib/opentimelineio_contrib/application_plugins/tests/test_rv_reader.py
jminor: Ok, I understand it better now. Let's leave the speed optimization for another time. Having it run correctly now is a great improvement.

rogernelson: Sounds good.  Do you want me to open an issue for a followup and assign it to myself?

PR #1151, opened 2021-11-18
AAF adapter: added support for keyframed properties

### LINKED ISSUE #1135: AAF adapter: animated parameters are omitted
## Feature Request

Animated AAF properties are omitted during translation. We would want do add them back in.

## Description

**Updated: adressed issue of non-linear values**

Currently the AAF adapter does not transcribe any animated parameters / properties from the source AAF file. If a parameter is animated `aaf2.misc.VaryingValue` (e.g. animated grid effect), the parameter will be omitted from being translated to OTIO. This results in very inconsistent results based on the animation state of each property.

This issue is easily fixed, but it poses a big problem when it comes to backwards compatibility. How should animated values be stored in OTIO metadata without breaking the existing property translation? Ideally animated parameters would have been included in design efforts from the beginning, with static parameters having identical representation - even just with one sample. Making this change now, would potentially break a lot of things. 

Instead we propose a halfway solution, where animated properties get wrapped in a dictionary that contains samples and interpolation type. To address the issue with non-linear interpolation curves, we can utilize `pyaaf2` integrated baking functionality - baking one sample per frame.

```json
# current parameter translation, omitting all animated values

"Parameters": {
    "AFX_GRID_TYPE": 0,
    "AFX_SCALE_Y_U": 74
}

# proposed parameter translation, including animated values
# non-animated parameters stay the same for back-compatibility

"Parameters": {
    "AFX_GRID_TYPE": 0,
    "AFX_SCALE_Y_U": 74
    "AFX_SCALE_Z_U": {
        "_aaf_keyframed_property": true,
        "keyframe_baked_values": [
            [
                0.0,
                22.0
            ],
            [
                1.0,
                22.5
            ],
            ...
            [
                254.0,
                100.0
            ]
       ],
        "keyframe_interpolation": "Linear",
        "keyframe_values": [
            [
                0.0,
                22.0
            ],
            [
                40.0,
                100.0
            ]
       ]
    }
}

```

timlehr:

Fixes #1135

**Summarize your change.**

Keyframed properties are not longer omitted and translated as a dictionary containing control points, interpolation type and optionally baked values. See issue #1135 for more details.

```json
# current parameter translation, omitting all animated values

"Parameters": {
    "AFX_GRID_TYPE": 0,
    "AFX_SCALE_Y_U": 74
}

# proposed parameter translation, including animated values
# non-animated parameters stay the same for back-compatibility

"Parameters": {
    "AFX_GRID_TYPE": 0,
    "AFX_SCALE_Y_U": 74
    "AFX_SCALE_Z_U": {
        "_aaf_keyframed_property": true,
        "keyframe_baked_values": [
            [
                0.0,
                22.0
            ],
            [
                1.0,
                22.5
            ],
            ...
            [
                254.0,
                100.0
            ]
       ],
        "keyframe_interpolation": "Linear",
        "keyframe_values": [
            [
                0.0,
                22.0
            ],
            [
                40.0,
                100.0
            ]
       ]
    }
}

```

`test_aaf_adapter.py`


jminor:
We did some review of this over video chat. We discussed some pros/cons of putting this into the AAF metadata & all agreed that this is a good step forward.

OTIO doesn't have any versioning mechanism for adapter-specific metadata (this is all within the "aaf" metadata dictionary) and we have not stated any guarantees about the stability of that metadata. Further this change is purely additive, so it should be harmless to anyone relying on the current AAF metadata.

In the future, having this as reference will help when we want to add official OTIO schema for animated parameters.

Having the toggle (off by default) for baking the values per frame is good. That computation might be slow & the result large, so it is good to be off by default, but having the option will be useful for pipelines where the consumer cannot do the keyframe math to compute the value.

Once the tests pass, we can accept this.

jminor (approved):
Once the CI passes, this looks ready to land.

timlehr:
@jminor CI passes now. Feel free to merge. :+1:

jminor:
Thanks @timlehr and @fabal 👍

PR #1161, opened 2021-11-27
C++ tests

### LINKED ISSUE #1153: Comprehensive C++ test suite
## Feature Request

Full C++ test suite

## Description

We currently rely on the Python tests to fully exercise the C++ implementation. We should thoroughly test the C++ implementation, via C++.

## Context

This is a prerequisite for a 1.0 release.

I propose that we port the C binding tests to C++ as they are thorough, and can be transliterated to C++ with a small effort.

darbyjohnston:
Fixes #1153

Add C++ tests using the CMake built-in CTest functionality.

Tests may be enabled/disabled with the CMake variable ```BUILD_TESTING``` (default is enabled).

The tests can be run with the ```test``` target (Makefile generators), or the ```RUN_TESTS``` target (IDE generators). Example output of running the tests:
```
Build started...
1>------ Build started: Project: RUN_TESTS, Configuration: Debug x64 ------
1>Test project C:/dev/otio/OpenTimelineIO-Debug
1>    Start 1: test_opentime
1>1/2 Test #1: test_opentime ....................   Passed    0.04 sec
1>    Start 2: test_clip
1>2/2 Test #2: test_clip ........................   Passed    0.02 sec
1>
1>100% tests passed, 0 tests failed out of 2
1>
1>Total Test time (real) =   0.07 sec
========== Build: 1 succeeded, 0 failed, 1 up-to-date, 0 skipped ==========
```

I have just added a couple of tests for now to get feedback before adding more. I've tried to follow the style and content of the Python tests to make comparison easy between the two languages.


tests/test_opentime.cpp
```diff
@@ -0,0 +1,39 @@
...
+main(int argc, char** argv) {
+    Tests tests;
+
+    tests.add_test("test_create", [] {
+        double              t_val = 30.2;
+        otime::RationalTime t(t_val);
+        assertEqual(t.value(), t_val);
+
+        t = otime::RationalTime();
+        assertEqual(t.value(), 0.0);
```
KarthikRIyer, line 17: Could we add an epsilon to this assert, to compare floating points?
In the C tests I use a value (suggested by Nick) of 5.82077e-11. Reason as mentioned in the [C test](https://github.com/OpenTimelineIO/OpenTimelineIO-C-Bindings/blob/main/tests/epsilon.h):
```
/**
 * We are not testing values outside of one million seconds.
 * At one million second, and double precision, the smallest
 * resolvable number that can be added to one million and return
 * a new value one million + epsilon is 5.82077e-11.
 *
 * This was calculated by searching iteratively for epsilon
 * around 1,000,000, with epsilon starting from 1 and halved
 * at every iteration, until epsilon when added to 1,000,000
 * resulted in 1,000,000.
 */
```

Also, I could help writing a few tests if you'd like.

darbyjohnston: Good suggestion about the epsilon, I'll add that to specialized versions of ```assertEqual()``` and ```assertNotEqual()``` for double types.

Helping out with tests would be great, thanks! I just added a couple of tests to get feedback about the style first.

One thing I was wondering, does there need to be a single ```main()``` function for the tests to get the code coverage to work?

darbyjohnston: OK, I added the epsilon for double comparison, check it out and see what you think. I hope you don't mind, I copied your comment about the epsilon value.

KarthikRIyer: This looks good!

W.r.t single main(), I'm not sure because I don't have experience using coverage with C/CPP. I looked at OCIO. Correct me if I'm wrong, I think they are using a single main().

https://github.com/AcademySoftwareFoundation/OpenColorIO/tree/main/tests/cpu

meshula:
This is a great start. I'm inclined to land it to make it easy for people to dogpile and get it done.

Also, does enabling running the tests via CI require a tweak to a configuration setting somewhere?

darbyjohnston:
Thanks! I'm about to commit another change to use a single ```main()``` entry point for the tests, I'm pretty sure that's needed with static linking for the code coverage.

For the CI, it should just require adding another step that builds the ```test``` target.

darbyjohnston:
OK, I modified the code to use a single ```main()``` entry point, sort of similar to OCIO but without the macros.

darbyjohnston:
I modified the GitHub actions to run the C++ tests in the "cpp_build" job. I also used static linking to avoid path issues with DLLs on Windows when running the tests.

I still need to hook up the code coverage... This CMake modules comes up when searching for "CMake code coverage":
https://github.com/bilke/cmake-modules/blob/master/CodeCoverage.cmake

Having the code coverage integrated in CMake seems nice?

tests/test_opentime.cpp
darbyjohnston: It looks like the coverage accumulates, so a single main() is not needed:
"The execution counts are cumulative. If the example program were executed again without removing the .gcda file, the count for the number of times each line in the source was executed would be added to the results of the previous run(s). This is potentially useful in several ways. For example, it could be used to accumulate data over a number of program runs as part of a test verification suite, or to provide more accurate long-term information over a large number of program runs."
https://linux.die.net/man/1/gcov

darbyjohnston:
It looks like the coverage accumulates, so a single main() is not needed:
"The execution counts are cumulative. If the example program were executed again without removing the .gcda file, the count for the number of times each line in the source was executed would be added to the results of the previous run(s). This is potentially useful in several ways. For example, it could be used to accumulate data over a number of program runs as part of a test verification suite, or to provide more accurate long-term information over a large number of program runs."
https://linux.die.net/man/1/gcov

meshula:
Looking at that module, it seems complex enough that I wouldn't want to conflate work on automating coverage through cmake with getting c++ tests up and running...!

darbyjohnston:
Agreed! After looking into it more I don't think that module is necessary, instead I was able to get something working similar to what's in the current Makefile. This is what the coverage looks like with the first two tests:
```
Reading tracefile coverage.filtered.info
                                               |Lines      |Functions|Branches  
Filename                                       |Rate    Num|Rate  Num|Rate   Num
================================================================================
[src/]
opentime/rationalTime.h                        |33.3%    69|40.9%  22|    -    0
opentime/stringPrintf.h                        |57.1%     7|11.1%   9|    -    0
opentime/timeRange.h                           |40.6%    32|45.5%  11|    -    0
opentime/timeTransform.h                       | 0.0%    14| 0.0%   6|    -    0
opentimelineio/anyDictionary.h                 |82.4%    34|87.5%   8|    -    0
opentimelineio/anyVector.h                     |75.0%    16|80.0%   5|    -    0
opentimelineio/clip.cpp                        |67.7%    31|87.5%   8|    -    0
opentimelineio/composable.cpp                  |37.0%    27|40.0%  10|    -    0
opentimelineio/composable.h                    | 0.0%     3| 0.0%   2|    -    0
opentimelineio/composition.cpp                 | 0.0%   293| 0.0%  31|    -    0
opentimelineio/composition.h                   | 0.0%    27| 0.0%   2|    -    0
opentimelineio/deserialization.cpp             |43.1%   367|46.9%  64|    -    0
opentimelineio/effect.cpp                      | 0.0%    13| 0.0%   5|    -    0
opentimelineio/errorStatus.cpp                 | 0.0%    52| 0.0%   1|    -    0
opentimelineio/errorStatus.h                   |28.6%    21|33.3%   6|    -    0
opentimelineio/externalReference.cpp           | 100%    13| 100%   5|    -    0
opentimelineio/externalReference.h             | 100%     3| 100%   1|    -    0
opentimelineio/freezeFrame.cpp                 | 0.0%     5| 0.0%   3|    -    0
opentimelineio/gap.cpp                         | 0.0%    18| 0.0%   7|    -    0
opentimelineio/generatorReference.cpp          | 0.0%    14| 0.0%   5|    -    0
opentimelineio/imageSequenceReference.cpp      | 0.0%   130| 0.0%  11|    -    0
opentimelineio/item.cpp                        |24.1%    79|35.7%  14|    -    0
opentimelineio/item.h                          | 100%     6| 100%   3|    -    0
opentimelineio/linearTimeWarp.cpp              | 0.0%    13| 0.0%   5|    -    0
opentimelineio/marker.cpp                      | 0.0%    15| 0.0%   5|    -    0
opentimelineio/mediaReference.cpp              |86.7%    15|66.7%   6|    -    0
opentimelineio/mediaReference.h                | 100%     5| 100%   2|    -    0
opentimelineio/missingReference.cpp            |46.2%    13|50.0%   6|    -    0
opentimelineio/serializableCollection.cpp      | 0.0%    44| 0.0%  11|    -    0
opentimelineio/serializableCollection.h        | 0.0%    20| 0.0%   1|    -    0
opentimelineio/serializableObject.cpp          |42.3%    97|47.6%  21|    -    0
opentimelineio/serializableObject.h            |68.4%   136|40.0% 110|    -    0
opentimelineio/seriali...ObjectWithMetadata.cpp| 100%    15|80.0%   5|    -    0
opentimelineio/serializableObjectWithMetadata.h| 100%     2| 100%   2|    -    0
opentimelineio/serialization.cpp               |53.6%   401|29.1% 151|    -    0
opentimelineio/stack.cpp                       | 0.0%    49| 0.0%  11|    -    0
opentimelineio/stringUtils.cpp                 |22.2%    27|16.7%   6|    -    0
opentimelineio/timeEffect.cpp                  | 0.0%     6| 0.0%   3|    -    0
opentimelineio/timeline.cpp                    | 0.0%    34| 0.0%   9|    -    0
opentimelineio/track.cpp                       | 0.0%   122| 0.0%  14|    -    0
opentimelineio/track.h                         | 0.0%     1| 0.0%   1|    -    0
opentimelineio/transition.cpp                  | 0.0%    33| 0.0%   9|    -    0
opentimelineio/transition.h                    | 0.0%     2| 0.0%   2|    -    0
opentimelineio/typeRegistry.cpp                |57.4%   129|61.5%  13|    -    0
opentimelineio/typeRegistry.h                  | 100%    13|56.5%  46|    -    0
opentimelineio/unknownSchema.cpp               | 0.0%    18| 0.0%   7|    -    0
opentimelineio/vectorIndexing.h                | 0.0%     2| 0.0%   2|    -    0
================================================================================
                                         Total:|32.1%  2486|33.6% 687|    -    0
```

This gets printed in the output of each GitHub action.

 I think the CodeCov upload is also working, though currently the pages show: "Oops, something went wrong", so I still need to look into that.

darbyjohnston:
Sorry for the delay, I think this is ready for review now. I've only implemented a couple of tests but it might be nice to merge this so other developers can help out with the rest of the test coverage.

Also I disabled uploading coverage reports for the C++ tests. It looks like Codecov doesn't support multiple reports and was getting confused having both the Python and C++ uploads. In the meantime developers can either run lcov locally, or check the log file from GitHub action run which will show the lcov results.

darbyjohnston:
Hi @JeanChristopheMorinPerso! I wanted to follow up from the meeting, do you have any suggestions for integrating the Python and C++ tests with CodeCov?

meshula:
TSC meeting note, we'll land this after discussion between Darby and Jean-Christophe around how to proceed on codecov.

JeanChristopheMorinPerso:
Hey @darbyjohnston , I don't know if you already did, but I would recommend reading https://docs.codecov.com/docs/flags. I just went through it and it seems like Codecov should be able to achieve what we want. They also have a nice video explains the feature: https://youtu.be/ZYEsHHohgqo.

Generally speaking, flags can be seen as coverage groups. Which means we can have a group for `src/py-opentimelineio` and another one for `src/opentime` + `src/opentimelineio`.

Though, for the python stuff, we might have to filter out `src/opentime` and `src/opentimelineio` (based on what the section "One-to-One Relationship of Flags to Uploads" says). But I would first try without changing the python coverage reports and only do the additional filtering if required.

If you already tried all that, I'll be happy to provide a second eye and see if I can find out why it didn't work for us.

Let me know.

meshula:
@darbyjohnston the -54% coverage is due (ironically?) to enabling code coverage for C++ right? So turning it off again puts our coverage ratio back where it was? If yes, let's turn it off again, land the testing framework, and start a new issue to sort out coverage. They are complementary aspects of CI, but I don't see that coverage should gate having tests.

darbyjohnston:
@JeanChristopheMorinPerso, thanks! I actually hadn't seen the docs about the flags, I'll check them out.

@meshula, the CodeCov upload is currently commented out for the C++ tests. However I do run ```lcov``` as part of the GitHub action to print out the coverage to the log. I think the -54% is from a previous CodeCov upload when I was playing around with it.

meshula (approved):

meshula:
Opened a new issue for code coverage. https://github.com/PixarAnimationStudios/OpenTimelineIO/issues/1233

Thanks for creating these tests!

PR #1170, opened 2021-12-06
Replaced @pixar email addresses with @aswf email addresses.

jminor:
In preparation for transfer of OpenTimelineIO from Pixar to ASWF, we are changing the contact email addresses to the OTIO TSC private mailing list.


ssteinbach (approved):
LGTM

PR #1180, opened 2021-12-21
added support for frame rates: 48000/1001, 100 , 120000/1001 and 120

### LINKED ISSUE #393: Missing timecode rates?
Seem to be missing 48 and 47.95 in opentime.py

splidje:
```Fixes #393```

added support for frame rates: 48000/1001, 100 , 120000/1001 and 120

### DIFF
--- a/src/opentime/rationalTime.cpp
+++ b/src/opentime/rationalTime.cpp
@@ -10,15 +10,17 @@ namespace opentime { namespace OPENTIME_VERSION {
 
 RationalTime RationalTime::_invalid_time{ 0, RationalTime::_invalid_rate };
 
-static constexpr std::array<double, 4> dropframe_timecode_rates{ {
+static constexpr std::array<double, 6> dropframe_timecode_rates{ {
     // 23.976,
     // 23.98,
     // 23.97,
     // 24000.0/1001.0,
     29.97,
     30000.0 / 1001.0,
+    48000.0 / 1001.0,
     59.94,
     60000.0 / 1001.0,
+    120000.0 / 1001.0,
 } };
 
 // currently unused:
@@ -37,7 +39,7 @@ static constexpr std::array<double, 10> non_dropframe_timecode_rates
 }};
 */
 
-static constexpr std::array<double, 16> valid_timecode_rates{
+static constexpr std::array<double, 20> valid_timecode_rates{
     { 1.0,
       12.0,
       23.97,
@@ -49,11 +51,15 @@ static constexpr std::array<double, 16> valid_timecode_rates{
       29.97,
       30000.0 / 1001.0,
       30.0,
+      48000.0 / 1001.0,
       48.0,
       50.0,
       59.94,
       60000.0 / 1001.0,
-      60.0 }
+      60.0,
+      100,
+      120000.0 / 1001.0,
+      120 }
 };
 
 bool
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/opentime/rationalTime.cpp [hunk 1]
```diff
@@ -10,15 +10,17 @@ namespace opentime { namespace OPENTIME_VERSION {
...
     // 23.976,
     // 23.98,
     // 23.97,
     // 24000.0/1001.0,
     29.97,
     30000.0 / 1001.0,
+    48000.0 / 1001.0,
     59.94,
     60000.0 / 1001.0,
+    120000.0 / 1001.0,
```
meshula, line 23: Is there any hardware that records at this rate?

src/opentime/rationalTime.cpp
```diff
@@ -10,15 +10,17 @@ namespace opentime { namespace OPENTIME_VERSION {
...
 
-static constexpr std::array<double, 4> dropframe_timecode_rates{ {
+static constexpr std::array<double, 6> dropframe_timecode_rates{ {
     // 23.976,
     // 23.98,
     // 23.97,
     // 24000.0/1001.0,
     29.97,
     30000.0 / 1001.0,
+    48000.0 / 1001.0,
```
meshula, line 20: Sadly, field recorders also offer 48000 * 1001 / 1000 as a rate, so it should be added as well.

src/opentime/rationalTime.cpp
```diff
@@ -49,11 +51,15 @@ static constexpr std::array<double, 16> valid_timecode_rates{
...
       30.0,
+      48000.0 / 1001.0,
       48.0,
       50.0,
       59.94,
       60000.0 / 1001.0,
-      60.0 }
+      60.0,
+      100,
+      120000.0 / 1001.0,
```
meshula, line 61: same question, is this a "real" rate?

src/opentime/rationalTime.cpp
```diff
@@ -118,7 +124,14 @@ RationalTime::from_timecode(
         unsigned int last_pos = 0;
         for (unsigned int i = 0; i < 4; i++)
         {
-            fields[i] = timecode.substr(last_pos, 2);
+            if (i < 3)
+            {
+                fields[i] = timecode.substr(last_pos, 2);
```
meshula, line 129: hours can exceed 99? or is that disallowed? If we're going to catch high frame counts, probably we should think about the full parse, and do the split based on separators, not a two character heuristic, if 100+ hours is legal...

src/opentime/rationalTime.cpp
reinecke: That is the record speed, but what timecode do they write?

meshula: Right, right. These values are meant for rates with associated SMPTE time codes, which does not include audio data. For some reason I was thinking that all the NTSC related rates were part of this stew, but they aren't.

src/opentime/rationalTime.cpp
splidje: I've purely based this on experimenting with 120 fps with Nuke and FFmpeg.

`ffmpeg -i <path> -timecode "00:00:00:119" <path>`

FFmpeg has no issue with that, and reports that timecode back.

Also Nuke reports it in the metadata and moving 1 frame forward it jumps to `00:00:01:00` and so on.

splidje: can't help feeling that back-in-the-day they could have at least pre-empted over 100 fps and possibly over 99 hours. Doesn't seem unfathomable.

meshula: To utilize framerates over 99, we'd have to modify the timecode parser and formatter to allow it ~ it's hardcoded for two digit codes at the moment.

Issue #190, opened 2017-12-06
Rate should be int/int not float


jminor:
Currently RationalTime is represented as a float value and a float rate. Unfortunately there are cases where floating point arithmetic leads to precision issues, rounding and undesirable drift.

Some other systems represent rate as an int/int rational number. For example 30000/1001 vs 29.97.

We would like to experiment with this alternate representation of rate for OTIO. In order to do so, we propose two concrete steps:
1. Write a unit test that clearly demonstrates a flaw in using a floating point rate.
2. Create two different time classes (perhaps RationalTime and FractionalTime?) that can be used interchangeably in OTIO, but with different representations of rate.

By doing this we can ensure that we won't break anything by making this switch and we can clearly explain to everyone (ourselves included) the difference between the two and the reasons for choosing one over the other.

Comments?

meshula:
TIL that the Python % operator and the C % operator are not the same thing. So I withdraw my flag of a bug in the timecode conversion code.

Creating a unit test to expose issues sounds like a great first step.

meshula:
Here is some excellent background reading. 

https://randomascii.wordpress.com/2012/02/13/dont-store-that-in-a-float/

The conclusion is that double precision will store a time value with constant precision at the microsecond level for 136 years.

So what's left is to work out is at what frame does precision break down for what frame rate. Since NTSC rates are irrational (30000/1001 == 29.9700299700299700...) several values should be checked beyond the usual integral multiples of 24, 25, and 30 - at least 25000/1001, 30000/1001, 60000/1001, and 120000/1001.

Since float operations potentially lose a little significance after every operation, the next question is: at the point precision breaks down, how much significance is lost for every operation?

https://benjaminjurke.com/content/articles/2015/loss-of-significance-in-floating-point-computations/

It might turn out that double is good enough, and it's kind of feeling like it, but at least we would know definitively.

quelsolaar:
Float and int aren't the only options. The time implementation I'm working on (Slowly Sorry!) internally stores values as a 64 bit integer and a 8 bit comma positioning value. This means i can do mixing of 24/30/29.97 FPS with perfect precision, and no loss. I think its in my Github branch (Ive been busy with other stuff and learning Python so things have slowed down) 

[...]

PR #1181, opened 2021-12-22
Add nearest_valid_timecode_rate function

splidje:
Added useful function nearest_valid_timecode_rate in lieu of access to the list of valid timecode rates in Python.

Example situation:

You've for some reason lost a significant figure on a frame rate: 23.97602397602397

`is_valid_timecode_rate(23.97602397602397)` returns `False`

`nearest_valid_timecode_rate(23.97602397602397)` returns `23.976023976023978`


meshula (approved):
looks like a good addition.

jminor:
This looks useful and clear. Could you add a test and a doc string for the new function?

tests/test_opentime.py
```diff
@@ -685,6 +685,19 @@ def test_passing_ndf_tc_at_df_rate(self):
         t2 = otio.opentime.from_timecode(NDF_TC, 29.97)
         self.assertEqual(t2.value, frames)
 
+    def test_nearest_valid_timecode_rate(self):
+        invalid_rate = 23.97602397602397
+        nearest_valid_rate = 23.976023976023978
+
+        self.assertFalse(otio.opentime.RationalTime.is_valid_timecode_rate(invalid_rate))
```
JeanChristopheMorinPerso, line 692: Builds show a linting error here: https://github.com/PixarAnimationStudios/OpenTimelineIO/runs/4704135694?check_suite_focus=true#step:7:13

jminor:
Could you add test cases for the common "abbreviated" NTSC rates. Specifically, 29.97, 23.98, 23.976 to ensure that those snap to the desired NTSC rate. Those truncated rates often appear in text fields, labels, etc. so they might find their way into this function.

splidje:
29.97, 23.98 and 23.976 are already in the valid timecode rates list.

Did you mean a test which checks e.g. 29.971 snaps to 29.97?

reinecke:
Ahh, right. I think this is a place where we might want to apply the [Robustness Principal](https://en.wikipedia.org/wiki/Robustness_principle) - we want to be permissive in what we accept (slightly rounded SMPTE rates are okay), but make our functions yield the technically correct SMPTE rates.

splidje:
Ok, so I guess I could make a separate array of the standards, and an array of acceptable roundings, then the `nearest_` function uses the former array?

splidje:
ok, sweet. I'll do that.

splidje:
Just a thought. Would it make sense to have a type like `RationalRate` to accurately represent a drop frame rate like 3000 / 1001 with the numerator and denominator?

reinecke:
Moving RationalTime to be a true rational number is something we've discussed, #190 has some more detail.
@ssteinbach and @meshula are spearheading some work to rigorously define OTIO's underlying mathematical framework. I suspect once that solidifies a bit we'll focus on discussing "RationalTime 2" - we'd certainly love to hear your perspectives!

meshula:
Yes, we have defined some rigorous math, and we have a prototype implementation that correctly (according to the unit tests at least) implements an integer rational number class that guarantees a certain precision, and tells you when you will lose precision by combining rates.

We've also got documentation under way to nail down nomenclature, so that we can all mean the same thing when we use terms of art, such as pull down, and so on.

All of this will show up at a tsc and the github repo during the course of this year, I expect.

PR #1189, opened 2022-01-09
Fix code coverage report uploads

JeanChristopheMorinPerso:
I recently noticed that the code coverage report uploads were no more running. The cause was that we forgot to un-bake the `GH_COV_OS` in the workflow after we release 0.14.0.


meshula (approved):
ah yes, I remember the conversation where we said we would do this, thanks for finding this omission!

PR #1194, opened 2022-01-15
Replace deprecated Pybind11 get_type() usages by py::type::of()

JeanChristopheMorinPerso:
**Summarize your change.**

I noticed while working on the docs for the Python bindings that the code base had a reference to a deprecated Pybind11 method, `get_type`. As stated in https://pybind11.readthedocs.io/en/stable/upgrade.html#v2-6, the replacement is `py::type::of`.

Quote from the doc:
> The undocumented h.get_type() method has been deprecated and replaced by py::type::of(h).

This PR replaces the `get_type` call with `py::type::of`.


meshula (approved):
lgtm!

Issue #1177, opened 2021-12-16
SVG Adapter erroneously assumes `media_reference` always has `target_url`


jlskuz:
### Incorrect Functionality and General Questions

(@KarthikRIyer) The SVG Adapter erroneously assumes `media_reference` always has `target_url`. Eg. `GeneratorReference` does not have `target_url`

## To Reproduce

This can be reproduced with any otio (like [test.otio.zip](https://github.com/PixarAnimationStudios/OpenTimelineIO/files/7723737/test.otio.zip) that has a  `media_reference`  with `"OTIO_SCHEMA": "GeneratorReference.1"`

Run `otioconvert -i test.otio -o test.svg` and you will get the output in the **Logs** section.

OS: Linux KDE Neon (Ubunt)
Python: 3.8.10
OpenTimelineIO: 0.14.1

## Expected Behavior

`otioconvert -i test.otio -o test.svg` should create `test.svg` instead of showing the error

## Logs

```
Traceback (most recent call last):
  File "/home/julius/.local/bin/otioconvert", line 8, in <module>
    sys.exit(main())
  File "/home/julius/.local/lib/python3.8/site-packages/opentimelineio/console/otioconvert.py", line 298, in main
    otio.adapters.write_to_file(
  File "/home/julius/.local/lib/python3.8/site-packages/opentimelineio/adapters/__init__.py", line 194, in write_to_file
    return adapter.write_to_file(
  File "/home/julius/.local/lib/python3.8/site-packages/opentimelineio/adapters/adapter.py", line 212, in write_to_file
    result = self.write_to_string(input_otio, **adapter_argument_map)
  File "/home/julius/.local/lib/python3.8/site-packages/opentimelineio/adapters/adapter.py", line 303, in write_to_string
    return self._execute_function(
  File "/home/julius/.local/lib/python3.8/site-packages/opentimelineio/plugins/python_plugin.py", line 174, in _execute_function
    return (getattr(self.module(), func_name)(**kwargs))
  File "/home/julius/.local/lib/python3.8/site-packages/opentimelineio/adapters/svg.py", line 1137, in write_to_string
    return convert_otio_to_svg(input_otio, width=width, height=height)
  File "/home/julius/.local/lib/python3.8/site-packages/opentimelineio/adapters/svg.py", line 1127, in convert_otio_to_svg
    draw_item(timeline, svg_writer, ())
[...]

PR #1195, opened 2022-01-16
SVG adapter fixes

KarthikRIyer:
Apologies for the delay. I haven't been keeping well for a while.

#1177

**Summarize your change.**

- The SVG Adapter erroneously assumed media_reference always has target_url. Eg. GeneratorReference does not have target_url. I've added a check for the attribute before drawing it.

- The adapter failed with timelines with transitions at the end of a track. Added a fix for that.


meshula (approved):
Sorry to hear you haven't been keeping well. But, thanks for the fix to the adaptor!

PR #1198, opened 2022-01-21
Remove extra useless parameter from RationalTime.__copy__ method

JeanChristopheMorinPerso:
I noticed that the `RationalTime.__copy__` method had an argument that is not used. The `__copy__` special method is not supposed to accept arguments.

From https://docs.python.org/3/library/copy.html:

> In order for a class to define its own copy implementation, it can define special methods \_\_copy\_\_() and \_\_deepcopy\_\_(). __The former is called to implement the shallow copy operation; no additional arguments are passed__. The latter is called to implement the deep copy operation; it is passed one argument, the memo dictionary.


meshula (approved):

jminor:
Does `__deepcopy__` have the same issue?

jminor (approved):

PR #1212, opened 2022-02-08
AAF reader supports muted clips

### LINKED ISSUE #483: AAF/XML Reader: Clip enabled/disabled/muted flag?
There currently seems to be no "status" flag that comes with a clip object to convey its state on the timeline regarding whether it is active, has been disabled, muted, etc... 

I'm not sure how this is handled on the AAF side of things, but taking a quick look at the XML, there are the fields for each track and clip **<enabled></enabled>** that carry this information in a easily accessible place. Is there somewhere appropriate within the OTIO clip structure that this information could be stored? Having the enabled status of both clips and tracks is fairly important in trying to figure out what on the timeline actually needs to be considered. 

Have put together a quick test Premiere project to try the different combinations. Files here:

[ClipEnableMuteTesting_Examples_1904048.zip](https://github.com/PixarAnimationStudios/OpenTimelineIO/files/3052887/ClipEnableMuteTesting_Examples_1904048.zip)

jminor:
```Fixes #483```

Now that #1175 has landed, this PR updates the AAF reader to use the new `enabled` boolean when it reads muted clips. This replaces the previous workaround where the AAF reader would place a Gap with the muted clip as metadata. Now each muted clip will be a normal Clip object with `enabled=False`.

I haven't tested with muted tracks yet.

### DIFF
--- a/contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py
+++ b/contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py
@@ -768,8 +768,8 @@ def _transcribe(item, parents, edit_rate, indent=0):
         # a multi cam thing
         if alternates and len(alternates) == 1:
             # Our result is the clip itself, but marked as enabled=False (muted)
-            metadata = copy.deepcopy(alternates[0].metadata["AAF"])
             result = alternates[0]
+            metadata = copy.deepcopy(result.metadata["AAF"])
             result.enabled = False
         else:
             metadata['alternates'] = alternates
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by andrewmoore-nz** [comment]
Hey @jminor. Looks like a good start. I think the multi-cam side of things might trip this up when there's only one alternate. I think we might need to go a bit more in-depth with the logic to look at the actual objects in play here e.g. whether the "Selected" object is a Filler (making it a muted object) or a SourceClip/Sequence, etc...

Also, the track-level mute side of things does need to be handled in a different way. This is something that I've just had to look at for Matchbox, and we're rolling in a fix to our branch shortly. The "enabled" flag will need to be set on the parent (Track <- TimelineMobSlot ), and the "Alternate", in this case an OperationGroup, needs to be parsed as normal. 

I'm going to be working on shifting our parser over to using the new "enabled" functionality this weekend so I can take a proper look in to it and we could go over it at some point?

**Comment by andrewmoore-nz** [comment]
Hey @jminor. Thanks for taking some time to take a look at this with me. 

As for the example files I provided with #483 - they were created using Adobe Premiere pretty early on in my journey to understanding all this stuff... almost three years ago now. Taking a quick look under the hood, it looks like Premiere is handling its clip muting process in a different way to Media Composer, so it makes sense that it may not be looking right to you. If you look at the XML file that was sent at the same time, you'll see there are plenty of objects in there with their `enabled` flag set to false e.g. `<enabled>FALSE</enabled>`

Regarding the `_DISABLE_CLIP_FLAG`, that was something I was seeing coming from an AAF out of Media Composer, but as it looks to be travelling as object metadata, likely not something that's part of the protocol and therefore won't be included in exports from other systems. So not something we can rely on, unfortunately. Here's an example of what I'm seeing... A `TimelineMobSlot` holding a `Selector` which has a `ComponentAttributeList` that includes the `TaggedValue` **"_DISABLE_CLIP_FLAG = 1"**

MC_MutedTrackStructure_220214

This metadata is already collected earlier in the `_transcribe()` function before we get to the `Selector` specific part of the `elif` block, so that's why I'm seeing it as part of my general metadata dict. As for how this situation is handled by other NLEs, I'm guessing it's probably going to be different, if the Premiere export is anything to go by. 

For now, we are adding in a clause here to allow for this to be handled properly, as last week I was seeing that the actual objects on the muted track were coming through incorrectly, so that the timelines were differing lengths. This seems to be working for now, but I intend on building it out properly once we switch over to using the new `enabled` flag:

```
    elif isinstance(item, aaf2.components.Selector):
        # A brute force check to see if what we're dealing with is a muted track
        if isinstance(parents[-1], aaf2.mobslots.TimelineMobSlot) and \
                isinstance(item.getvalue('Selected'), aaf2.components.Filler):
            alt = item.getvalue('Alternates')[0]
            result = _transcribe(alt, parents + [item], editRate)
```

jminor:
This PR was superseded by #1277

PR #1214, opened 2022-02-11
On macOS build universal arm64 and x86_64

jminor:
On macOS, we now build libopentime and libopentimelineio as universal (both arm64 and x86_64)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,6 +15,8 @@ set(OTIO_AUTHOR       "Contributors to the OpenTimelineIO project")
 set(OTIO_AUTHOR_EMAIL "otio-discussion@lists.aswf.io")
 set(OTIO_LICENSE      "Modified Apache 2.0 License")
 
+set(CMAKE_OSX_ARCHITECTURES "arm64;x86_64" CACHE STRING "")
+
 project(OpenTimelineIO VERSION ${OTIO_VERSION} LANGUAGES C CXX)
 
 #------------------------------------------------------------------------------


JeanChristopheMorinPerso:
Nice! FYI, you will have to modify the CI to output `universal2` tagged wheels. See https://cibuildwheel.readthedocs.io/en/1.x/options/#archs. This can be done here: https://github.com/PixarAnimationStudios/OpenTimelineIO/blob/main/.github/workflows/python-package.yml#L117.

I just don't know if it's going to work for Python less than 3.8.

meshula:
@JeanChristopheMorinPerso I assume you mean, after this change lands, we can update the CI for universal tagged builds?

meshula:
Also, why 3.8 specifically? 3.7.x was when M1 support appeared?

JeanChristopheMorinPerso:
I guess wheel support could be added later...

As for why 3.8+, see https://cibuildwheel.readthedocs.io/en/stable/faq/#universal2

More specifically:
> Only CPython 3.8 and newer support universal2 and arm64 wheels.

But the whole section is worth a read.

JeanChristopheMorinPerso:
> I guess wheel support could be added later...

The triple dots were because I wasn't sure of the end result of having universal libs while having "non aware" wheels. But thinking about it more, all the tests are passing on all platforms and python versions. So everything will be fine.

Also I realize the tone of my initial comment kind of sounded like an order or authoritive. I'm really sorry for that. Thanks @meshula for pointing it out (kind of indirectly, but still).

The idea behind my comment was more to remind that it's something we shouldn't forget to do before the next release.

meshula:
No need to be sorry, there wasn't a tone problem, I was just looking for clarification :) I wanted to know if (1) we can land this change and it won't disrupt wheels (I understand from your comment that it won't) and (2) if the update to the wheel CI can be a second change (I understand from your comment that we can do it as a second step).

PR #1221, opened 2022-02-20
Port to PySide6

### LINKED ISSUE #1215: Upgrade otioview to PySide6
Currently otioview uses PySide2, which is quite old. In some newer versions of Python, or newer hardware (e.g. Mac M1) it is difficult to build or install PySide2. We should upgrade otioview to PySide6 (so named since it uses Qt 6).

At first glance, the API seems to have only minimal changes between PySide2 and PySide6, so maybe otioview could be made to work with both? If someone can do the port to PySide6, then we can judge if making it work on both is possible.

This might also be a good time to re-examine #616

### LINKED ISSUE #1220: PySide6 removes QRegExp
## Problem: PySide6 removes QRegExp

```
  File "miniforge3/envs/otio-py310/lib/python3.10/site-packages/opentimelineview/details_widget.py", line 87, in highlightBlock
    expression = QtCore.QRegExp("(\\{|\\}|\\[|\\]|\\:|\\,)")
AttributeError: module 'PySide6.QtCore' has no attribute 'QRegExp'
```

QRegularExpression exists, but the regex syntax and class usage has changed substantially: https://doc.qt.io/qt-5/qregularexpression.html#notes-for-qregexp-users

It makes sense to port to the python standard `re` library, so that the code works in PySide2 and PySide6 without some sort of shim layer, or version selection code.

https://docs.python.org/3/library/re.html

Help welcome if anyone has time to look at this task.

See umbrella task: https://github.com/PixarAnimationStudios/OpenTimelineIO/issues/1215

darbyjohnston:
Fixes #1215
Fixes #1220

This PR upgrades opentimelineview to work with PySide6. I believe most of these changes are also compatible with PySide2 except for these:
* from PySide6
* QtGui.QAction

I'm not very familiar with Python so I'm not sure if there is a way to support both PySide2 and PySide6?

A screenshot of otioview running with PySide6 on an Apple M1:
Screen Shot 2022-02-20 at 3 31 07 PM

### DIFF
--- a/docs/tutorials/quickstart.md
+++ b/docs/tutorials/quickstart.md
@@ -8,10 +8,10 @@ This is for users who wish to get started using the "OTIOView" application to in
 
 OTIOView has an additional prerequisite to OTIO:
 
-- Try `python -m pip install PySide2`
-- If that doesn't work, try downloading PySide2 here: <a href="https://wiki.qt.io/Qt_for_Python" target="_blank">https://wiki.qt.io/Qt_for_Python</a>
+- Try `python -m pip install PySide2` or `python -m pip install PySide6`
+- If that doesn't work, try downloading PySide here: <a href="https://wiki.qt.io/Qt_for_Python" target="_blank">https://wiki.qt.io/Qt_for_Python</a>
 
-You probably want the prebuilt binary for your platform.  PySide2 generally includes a link to the appropriate version of Qt as well.
+You probably want the prebuilt binary for your platform.  PySide generally includes a link to the appropriate version of Qt as well.
 
 ## Install OTIO
 
--- a/setup.py
+++ b/setup.py
@@ -356,7 +356,8 @@ setup(
             'urllib3>=1.24.3'
         ],
         'view': [
-            'PySide2~=5.11'
+            'PySide2~=5.11; platform.machine=="x86_64"',
+            'PySide6~=6.2; platform.machine=="aarch64"'
         ]
     },
 
--- a/src/opentimelineview/console.py
+++ b/src/opentimelineview/console.py
@@ -28,7 +28,12 @@
 import os
 import sys
 import argparse
-from PySide6 import QtWidgets, QtGui
+try:
+    from PySide6 import QtWidgets, QtGui
+    from PySide6.QtGui import QAction
+except ImportError:
+    from PySide2 import QtWidgets, QtGui
+    from PySide2.QtWidgets import QAction
 
 import opentimelineio as otio
 import opentimelineio.console as otio_console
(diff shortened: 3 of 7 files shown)


meshula:
That's awesome! I was much closer to the finish line than I thought, although I was stumped on the regexp. Unfortunately, I think we are stuck with pyside2 for a while for some users, I wonder what a strategy might be to work with both? Maybe we could land this on a pyside6 branch so we can start exercising it while we discuss what to do about 2?

darbyjohnston:
Yeah, the regexp stuff was a bit confusing, though I do have to say Qt did a nice job documenting the other stuff that was deprecated.

I did a bit of searching and it looks like this can be used to support both PySide versions:
```
try:
    from PySide6 import QtWidgets, QtGui, QtCore
except ImportError:
    from PySide2 import QtWidgets, QtGui, QtCore
```

Do you know if something similar would work for the ```QtGui.QAction``` statements?

darbyjohnston:
I added the try/except for the PySide2 and PySide6 imports, and realized I could also do the same for QAction. I think this will work now for both versions, but I haven't tried it with PySide2 yet since that doesn't seem to be available on my mac.

meshula:
I think that try/except business is the right approach.

That just leaves QRegExpr, vs QRegularExpression, since I assume QRegularExpression doesn't exist in PySide2? Perhaps just using re is the right answer there?

darbyjohnston:
I tried with PySide2 on my Linux box and it looks like QRegularExpression is supported. According to the Qt docs QRegularExpression was added in Qt 5.0.

I also added a small update to the quickstart docs about supporting PySide2 and PySide6. It looks like README.md and setup.py also reference PySide2, but maybe that's OK for now since PySide2/Qt5 is probably the main use case?

meshula:
If it's possible ~ for setup.py to default (for now) to PySide2/Qt5 except on M1, where it should use PySide6/QT6 ~ I think we'd have the best out of the box experience.

meshula:
I wonder what notes would be most useful in the README? Should setup.py also take a flag to allow PySide6 as an installation preference?

darbyjohnston:
I'm not familiar with setup.py, is that specific to OTIO or part of Python?

meshula:
It's the install script for Python that exists at the root of this repo. It specifies which PySide to fetch here:

https://github.com/PixarAnimationStudios/OpenTimelineIO/blob/8f93157e2777536c2f90a871b21edbc492698a29/setup.py#L359

It's typical to invoke it via something like `python setup.py install`

JeanChristopheMorinPerso:
@meshula something like this could do it:
```
[
    'PySide2; platform.machine=="x86_64"',
    'PySide2; platform.machine=="aarch64"'
]
```

Potentially more complex markers could be used.

meshula:
I think that'll cover the use cases we know about!

darbyjohnston:
I modified ```setup.py``` and was able to run ```python -m pip install .``` on both my x64 Linux workstation and M1 laptop.

meshula (approved):
Looks good! Let's land this as soon as a couple of people have tried it out locally!

PR #1236, opened 2022-03-09
Don't skip asserts in unit test release builds

### LINKED ISSUE #1235: C++ Unit tests always pass in release builds

When using a release build the c++ unit tests asserts will never be triggered.

## To Reproduce

1. Operating System: Any
2. Python version: N/A
3. Example snippet that demonstrates the issue - if it's a build issue: see below
4. OpenTimelineIO release version or commit hash: 0450ef461d52cebd397227a9d966352781a9353d
5. Compiler information:
    on Mac, type `clang -v`, and paste the results here
Apple clang version 13.0.0 (clang-1300.0.29.30)
Target: arm64-apple-darwin21.3.0

For example, add the following to any test:
```
assertEqual(true, false);
```
Then on the command-line:
make 
ctest

The test will pass.

## Expected Behavior

The assert should be triggered.

rogernelson:
Fixes #1235

**Summarize your change.**

This branch undefines NDEBUG when using unit tests.  This is to ensure the asserts used in assertEqual and assertNotEqual, etc fail the test even in release builds.  Since this file is only included by the unit tests it won't have any impact outside of that scope.

test_clip.cpp and test_opentime.cpp

### DIFF
--- a/tests/utils.h
+++ b/tests/utils.h
@@ -1,4 +1,5 @@
 #pragma once
+#undef NDEBUG
 
 #include <algorithm>
 #include <cassert>


rogernelson:
I can't modify the list of reviewers, but I'd like to request that @darbyjohnston review this.

darbyjohnston:
Seems like a reasonable change, I guess I had assumed the tests would always be run in Debug mode.

meshula (approved):
Thanks for the catch! Over time we should move away from cstd assertions, and towards a framework of some sort. If anyone is thinking about that, lighter is better, speaking for myself, I'd prefer a homegrown framework, or at least a framework that is not larger than OTIO itself ;)

PR #1237, opened 2022-03-11
otioview track and window fit improvements

### LINKED ISSUE #1207: otioview scene rect clips items at the end of track

When adding a new Stack to otioview, `frame_all` is run to scale the items in each track to fit the visible scene rect. The scaling function works well, but it doesn't seem to account for the width `TrackNameItem` adds to the scene. As a result, items toward the end of the track are clipped, so that when you are zoomed into the track (and sometimes upon the initial `frame_all` fit), scrolling right to see items at the end of the track is not possible, as they extend beyond the scroll area, and you must instead zoom out or widen the window to see them.

I dug into the code to investigate this, hoping it would be a simple fix (by adding `TRACK_NAME_WIDGET_WIDTH` to the initial scene rect calculation, for example), but the solution was more complicated since it must account for the various `counteract_zoom` implementations. What I did find was that if you modify these two lines in `track_widgets.py` (removing the `TrackNameItem` offset), the fit and scroll area are pixel perfect:

In `BaseItem.__init__`:
```
self.current_x_offset = TRACK_NAME_WIDGET_WIDTH
```
-> 
```
self.current_x_offset = 0
```

In `Track._populate`:
```
    track_name_rect = QtCore.QRectF(
        0,
        0,
        TRACK_NAME_WIDGET_WIDTH,
        TRACK_HEIGHT
    )
```
->
```
    track_name_rect = QtCore.QRectF(
        0,
        0,
        0,
        TRACK_HEIGHT
    )
```

Those won't work for fixing it of course, but confirmed my suspicion that the `TrackNameItem` was the underlying cause. I think long-term, the fix here would be to move the track names to a separate widget to the left of the timeline, which would also replicate the behavior of most non-linear editors where scrolling the timeline doesn't effect the track list. Short-term, this could probably be resolved with some tweaks to the sizing math, but those tweaks felt non-trivial at first glance.

## To Reproduce

I'm seeing this on: Windows 10, Python 3.7, OTIO 0.14

To reproduce, open a timeline in otioview, and reduce the window size (or move the splitter left) so that the horizontal scroll bar appears. Scroll this to the right. You shouldn't be able to see all the items at the end of the track. The width of the clipped area is about equivalent to the track name item width.

michdolan:
Fixes #1207 

This resolves an issue where `otioview` didn't take into account `TrackNameItem` width when framing the timeline within the current graphics scene. This resulted in track ends going offscreen without the ability to scroll to them until the window was expanded. It will now factor in the largest name item width to adapt to normal and expanded item states.

This PR also resolves a related issue with the initial size and position of the `otioview` window. Previously the window was centered to a default constructed `QScreen`, which in my debugging on Windows could contain garbage size data and put the otioview window far off screen. It also didn't take into account small screens (like the laptop I am contributing from), window title bars, and OS-used portions of the screen for things like a taskbar. I've updated the method to get the screen from the QApplication, to resize the window to fit the screen if necessary, and to make sure it is centered in the available space provided by the OS. This reduces the need for users to move or resize the window to work in `otioview`, preserving the initial track fit results.

Lastly, there is a one line change that fixes a Python 3 incompatibility. I've been testing this in Python 3.7 and 3.9 on Windows and have not run into issues. It would be good to verify the results of these fixes on Linux and macOS prior to merging this.

### DIFF
--- a/src/opentimelineview/timeline_widget.py
+++ b/src/opentimelineview/timeline_widget.py
@@ -197,7 +197,7 @@ class CompositionWidget(QtWidgets.QGraphicsScene):
         self.setSceneRect(
             start_time * track_widgets.TIME_MULTIPLIER,
             0,
-            duration * track_widgets.TIME_MULTIPLIER + 
+            duration * track_widgets.TIME_MULTIPLIER +
             self.track_name_width() * zoom_level,
             height
         )
--- a/src/opentimelineview/timeline_widget.py
+++ b/src/opentimelineview/timeline_widget.py
@@ -128,7 +128,28 @@ class CompositionWidget(QtWidgets.QGraphicsScene):
 
         self._data_cache = self._cache_tracks()
 
-    def _adjust_scene_size(self):
+    def track_name_width(self):
+        # Choose track name item from the longest track, since that track will
+        # define the composition's scene width.
+        width = track_widgets.TRACK_NAME_WIDGET_WIDTH
+        longest_track = None
+        max_end_time = 0
+
+        for item in self.items():
+            if isinstance(item, track_widgets.Track):
+                track_range = item.track.trimmed_range_in_parent()
+                end_time = track_range.end_time_inclusive().to_seconds()
+                if end_time > max_end_time:
+                    max_end_time = end_time
+                    longest_track = item
+
+        if longest_track is not None:
+            if longest_track.track_name_item is not None:
+                width = longest_track.track_name_item.rect().width()
+
+        return width
+
+    def _adjust_scene_size(self, zoom_level=1.0):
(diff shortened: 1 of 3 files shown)


src/opentimelineview/timeline_widget.py
```diff
@@ -83,7 +83,7 @@ def get_nav_menu_data():
 
 def get_filters(filter_dict, bitmask):
     filters = list()
-    for item in filter_dict.itervalues():
+    for item in filter_dict.values():
```
michdolan, line 86: Note: This is the Python 3 fix

meshula (approved):
This is more complex than I expected, but it all looks good. We should keep an eye on behavior, but this looks like an improvement overall.

PR #1238, opened 2022-03-11
Install full OTIO Python env with CMake

michdolan:
Currently when setting `OTIO_PYTHON_INSTALL=ON` in CMake, only the compiled Python bindings are installed, which can't be used on their own. This update expands this to match the OTIO PyPI package in the CMake install location, including the three pure-python libs: `opentimelineio`, `opentimelineview`, and `opentielineio_contrib`. The result is that the CMake install is production-ready in a Python env.

While on by default when building the Python bindings, `opentimelineview` is also gated behind the existing (and currently unused) `OTIO_INSTALL_COMMANDLINE_TOOLS`, and `opentielineio_contrib` behind a new `OTIO_INSTALL_CONTRIB` option, so both can be turned off if only the core lib is needed.

### DIFF
--- a/src/py-opentimelineio/CMakeLists.txt
+++ b/src/py-opentimelineio/CMakeLists.txt
@@ -1,2 +1,21 @@
+#------------------------------------------------------------------------------
+# py-opentimelineio/CMakeLists.txt
+
 add_subdirectory(opentime-bindings)
 add_subdirectory(opentimelineio-bindings)
+
+# Install pure-python OTIO packages to match PyPI wheel structure
+install(DIRECTORY "${PROJECT_SOURCE_DIR}/src/py-opentimelineio/opentimelineio/"
+    DESTINATION "${OTIO_RESOLVED_PYTHON_INSTALL_DIR}/opentimelineio")
+
+if(OTIO_INSTALL_COMMANDLINE_TOOLS)
+    install(DIRECTORY "${PROJECT_SOURCE_DIR}/src/opentimelineview"
+        DESTINATION "${OTIO_RESOLVED_PYTHON_INSTALL_DIR}")
+endif()
+
+if(OTIO_INSTALL_CONTRIB)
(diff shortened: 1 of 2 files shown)


src/py-opentimelineio/CMakeLists.txt
```diff
@@ -1,2 +1,21 @@
+#------------------------------------------------------------------------------
+# py-opentimelineio/CMakeLists.txt
+
 add_subdirectory(opentime-bindings)
 add_subdirectory(opentimelineio-bindings)
+
+# Install pure-python OTIO packages to match PyPI wheel structure
+install(DIRECTORY "${PROJECT_SOURCE_DIR}/src/py-opentimelineio/opentimelineio/"
+    DESTINATION "${OTIO_RESOLVED_PYTHON_INSTALL_DIR}/opentimelineio")
+
```
meshula, line 10:
```suggestion
if(OTIO_PYTHON_INSTALL)
    install(DIRECTORY "${PROJECT_SOURCE_DIR}/src/py-opentimelineio/opentimelineio/"
        DESTINATION "${OTIO_RESOLVED_PYTHON_INSTALL_DIR}/opentimelineio")
endif()
```

meshula: Did you mean to guard this new bit, as I've suggested above?

meshula (commented):
This looks good to me, thanks! I have one question.

src/py-opentimelineio/CMakeLists.txt
michdolan: This wasn't needed because this module is guarded in the root CMake by  `OTIO_PYTHON_INSTALL`, so all three of these install statements are implicitly guarded by it:

```
if(OTIO_PYTHON_INSTALL)
    add_subdirectory(src/py-opentimelineio)
endif()
```

meshula: ah gotcha, forget to expand the context

meshula (approved):

PR #1239, opened 2022-03-11
Keep markers at the track if `child_at_time()` found something that does not hold markers

jchen9:
The issue happens when `track.child_at_time()` found something the does not have markers property, for example, `Transition`. 

It triggered `AttributeError: 'opentimelineio._otio.Transition' object has no attribute 'markers'` in the line:

```
                # attach marker to target item
                target_item.markers.append(marker)
```

The change now checks both `None` and whether the `target_item` has the `markers` attribute.
An example AAF composition and test case demonstrating the issue is also added.


meshula:
@reinecke I don't think there's a reason to break up a print statement to satisfy a line length rule in this case, such a split won't make the code more readable. Is there a tag to suppress the linter for that line?

meshula (approved):

contrib/opentimelineio_contrib/adapters/tests/test_aaf_adapter.py
```diff
@@ -1100,6 +1105,26 @@ def test_aaf_transcribe_log(self):
...
+
+        timeline = None
+
+        try:
+            timeline = otio.adapters.read_from_file(
+                MARKER_OVER_TRANSITION_PATH
+            )
+
+        except Exception as e:
+            print('[ERROR] Transcribing test sample data `{}` caused an exception: {}'.format(
```
meshula, line 1121:
```suggestion
            print('[ERROR] Transcribing test sample data `{}` caused an exception: {}'.format( #noqa
```

contrib/opentimelineio_contrib/adapters/tests/test_aaf_adapter.py
```diff
@@ -1100,6 +1105,26 @@ def test_aaf_transcribe_log(self):
...
+
+        timeline = None
+
+        try:
+            timeline = otio.adapters.read_from_file(
+                MARKER_OVER_TRANSITION_PATH
+            )
+
+        except Exception as e:
+            print('[ERROR] Transcribing test sample data `{}` caused an exception: {}'.format( #noqa
```
meshula, line 1121:
```suggestion
            print('[ERROR] Transcribing test sample data `{}` caused an exception: {}'.format(  # noqa
```

PR #1241, opened 2022-03-14
Media Multi-Reference Feature

### LINKED ISSUE #630: Support Multiple Media Reference on clip
In OTIO, `MediaReference` represents a specific asset that can be fetched and operated on. Among others these include movie files on disk, streaming proxies, and image sequences.
OTIO `Clip` represents an instance of usage of that media in the timeline's composition.

There are many contexts where it is useful to express there are multiple assets available providing media for a given clip, in general this would be to express that a clip has multiple "representations" available. Applications could then select from a Clip's set of MediaReferences based on that application's context. 
Some use cases include:

- Provide information allowing review tool users to switch between a reduced quality streaming proxy and full quality EXRs.
- Media being delivered to editorial has metadata described in a `SerializableCollection` of `Clip` instances where each clip contains references to both the editorial proxies and original camera files. This allows that metadata to later be used to locate media for final conform.

Considerations:
- Do authors of files want to express that a given media reference is the "Preferred" media reference?
- Should there be tooling to aid applications on selecting the "best" media reference for their use case?
- Would it be possible to implement this in a non-breaking way? (could you still assign a single media reference to a clip? Which media reference do you get when accessing the `media_reference` property?)
- Rather than making a clip have multiple media references, would it be more natural to create a new `MediaReference` type called something like `MultiReference` that then refers to all the separate references?

rogernelson:
Fixes #630 

**Summarize your change.**

This branch is an initial prototype for the changes and discussion initially proposed in this [working document](https://docs.google.com/document/d/13jrl2qtWBTSFqMTVQh28LWc16uYvQ79w-7TfX92JaIM/edit?usp=sharing).

To summarize, there are often uses cases when a single conceptual clip can have multiple different media representations.  A common example is high-resolution and proxy-resolution representations of the same media where the high-resolution is used when users have access to the file path and hardware appropriate for playback and proxy-resolution for when they do not.  In these situations it's convenient for the Clip to have a reference to each media representation so we can easily toggle between them based on our needs.

To accommodate these workflows, this branch transforms the single media reference contained within a clip into a map/dictionary of media references. Below is what the difference looks like when serialized.

before:
```
"media_reference" : {
    "OTIO_SCHEMA" : "MissingReference.1",
    "available_range" : null,
    "available_image_bounds" : null,
    "metadata" : {},
    "name" : ""
}
```
after:
```
"active_media_reference": "DEFAULT_MEDIA",
"media_references" : {
    "DEFAULT_MEDIA": {
        "OTIO_SCHEMA" : "MissingReference.1",
        "available_range" : null,
        "available_image_bounds" : null,
        "metadata" : {},
        "name" : ""
    }
}
```
As can be seen above, the later format is extensible allowing new media references to be added to the `media_references` map.  The key set in the `active_media_reference` will be the media reference returned using the existing `media_reference` properties.   To change the active reference, the user can simply just pass the name of the key of the new active reference to the new `active_media_reference` property. 

test_clip.{py,cpp} were modified to test the new functionality.  All other existing tests access that media_reference (test_clip, test_track, test_stack, etc) ensure the existing functionality continues to work.


src/opentimelineio/clip.h
```diff
@@ -9,10 +9,18 @@ namespace opentimelineio { namespace OPENTIMELINEIO_VERSION {
 class Clip : public Item
 {
 public:
+    struct MediaRepresentation
```
rogernelson, line 12: Naming check: everyone happy with MediaRepresentation? Any other suggestions?

tests/utils.cpp
```diff
@@ -38,6 +39,9 @@ Tests::run(int argc, char** argv)
                 std::find(filter.begin(), filter.end(), test.first);
             run_test = filter_it != filter.end();
         }
+
+        std::cout << (run_test ? "Running" : "Skipping") << " test "
```
rogernelson, line 43: The stdout/stderr output is only shown when running with `ctest -VV` or invoking the resulting test executable directly otherwise its suppressed. This will show the subtest running like the below:
```
Running test test_cons
Running test test_ranges
Running test test_clip_v1_to_v2_null
Running test test_clip_v1_to_v2
Running test test_clip_media_representation
```

I thought it was useful to add since when you get a failure you can run with the verbose option to know which sub-test failed.

src/opentimelineio/clip.cpp
```diff
@@ -19,33 +25,74 @@ Clip::~Clip()
...
 {
-    return _media_reference;
+    auto active = _media_references.find(_active_media_reference);
+    return active == _media_references.end() || !active->second
+               ? nullptr
+               : active->second;
+}
+
+Clip::MediaReferences
+Clip::media_references() const noexcept
```
rogernelson, line 35: Although I think it would be more useful to be able to return a mutable map the Python bindings make this very tricky to do.  The only way to do it currently is to use an AnyDictionary (like the `metadata` property on a SerializableObject), but I don't think that's wise to do here as we don't want the value of the map to be anything other than a MediaReference. 

Thus, I just use a simple getter/setter instead so minimal Python glue is needed.  The only real requirement is to convert between Retainer<> and raw pointers, which is the reason these methods don't just do a simple `=`.

src/opentimelineio/clip.h
rogernelson: Also, I followed the model of Schema and Marker::Color which use a struct of strings rather than an enum with conversion functions between enum and string. I assume this is used because it makes the de/serialization easier.  If there's a better approach, let me know.

tests/CMakeLists.txt
```diff
@@ -8,13 +8,21 @@ foreach(test ${tests_opentime})
 	add_executable(${test} utils.h utils.cpp ${test}.cpp)
 	target_link_libraries(${test} opentime)
 	set_target_properties(${test} PROPERTIES FOLDER tests)
-	add_test(${test} ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/${test}${CMAKE_EXECUTABLE_SUFFIX})
+	add_test(NAME ${test} 
+           COMMAND ${test}
+           # Set the pwd to the source directory so we can load the samples
+           # like the python tests do
+           WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/sample_data)
```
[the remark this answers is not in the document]
rogernelson: @darbyjohnston just want to make sure you're okay with some of the changes I made to the cpp tests.

src/opentimelineio/clip.h
```diff
@@ -9,10 +9,18 @@ namespace opentimelineio { namespace OPENTIMELINEIO_VERSION {
...
+    {
+        static char constexpr default_media[]         = "DEFAULT_MEDIA";
+        static char constexpr high_resolution_media[] = "HIGH_RESOLUTION_MEDIA";
+        static char constexpr proxy_resolution_media[] =
+            "PROXY_RESOLUTION_MEDIA";
+    };
```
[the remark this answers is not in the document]
apetrynet: Hi @rogernelson!

I'm sorry if this is obvious in the code, but will users be able to assign their own keys or is this going to be _the_ list of available keys?

May I also suggest we consider something like "high_quality_media" and "low/proxy_quality_media" instead of "*_resolution_media"? Codecs considered "proxy" may be of high resolution (At least in the pixel dimensions sense).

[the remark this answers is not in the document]
rogernelson: There's nothing preventing users from using their own strings.  I modelled it after the Marker::Color, so there's a list of pre-defined strings to use, but any string they pass in is going to work -- unless we wanted to prevent that, but I don't think we do.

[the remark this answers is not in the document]
rogernelson: I definitely don't mind `quality` over `resolution`.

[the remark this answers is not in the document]
apetrynet: Nice! 
I totally agree with leaving it open for users to add their own :)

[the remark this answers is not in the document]
rogernelson: The risk of using something not in the list is basically interchange between different applications. Apps can easily add support for whatever is in the pre-defined list, but they can't guess at all the different strings the users could put in there, so if you do use your own strings it might not be understood outside of your own context (which is of course fine if you only intended to use the otio file within your own context).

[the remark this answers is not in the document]
rogernelson: Potentially, but I imagine it could be possible to have high/proxy quality media available both locally and in the cloud.  I was hoping to get more suggestions for the list here anyway, so I could change the list to `disk_high_quality`, `disk_proxy_quality`, `cloud_high_quality`, `cloud_proxy_quality` if that makes sense.

src/opentimelineio/clip.cpp
```diff
@@ -19,33 +25,74 @@ Clip::~Clip()
...
+    for (auto const& m: _media_references)
+    {
+        result.insert(
+            { m.first, dynamic_retainer_cast<MediaReference>(m.second) });
+    }
+    return result;
+}
+
+void
+Clip::set_media_references(MediaReferences const& media_references) noexcept
```
rogernelson, line 47: There exists potential here to set a map where none of the keys match the `active_reference`.  Currently, if this is the case when they call `media_reference`, it will return a MissingReference. Is that sufficient for error handling? Or do we want to take an ErrorStatus parameter here like some other methods do? Or throw an exception?

src/opentimelineio/clip.cpp
```diff
@@ -19,33 +25,74 @@ Clip::~Clip()
...
+}
+
+std::string
+Clip::active_media_reference() const noexcept
+{
+    return _active_media_reference;
+}
+
+void
+Clip::set_active_media_reference(std::string const& new_active) noexcept
```
rogernelson, line 63: The corollary to my comment above is that someone can also set a key here that does not match any in the map, so whatever error handling we decide on we should do the same here.

src/opentimelineio/clip.h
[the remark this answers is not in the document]
reinecke: Hot take: What if we didn't have a "standard" set of keys defined. The representations that will be available and the heuristic for choosing the appropriate one will be highly pipeline/use-case dependent. I think having "standard" keys encourages trying to encode too much meaning in them and the set will infinitely grow with use cases - it'll become a really unmanageable list. I think it's out of scope for OpenTimelineIO to be trying to define workflow meanings for representations.

In a lot of ways this lines up with adaptive streaming. In that case a player chooses the appropriate representation using an algorithm that takes into account a user's measured connection speed, characteristics of the display device (resolution, bit depth, etc.), how important it is for playback to start immediately vs. prioritizing quality, supported codecs, whether or not there is time to pre-load, etc. The manifest [encodes relevant unopinionated info](https://datatracker.ietf.org/doc/html/draft-pantos-hls-rfc8216bis#section-4.4.6.2) to feed that decision process, but it's up to the player to decide what is important based on the context.

What if, instead, we thought of the keys as being a simple human-friendly selector that can be chosen based on what makes sense for the workflow. Then, for the more advanced programmatic heuristics we could encourage the use of metadata on the media reference.

If there are other bits of meaning we want to encode to aid in choosing an appropriate rep, we should talk about ways to explicitly encode those.

src/opentimelineio/clip.cpp
reinecke: We could also consider doing the mutable mapping wrapping of these functions in the python layer for that api convenience.
I can take a pass at trying to implement that.

tests/utils.cpp
darbyjohnston: That definitely helps, I wonder if we should enable the verbose mode in the GitHub Actions.

tests/CMakeLists.txt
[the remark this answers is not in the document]
darbyjohnston: The extra assert functions look good and I like the change for printing out the test names.

The ```WORKING_DIRECTORY``` also seems like a good change if that's how the Python tests work. The only concern I have is that if a test writes out a temporary file to the current directory it would end up in the source tree, but I guess that's not an issue if the Python tests aren't doing that.

src/opentimelineio/clip.cpp
reinecke: Checking the current behavior of `Clip` in python, it appears that setting `media_reference` to `None` results in it being set to a blank `MissingReference` - so I think returning `MissingReference` feels pretty consistent with our current behavior.

src/opentimelineio/clip.h
```diff
@@ -21,12 +34,21 @@ class Clip : public Item
...
+    void            set_media_reference(MediaReference* media_reference);
     MediaReference* media_reference() const noexcept;
 
+    using MediaReferences = std::map<std::string, MediaReference*>;
+
+    MediaReferences media_references() const noexcept;
+    void set_media_references(MediaReferences const& media_references) noexcept;
+
+    std::string active_media_reference() const noexcept;
+    void set_active_media_reference(std::string const& new_active) noexcept;
```
reinecke, line 50: This might be a bit nitpick, but `active_media_reference` reads as if the value should be a `MediaReference` instead of a string. If we were being totally explicit, we might call it `active_media_reference_key`.

Do others have thoughts on this?

reinecke (commented):
At the moment, the ergonomics look like they'd be pretty good on this!

I just had a few comments for discussion about having a standard set of reference keys and bike shedding the name of `active_media_reference.

tests/test_clip.py
```diff
@@ -181,6 +181,70 @@ def test_ref_default(self):
...
+        cl.active_media_reference = \
+            otio.schema.ClipMediaRepresentation.DISK_HIGH_QUALITY_MEDIA
+        self.assertIsOTIOEquivalentTo(
+            cl.media_reference,
+            otio.schema.GeneratorReference()
+        )
+        self.assertEqual(
+            otio.schema.ClipMediaRepresentation.DISK_HIGH_QUALITY_MEDIA,
+            cl.active_media_reference
+        )
```
reinecke, line 246: We may just want to add some test cases around behavior when `active_media_reference` has no appropriate equivalent in the `media_references` dictionary.

tests/CMakeLists.txt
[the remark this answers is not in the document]
rogernelson: That's true, good point. The convention I am used to for unit test output is to write to the system temp directory, and that seems to be what the python tests are doing as well.  As long as the c++ tests follow the same convention we should be alright.

src/opentimelineio/clip.cpp
rogernelson: A mutable mapping is what I originally wanted to do and took a brief shot at it, but it is indeed a little tricky.  If it were available in the api I would certainly use it! 

We discussed this a bit on slack already in case you didn't see it:
https://academysoftwarefdn.slack.com/archives/CMQ9J4BQC/p1647149228172289

src/opentimelineio/clip.h
rogernelson: Not a nitpick, naming is important. I was hoping to go over all the naming at the TSC meeting on Thursday to make sure everyone is on board with the chosen names.

src/opentimelineio/clip.cpp
```diff
@@ -19,33 +27,74 @@ Clip::~Clip()
...
+    return result;
+}
+
+void
+Clip::set_media_references(MediaReferences const& media_references) noexcept
+{
+    _media_references.clear();
+    for (auto const& m: media_references)
+    {
+        _media_references[m.first] = m.second;
```
rogernelson, line 54: Should we do this instead?
```
_media_references[m.first] = m.second ? m.second : new MissingReference
```

So that this:
```
clip.set_media_reference(nullptr);
```
behaves the same as this:
```
clip.set_media_references({"default", nullptr});
```

meshula: What's the precedent elsewhere, within OTIO? 

As a C++ programmer, I intuitively expect set(nullptr) means whatever is set, is going to be cleared, in the general case.
If a new map is going to be created, I'd want that to consistently match other interfaces. @davidbaraff Have you got thoughts on this?

meshula:
@rogernelson This proposal has iterated greatly in a good direction, since the original document 

https://docs.google.com/document/d/13jrl2qtWBTSFqMTVQh28LWc16uYvQ79w-7TfX92JaIM/edit?usp=sharing

and a great many decisions have been taken since then, making it hard to parse the implementation versus the design intention.

As I read through the comments and changes, I keep thinking I wish I could refer the current state against a technical design document. Could the current state of the design and its behaviors be added to the existing document? I think a paragraph summary of the addition to the schema, and the current state of the design would suffice.

I added a Current Status block at the document, pointing to this PR, as a place where I thought it could go.

src/opentimelineio/clip.h
meshula: Do we have a consistent nomenclature throughout OTIO for "keys" or "tags" in general? I do agree that `active_media_reference` sounds like it's going to give me access to a `MediaReference` object in some way.

src/opentimelineio/clip.cpp
meshula: I think set_media_reference has to clear the active_reference. The reason is that we are REPLACING active media_references, not merging them. So if before/after maps both have a "fizzbin" tag, and "fizzbin" was the active reference, the correspondence is superficial. 

In general, setting the media references, and thereby introducing a potential consistency problem (before has "fizzbin", after does not), seems like an undesirable behavior.

tests/CMakeLists.txt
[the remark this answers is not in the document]
meshula: `WORKING_DIRECTORY` seems good to me as long as there's no files written to it.

src/opentimelineio/clip.h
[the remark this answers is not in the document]
meshula: I have to say I'm biased in favor of the "hot take" no standard set of keys. I think setting active, and having a rule for what a viewer should display in the case there is no active, should suffice.

src/opentimelineio/clip.cpp
rogernelson: If you set a nullptr (or None via the Python bindings) with `set_media_reference` on a Clip, it will result in the Clip's media reference being set to a MissingReference.  Personally, I find this a bit misleading and prefer explicit behaviour (ie if I wanted a MissingReference I would have done `clip.set_media_reference( new MissingReference() );`). 

Since set_media_reference is also called in the ctor, this effectively means a Clip can never have a null MediaReference, perhaps there is some historical reason for this?  Currently, all the code I can find is properly checking for nullptr even though it will never be the case.  For example in Clip::available_range:

```
    auto active_media = media_reference();
    if (!active_media)
    {
        if (error_status)
        {
            *error_status = ErrorStatus(
                ErrorStatus::CANNOT_COMPUTE_AVAILABLE_RANGE,
                "No media reference set on clip",
                this);
        }
        return TimeRange();
    }

    if (!active_media->available_range())
    {
        if (error_status)
        {
            *error_status = ErrorStatus(
                ErrorStatus::CANNOT_COMPUTE_AVAILABLE_RANGE,
                "No available_range set on media reference on clip",
                this);
        }
        return TimeRange();
    }
```

src/opentimelineio/clip.h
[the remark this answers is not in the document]
rogernelson: Alright, I'm okay with it too. I think the one exception being 'default_media', correct? When we convert a Clip.1 to a Clip.2 we will need to set its key to something, and we can also use this value and the default argument in the ctor.

src/opentimelineio/clip.cpp
rogernelson: I think you mean `set_media_references` (plural, not singular, correct?). 

Assuming that's the case, we could potentially modify the signature to `set_media_references( MediaReferences, std::string);` forcing the caller to tell us which reference in the new map is the active one.

src/opentimelineio/clip.h
rogernelson: We use `key` in SerializableObject::Reader:
```
 bool read(std::string const& key, bool* dest);
 ```
 
 Since AnyDictionary inherits from std::map, it also uses the term `key`:
 https://en.cppreference.com/w/cpp/container/map
 
 The term `tag` is only really used by rapidjson whose API is not directly exposed to the end user.

**Comment by rogernelson** [comment]
> I added a Current Status block at the document, pointing to this PR, as a place where I thought it could go.

Done. I listed the API additions as well as describing their intended usage.  I also tried to summarize the various points of discussion that have come up here.  Please let me know if there's anything else you think should be in there.

File: src/opentimelineio/clip.h (hunk 5 above)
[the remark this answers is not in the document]
Comment (reinecke) [reply src/opentimelineio/clip.h:23]: Good call, it would probably be a good idea to standardize the "default" key. Over time other standardized keys may also emerge, it'll be a good thing to check in with the community about.

On the point of providing intended use case examples - I think that's a reasonable point. Perhaps we can address it with some good code/sample file examples. Additionally, we can look at adding some more documentation to clarify intended use cases.

File: src/opentimelineio/clip.cpp (hunk 10 above)
Comment (meshula) [reply src/opentimelineio/clip.cpp:54]: I don't know the history, but it's consistent, thanks for looking that up.

File: src/opentimelineio/clip.h [hunk 11]
```diff
@@ -9,24 +9,39 @@ namespace opentimelineio { namespace OPENTIMELINEIO_VERSION {
...
 
+    void            set_media_reference(MediaReference* media_reference);
     MediaReference* media_reference() const noexcept;
 
+    using MediaReferences = std::map<std::string, MediaReference*>;
+
+    MediaReferences media_references() const noexcept;
+    void            set_media_references(
+                   MediaReferences const& media_references,
+                   std::string const&     new_active_key = std::string(),
```
rogernelson, line 37: The one remaining question I have after the TSC meeting.  I asked it already on slack, so I'll repost here:

 I have a brief followup question from yesterday's TSC meeting.  When we discussed modifying the media_refererence and media_reference_key setters so that they will enforce that the active key is never invalidated, something that occurred to me after the meeting is that this would make it very difficult to remove the active key.  For example, if my keys were ['a', 'b'] with 'a' being active and I wanted to change for ['c', 'd'] with 'c' being active, I would need to do leave the active key in there temporarily, then change the active key, then reset the values with the old active key removed. Something like this:
clip.set_media_references( {{"a": dummy_ref},{"b", ref1}, {"c", ref2}} );
clip.set_active_media_reference_key("b");
clip.set_media_references( {{"b", ref1}, {"c", ref2}} );
Needless to say that is quite tedious.  I think the best solution for this is to allow the user to pass a new active_key to the set_media_reference call.  Something like this:
clip.set_media_references( {{"b", ref1}, {"c", ref2}}, "b");
I am wondering the second parameter should be optional or not and if unset we leave the key as is?
If we make it mandatory it would be trivial but maybe a bit tedious to leave it as is:
clip.set_media_refererences( {{"a", ref1}, {"b", ref2}}, clip.get_active_media_reference_key() ); 
And if we do leave it optional is the behaviour of defaulting to an empty string make the intent clear enough?
void set_media_references(MediaReferences const& media_references,
                          std::string const& new_active_key = std::string(),
                          ErrorStatus* error_status = nullptr);
This would of course assign special meaning to empty string as well, preventing the user from using empty string as a key in the dictionary that they're setting.  Obviously, an edge-case but something to consider. I think using std::optional<std::string> is a more clear way of doing it but that's not done anywhere in the API currrently so there's no mechanism for serializing that at the moment (I could add it if desired).  Alternatively, it could be done the old-school way with a pointer to std::string defaulting to nullptr.

rogernelson:
The python 2.7 build is failing because of this:
>- - *media_references*: media_references(self: opentimelineio._otio.Clip) -> Dict[str, opentimelineio._otio.MediaReference]
?                                                                                 ^^^
+ - *media_references*: media_references(self: opentimelineio._otio.Clip) -> Dict[unicode, opentimelineio._otio.MediaReference]
?       

I build the python bindings with python3, so that probably explains why 2.7 is failing.  But I am curious, why is it putting in the function signature for `media_references`? It doesn't seem to do that for anything else...

src/opentimelineio/clip.h
meshula: For a first draft, I lean to simple-as-possible. Why don't we try making it non-optional and discover whether it's actually too clumsy in practice? This looks perfectly fine to me: `clip.set_media_references( {{"b", ref1}, {"c", ref2}}, "b");`

meshula: I am presuming that the simplest case is actually the common one...

tests/utils.cpp
meshula: Best to open an issue, so someone with the requisite knowledge will see it.

src/py-opentimelineio/opentimelineio/console/autogen_serialized_datamodel.py
```diff
@@ -200,7 +200,8 @@ def _generate_model_for_module(mod, classes, modules):
 
             for fetcher in PROP_FETCHERS:
                 try:
-                    model[cl][k] = fetcher(cl, k)
+                    model[cl][k] = fetcher(cl, k) if isinstance(
```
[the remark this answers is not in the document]
reinecke: Yeah, I think this is sensible. I might recommend you add a comment here just to document why this check is done.

jminor (approved):

src/opentimelineio/clip.cpp
```diff
@@ -19,33 +23,117 @@ Clip::~Clip()
...
+
+Clip::MediaReferences
+Clip::media_references() const noexcept
+{
+    MediaReferences result;
+    for (auto const& m: _media_references)
+    {
+        result.insert(
+            { m.first, dynamic_retainer_cast<MediaReference>(m.second) });
+    }
```
meshula, line 40: To be consistent with `media_reference`'s logic, shouldn't m.second be tested against nullptr and only added if the dynamic_cast is not null?

src/opentimelineio/clip.cpp
```diff
@@ -19,33 +23,117 @@ Clip::~Clip()
...
+        if (error_status)
+        {
+            *error_status = ErrorStatus(
+                ErrorStatus::MEDIA_REFERENCES_MISSING_ACTIVE_KEY,
+                "The media references do not contain the active key",
+                this);
+        }
+        return false;
+    }
+    return true;
```
meshula, line 63: Should there be a second error status of `Active key exists but has no media references`?

src/opentimelineio/clip.cpp
```diff
@@ -19,33 +23,117 @@ Clip::~Clip()
...
+    if (!check_for_valid_media_reference_key(
+            new_active_key, media_references, error_status))
+    {
+        return;
+    }
+
+    _media_references.clear();
+    for (auto const& m: media_references)
+    {
+        _media_references[m.first] = m.second ? m.second : new MissingReference;
```
meshula, line 81: should the cases I noted above also be tested against MissingReference?

src/opentimelineio/clip.cpp
```diff
@@ -19,33 +23,117 @@ Clip::~Clip()
...
+}
+
+void
+Clip::set_active_media_reference_key(
+    std::string const& new_active_key, ErrorStatus* error_status) noexcept
+{
+    if (!check_for_valid_media_reference_key(
+            new_active_key, _media_references, error_status))
+    {
+        return;
```
meshula, line 103: augment the error status perhaps to indicate set couldn't be performed?

src/opentimelineio/clip.cpp
```diff
@@ -19,33 +23,117 @@ Clip::~Clip()
...
+    {
+        if (error_status)
+        {
+            *error_status = ErrorStatus(
+                ErrorStatus::MEDIA_REFERENCES_MISSING_ACTIVE_KEY,
+                "The media references do not contain the active key",
+                this);
+        }
+        return false;
+    }
```
meshula, line 62: is MissingReference to be considered valid or invalid? Either way, there should be a line  of doc, methinks

src/opentimelineio/clip.cpp
```diff
@@ -19,33 +23,117 @@ Clip::~Clip()
...
+    writer.write("media_references", _media_references);
+    writer.write("active_media_reference_key", _active_media_reference_key);
 }
 
 TimeRange
 Clip::available_range(ErrorStatus* error_status) const
 {
-    if (!_media_reference)
+    auto active_media = media_reference();
+    if (!active_media)
```
meshula, line 136: does a special MissingReference check need to be done here or is it accommodated sufficiently by available_range below?

meshula (commented):
This looks good to me, I left some questions about corner cases.

src/opentimelineio/clip.cpp
```diff
@@ -19,33 +23,117 @@ Clip::~Clip()
...
+    return true;
+}
+
+void
+Clip::set_media_references(
+    MediaReferences const& media_references,
+    std::string const&     new_active_key,
+    ErrorStatus*           error_status) noexcept
+{
+    if (!check_for_valid_media_reference_key(
```
meshula, line 72: Shouldn't the check for empty new_active_key be here, and early exit if the string is empty? Rather than checking for empty below. Most of the logic here depends on the key being good.

src/opentimelineio/clip.cpp
rogernelson: This is the getter code.  The old `media_reference` code was simply doing:
```
return _media_reference;
```
So it could also return a nullptr.  

It might look like we're setting it, since we're copying into a MediaReference map, but the reason for that is because we can't rely on the return value being implicitly converted from a `Retainer<MediaReference>` to a MediaReference* anymore -- we need to do it explicitly by copying into the returned map.

rogernelson: That said, it shouldn't really ever be nullptr since the setter is converting nullptr into MissingReference, same as `set_media_reference` was doing.

src/opentimelineio/clip.cpp
rogernelson: It's valid (as it was in the original code as well).  While we do convert nullptr into MissingReference, a MissingReference could be explicitly added by a user, and I don't think we want to make presumptions about that being invalid.

src/opentimelineio/clip.cpp
rogernelson: Hmm. The find on line ensure its has some kind of value (since a key must have a value), and a nullptr will be converted into a MissingReference later on in the setter.  Is there another type of invalid value we should be checking for that I am not thinking of?

src/opentimelineio/clip.cpp
meshula: Since it should always be MissingReference, it's good

src/opentimelineio/clip.cpp
meshula: ok

src/opentimelineio/clip.cpp
meshula: MissingReference will be returned

src/opentimelineio/clip.cpp
meshula: that sounds like a good strategy

src/opentimelineio/clip.cpp
rogernelson: A MissingReference will give an unset optional<TimeRange> in C++ or a None in Python. 

This is the same as it would have done before with a MissingReference as the sole media reference, but we can take this opportunity to do something better if you have a better suggestion.

src/opentimelineio/clip.cpp
```diff
@@ -69,13 +157,14 @@ Clip::available_range(ErrorStatus* error_status) const
...
-    return _media_reference->available_range().value();
+    return active_media->available_range().value();
 }
 
 optional<Imath::Box2d>
 Clip::available_image_bounds(ErrorStatus* error_status) const
 {
-    if (!_media_reference)
+    auto active_media = media_reference();
+    if (!active_media)
```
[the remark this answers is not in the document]
rogernelson: same thing here (except with unset optional<Imath::Box2d> in C++ and None in Python).

src/opentimelineio/clip.cpp
meshula: sounds good to me.

src/opentimelineio/clip.cpp
rogernelson: Sorry, I am a bit lost with the resolved conversations now... are those cases MissingReference with `available_range` and `available_image_bounds`?

meshula: You mentioned that we'll never return nullptrs, but MissingReferences, and the MissingReferences have well defined semantics through the optional. So this one is resolved as well.

src/opentimelineio/clip.cpp
```diff
@@ -19,33 +23,117 @@ Clip::~Clip()
...
+void
+Clip::set_media_references(
+    MediaReferences const& media_references,
+    std::string const&     new_active_key,
+    ErrorStatus*           error_status) noexcept
+{
+    if (!check_for_valid_media_reference_key(
+            new_active_key, media_references, error_status))
+    {
+        return;
```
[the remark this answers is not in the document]
rogernelson: I left the MEDIA_REFERENCES_SET part out of the ErrorStatus name since the validation method is also called by `set_active_media_reference`, and I passed the caller name as a parameter so we can add that to the string error message.  Otherwise, I followed your suggestion.

meshula (approved):
Thanks for the last minute touch ups! lgtm

jminor:
Thanks for the contribution @rogernelson !!!

PR #1248, opened 2022-03-22
Add ALE adapter argument `ale_name_column_key`

jchen9:
This gives caller an option to specify which column to use as a clip’s name.


meshula (approved):

meshula:
It's strange that existing code popped the metadata upon reading.

PR #1251, opened 2022-03-23
Fix missing init metadata

### LINKED ISSUE #1250: module level dunder names not appearing in __init__.py
## Bug Report

```
Python 3.10.2 | packaged by conda-forge | (main, Mar  8 2022, 16:02:23) [Clang 11.1.0 ] on darwin
>>> otio.__version__
AttributeError: module 'opentimelineio' has no attribute '__version__'
```

This is a regression from 0.14, see here: https://github.com/PixarAnimationStudios/OpenTimelineIO/issues/1224

## Expected Behavior

[Description of the expected behavior.
](https://peps.python.org/pep-0008/#module-level-dunder-names)

JeanChristopheMorinPerso:
Fixes #1250

#1238 introduced a regression where it overwrote Python modules where modified by setuptools.

This PR disables copying python files from CMake if `OTIO_PYTHON_INSTALL_DIR` is not set. AFAIK, `OTIO_PYTHON_INSTALL_DIR` is only used when running from setuptools. Or I think.

I've also set `BUILD_TESTING` to OFF when building from setuptools since there is no point of building the C++ tests when building the Python extension.


CMakeLists.txt
```diff
@@ -75,7 +75,7 @@ if(OTIO_PYTHON_INSTALL)
         message(STATUS "OTIO Defaulting Python install to ${OTIO_RESOLVED_PYTHON_INSTALL_DIR}")
     else()
         # either python_install or install_prefix have been set
-        if(OTIO_PYTHON_INSTALL_DIR_INITIALIZED_TO_DEFAULT)
+        if(NOT OTIO_PYTHON_INSTALL_DIR)
```
JeanChristopheMorinPerso, line 78: `OTIO_PYTHON_INSTALL_DIR_INITIALIZED_TO_DEFAULT` seemed to always evaluate to false and I couldn't find anything in the CMake docs referencing `_INITIALIZED_TO_DEFAULT` as a suffix. But let me know if I missed something.

src/py-opentimelineio/CMakeLists.txt
```diff
@@ -4,18 +4,20 @@
 add_subdirectory(opentime-bindings)
 add_subdirectory(opentimelineio-bindings)
 
-# Install pure-python OTIO packages to match PyPI wheel structure
-install(DIRECTORY "${PROJECT_SOURCE_DIR}/src/py-opentimelineio/opentimelineio/"
-    DESTINATION "${OTIO_RESOLVED_PYTHON_INSTALL_DIR}/opentimelineio")
+if(NOT OTIO_PYTHON_INSTALL_DIR)
```
JeanChristopheMorinPerso, line 7: I feel like this is fragile. I was thinking of adding a new variable like `CALLED_FROM_SETUPTOOLS` or something like that which would be more reliable. Opinions?

CMakeLists.txt
meshula: OTIO_PYTHON_INSTALL_DIR_INITIALIZED_TO_DEFAULT appears to be vestigial from an earlier iteration. 

That said, the previous clause checks for OTIO_PYTHON_INSTALL_DIR is "", and the docs say that is supposed to mean the install dir was not supplied from the command line. So how could we actually end up in this clause with it set to NO?

setup.py
```diff
@@ -91,6 +91,7 @@ def generate_cmake_arguments(self):
             '-DOTIO_SHARED_LIBS:BOOL=OFF',
             '-DCMAKE_BUILD_TYPE=' + self.build_config,
             '-DOTIO_PYTHON_INSTALL_DIR=' + install_dir,
+            '-DBUILD_TESTING:BOOL=OFF'
```
meshula, line 94:
```suggestion
            # turn off the C++ tests during a Python build
            '-DBUILD_TESTING:BOOL=OFF'
```

src/py-opentimelineio/CMakeLists.txt
meshula: maybe `OTIO_INVOKED_FROM_SETUPTOOLS` to keep with the `OTIO_` pattern for outside visibility?

meshula: This seems to be the core bug here, so flagging the case explicitly as you suggest seems like the right answer, and maybe that can help clarify the logic I was asking about above.

meshula (commented):
Thanks for figuring out where the problem lay! Some comments in line

CMakeLists.txt
JeanChristopheMorinPerso: We can endup here if `OTIO_PYTHON_INSTALL_DIR` is not set and `CMAKE_INSTALL_PREFIX` was not provided by the user.

So `cmake -DOTIO_PYTHON_INSTALL=ON` will get us here.

JeanChristopheMorinPerso: Wait no. You can get here if you set `CMAKE_INSTALL_PREFIX`. Sorry for the confusion.

So `cmake -DCMAKE_INSTALL_PREFIX=/tmp/otio -DOTIO_PYTHON_INSTALL` for example.

src/py-opentimelineio/CMakeLists.txt
meshula: ok, now that the above is resolved, maybe this clause is fine now?

JeanChristopheMorinPerso: I still feel like it's fragile. It relies on logics while I think a dedicated variable would be better. Also, what do you think of `OTIO_PYTHON_MODULES_INSTALL` instead of `OTIO_INVOKED_FROM_SETUPTOOLS`. It would be slightly more generic.

meshula (approved):

meshula:
Thanks for the fix!

JeanChristopheMorinPerso:
It's a pleasure :)

PR #1253, opened 2022-03-24
Support OTIO_PLUGIN_MANIFEST_PATH being set to an empty string

### LINKED ISSUE #1252: Setting OTIO_PLUGIN_MANIFEST_PATH to "" causes load_manifest() error

Setting `OTIO_PLUGIN_MANIFEST_PATH` to `''` will cause `otio.plugins.manifest.load_manifest()` to error trying to load an empty file. This is a regression caused by !945 that I missed. Previously in this case the code checked `os.path.exists('')` which would be `False`, but now checks the existence of the `os.path.abspath('')` which is `True` (and incorrect).

## To Reproduce

1. Operating System - Linux
2. Python version - 2.7.5
4. OpenTimelineIO release version or commit hash - 0.15
5. Compiler information - gcc version 4.8.5 20150623 (Red Hat 4.8.5-44) (GCC) 

```
> env OTIO_PLUGIN_MANIFEST_PATH='' python -c "import opentimelineio as otio; otio.plugins.manifest.load_manifest()"
  File "<string>", line 1, in <module>
  File "/home/rnr/venvs/otio/lib/python2.7/site-packages/opentimelineio/plugins/manifest.py", line 253, in load_manifest
    result.extend(manifest_from_file(json_path))
  File "/home/rnr/venvs/otio/lib/python2.7/site-packages/opentimelineio/plugins/manifest.py", line 62, in manifest_from_file
    result = core.deserialize_json_from_file(filepath)
ValueError: JSON parse error while reading: JSON parse error on input string: The document is empty. (line 1, column 0)
```

## Expected Behavior

This should not error and the default manifests should only be loaded.

avrata:
This restores the ability to set OTIO_PLUGIN_MANIFEST_PATH to an empty string
to indicate no local custom manifest file.

Fixes #1252

**Summarize your change.**

This changes a truthy check on the env var OTIO_PLUGIN_MANIFEST_PATH to skip any false value and not explicitly `is None`. This will allow OTIO_PLUGIN_MANIFEST_PATH='' to be equivalent to that env var not being set. 

Added a test specifically for this. Also updated all tests in `test_adapter_plugin.py` to always un-set custom env vars or manifests on exceptions due to test failures so as to not pollute subsequent tests.

### DIFF
--- a/src/py-opentimelineio/opentimelineio/plugins/manifest.py
+++ b/src/py-opentimelineio/opentimelineio/plugins/manifest.py
@@ -232,9 +232,10 @@ def load_manifest():
 
     # Read plugin manifests defined on the $OTIO_PLUGIN_MANIFEST_PATH
     # environment variable.  This variable is an os.pathsep separated list of
-    # file paths to manifest json files.
+    # file paths to manifest json files. Can be set to "" to indicate no
+    # local custom manifest path.
     _local_manifest_path = os.environ.get("OTIO_PLUGIN_MANIFEST_PATH", None)
-    if _local_manifest_path is not None:
+    if _local_manifest_path:
         for src_json_path in _local_manifest_path.split(os.pathsep):
             json_path = os.path.abspath(src_json_path)
             if (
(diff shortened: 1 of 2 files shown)


meshula (approved):
lgtm

PR #1256, opened 2022-03-26
Add Python 3.10 to CI

### LINKED ISSUE #1224: Support for Python 3.10?
## Feature Request

Is this a New Feature, or a Proposed Change to existing Behavior? 
This is the continuity of existing feature.

## Description

Describe the new feature, or describe the modification to existing behavior, and provide context as necessary to help with understanding how the feature relates to a workflow or functionality.
I am the developer of [Shot Manager](https://github.com/ubisoft/shotmanager), a Blender add-on that provides many features to manipulate shots in the 3D space of Blender but also to import and export edits with other software. To do so I rely on OpenTimelineIO.
Thing is Blender is changing regularly (or irregularly I should say) the version of Python they support, this in order to stick to the latest version. An in March Blender 3.1 will be on Python 3.10.

Would it be possible to have a version of OTIO for Python 3.10? That would be of a great help!
Thank you

JeanChristopheMorinPerso:
Fixes #1224.

Add official support for Python 3.10.

Note that I had to update `cibuildwheel` to the latest because 1.12.0 doesn't support Python 3.10. And while being there, I also made sure to use the latest for all Python versions except Python 2.7. You will also notice that I force `manulinux2010` on Linux. This is because this is the version that was previously used and the latest cibuildwheel defaults to manulinux2014. It's important to stick to manylinux2010 because of the VFX platform. See https://github.com/PixarAnimationStudios/OpenTimelineIO/issues/1094 for more details.

I invite anyone interested to test to download the wheels artifact from the job and test locally on your machine.

If anyone is curious, it's a pretty big feat to be able to support 2.7 and 3.10 at the same time in 2022 :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@ OpenTimelineIO
 ==============
 
 [![Supported VFX Platform Versions](https://img.shields.io/badge/vfx%20platform-2018--2021-lightgrey.svg)](http://www.vfxplatform.com/)
-![Supported Versions](https://img.shields.io/badge/python-2.7%2C%203.7%2C%203.8%2C%203.9-blue)
+![Supported Versions](https://img.shields.io/badge/python-2.7%2C%203.7%2C%203.8%2C%203.9%2C%203.10-blue)
 [![Build Status](https://github.com/PixarAnimationStudios/OpenTimelineIO/actions/workflows/python-package.yml/badge.svg)](https://github.com/PixarAnimationStudios/OpenTimelineIO/actions/workflows/python-package.yml)
 [![codecov](https://codecov.io/gh/PixarAnimationStudios/OpenTimelineIO/branch/main/graph/badge.svg)](https://codecov.io/gh/PixarAnimationStudios/OpenTimelineIO)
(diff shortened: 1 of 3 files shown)


JeanChristopheMorinPerso:
Can someone from the TSC re-run the build please? The Codecov upload failed with a 503 error...

README.md
```diff
@@ -52,8 +52,7 @@ Supported VFX Platforms
 -----------------
 The current release supports:
 - VFX platform 2021, 2020, 2019, 2018
-- Python 2.7 - 3.9
-- Notice that Python 2.7 is deprecated & we plan to drop it in OTIO release 0.15
```
JeanChristopheMorinPerso, line 56: Plans kind of changed, so I thought we should remove this until we have a clear timeline on when 2.7 support will be dropped before adding a deprecation notice.

meshula: We do have a clear timeline ~ as soon as CY2023 is specified, CY2019 will drop off and with it, py27.

meshula (approved):

PR #1258, opened 2022-03-30
Fix #1257 missing each_clip on Clip

### LINKED ISSUE #1257: each_clip no longer exists on Clip
This change https://github.com/PixarAnimationStudios/OpenTimelineIO/pull/926/files caused a regression, by inadvertently removing this method:

```py
@add_method(_otio.Clip)
 def each_clip(self, search_range=None):
     yield self
```

we need to add it back in

meshula:

```Fixes #1257```

**Summarize your change.**

Pull #926 removed each_clip inadvertently from Clip. This change reintroduces it.

### DIFF
--- a/src/py-opentimelineio/opentimelineio/schema/clip.py
+++ b/src/py-opentimelineio/opentimelineio/schema/clip.py
@@ -27,3 +27,8 @@ def __repr__(self):
             repr(self.metadata),
         )
     )
+
+
+@add_method(_otio.Clip)
+def each_clip(self, search_range=None):
+    yield self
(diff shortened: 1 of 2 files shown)


darbyjohnston:
Yup, looks like the code I removed, sorry! Not necessarily as part of this change, but should there also be a C++ equivalent?

meshula:
Probably it should exist for C++ as well, although my mind was only on the regression.

darbyjohnston:
LGTM for fixing the regression.

reinecke (approved):
LGTM!

PR #1259, opened 2022-04-02
Add a test case and fix for transcribing markers over a single audio AAF file

jchen9:
This PR fixes a problem that is only encountered when a marker is found in a media AAF (e.g. one with audio or video data, and only a single clip as the composition.)


meshula (approved):
lgtm

contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py
```diff
@@ -1246,22 +1246,30 @@ def _attach_markers(collection):
...
+                        marker.marked_range.start_time, target_item
+                    )
+
+                    marker.marked_range = otio.opentime.TimeRange(
+                        start_time=marked_start_local,
+                        duration=marker.marked_range.duration
+                    )
+
+                except otio.exceptions.CannotComputeAvailableRangeError:
                     target_item = target_track
```
jminor, line 1268: Can you leave some comments here explaining when the exception is expected?

contrib/opentimelineio_contrib/adapters/tests/test_aaf_adapter.py
```diff
@@ -1125,6 +1130,26 @@ def test_aaf_marker_over_transition(self):
...
+                MARKER_OVER_AUDIO_PATH
+            )
+
+        except Exception as e:
+            print('[ERROR] Transcribing test sample data `{}` caused an exception: {}'.format(  # noqa
+                os.path.basename(MARKER_OVER_AUDIO_PATH),
+                e)
+            )
+
+        self.assertIsNotNone(timeline)
```
jminor, line 1151: Can you also check that the markers are in the right place (on the right item, and the right time)?

jminor (changes requested):

contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py
jchen9: Added

contrib/opentimelineio_contrib/adapters/tests/test_aaf_adapter.py
jchen9: Added

contrib/opentimelineio_contrib/adapters/tests/test_aaf_adapter.py
```diff
@@ -1125,6 +1130,63 @@ def test_aaf_marker_over_transition(self):
...
+
+        # Verify markers
+        # We expect 1 track with 3 markers on it from the test data.
+        self.assertTrue(1 == len(timeline.tracks))
```
jminor, line 1155: `assertEqual` is a better choice, since any failure will be reported more clearly.

jminor (changes requested):

contrib/opentimelineio_contrib/adapters/tests/test_aaf_adapter.py
jchen9: Switched to assertEqual.

jminor (approved):

PR #1266, opened 2022-04-06
Copy core sample data used by contrib tests into the contrib area to better isolate contrib tests

avrata:
This allows the contrib tests to such that they can run in a self contained way and allows for a future where they are split out.

This was done to allow testing of the `contrib/opentimelinio_contrib/adapters/tests` on it's own without the need for assumptions of the current file structure.

test_rvsession.py and test_maya_sequencer.py


jminor (approved):
This looks great.
The only note I have is unrelated to the change itself. It seems slightly odd that these tests are relying on the EDL adapter to function. It would be cleaner if they only needed OTIO files. That doesn't need to happen now, but could be done later when we split contrib out.

PR #1267, opened 2022-04-07
Support python 3 in the "rv_session" adapter.

### LINKED ISSUE #1264: rv_session adapter does not work with python 3+

While testing the rv adapter (not-plugin), I noticed that the subprocess is unable to properly handle str vs bytes in python 3 leading to:

```
In [4]: otio.adapters.write_to_file(timeline, '/tmp/out.rv')
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
...

<otiolocation>/lib/python3.7/site-packages/opentimelineio_contrib/adapters/rv.py in write_to_file(input_otio, filepath)
     90     # helpful message from the subprocess.
     91     try:
---> 92         proc.stdin.write(json.dumps(simplified_data))
     93     except IOError:
     94         pass

TypeError: a bytes-like object is required, not 'str'
```
It looks like we just need to account for this behavior difference between python 2 and 3. We should probably check to make sure other subprocess cases that are piping input/output are also handling this as well.

## Expected Behavior

The "rv_session" adapter should properly export files in python 3 via a subprocess pipe

avrata:
This can be used with rv versions that support python 3

Fixes #1264

**Summarize your change.**

The rv_session adapter was written in a time before any version of rv supported python 3. That's now a supported version of python so this change enables python 3 for this adapter.

There was one issue with testing where the connections list was not stable between python versions. I think this is due to dict order differences used in the rvSession.py's nodes dict.

### DIFF
--- a/contrib/opentimelineio_contrib/adapters/extern_rv.py
+++ b/contrib/opentimelineio_contrib/adapters/extern_rv.py
@@ -60,7 +60,9 @@ def main():
 
     output_fname = sys.argv[1]
 
-    simplified_data = _remove_unicode(json.loads(sys.stdin.read()))
+    simplified_data = json.loads(sys.stdin.read())
+    if sys.version_info.major <= 2:
+        simplified_data = _remove_unicode(simplified_data)
 
     result = execute_rv_commands(simplified_data, session_file)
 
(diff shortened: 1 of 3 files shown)


jminor (approved):
Thanks for the fix!

PR #1273, opened 2022-04-18
Fix for external Imath, must set USE_DEPS_IMATH OFF

### LINKED ISSUE #1272: Use of external Imath is not set up correctly
### Build Problem

Using an external Imath doesn't work because a flag forcing the use of the interned Imath is always YES.

meshula:

Fixes #1272

**Summarize your change.**

If an external IMath is indicated via `OTIO_FIND_IMATH`, `USE_DEPS_IMATH` should be OFF.

Also, make the Imath directory public so the tests can see it, and make it system because we use angle brackets to access Imath headers in the code.


CMakeLists.txt
```diff
@@ -247,8 +247,20 @@ if (OTIO_FIND_IMATH)
         message(STATUS "Using Imath::Imath specified by ${Imath_CONFIG}")
     endif()
 else()
-    message(STATUS "Using Imath in OpenTimelineIO/src/deps")
-    set(USE_DEPS_IMATH ON)
+    if ("${ILMBASE_LIBRARIES}" STREQUAL "")
```
JeanChristopheMorinPerso, line 250: I am curious, how/what is going to set this variable?

meshula: It's used in src/deps/CMakeLists.txt, where the condition reads as below. The idea is that if Imath is provided externally, for example, if OTIO is part of a larger build, we need to respect the external Imath, and not use the one included via `add_subdirectory(Imath)`.

```
if (USE_DEPS_IMATH)
    # preserve BUILD_SHARED_LIBS options for this project, but set it off for Imath
    option(BUILD_SHARED_LIBS "Build shared libraries" ON)
    set(BUILD_SHARED_LIBS OFF)

    # If we do not want Imath to install headers and CMake files use the EXCLUDE_FROM_ALL option
    if(OTIO_CXX_INSTALL AND OTIO_DEPENDENCIES_INSTALL)
        add_subdirectory(Imath)
    else()
        add_subdirectory(Imath EXCLUDE_FROM_ALL)
    endif()
endif()
```

meshula: `ILMBASE_LIBRARIES` is the variable we use for Imath 2, to set up the dependency:

```
if (USE_DEPS_IMATH)
    # preserve BUILD_SHARED_LIBS options for this project, but set it off for Imath
    option(BUILD_SHARED_LIBS "Build shared libraries" ON)
    set(BUILD_SHARED_LIBS OFF)

    # If we do not want Imath to install headers and CMake files use the EXCLUDE_FROM_ALL option
    if(OTIO_CXX_INSTALL AND OTIO_DEPENDENCIES_INSTALL)
        add_subdirectory(Imath)
    else()
        add_subdirectory(Imath EXCLUDE_FROM_ALL)
    endif()
endif()
```

and then finally:

```
target_link_libraries(opentimelineio 
    PUBLIC opentime ${OTIO_IMATH_TARGETS})
```

JeanChristopheMorinPerso: So it `ILMBASE_LIBRARIE` would be externally provided?

meshula: yes.

- OTIO_FIND_MATH if NO (which is the default)
    - if ILMBASE_LIBRARIES is supplied, use that
    - if not supplied, use src/deps/Imath
- else OTIO_FIND_IMATH is YES
   - use cmake's find_package
       - prefer Imath 3, because Imath 2 is deprecated
       - if Imath 3 not found, look for Imath 2
       - if Imath 2 not found, use src/deps/Imath

meshula: we could add yet another variable ~ OTIO_IMATH_LIBRARIES ~ instead of using ILMBASE_LIBRARIES...

meshula: @JeanChristopheMorinPerso Okay, I rewrote it a little per this conversation to make it obvious and easy to configure.

JeanChristopheMorinPerso: Indeed, it's easier to read!

JeanChristopheMorinPerso (approved):

jminor (approved):

PR #1275, opened 2022-04-19
Fix MediaReference per issues identified in latest Xcode

### LINKED ISSUE #1274: xcode 13.3 doesn't like the use of the keyword final when 'virtual` is not defined on the function
## Bug Report

xcode 13.3 doesn't like the use of the keyword final when 'virtual` is not defined on the function. This exposed a small chain reaction of other small polymorphic issues.

meshula:

Fixes #1274

xcode 13.3 doesn't like the use of the keyword final when 'virtual` is not defined on the function. Fixing that revealed a small cascade of related C++ polymorphism issues.


src/opentimelineio/multiMediaReference.cpp
```diff
@@ -1,23 +1,22 @@
...
-    , _available_range(available_range)
 {}
 
 MultiMediaReference::~MultiMediaReference()
 {}
 
 optional<TimeRange>
-MultiMediaReference::available_range() const noexcept final
+MultiMediaReference::available_range() const noexcept
 {}
```
jminor, line 15: Isn't this method body missing a return statement?

rogernelson: I think this whole file can just be removed! Media Multi-Reference is handled at the clip level.  It looks like was an early prototype I was working on that accidentally got checked in with my spatial coordinates branch in commit 8f93157e.  Sorry about that!

rogernelson: It's odd that this file was being compiled, right? It's not in the CMakeFile ...

meshula: yes. Now that you mention it, I believe the intent was that these functions don't do anything. Weird that the compiler doesn't flag it.

meshula: @rogernelson The Swift Bindings compile all the cpp files. That explains why it's never shown up as an error before, since I have updated the Swift Bindings to latest.

meshula: Oh, it's mean to be removed.

meshula:
The only change necessary for the Swift Bindings is the include of ImathBox. I'll open a new PR

PR #1276, opened 2022-04-19
Change angle include Imath to quote include, remove multiMediaReference prototype

meshula:

Part of the fix for https://github.com/PixarAnimationStudios/OpenTimelineIO/issues/1272

**Summarize your change.**

The Swift Package Manager does not allow setting include paths that require angle brackets. The rest of the OTIO code already uses quote includes, changing the Imath include works for both interned Imath and for system Imath, and resolves the inability of the Swift Package Manager to build OTIO.


meshula:
@rogernelson @jminor After the removal of multiMediaReference.*, the only outstanding issue for the Swift Package Manager is this.

reinecke (approved):
Looks good!

PR #1277, opened 2022-04-20
AAF Clip Enabled/Muting Update

andrewmoore-nz:
An extension of the AAF clip enabled/muting process, start in #1212

This covers the currently known cases of how clip mutes are being handled in the AAF, and the adjacent logic for the closely related scenarios of `ScopeReferece` objects, and multi-cam clip objects.

@jminor In updating/running the tests for this, it looks like the test AAF file is actually not correct for this. If the `Selected` attribute in the `Selector` object is a `ScopeReference`, as it is in that test file, I don't believe that specifically means the clip is muted. In all situations I've seen, where the clip is muted, the `Selected` attribute should be a `Filler` object. 

Is there anyone else who can weigh in on this? Looking at the AAF documentation, there's no mention of the `ScopeReference` specifically saying that it is to be considered muting the object.

### DIFF
--- a/contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py
+++ b/contrib/opentimelineio_contrib/adapters/advanced_authoring_format.py
@@ -759,17 +759,17 @@ def _transcribe(item, parents, edit_rate, indent=0):
         alternates = item.getvalue('Alternates', None)
 
         # First we check to see if the Selected component is either a Filler
-        # or ScopeReference object, meaning we have to grab the Alternate instead
+        # or ScopeReference object, meaning we have to use the alternate instead
         if isinstance(selected, aaf2.components.Filler) or \
                 isinstance(selected, aaf2.components.ScopeReference):
 
-            # Get the first element of the alternates list and transcribe
-            if len(alternates) != 1:
+            # Safety check of the alternates list, then transcribe first object -
+            # there should only ever be one alternate in this situation
+            if alternates is None or len(alternates) != 1:
                 err = "AAF Selector parsing error: object has unexpected number of " \
                       "alternates - {}".format(len(alternates))
                 raise AAFAdapterError(err)
-            alt = alternates[0]
-            result = _transcribe(alt, parents + [item], edit_rate, indent + 2)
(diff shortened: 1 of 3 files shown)


jminor:
Let me open that test AAF in Media Composer and/or make a fresh one so we can be clear. If you can make a test AAF from Premiere or Resolve that might help to illustrate the differences we saw previously.

jminor:
The AAF from this PR comes up as a muted clip in Media Composer. I made a new AAF with a slightly more complex use case for clarity. It has three clips, with the middle one muted in all tracks (V1, A1, A2). I also muted one of the audio tracks in case we want to handle that later. It looks like this in MC:
Screen Shot 2022-04-21 at 2 43 58 PM
 
With this PR as-is, OTIO sees that the audio clips are muted, but not the video clip:
Screen Shot 2022-04-21 at 2 47 33 PM

jminor:
Here's the new AAF to match those screenshots:
[Testing Muted Clips and Tracks.zip](https://github.com/PixarAnimationStudios/OpenTimelineIO/files/8535302/Testing.Muted.Clips.and.Tracks.zip)

andrewmoore-nz:
@jminor Thanks for looking in to that. I was wrong in my assumption there, as I think I'd misinterpreted some examples I was seeing a few weeks back. 

`Filler` components are being used for muted/disabled audio clips, whereas `ScopeReference` components appear to be used for muted/disable video clips. Looking at the AAF Edit Protocol, this makes a bit more sense now, as I think this is how it handles the fall back to an underlying track/clip when a video clip is muted, making sure the hierarchy of which clips should be shown is maintained. 

Interestingly enough, neither Premiere or Resolve are actually handling this correctly, or consistently. Resolve is creating a `Selector` component with `Filler` selected for an audio clip, but not for a video clip - it just comes through as a regular `SourceClip`. Premiere is not creating a `Selector` object for either, meaning that the clip mutes don't appear to be represented in the AAF at all, from what I can see. 

Anyway... I've adjusted this PR and it's now passing the tests. I'll update our Matchbox branch to reflect this as well.

jminor (approved):

jminor:
Thanks @andrewmoore-nz !

PR #1283, opened 2022-04-28
Write color decision list to ale

MichaelPlug:
### Issue https://github.com/PixarAnimationStudios/OpenTimelineIO/issues/1282

During the adapting from .otio to .ale, I read the color decision list (ASC_SAT, ASC_SOP) from the .otio file and I save it in the .ale file as a column.

In my opinion this changes can be useful to reduce the loss of data during the export of files.


contrib/opentimelineio_contrib/adapters/ale.py
```diff
@@ -324,7 +324,7 @@ def write_to_string(input_otio, columns=None, fps=None, video_format=None):
                     columns.append(key)
 
     # Always output these
-    for c in ["Duration", "End", "Start", "Name", "Source File"]:
+    for c in ["Duration", "End", "Start", "Name", "Source File", "ASC_SOP", "ASC_SAT"]:
```
reinecke, line 327: I think we may not want to add `ASC_SOP` and `ASC_SAT` here - we only want to include those columns if one of the clips uses them.

contrib/opentimelineio_contrib/adapters/ale.py
```diff
@@ -362,6 +362,27 @@ def val_for_column(column, clip):
...
+            asc_sop = ""
+            for i in range(3):
+                asc_sop += "("
+                asc_sop += str(asc_sop_arr[i][0]) + " "
+                asc_sop += str(asc_sop_arr[i][1]) + " "
+                asc_sop += str(asc_sop_arr[i][2]) + ")"
+            return asc_sop
+        elif column == "ASC_SAT":
+            asc_sat = clip.metadata.get("cdl", {}).get("asc_sat", None)
+            if clip.metadata.get is not None:
```
reinecke, line 382: This is checking that the `get` method `is not None` on the clip's `metadata` dictionary - that will always be `True` since methods are not `None`.
Did you mean to check `if asc_sat is not None`?

reinecke (changes requested):
Thank you for the contribution!
I made a few notes in the PR, also we'd like to see unittests that cover the new functionality and make sure it doesn't get broken in the future.
You may also want to check on #1054 - I think it may implement the same functionality.

MichaelPlug:
Thank you for you're help and your suggestions. These two PR are very similar, I miss it, I think that we are a the same point, more or less.
Do you think that this part is ok? The next step is to implement one or more test about this new functionality? In the other PR this wasn't done, right?

MichaelPlug:
I'm sorry, I don't get what are the next steps to finish the job

MichaelPlug:
I'm sorry  @reinecke , I didn't get what the next steps are to finish the job

jminor:
Now that the OTIO adapters have each moved to their own repository, this PR should be continued here: https://github.com/OpenTimelineIO/otio-ale-adapter/

Also, this might overlap this PR: https://github.com/AcademySoftwareFoundation/OpenTimelineIO/pull/1054

PR #1292, opened 2022-05-09
Fixes #1188 BREAKING CHANGE: change "effect" argument to Gap constructor to "effects"

### LINKED ISSUE #1188: Gap python constructors have "effects" list mislabeled "effect"

The python [gap constructors](https://github.com/PixarAnimationStudios/OpenTimelineIO/blob/990a54ccbe6488180a93753370fc87902b982962/src/py-opentimelineio/opentimelineio-bindings/otio_serializableObjects.cpp#L373-L394) accept a keyword argument `effect` which is a list of `Effect` instances:
```python
1. __init__(self: opentimelineio._otio.Gap, name: str = '', source_range: opentimelineio._opentime.TimeRange = otio.opentime.TimeRange(start_time=otio.opentime.RationalTime(value=0, rate=1), duration=otio.opentime.RationalTime(value=0, rate=1)), effect: object = None, markers: object = None, metadata: object = None) -> None

2. __init__(self: opentimelineio._otio.Gap, name: str = '', duration: opentimelineio._opentime.RationalTime = otio.opentime.RationalTime(value=0, rate=1), effect: object = None, markers: object = None, metadata: object = None) -> None
```

## To Reproduce

Using the python REPL, the following shows the issue:

```python
>>> otio.schema.Gap(effects=[])
TypeError: __init__(): incompatible constructor arguments. The following argument types are supported:
    1. opentimelineio._otio.Gap(name: str = '', source_range: opentimelineio._opentime.TimeRange = otio.opentime.TimeRange(start_time=otio.opentime.RationalTime(value=0, rate=1), duration=otio.opentime.RationalTime(value=0, rate=1)), effect: object = None, markers: object = None, metadata: object = None)
    2. opentimelineio._otio.Gap(name: str = '', duration: opentimelineio._opentime.RationalTime = otio.opentime.RationalTime(value=0, rate=1), effect: object = None, markers: object = None, metadata: object = None)

Invoked with: kwargs: effects=[]
>>> otio.schema.Gap(effect=[])
otio.schema.Gap(name='', source_range=otio.opentime.TimeRange(start_time=otio.opentime.RationalTime(value=0, rate=1), duration=otio.opentime.RationalTime(value=0, rate=1)), effects=[], markers=[], metadata={})
```

## Expected Behavior

The library convention is to make the constructor arg match the property name (`effects` in this case).

## Additional Context

Is the `effects` list useful to specify in a Gap constructor? What is the semantic meaning of an effect on a Gap? Is it possible for an effect on a gap to try and convert it to something other than 100% transparent or somehow yielding audio?

visajshah:
This PR fixes #1188.

The argument `effect` in the Gap constructor has been changed to `effects`.

### DIFF
--- a/src/py-opentimelineio/opentimelineio-bindings/otio_serializableObjects.cpp
+++ b/src/py-opentimelineio/opentimelineio-bindings/otio_serializableObjects.cpp
@@ -390,7 +390,7 @@ static void define_items_and_compositions(py::module m) {
                                          py_to_any_dictionary(metadata)); }),
              py::arg_v("name"_a = std::string()),
              "source_range"_a = TimeRange(),
-             "effect"_a = py::none(),
+             "effects"_a = py::none(),
              "markers"_a = py::none(),
              py::arg_v("metadata"_a = py::none()))
        .def(py::init([](std::string name, RationalTime duration, py::object effects,
@@ -401,7 +401,7 @@ static void define_items_and_compositions(py::module m) {
                                          py_to_any_dictionary(metadata)); }),
              py::arg_v("name"_a = std::string()),
              "duration"_a = RationalTime(),
-             "effect"_a = py::none(),
+             "effects"_a = py::none(),
              "markers"_a = py::none(),
              py::arg_v("metadata"_a = py::none()));
 


reinecke:
@visajshah Thanks for the fix!

@ssteinbach and @jminor - I have two questions I'd like your opinions on:

1. There apparently aren't any tests that are covering any part of the gap constructor, should there be? Do we want to gate merging this PR on that or take that as a separate task?
2.  This is technically a breaking change, but the impact seems pretty low given that effects on a `Gap` are a somewhat odd case. Are we concerned about that?

My personal opinion is that since this fixes what was likely a typo, it brings the constructor arg on `Gap` in-line with `Item`, and the keyword arg is likely rarely used, the breaking change is justified.
On the tests question, I'm a bit more up in the air.

ssteinbach:
> @visajshah Thanks for the fix!
> 
> @ssteinbach and @jminor - I have two questions I'd like your opinions on:
> 
> 1. There apparently aren't any tests that are covering any part of the gap constructor, should there be? Do we want to gate merging this PR on that or take that as a separate task?
I'm not against adding a test for the gap constructor, but the gap constructor *is* called, for example: 
https://github.com/PixarAnimationStudios/OpenTimelineIO/blob/1ec6f07f1af525ba4ca0aa91e01e5939d6237f01/tests/test_composition.py#L382
Unless someone feels really strongly that it needs to be added, I don't think this PR needs to add a test, unless the author wants to.  It would entail adding a test_gap.py with a call to the constructor and pass an argument into the effect parameter.

> 2. This is technically a breaking change, but the impact seems pretty low given that effects on a `Gap` are a somewhat odd case. Are we concerned about that?
I think this is ok, I would vote it just needs to be called out in the release notes for the release.  We've done similar things before.

> My personal opinion is that since this fixes what was likely a typo, it brings the constructor arg on `Gap` in-line with `Item`, and the keyword arg is likely rarely used, the breaking change is justified. On the tests question, I'm a bit more up in the air.
My vote is: accept as is, but change the title of the PR to be more representative of the change (which I just did, but feel free to adjust however).  Make sure to call out in release notes.

ssteinbach (approved):
Pending consensus, looks good.

reinecke:
I think @ssteinbach and I are in alignment on this.
@visajshah We can go ahead and merge this, just let us know whether or not you'd like to add any tests.

reinecke (approved):

visajshah:
@reinecke, I think we can merge this without adding any new tests.

PR #1297, opened 2022-05-14
Fix parsing of time_strings lacking leading zeroes

meshula:

#1293

@jchen9 Hi Julian, here's a rework of the existing routine to address the problem you found, and introduce a little more rigor to what is or is not a well-formed time string.

```Fixes #1293```

**Summarize your change.**

- fixes parsing of time strings without leading zeroes.
- enforces that a negative sign can only appear in the left most position
- implementation does not allocate memory or copy strings
- implementation does not allow exponential notation and other things that std does allow but are inappropriate for time strings
-  compatible with strings produced by ffprobe

adds C based tests corresponding to the existing Python based tests.

### DIFF
--- a/src/opentime/rationalTime.cpp
+++ b/src/opentime/rationalTime.cpp
@@ -311,8 +311,9 @@ set_error(std::string const& time_string,
         *err = ErrorStatus(
             code,
             string_printf(
-                "Input time string '%s' is an invalid time string",
-                time_string.c_str()));
+                "Error: '%s' - %s",
+                time_string.c_str(),
+                ErrorStatus::outcome_to_string(code).c_str()));
     }
 }
 
--- a/src/opentime/rationalTime.cpp
+++ b/src/opentime/rationalTime.cpp
@@ -122,7 +122,14 @@ parseFloat(char const* pCurr, char const* pEnd, bool allow_negative, double* res
     }
 
     // get integer part
-    uint32_t uintPart = 0;
+    //
+    // Note that uint64_t is used because overflow is well defined for
+    // unsigned integers, but it is undefined behavior for signed integers,
+    // and floating point values are couched in the specification with
+    // the caveat that an implementation may be IEEE-754 compliant, or only
+    // partially compliant.
+    //
+    uint64_t uintPart = 0;
     while (pCurr < pEnd) 
     {
         char c = *pCurr;
@@ -130,7 +137,7 @@ parseFloat(char const* pCurr, char const* pEnd, bool allow_negative, double* res
         {
             break;
         }
-        uint32_t accumulated = uintPart * 10 + c - '0';
+        uint64_t accumulated = uintPart * 10 + c - '0';
         if (accumulated < uintPart)
         {
             // if there are too many digits, resulting in an overflow, fail
@@ -142,6 +149,13 @@ parseFloat(char const* pCurr, char const* pEnd, bool allow_negative, double* res
     }
 
     ret = (double) uintPart;
+    if (uintPart != (uint64_t) ret)
+    {
+        // if the double cannot be casted precisely back to uint64_t, fail
+        // A double has 15 digits of precision, but a uint64_t can encode more.
+        *result = 0.0;
+        return false;
+    }
 
     // check for end of string or delimiter
     if (pCurr == pEnd || *pCurr == '\0') 
--- a/src/opentime/rationalTime.cpp
+++ b/src/opentime/rationalTime.cpp
@@ -128,10 +128,16 @@ parseFloat(char const* pCurr, char const* pEnd, bool allow_negative, double* res
 
     ret = (double) uintPart;
 
-    // check for fractional part
-    if (pCurr == pEnd || *pCurr != '.') {
+    // check for end of string or delimiter
+    if (pCurr == pEnd || *pCurr == '\0') {
         *result = sign * ret;
-        return true;   // no decimal, not a float, stop parsing
+        return true;
+    }
+
+    // if the next character is not a decimal point, the string is malformed.
+    if (*pCurr != '.') {
+        *result = 0.f; // zero consistent with earlier error condition
+        return false;
     }
 
     ++pCurr; // skip decimal
--- a/src/opentime/rationalTime.cpp
+++ b/src/opentime/rationalTime.cpp
@@ -97,7 +97,8 @@ is_dropframe_rate(double rate)
 static bool
 parseFloat(char const* pCurr, char const* pEnd, bool allow_negative, double* result)
 {
-    if (pCurr >= pEnd || !pCurr) {
+    if (pCurr >= pEnd || !pCurr)
+    {
         *result = 0.0;
         return false;
     }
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: tests/test_opentime.cpp [hunk 1]
```diff
@@ -38,6 +38,75 @@ main(int argc, char** argv)
         assertFalse(t1 != t3);
     });
 
+    tests.add_test("test_from_time_string", [] {
+         std::string time_string = "0:12:04";
+         auto t = otime::RationalTime(24 * (12 * 60 + 4), 24);
+         auto time_obj = otime::RationalTime::from_time_string(time_string, 24);
+         assertTrue(t.almost_equal(time_obj, 0.001));
```
jminor, line 45: Could you add a test for negative values, and another to test an invalid input?

jminor (changes requested):

src/opentime/rationalTime.cpp
```diff
@@ -94,6 +94,63 @@ is_dropframe_rate(double rate)
...
+        uintPart = uintPart * 10 + c - '0';
+        ++pCurr;
+    }
+
+    ret = (double) uintPart;
+
+    // check for fractional part
+    if (pCurr == pEnd || *pCurr != '.') {
+        *result = sign * ret;
+        return true;   // no decimal, not a float, stop parsing
```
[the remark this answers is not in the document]
meshula: That's a good point.

tests/test_opentime.cpp
meshula: I should add assertions that `assertTrue` for invalid time returns for all the error conditions that are tested for in the parser...

src/opentime/rationalTime.cpp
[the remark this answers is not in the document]
meshula: fixed

tests/test_opentime.cpp
meshula: added a test to raise a value error for a negative time code, and also pink elephants.

src/opentime/rationalTime.cpp
```diff
@@ -94,6 +94,69 @@ is_dropframe_rate(double rate)
...
+    }
+
+    double ret = 0.0;
+    double sign = 1.0;
+
+    if (*pCurr == '+')
+        ++pCurr;
+    else if (*pCurr == '-') {
+        if (!allow_negative) {
+            *result = 0.f;
```
jminor, line 112: Some places in this code use `0.f` and others `0.0` is there a meaningful difference?

src/opentime/rationalTime.cpp
```diff
@@ -94,6 +94,69 @@ is_dropframe_rate(double rate)
...
+    if (pCurr >= pEnd || !pCurr) {
+        *result = 0.0;
+        return false;
+    }
+
+    double ret = 0.0;
+    double sign = 1.0;
+
+    if (*pCurr == '+')
+        ++pCurr;
```
jminor, line 109: Do we have a code convention/opinion on mixing braces or non-braces?

src/opentime/rationalTime.cpp
```diff
@@ -94,6 +94,69 @@ is_dropframe_rate(double rate)
...
+
+    // if the next character is not a decimal point, the string is malformed.
+    if (*pCurr != '.') {
+        *result = 0.f; // zero consistent with earlier error condition
+        return false;
+    }
+
+    ++pCurr; // skip decimal
+
+    double scaler = 0.1;
```
jminor, line 145: On first read, I was stuck on whether this is a "scalar" or a scale factor. I see that it is the order of magnitude, or place-value scale. Maybe just a comment would help others avoid the confusion?

src/opentime/rationalTime.cpp
```diff
@@ -205,55 +268,88 @@ RationalTime::from_timecode(
...
+        if (*current == ':') {
+            parse_end = current + 1;
+            char c = *parse_end;
+            if (c != '\0' && c != ':') {
+                if (c< '0' || c > '9') {
+                    set_error(time_string, ErrorStatus::INVALID_TIME_STRING, error_status);
+                    return RationalTime::_invalid_time;
+                }
+                double val = 0.0;
+                parseFloat(parse_end, prev_parse_end + 1, false, &val);
```
jminor, line 318: The success/fail return value of parseFloat should be checked and turned into a set_error call, right? Same thing for the 2nd use of parseFloat below.

jminor (changes requested):

src/opentime/rationalTime.cpp
```diff
@@ -94,6 +94,69 @@ is_dropframe_rate(double rate)
...
+    }
+
+    // get integer part
+    uint32_t uintPart = 0;
+    while (pCurr < pEnd) {
+        char c = *pCurr;
+        if (c < '0' || c > '9')
+            break;
+        uintPart = uintPart * 10 + c - '0';
+        ++pCurr;
```
markreidvfx, line 126: shouldn't it also be a error here if `uintPart` this overflows?

src/opentime/rationalTime.cpp
meshula: 0.f is a floating point constant, 0.0 is a double. It should be 0.0 here, although it won't affect codegen.

src/opentime/rationalTime.cpp
meshula: the rest of the file is as bracey as possible, and new lines before braces. I'll update

src/opentime/rationalTime.cpp
meshula: I've renamed it "position_scale" for unambiguity.

src/opentime/rationalTime.cpp
meshula: done

src/opentime/rationalTime.cpp
meshula: That had not occurred to me. I added an overflow check

src/opentime/rationalTime.cpp
```diff
@@ -205,55 +288,108 @@ RationalTime::from_timecode(
...
+static void
+set_error(std::string const& time_string,
+          ErrorStatus::Outcome code,
+          ErrorStatus* err)
+{
+    if (err) {
+        *err = ErrorStatus(
+            code,
+            string_printf(
+                "Input time string '%s' is an invalid time string",
```
ssteinbach, line 300: Should either this message be more generic, or this specific message passed in?  Or maybe the function name made more specific?  It seems very focused on invalid time string but has the name `set_error`

ssteinbach (commented):
Looks good. One little question about naming. Like to see those c++ tests start to fill in!

src/opentime/rationalTime.cpp
rogernelson: There's a `.clang-format` file in the root of the repo.  Running clang-format should confirm everything to the same standard.

src/opentime/rationalTime.cpp
```diff
@@ -94,6 +94,103 @@ is_dropframe_rate(double rate)
...
+        {
+            // if there are too many digits, resulting in an overflow, fail
+            *result = 0.0;
+            return false;
+        }
+        uintPart = accumulated;
+        ++pCurr;
+    }
+
+    ret = (double) uintPart;
```
rogernelson, line 151: You might want to consider a static_cast here instead of a C cast -- static_casts are verified at compile team and C casts at runtime (and can thus fail).

src/opentime/rationalTime.cpp
meshula: ok, made set_error generic

src/opentime/rationalTime.cpp
meshula: sounds good, done

ssteinbach (approved):
Lgtm, thanks!

jminor (approved):

PR #1305, opened 2022-05-18
Pass correct element when generating nested sequences in FCP XML adapter

reinecke:
**Summarize your change.**

When adding a nested sequence in an FCP XML timeline, the incorrect element was being passed causing nested stacks to be generated incorrectly.

** TODO **

- [ ] Add tests including an FCP XML file exhibiting the issue


reinecke:
The adapter has moved out to [OpenTimelineIO/otio-fcp-adapter](https://github.com/OpenTimelineIO/otio-fcp-adapter) - this change should move there

PR #1312, opened 2022-05-21
Update Pybind11 to 2.9.2

### LINKED ISSUE #1310: Update Pybind11 to 2.9.2
## Feature Request

Update to Pybind11 to the latest version that supports Python 2.7 (which as of now is 2.9.2)

## Description

We are lagging behind a little bit in terms of Pybind11 versions. We are currently on 2.8.x while the latest is 2.9.2. The 2.9 series is the last to support Python 2.7. Pybind11 2.10 will drop support for Python 2.7.

## Context

Updating would allow us to have a better Sphinx documentation process as it would get rid of a Sphinx warning. The warning in question is:
```
docstring of opentimelineio._otio.Item:: WARNING: py:class reference target not found: bool_
```
and we can't get rid of it without updating pybind11. Once https://github.com/AcademySoftwareFoundation/OpenTimelineIO/pull/1209 gets merged, it will be the last remaining Sphinx warning.

If this warning goes away, we will be able to add a GitHub action job that would ensure our Sphinx docs are warning free, which guarantees that all references are working, etc.

JeanChristopheMorinPerso:
Fixes #1310 .

This PR updates the Pybind11 version to 2.9.2.

I cd'ed into `src/deps/pybind11` and `git checkout v2.9.2`.


meshula:
Does this cover the same thing as dependabot, or are you after a specific commit? https://github.com/AcademySoftwareFoundation/OpenTimelineIO/pull/1315

JeanChristopheMorinPerso:
@meshula #1315 updates to the tip of pybind11's main branch while my PR specifically picks the commit that corresponds to 2.9.2.

The tip of pybind11's main branch is no more compatible with Python 2.

JeanChristopheMorinPerso:
For reference, Dependabot doesn't support git tags, see https://github.com/dependabot/dependabot-core/issues/1639. It will only update to the latest commit.

meshula:
Ah, right. Could you amend the commit comment with some notes along the lines in the issue? Something about "We are lagging behind a little bit in terms of Pybind11 versions. The 2.9 series is the last to support Python 2.7. Pybind11 2.10 will drop support for Python 2.7."

JeanChristopheMorinPerso:
Done. I went with "Manually update Pybind11 to 2.9.2 since dependabot doesn't support git tags and updates to the latest commit of pybind11. The 2.9 series is the last to support Python 2.7. Pybind11 will drop support for Python 2.7."

meshula (approved):

meshula:
Great, thanks!

PR #1336, opened 2022-06-22
AVB Adaptor

markreidvfx:
This PR adds a adapter for reading Avid Bin Files (AVB). 

AVB format is a very similar format to AAF but unfortunately undocumented. 

[pyavb ](https://github.com/markreidvfx/pyavb)is the python module being used to add read support.  The module is pure python zero dependency like pyaaf2 but also has an optional cython acceleration module that needs a compiler to build. Wheels are available on [pypi ](https://pypi.org/project/pyavb/) and on [github](https://github.com/markreidvfx/pyavb/releases) that include the cython module pre-built.

The initial adapter is pretty much a copy/paste of the AAF Adapter and is still a little rough around the edges. I'm currently not happy with the amount of metadata being exported, large timelines can result in very large files. I have some idea on how to reduce file size by not storing properties set to their default values. Also a lot of the metadata probably isn't useful in its current state, it might be better to translate the metadata to closer match AAF.

I haven't measured performance, but large timelines do feel slow. I have some hunches of things to improve perf, but further investigation is needed.

Further work is also required for retimes and effects. The metadata for retimes exported as is and is very similar to AAF. 

I'm still in the progress of converting AAF tests to AVB tests

Looking forward to feedback!


markreidvfx:
I going to open a new pull request for this that isn't off my main branch

PR #1346, opened 2022-07-07
Prevent divide by zero and print warning if width is zero

markreidvfx:
When working on an adapter I found this potential divide by zero error, causing nothing to show up at all in the timeline viewer.


meshula (approved):
lgtm!

jminor (approved):

PR #1353, opened 2022-07-15
Add unit tests in preparation for moving the aaf adapter

markreidvfx:
This pull request should keep all the coverage lost if #1348 was merged


meshula (approved):
Love it :)

PR #1354, opened 2022-07-15
Add basic mingw_x86_64 build support

markreidvfx:
This is the minimal amount of changes required to add mingw_x86_64 support on windows and pass all the tests.

This pull request is mainly for discussion as to if this is a platform OTIO wishes to support. I haven't looked into adding the platform to CI yet, but feels like it might be a bit of a headache to get working. 

I also haven't been able to get it to build in 32-bit mode yet, just in 64-bit.

Not being able to build on mingw has been mentioned here https://github.com/AcademySoftwareFoundation/OpenTimelineIO/discussions/1342#discussioncomment-3088730

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -131,6 +131,36 @@ jobs:
         name: py-opentimelineio-codecov
         fail_ci_if_error: true
 
+  py_build_test_mingw64:
+    runs-on: windows-latest
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        submodules: 'recursive'
+    - name: Setup MSYS2
+      uses: msys2/setup-msys2@v2
+      with:
+        msystem: mingw64
+        install: >-
+          mingw-w64-x86_64-python
+          mingw-w64-x86_64-python-setuptools
+          mingw-w64-x86_64-python-pip
+          mingw-w64-x86_64-python-wheel
+          mingw-w64-x86_64-python-build
+          mingw-w64-x86_64-gcc
+          mingw-w64-x86_64-cmake
+          make
+    - name: Build Wheel
+      shell: msys2 {0}
+      run: |
+        python --version
+        python -m build -w .
+    - name: Install and Test
+      shell: msys2 {0}
+      run: |
+        python -m pip install -I dist/*.whl
+        make test
+
   package_wheels:
     needs: py_build_test
     runs-on: ${{ matrix.os }}
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -88,6 +88,15 @@ jobs:
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
         python-version: ['2.7', '3.7', '3.8', '3.9', '3.10']
+        include:
+          - { os: ubuntu-latest, shell: bash }
+          - { os: macos-latest, shell: bash }
+          - { os: windows-latest, shell: pwsh }
+          - { os: windows-latest, shell: msys2, python-version: 'mingw64' }
+
+    defaults:
+      run:
+        shell: '${{ matrix.shell }} {0}'
 
     env:
       OTIO_CXX_COVERAGE_BUILD: ON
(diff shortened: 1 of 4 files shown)


meshula (commented):
This change looks easy to maintain, especially if there's a way to add it to CI.

Assuming we can sort CI, I'd suggest we factor the platform tests, as the sort of logic we need here has a tendency to decay over time...

In other words, instead of 

```
if platform.system() == "Windows" and not self.plat_name.startswith('mingw'):
```
this would probably be more maintainable, assuming we write the needed functions:
```
if WIndows() and MinGW():
```

JeanChristopheMorinPerso:
I'm not entirely sure what we should test in CI. Maybe just test that it compiles and that it can be imported? Or maybe run the whole thing but just on one version of Python...

markreidvfx:
@JeanChristopheMorinPerso I wasn't sure what the best approach would be either.  I guess we would want to pin the mingw version and run the unit test for its version of python.

I was going to see what other projects might be doing but haven't found anything yet.  It doesn't seem to be a cibuildwheel supported platform either. I'll do a little more investigation.

Supporting this platform is probably not be worth the effort if there is no demand for it.

meshula:
The reason I mentioned CI is that I don't think we have expertise to maintain it on the team, so CI would be the only way we'd know if something needs addressing, unless there was an active community around the target. Since the change is mostly detecting the platform of picking a generator, I wouldn't expect it to be overwhelming to maintain as long as otioview isn't part of it, and as long as it's as Mark suggested pinned.

As a counterpoint, OpenEXR is an example project that is known to work with MinGW, and the community has been gracious enough to do the maintenance over the years. So perhaps a non-rigorous approach is fine, as long as we aren't making an over arching promise that it always works. 

This could be a good topic to raise at an upcoming TSC meeting to take the temperature?

markreidvfx:
Yeah we can discuss it during the next TSC meeting.

markreidvfx:
I have a rough github actions building / testing examples here.
https://github.com/markreidvfx/otio-MinGW-w64-test

Getting it integrated into otio's build matrix might still be tricky, but I'll keep chipping away at it.

.github/workflows/python-package.yml
```diff
@@ -131,6 +131,36 @@ jobs:
...
+    - uses: actions/checkout@v3
+      with:
+        submodules: 'recursive'
+    - name: Setup MSYS2
+      uses: msys2/setup-msys2@v2
+      with:
+        msystem: mingw64
+        install: >-
+          mingw-w64-x86_64-python
+          mingw-w64-x86_64-python-setuptools
```
JeanChristopheMorinPerso, line 146: Only python, pip, make and GCC should be required. Like that. `build` can be installed with `python -m pip install build` (which I trust more i think than the one from mingw) and the rest is defined in the `pyproject.toml` and will be installed in a virtualenv by `build`.

.github/workflows/python-package.yml
```diff
@@ -131,6 +131,36 @@ jobs:
         name: py-opentimelineio-codecov
         fail_ci_if_error: true
 
+  py_build_test_mingw64:
```
JeanChristopheMorinPerso, line 134: I'm wondering if this could just be integrated into the `py_build_test` matrix using an [include](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#example-adding-configurations) with an `if` on the msys2 action. Would that work instead of creating a new job?

.github/workflows/python-package.yml
markreidvfx: Thanks for taking a look at this! I'll take a look at reducing the these dependencies.

.github/workflows/python-package.yml
markreidvfx: I started doing it that way, but it got really messy.  This was much simpler to get working.  I can take another crack at it if that would be preferred.

I'm also not uploading the artifacts as they are not supported by pypi.

JeanChristopheMorinPerso:
Your last commit looks much better @markreidvfx !

markreidvfx:
I'm happier with it too! The only thing I haven't figured out is how to pin the MSYS2 version.  There is some discussion of being able to do that here: https://github.com/msys2/setup-msys2/issues/180, but the feature has not been implemented.

ssteinbach:
@markreidvfx does the MSYS2 pinning represent a blocking fix?  Is this otherwise ready to land?  Thanks!

markreidvfx:
Pining the MSYS2 version is the only thing I haven't been able to solve. 
From what I understand MSYS2 only updates their base packages once a year. Unless you run a pacman sync command to get updates on top of that (which I don't) you get the same base version for that period of time.  
I personally think that it should be stable enough to test against and if it proves to be problematic, it is pretty easy to turn off the CI.

meshula (approved):

meshula (approved):
Discussed; seems ready to land. Further improvements are possible pending community feedback.

PR #1360, opened 2022-07-26
Added SVG and PDF vector logos.

jminor:
We already had PNG logos for OpenTimelineIO in a variety of sizes, but we did not have vector logos available in the repo. This PR adds both SVG and PDF format vector logos.


reinecke (approved):
Awesome! I was wondering if these were floating around somewhere!

JeanChristopheMorinPerso:
I was also in need of the SVG version not long ago!

ssteinbach:
Should we put these in a more accessible place in the repo?

ssteinbach (approved):

PR #1363, opened 2022-07-26
Close Temporary file handles in tests

markreidvfx:
I noticed these file handles were leaking when running tests on mingw64.  
I matched the style that's in the files, but I think it would be better to use a `TemporaryDirectory` and cleanup when the test is done.  Some tests are doing that, but it is a bit mixed.

### DIFF
--- a/tests/test_otiod.py
+++ b/tests/test_otiod.py
@@ -102,7 +102,8 @@ class OTIODTester(unittest.TestCase, otio_test_utils.OTIOAssertions):
             self.assertEqual(len(manifest[fname]), count)
 
     def test_round_trip(self):
-        tmp_path = tempfile.NamedTemporaryFile(suffix=".otiod").name
+        with tempfile.NamedTemporaryFile(suffix=".otiod") as bogusfile:
+            tmp_path = bogusfile.name
         otio.adapters.write_to_file(self.tl, tmp_path)
         self.assertTrue(os.path.exists(tmp_path))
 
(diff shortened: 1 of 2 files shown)


jminor:
This looks like a good step forward to me. I always find Python's scoping rules for file handles confusing, and in this case the code really just wants a unique filename, not an actual file handle. Is there a standard pattern for cleaning up temp files in unit tests that we should be using here?

markreidvfx:
Currently some tests use `unittest.TestCase` `setUp` and `tearDown` methods

```python
    def setUp(self):
        self.tmpDir = tempfile.mkdtemp()

    def tearDown(self):
        shutil.rmtree(self.tmpDir)
```

Some tests are using `tempfile.TemporaryDirectory`

```python
with tempfile.TemporaryDirectory() as temp_dir:
    temp_file = os.path.join(temp_dir, "test_basic.otio")
```

Some tests are not cleaning up at all.

In `pyaaf2` I write all test results to known location instead of temp. I also don't delete them so you can inspect them if a failure happens.

JeanChristopheMorinPerso:
>In pyaaf2 I write all test results to known location instead of temp. I also don't delete them so you can inspect them if a failure happens.

If we were using Pytest (and here not just as a runner), we would get all these features. But that's for another day.

jminor:
There is some other discussion about using pytest here: https://github.com/AcademySoftwareFoundation/OpenTimelineIO/pull/1364

I'm fine accepting this PR as-is, and then we can look into the improvements that pytest could give us as a separate change.

jminor (approved):

Issue #1228, opened 2022-03-03
BDD Testing


KarthikRIyer:
## Feature Request

Proposal to add BDD (Behaviour Driven Development) tests to the project.

## Description

I'm thinking of migrating to BDD tests in the Java bindings using [spock](https://spockframework.org/), a groovy framework. I thought it might be helpful to do the same for the python tests.

The idea of BDD testing is to write tests in a simple user/business friendly manner. The tests become a documentation in themselves.  For example:

If we want to write a simple test to see if disabling an Item works...

The BDD specification would be like:

```feature
given: "An item is created"
when: "Item is disabled"
then: "Item does not contribute to the composition"
```

And the code in groovy would be like:
```groovy
given: "An item is created"
def item = new Item()
verify {
    item.enabled() == true
}

when: "Item is disabled"
item.setEnabled(false)

then: "Item does not contribute to the composition"
verifyAll {
        item.isEnabled() == false
}
```

It might be a lot of effort in the beginning but once you get used to it, it becomes simpler and clearer. It'll also be helpful to community members new to the project. The current tests might not be very clear to them.

There are a couple of BDD testing frameworks in python ([framework comparison](https://automationpanda.com/2019/04/02/python-bdd-framework-comparison/)). The most popular and powerful seems to be [behave](https://behave.readthedocs.io/en/stable/). There's also [pytest-bdd](https://pypi.org/project/pytest-bdd/) which is compatible with pytest. Does anybody have an opinion or idea on what could be used? I tried using behave and [implemented](https://github.com/KarthikRIyer/behave-tut) a tutorial. Seems simple enough.

Any thought on taking this forward? Or should we continue writing unit test as we do right now?

## Context

Here's the Java bindings PR where I discussed with Eric about this: [#57](https://github.com/OpenTimelineIO/OpenTimelineIO-Java-Bindings/pull/57)

reinecke:
I’ve long been a fan of pytest and would support us switching over. It seems to have pretty wide adoption at this point and I think it would provide some tools that would simplify a lot of our test code.

While BDD testing in Spock is pretty elegant, the BDD solutions for python feel to me like they’d be heavy to maintain. Every test requires keeping two files in sync with one another and there is a lot of boilerplate (in separate functions with specific decorators) required to implement them.

My vote at the moment would be to migrate to pytest, but defer on BDD unless a more elegant solution shows up.

reinecke:
[...]

PR #1364, opened 2022-07-29
Added pytest assertions

GmoZen:

**This pull request was initiated as a part of a CodeDay mentorship program.**

```Partially addresses https://github.com/AcademySoftwareFoundation/OpenTimelineIO/issues/1228```

**Summarize your change.**

Replaced unittest assertions with Pytest assertions (except those for multi-line strings, as unittests assertions were more descriptive). We're migrating unit tests from vanilla unittest to Pytest to better be inline with the greater Python community.

test_clip.py

### DIFF
--- a/tests/test_clip.py
+++ b/tests/test_clip.py
@@ -2,6 +2,7 @@
 # Copyright Contributors to the OpenTimelineIO project
 
 import unittest
+import pytest
 
 import opentimelineio as otio
 import opentimelineio.test_utils as otio_test_utils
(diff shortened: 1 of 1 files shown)


jminor:
I like the improved readability of the new test cases.

The CI is complaining about linting errors. You can run the linter locally to ensure that your change will pass that check, via `make lint`. After resolving those, you can push your changes to the same branch you've been using and this PR will update & re-run the CI checks.

Also, let us know if you need help with DCO or EasyCLA. We know those can be confusing for new contributors & we're happy to explain what those are all about.

jminor:
I think the remaining linter issues are not your fault. See this: https://github.com/AcademySoftwareFoundation/OpenTimelineIO/issues/1369

PR #1379, opened 2022-08-09
Changed assertions and fixtures

1rvine:

```CodeDay- Changing to from unit test to pytest```

I was working on pytest unit testing, I changed self.assertions to assert and created pytest fixtures.

test_core.py & test_opentime.py

### DIFF
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -5,17 +5,17 @@ import sys
 import shutil
 import tempfile
 import unittest
+import pytest
 
 import opentimelineio as otio
 
 
 class TestCoreFunctions(unittest.TestCase):
-    def setUp(self):
-        self.tmpDir = tempfile.mkdtemp()
-
-    def tearDown(self):
-        shutil.rmtree(self.tmpDir)
-
+    
+    @pytest.fixture()
+    def dir():
+        return tempfile.mkdtemp()
+    
     def test_deserialize_json_from_file_errors(self):
         """Verify that the bindings return the correct errors based on the errno"""
         if sys.version_info[0] < 3:
@@ -25,7 +25,8 @@ class TestCoreFunctions(unittest.TestCase):
 
         with self.assertRaises(excType) as exc:
             otio.core.deserialize_json_from_file('non-existent-file-here')
-        self.assertIsInstance(exc.exception, excType)
+        assert (exc.exception, excType)
+
 
     @unittest.skipUnless(not sys.platform.startswith("win"), "requires non Windows sytem")  # noqa
     def test_serialize_json_to_file_errors_non_windows(self):
@@ -36,8 +37,9 @@ class TestCoreFunctions(unittest.TestCase):
             excType = IsADirectoryError  # noqa: F821
 
         with self.assertRaises(excType) as exc:
-            otio.core.serialize_json_to_file({}, self.tmpDir)
-        self.assertIsInstance(exc.exception, excType)
+            otio.core.serialize_json_to_file({}, tempfile.mkdtemp())
+        assert (exc.exception, excType)
+
 
     @unittest.skipUnless(sys.platform.startswith("win"), "requires Windows")
     def test_serialize_json_to_file_errors_windows(self):
@@ -48,5 +50,7 @@ class TestCoreFunctions(unittest.TestCase):
             excType = PermissionError  # noqa: F821
 
         with self.assertRaises(excType) as exc:
-            otio.core.serialize_json_to_file({}, self.tmpDir)
-        self.assertIsInstance(exc.exception, excType)
+            otio.core.serialize_json_to_file({}, dir())
+        assert (exc.exception, excType)
+def myTestCoreFunctions():
+    return 
--- a/tests/test_opentime.py
+++ b/tests/test_opentime.py
@@ -14,63 +14,64 @@ class TestTime(unittest.TestCase):
     def test_create(self):
         t_val = 30.2
         t = otio.opentime.RationalTime(t_val)
-        self.assertIsNotNone(t)
-        self.assertEqual(t.value, t_val)
+        assert (t is not None) 
+        assert (t.value == t_val)
 
         t = otio.opentime.RationalTime()
-        self.assertEqual(t.value, 0)
-        self.assertEqual(t.rate, 1.0)
+        assert  (t.value ==  0)
+        assert (t.rate == 1.0)
 
     def test_equality(self):
         t1 = otio.opentime.RationalTime(30.2)
-        self.assertEqual(t1, t1)
+        assert  (t1 == t1)
         t2 = otio.opentime.RationalTime(30.2)
-        self.assertTrue(t1 is not t2)
-        self.assertEqual(t1, t2)
+        assert (t1 is not t2)
+        assert  (t1 == t2)
 
     def test_inequality(self):
         t1 = otio.opentime.RationalTime(30.2)
-        self.assertEqual(t1, t1)
+        assert (t1 == t1)
         t2 = otio.opentime.RationalTime(33.2)
-        self.assertTrue(t1 is not t2)
-        self.assertNotEqual(t1, t2)
+        assert (t1 is not t2)
+        assert  (t1 != t2)
+        
         t3 = otio.opentime.RationalTime(30.2)
-        self.assertTrue(t1 is not t3)
-        self.assertFalse(t1 != t3)
+        assert (t1 is not t3)
+        assert (t1 != t3) == False
 
     def test_comparison(self):
         t1 = otio.opentime.RationalTime(15.2)
         t2 = otio.opentime.RationalTime(15.6)
-        self.assertTrue(t1 < t2)
-        self.assertTrue(t1 <= t2)
-        self.assertFalse(t1 > t2)
-        self.assertFalse(t1 >= t2)
+        assert (t1 < t2)
+        assert (t1 <= t2)
+        assert (t1 > t2) == False
+        assert (t1 >= t2) == False
 
         # Ensure the equality case of the comparisons works correctly
         t3 = otio.opentime.RationalTime(30.4, 2)
-        self.assertTrue(t1 <= t3)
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: tests/test_core.py [hunk 1]
```diff
@@ -48,5 +50,7 @@ def test_serialize_json_to_file_errors_windows(self):
             excType = PermissionError  # noqa: F821
 
         with self.assertRaises(excType) as exc:
-            otio.core.serialize_json_to_file({}, self.tmpDir)
-        self.assertIsInstance(exc.exception, excType)
+            otio.core.serialize_json_to_file({}, dir())
+        assert (exc.exception, excType)
```
reinecke, line 54: I think this should be:
`assert isinstance(exc.exception, excType)`

This line is doing an `assert` on the `tuple`: `(exc.exception, excType)` - which will always pass because a non-empty tuple is "[truthy](https://www.freecodecamp.org/news/truthy-and-falsy-values-in-python/)"

What this assertion is meant to do is test that the exception raised is of the correct type (based on whether it's python 2 or 3) - The [`isinstance` function](https://docs.python.org/3/library/functions.html#isinstance) will check if a given instance is of the provided type (or one of it's sub-types).

You can try this out in an interactive python interpreter:
```python
>>> # Here is an exception instance, and a desired exception type
>>> exception = Exception("This is not an OSError")
>>> excType = OSError
>>> # Let's try asserting the tuple of those two
>>> assert (exception, excType)
<stdin>:1: SyntaxWarning: assertion is always true, perhaps remove parentheses?
>>> # Note that the python interpreter gives us a warning that our assertion will pass no matter what, oh no!
>>> # Let's try it now using the isinstance function
>>> assert isinstance(exception, excType)
AssertionError
>>> # Now we get an expected AssertionError because Exception is not of type OSError
>>> # Let's try it again where exception is of the type OSError
>>> exception = OSError("This is an OSError instance!")
>>> assert isinstance(exception, excType)
>>>
>>> # Note that there is no AssertionError raised in the terminal because the assertion passed as expected
```

tests/test_core.py
```diff
@@ -25,7 +25,8 @@ def test_deserialize_json_from_file_errors(self):
 
         with self.assertRaises(excType) as exc:
             otio.core.deserialize_json_from_file('non-existent-file-here')
-        self.assertIsInstance(exc.exception, excType)
+        assert (exc.exception, excType)
```
reinecke, line 28: This should use `isinstance` - see the note on line 54 for more detail.

tests/test_core.py
```diff
@@ -36,8 +37,9 @@ def test_serialize_json_to_file_errors_non_windows(self):
             excType = IsADirectoryError  # noqa: F821
 
         with self.assertRaises(excType) as exc:
-            otio.core.serialize_json_to_file({}, self.tmpDir)
-        self.assertIsInstance(exc.exception, excType)
+            otio.core.serialize_json_to_file({}, tempfile.mkdtemp())
+        assert (exc.exception, excType)
```
reinecke, line 41: This should use `isinstance` - see the note on line 54 for more detail.

tests/test_core.py
```diff
@@ -5,17 +5,17 @@
 import shutil
 import tempfile
 import unittest
```
reinecke, line 7: Since this PR is aiming to switch from unittest to pytest, I'd look at the uses of `unittest` within this module and see if you can replace them with pytest equivalents.

One example I see is the usage of the `@unittest.skipUnless` decorator - this can be replaced by "inverting" the logic on the test case and using `@pytest.mark.skipif` instead.

tests/test_opentime.py
```diff
@@ -524,95 +538,96 @@ def test_frames_with_int_fps(self):
...
         t1 = otio.opentime.from_seconds(s1)
-        self.assertEqual(t1.value, 1834)
-        self.assertEqual(t1.rate, 1)
+        assert (t1.value == 1834)
+        assert (t1.rate == 1)
         t1_as_seconds = otio.opentime.to_seconds(t1)
-        self.assertEqual(t1_as_seconds, s1)
-        self.assertAlmostEqual(float(t1.value) / t1.rate, s1)
+        assert (t1_as_seconds == s1)
+        assert (float(t1.value) / t1.rate, s1)
```
reinecke, line 556: These assertions don't quite match, the unittest version was using `assertAlmostEqual`, the updated version is asserting a tuple with two elements, which will always pass.

This should probably use [`pytest.approx`](https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest-approx):
```python
assert (float(t1.value) / t1.rate) == pytest.approx(s1)
```

tests/test_opentime.py
```diff
@@ -524,95 +538,96 @@ def test_frames_with_int_fps(self):
...
         t2 = otio.opentime.from_seconds(s2)
-        self.assertAlmostEqual(t2.value, s2)
-        self.assertAlmostEqual(t2.rate, 1.0)
+        assert (t2.value, s2)
+        assert (t2.rate, 1.0)
         t2_as_seconds = otio.opentime.to_seconds(t2)
-        self.assertAlmostEqual(s2, t2_as_seconds)
-        self.assertAlmostEqual(float(t2.value) / t2.rate, s2)
+        assert (s2, t2_as_seconds)
+        assert (float(t2.value) / t2.rate, s2)
```
reinecke, line 564: see above note about `pytest.approx`

tests/test_opentime.py
```diff
@@ -524,95 +538,96 @@ def test_frames_with_int_fps(self):
...
 
         v3 = 3459
         r3 = 24
         s3 = float(3459) / 24
         t3 = otio.opentime.RationalTime(v3, r3)
         t4 = otio.opentime.from_seconds(s3)
-        self.assertAlmostEqual(otio.opentime.to_seconds(t3), s3)
-        self.assertAlmostEqual(otio.opentime.to_seconds(t4), s3)
+        assert (otio.opentime.to_seconds(t3), s3)
+        assert (otio.opentime.to_seconds(t4), s3)
```
reinecke, line 572: see above note about `pytest.approx`

tests/test_opentime.py
```diff
@@ -649,20 +664,20 @@ def test_passing_ndf_tc_at_df_rate(self):
...
             otio.opentime.RationalTime(frames, 29.97),
             29.97,
             drop_frame=False
         )
-        self.assertEqual(tc2, NDF_TC)
+        assert (tc2 == NDF_TC)
 
         t1 = otio.opentime.from_timecode(DF_TC, 29.97)
-        self.assertEqual(t1.value, frames)
+        assert (t1.value, frames)
```
reinecke, line 677: This is asserting a two-element tuple, which will always pass. It should likely be:
`assert t1.value == frames`

tests/test_opentime.py
```diff
@@ -819,15 +834,15 @@ def test_extended_by(self):
...
 
     def test_end_time(self):
         # test whole number duration
         rt_start = otio.opentime.RationalTime(1, 24)
         rt_dur = otio.opentime.RationalTime(5, 24)
         tr = otio.opentime.TimeRange(rt_start, rt_dur)
-        self.assertEqual(tr.duration, rt_dur)
-        self.assertEqual(tr.end_time_exclusive(), rt_start + rt_dur)
+        assert  (tr.duration == rt_dur)
+        assert (tr.end_time_exclusive() == rt_start + rt_dur)
```
reinecke, line 845: For clarity, I'd suggest adjusting how parens are used in this assertion:
```python
assert tr.end_time_exclusive() == (rt_start + rt_dur)
```

tests/test_opentime.py
```diff
@@ -836,7 +851,7 @@ def test_end_time(self):
         # test non-integer duration value
         rt_dur = otio.opentime.RationalTime(5.5, 24)
         tr = otio.opentime.TimeRange(rt_start, rt_dur)
-        self.assertEqual(tr.end_time_exclusive(), rt_start + rt_dur)
+        assert (tr.end_time_exclusive() ==  rt_start + rt_dur)
```
reinecke, line 854: For clarity, I'd suggest adjusting how parens are used in this assertion:
```python
assert tr.end_time_exclusive() == (rt_start + rt_dur)
```

tests/test_opentime.py
```diff
@@ -959,49 +975,50 @@ def test_overlaps_timerange(self):
         tdur = otio.opentime.RationalTime(3, 25)
         tr_t = otio.opentime.TimeRange(tstart, tdur)
 
-        self.assertFalse(tr.overlaps(tr_t))
+        assert(tr.overlaps(tr_t)) == False
```
reinecke, line 978: Blanket note: Expressions like this should be written:
```python
assert tr.overlaps(tr_t) == False
```

tests/test_core.py
```diff
@@ -5,17 +5,17 @@
...
+import pytest
 
 import opentimelineio as otio
 
 
 class TestCoreFunctions(unittest.TestCase):
```
reinecke, line 13: Since we're converting to pytest, this no longer needs to be a subclass of `unittest.TestCase`, it can be converted to:
```python
class TestCoreFunctions:
```

reinecke (changes requested):
Looks like it's coming along!

In addition to the specific notes I gave, here are some more general notes to apply broadly across the PR:

- It looks like the [build failed](https://github.com/AcademySoftwareFoundation/OpenTimelineIO/runs/7756673102?check_suite_focus=true) mostly due to linting errors. Check out [PEP 8](https://peps.python.org/pep-0008/) and [The Black code style](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html) to get familiar with python code style guidelines.
- Simple assertions (like `assert (t.value == 0)` or `assert(tr.contains(tstart))`) should not include the parens, they can just be written as `assert t.value == 0` or `assert tr.contains(tstart)`
- Since the goal of the PR is to move from unittest to pytest, try removing the `import unittest` from the top of the files. What errors pop up when you run the tests? Those are all things that moving to pytest's standards will help address.

Issue #1295, opened 2022-05-12
Backwards and forwards compatibility


jminor:
Currently OTIO can read old `.otio` files that were serialized with older schema versions, because there is a built-in schema upgrade system. However, if an `.otio` file is written with a brand new schema revision, then software linked with an older version of OTIO will not be able to read those files. If you attempt this, a SCHEMA_VERSION_UNSUPPORTED error occurs.

In a one-way pipeline, this is manageable, as long as you are careful to upgrade the OTIO version in your software starting with the most downstream end of the pipeline, working your way back to the start. For example a pipeline which passes `.otio` files from: A to B to C is upgraded C, then B, then A.

However, one of OTIO's stated goals is to support round-trip pipelines. Even if a pipeline is only two pieces of software, A and B, they must be upgraded in lock step anytime a OTIO schema version change is introduced.

To solve this problem, we must do one of the following:
- Allow modern OTIO software to write out old OTIO schema (controlled by a configuration option).
- Allow old OTIO software to read modern OTIO schema (via some runtime plugin).
- Modify our strategy for schema changes to ensure that revisions can still be read by old software (within some range of versions). 
- Never change the schema.

Are there other options? Can we look at other file formats for inspiration or guidance on this?

meshula:
There is quite a body of work around schema migration for XML (c.f. https://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.174.6520&rep=rep1&type=pdf)

with xslt (for xml) one would set up a matcher for rev A, with a corresponding translation rule to rev B. so,

`new = xform(old, A_B_migrate)`

one could also set up a matcher for rev B, with a translation rule to A.

`old = xform(new, B_A_migrate)`

now, this transform is not transitively idempotent. i.e.

`new != xform(xform(new, B_A), A_B)`

so we introduce a further transform for elements that do not exist in A, such that the B_A migration rule moves the A-non-existent data into a B_A_migration metadata section.

finally, we modify the A_B migration to recognize the existence of B_A_migration metadata, and use it to fully reconstitute the B fields, and we achieve

`new == xform(xform(new, B_A), A_B)`

A strong candidate for the json equivalent of xslt is jq, which has a zero dependency C implementation. https://stedolan.github.io/jq/

jp is also nice, but the Golang dependency puts it out of scope for our purposes, where our current core language requirements are C/C++ and Python. https://github.com/go-jsonfile/jp/

[...]

PR #1387, opened 2022-08-25
Schema Downgrading System

ssteinbach:
## Overview

As OpenTimelineIO makes its way into vendor tools, it is likely that multiple versions of OTIO will exist in different applications that would like to interoperate bidirectionally within a workflow.  Therefore, we are adding a system that enables a newer version of the OpenTimelineIO library to write files with schemas that are compatible with older versions of the library.

See also: #1295.

## C++

- Introduces the concept of a `schema_version_map`, a mapping of schema names to desired schema versions
- Adds the `CORE_VERSION_MAP`, a compiled in constant mapping of "label" to `schema_version_map`.  Labels are currently associated with releases of the OpenTImelineIO library (ie. "0.15.0").  Intended to describe sets of schema versions that are compatible with specific releases of OpenTimelineIO.
- Adds `register_downgrade_function` to pair with `register_upgrade_function`.  Like upgrade functions, downgrade functions take an AnyDictionary and operate on it in place.
- Adds `schema_version_targets` arguments to `to_json_string` and `to_json_file` family of functions/methods.  These are `optional<schema_version_map>` and allow the user to optionally provide a set of schema version targets for downgrading during serialization
- during serialization, if a downgrade needs to happen, uses the `CloningEncoder` to build an `AnyDictionary` of the object, which is then passed through the downgrade functions and into the serializer.
- Adds `type_version_map` method to `TypeRegistry` for querying the schema names and versions of all currently registered types
- Adds the `io_perf_test` and `upgrade_downgrade_example` example C++ programs
- Adds some python-like convenience methods to `AnyDictionary`: `get_default`, `set_default` and `has_key`
- use `std::unordered_map` instead of `std::map` inside the writer for a number of internal data structures where order is not relevant
- when converting built-in but schema'd types to `AnyDictionary`, like `RationalTime`, don't store them as concrete types but store them as `AnyDictionary`.  Otherwise they can't be downgraded.
- add `override` tagging to Encoder/Writer virtual methods
- serialization performance improvements when not downgrading.  Avoided copies, `const&` etc.  20mb OTIO serializes with no downgrading in 0.03s (was 0.04s).

## Python

- Adds python bindings for registering downgrade functions, exposed via the `downgrade_function_from` decorator, to echo the `upgrade_function_for` function.
- Adds python bindings for querying the full schema/version map and CORE_VERSION_MAP
- Adds `version_manifest` field to the plugin manifest system, allowing you to define custom family/label/schema version sets to target
- Adds `OTIO_DEFAULT_TARGET_VERSION_FAMILY_LABEL` environment variable for telling the python API that you want to use that as a default downgrade target.
- Adds `autogen_version_map` python script that generates the `CORE_VERSION_MAP.cpp` file for the c++ api

## Performance

There is a performance cost to using the downgrading system.  On a 20mb OTIO file where all clips are downgraded, the difference in serialization time was a slowdown of about 4x (0.03s -> 0.12s with a release build on an M1 Mac).  Deserialization is not impacted at all.

As noted above, the baseline serialization performance was also improved, even when not downgrading.

### Performance Scaling With File Size

- all times in seconds
- dg = with downgrading
- →str = to string
- →file = to file

| OTIO size | read |  →str  |  dg→str  |  →file  |  dg→file  |
|------|------|------|------|------|------|
| 20mb | 0.18 | 0.04 | 0.14 | 0.14 | 0.25 |
| 86mb | 0.77 | 0.15 | 0.59 | 0.64 | 1.02 |
| 257mb| 2.35 | 0.47 | 1.75 | 1.83 | 3.07 |
| 1.0gb| 9.72 | 1.85 | 7.06 | 7.71 | 12.3 |

## Follow up issues
- Have a discussion about making the registration functions block double registration: #1391 
- Adding bread crumbing support to the upgrade/downgrade functions: #1392
    
## Further Design Discussions

- In the current system, if adding a new attribute to a schema and the new attribute's default value is the correct one with no further questions if it isn't present, then we don't add an upgrade function and do not increment the schema version (for example: the `enabled` flag).  For downgrading, its true that the parser ignores extra keys, so we could leave enabled in objects and read them in old versions, *however* if someone goes in and edits the value of `enabled` the old APIs won't know what to do with them and it is possible to create data that would be interpreted differently in the old system (which neither models nor reads the .enabled flag and therefore assumes everything is enabled) and the newer system, which would read that as being enabled false.
- The existing system (in particular the unit test system, but the plugin system should be tested) is built around the fact that if you double register something, it is a no-op instead of an error.   Long term we should probably address this, but it might be a bit of a retrofit for the `TypeRegistry` architecture
- Right now the core schemas are everything that is present *just* in the C++ core, and the version is tied to the release of the software library.  We probably want to separate the two versions of things and establish an "Otio interoperability version" which is a set of schema versions, independent from the versions of the library, and also discuss how schemas are included or not in that list.
- Should examples be compiled by default?  Right now they are not.  It could help to ensure that they stay current.  Maybe even better than that would be to move forward with the discussed-previously idea of splitting the C++ and python projects up and setting the C++ library up with its own unit test suite.
- Should we offer a mode where the schema/up/downgrade functions are internal only and not exposed as a public interface?  Would vendors want something like that?


examples/io_perf_test.cpp
```diff
@@ -36,12 +36,35 @@ print_elapsed_time(
...
+    for (auto kv_fam : fammap)
+    {
+        std::cerr << kv_fam.first << std::endl;
+        for (auto kv_lbl: kv_fam.second)
+        {
+            std::cerr << "  " << kv_lbl.first << std::endl;
+            for (auto kv_schema_version : kv_lbl.second)
+            {
+                std::cerr << "    \"" << kv_schema_version.first << "\": ";
+                std::cerr << kv_schema_version.second << std::endl;
```
meshula: I assume there's no performance constraint here? i.e. print_version_map doesn't have to be fast? because std::endln flushes io in addition to inserting a "\n" in the stream.

ssteinbach: io_perf_test is just an example file - I have it dumping the core schemas in its startup to stdout just for info/debugging.  I can remove that function if you think its a distraction.  It was more useful when there was code allowing the family/labels to be set in the core library.

src/opentimelineio/typeRegistry.cpp
```diff
@@ -113,6 +111,33 @@ TypeRegistry::TypeRegistry()
...
+            active_ref = media_refs.get_default(active_reference_key, AnyDictionary());
+        }
+
+        (*d)["media_reference"] = active_ref;
+
+        auto downgrade_md = AnyDictionary();
+        downgrade_md["media_references"] = media_refs;
+        downgrade_md["active_media_reference_key"] = active_reference_key;
+
+        d->set_default(
```
rogernelson, line 132: If information that is discarded during downgrading is going to preserved, I wonder if it would be worth defining a schema for that purpose so there's a standardized way to doing it.

```
"metadata": {}.
"downgrades" : [ 
   {
    "OTIO_SCHEMA": "Downgrade.1",
    "from": "Clip.2",
    "to": "Clip.1",
    "discarded": {
      "media_refererences": media_references
      "active_media_reference": active_media_reference,
      "enabled": true
    }
  }
]
```
If we ever wanted to do something with the discarded data in the future (reconstructing the upgraded version), something well defined would make it easier to do so imo rather than relying on the metadata which would be mixed in with any client metadata.

Just some food for thought.

ssteinbach: I think initially I want to start with some simpler breadcrumbs (a log in the metadata saying "this was downgraded from version X on day Y").  I'm open to trying this approach but I suspect it might be problematic and not straightforward to reconstruct versions (if there is a conflict between something new authored and old-out of date data that has been kept along, etc.).

tests/test_serializable_object.py
```diff
@@ -177,5 +136,131 @@ def test_cycle_detection(self):
...
+        f.foo_two = "a thing here"
+
+        # downgrade_target = otio.core.fetch_version_map(
+        #         "OTIO_CORE",
+        #         otio.__version__
+        # )
+
+        downgrade_target = {"FakeThingToDowngrade": 1}
+
+        result = eval(
```
JeanChristopheMorinPerso, line 252: Eval is usually frown upon. Would a `json.loads` or even a `ast.literal_eval` do the trick?

src/py-opentimelineio/opentimelineio/adapters/otio_json.py
```diff
@@ -4,11 +4,13 @@
...
 from .. import (
-    core
+    core,
+    versioning
 )
 
+import os
 
-# @TODO: Implement out of process plugins that hand around JSON
+DEFAULT_VERSION_ENVVAR = "OTIO_DEFAULT_TARGET_VERSION_FAMILY_LABEL"
```
JeanChristopheMorinPerso, line 13: You might want to add this environment variable to https://github.com/AcademySoftwareFoundation/OpenTimelineIO/blob/main/docs/tutorials/otio-env-variables.md.

src/opentimelineio/serializableObject.cpp
```diff
@@ -112,18 +113,34 @@ SerializableObject::is_unknown_schema() const
 }
 
 std::string
-SerializableObject::to_json_string(ErrorStatus* error_status, int indent) const
+SerializableObject::to_json_string(
+        ErrorStatus* error_status,
+        optional<const schema_version_map*> schema_version_targets,
```
ssteinbach, line 118: `optional<>` -> `const schema_version_map*`

src/opentimelineio/anyDictionary.h
```diff
@@ -122,6 +123,53 @@ class AnyDictionary : private std::map<std::string, any>
         map::swap(other);
     }
 
+    /// @TODO: need the right policy for accessing/returning membors
+    template<typename containedType>
+    containedType
+    get_default(
```
ssteinbach, line 129: -> `get_or_default()`

src/opentimelineio/anyDictionary.h
```diff
@@ -122,6 +123,53 @@ class AnyDictionary : private std::map<std::string, any>
...
+            const std::string& key,
+            const containedType& default_value
+    )
+    {
+        const auto& it = this->find(key);
+
+        return (
+                 it != this->end()
```
ssteinbach, line 137: add parens()

src/opentimelineio/anyDictionary.h
```diff
@@ -122,6 +123,53 @@ class AnyDictionary : private std::map<std::string, any>
...
+    get_default(
+            const std::string& key,
+            const containedType& default_value
+    )
+    {
+        const auto& it = this->find(key);
+
+        return (
+                 it != this->end() 
+                 && it->second.type().hash_code() == typeid(containedType).hash_code()
```
ssteinbach, line 138: add () outside of the &&

src/opentimelineio/anyDictionary.h
```diff
@@ -122,6 +123,53 @@ class AnyDictionary : private std::map<std::string, any>
...
 
+    /// @TODO: need the right policy for accessing/returning membors
+    template<typename containedType>
+    containedType
+    get_default(
+            const std::string& key,
+            const containedType& default_value
+    )
+    {
+        const auto& it = this->find(key);
```
ssteinbach, line 134: -> auto

tests/test_serializable_object.py
```diff
@@ -177,5 +137,131 @@ def test_cycle_detection(self):
...
+
+        f = FakeThing()
+        f.foo_two = "a thing here"
+
+        # downgrade_target = otio.core.fetch_version_map(
+        #         "OTIO_CORE",
+        #         otio.__version__
+        # )
+
+        downgrade_target = {"FakeThingToDowngrade": 1}
```
ssteinbach, line 251: delete dead comment

src/opentimelineio/anyDictionary.h
ssteinbach: switched it around to `get_if_set` but open to suggestions on this name (if we keep it, which -- also borderline on that)

docs/tutorials/versioning-schemas.md
```diff
@@ -1,16 +1,209 @@
...
+C++ Example (can be viewed/run in `examples/upgrade_downgrade_example.cpp`):
+
+```cpp
+class SimpleClass : public otio::SerializableObject
+{
+public:
+    struct Schema
+    {
+        static auto constexpr name   = "SimpleClass";
+        static int constexpr version = 2;
```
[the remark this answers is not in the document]
ssteinbach: -> separate PR

examples/io_perf_test.cpp
```diff
@@ -0,0 +1,176 @@
...
+    bool TO_JSON_STRING              = true;
+    bool TO_JSON_STRING_NO_DOWNGRADE = true;
+    bool TO_JSON_FILE                = true;
+    bool TO_JSON_FILE_NO_DOWNGRADE   = true;
+    bool CLONE_TEST                  = true;
+    bool SINGLE_CLIP_DOWNGRADE_TEST  = true;
+} RUN_STRUCT ;
+
+/// utility function for printing std::chrono elapsed time
+const void
```
darbyjohnston, line 26: Minor point, the "const" is probably not necessary.

examples/io_perf_test.cpp
```diff
@@ -0,0 +1,176 @@
...
+            return 1;
+        }
+        print_elapsed_time("serialize_json_to_string [no downgrade]", begin, end);
+    }
+
+    if (RUN_STRUCT.TO_JSON_FILE)
+    {
+        begin = std::chrono::steady_clock::now();
+        timeline.value->to_json_file(
+                "/var/tmp/io_perf_test.otio",
```
darbyjohnston, line 159: In the "util.h" header file there is a function "std::string create_temp_dir();" that creates a temporary directory in a portable way.

examples/io_perf_test.cpp
```diff
@@ -0,0 +1,176 @@
+#include <iostream>
```
darbyjohnston, line 1: It seems like "io_perf_test.cpp" is maybe more of a test than an example? There is the beginning of some C++ tests in the "tests" directory.

examples/upgrade_downgrade_example.cpp
```diff
@@ -0,0 +1,109 @@
...
+{
+public:
+    struct Schema
+    {
+        static auto constexpr name   = "SimpleClass";
+        static int constexpr version = 2;
+    };
+
+    void set_new_field(int64_t val) { _new_field = val; }
+    const int64_t new_field() const { return _new_field; }
```
darbyjohnston, line 21: Minor point, the "const" in "const int64_t" is probably not necessary.

src/opentimelineio/typeRegistry.h
```diff
@@ -9,14 +9,25 @@
 #include <algorithm>
 #include <functional>
 #include <map>
+#include <map>
```
darbyjohnston, line 12: Minor point, looks like there is already an "#include <map>"?

src/opentimelineio/typeRegistry.h
```diff
@@ -9,14 +9,25 @@
...
+class Encoder;
 class AnyDictionary;
 
+// typedefs for the schema downgrading system
+// @TODO: should we make version an int64_t?  That would match what we can
+//        serialize natively, since we only serialize 64 bit signed ints.
+using schema_version_map = std::unordered_map<std::string, int64_t>;
+using label_to_schema_version_map = std::unordered_map<std::string, schema_version_map>;
+
+extern const label_to_schema_version_map CORE_VERSION_MAP;
```
darbyjohnston, line 29: What is the "extern" here for?

darbyjohnston:
I'm not qualified to review the schema logic, just left a couple notes on the C++ code.

From your list of questions:

> Performance

Is it bad? Have you tried profiling it?

> discussed-previously idea of splitting the C++ and python projects

I like that idea.

examples/io_perf_test.cpp
ssteinbach: Well, except that it requires a path to an .otio file, and then profiles the time to read it in and out with and without downgrading. So it's a bit between example, profile harness, and test. The upgrade/downgrade example program is a more straightforward example, I agree. But I've been using it as a mix of correctness and performance test with big production files.

src/opentimelineio/typeRegistry.h
ssteinbach: CORE_VERSION_MAP is a global const data structure that is defined in CORE_VERSION_MAP.cpp - the `extern` allows anything importing typeRegistry to access it.

ssteinbach:
Yes, currently performance is about a 10x slowdown when downgrading.  I have some leads on improving it though, so stay tuned for later in the week!

examples/io_perf_test.cpp
ssteinbach: I think I've addressed these notes.  Thanks @darbyjohnston -- let me know if you see anything else or if I missed anything!

docs/tutorials/versioning-schemas.md
```diff
@@ -1,16 +1,209 @@
...
+
+OpenTimelineIO can still interoperate with older and newer versions of the library by way of the schema upgrading/downgrading system.  As OpenTimelineIO deserializes json from a string or disk, it will upgrade the schemas to the version supported by the library before instantiating the concrete in-memory object.  Similarly, when serializing OpenTimelineIO back to disk, the user can instruct OpenTimelineIO to downgrade the JSON to older versions of the schemas.  In this way, a newer version of OpenTimelineIO can read files with older schemas, and a newer version of OpenTimelineIO can generate JSON with older schemas in it.
+
+## Schema Upgrading
+
+Once a type is registerd to OpenTimelineIO, deveopers may also register upgrade functions.  In python, each upgrade function takes a dictionary and returns a dictionary.  In C++, the AnyDictionary is manipulated in place.  Each upgrade function is associated with a version number - this is the version number that it upgrades to.
+
+C++ Example (can be viewed/run in `examples/upgrade_downgrade_example.cpp`):
```
JeanChristopheMorinPerso, line 21: It could be nice if this was actually linking to the file in the repo. Or maybe we could render this example directly in the documentation?

src/py-opentimelineio/opentimelineio/adapters/otio_json.py
```diff
@@ -46,13 +81,39 @@ def write_to_string(input_otio, indent=4):
...
+
     Returns:
         str: A json serialized string representation
-    """
-    return core.serialize_json_to_string(input_otio, indent)
 
+    Raises:
+        otio.exceptions.InvalidEnvironmentVariableError: if there is a problem
+        with the default environment variable
+        "OTIO_DEFAULT_TARGET_VERSION_FAMILY_LABEL".
```
JeanChristopheMorinPerso, line 95:
```suggestion
    :raises otio.exceptions.InvalidEnvironmentVariableError: if there is a problem
        with the default environment variable
        ``OTIO_DEFAULT_TARGET_VERSION_FAMILY_LABEL``.
```

See https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#info-field-lists for the syntax.

Or is it better to keep the current style because it's easier to display in the documentation?

src/py-opentimelineio/opentimelineio/core/__init__.py
```diff
@@ -134,6 +155,45 @@ def wrapped_update(data):
...
+            return {"old_attr": data["new_attr"]}
+
+    This will get called to downgrade a schema of MyClass from version 5 to
+    version 4. MyClass must be a class deriving from
+    :class:`~SerializableObject`.
+
+    The downgrade function should take a single argument - the dictionary to
+    downgrade, and return a dictionary with the fields downgraded.
+
+    :param type cls: class to downgrade
```
JeanChristopheMorinPerso, line 175: Should the type be `typing.Type[SerializableObject]`? (see https://docs.python.org/3/library/typing.html#typing.Type for the documentation on `typing.Type`)

src/py-opentimelineio/opentimelineio/versioning.py
```diff
@@ -0,0 +1,77 @@
...
+
+def fetch_map(family, label):
+    """Fetch the version map for the given family and label.  OpenTimelineIO
+    includes a built in family called "OTIO_CORE", this is compiled into the
+    C++ core and represents the core interchange schemas of OpenTimelineIO.
+
+    Users may define more family/label/schema:version mappings by way of the
+    version manifest plugins.
+
+    Returns a dictionary mapping Schema name to schema version, like:
```
JeanChristopheMorinPerso, line 60: If these functions are part of the public API, they could benefit from actual sphinx docs:

```
:returns: dictionary mapping Schema name to schema version
:rtype: dict[str, int]
```

src/opentimelineio/serialization.cpp
```diff
@@ -819,8 +1113,16 @@ SerializableObject::Writer::write(
 }
 
 void
-SerializableObject::Writer::write(std::string const& key, any const& value)
+SerializableObject::Writer::write(
+        std::string const& key,
+        any const& value)
 {
+    // if (value.empty())
```
JeanChristopheMorinPerso, line 1120: Dead code?

src/py-opentimelineio/opentimelineio/core/__init__.py
```diff
@@ -22,9 +22,12 @@
     instance_from_schema,
     register_serializable_object_type,
     register_upgrade_function,
+    register_downgrade_function,
     set_type_record,
     _serialize_json_to_string,
     _serialize_json_to_file,
+    type_version_map,
+    release_to_schema_version_map,
```
JeanChristopheMorinPerso, line 30: Should this one be exposed in `__all__`?

tests/test_builtin_adapters.py
```diff
@@ -68,6 +68,14 @@ def test_disk_vs_string(self):
             with open(temp_file, 'r') as f:
                 on_disk = f.read()
 
+            self.maxDiff = None
+
+            with open("/var/tmp/in_memory.otio", "w") as fo:
```
JeanChristopheMorinPerso, line 73: Should this use the `tempfile` library? Also I'm struggling to understand why you write into a file without reading it later in the test.

src/opentimelineio/typeRegistry.h
darbyjohnston: Ah, I guess since the data is const it's OK to be global. On Windows there's an extra wrinkle with static data and dynamic libraries, where the app and library each get their own copy of the data. But again since it's const that seems OK.

examples/io_perf_test.cpp
darbyjohnston: Hmm, still sounds like a test to me but fair enough. :) Probably a good idea to add the SPDX headers to this and the other example.

ssteinbach: Good note!  My general policy is to try and capture these kinds of interstitial programs and put them somewhere in the repo, especially since (As you had previously noted) we don't have a ton of C++ example code.

src/opentimelineio/typeRegistry.h
ssteinbach: Good to know!

tests/test_builtin_adapters.py
ssteinbach: Diagnostic code that was left in.  Removed!

src/py-opentimelineio/opentimelineio/core/__init__.py
ssteinbach: good catch!

src/py-opentimelineio/opentimelineio/adapters/otio_json.py
ssteinbach: @reinecke you have your ear to the ground with this stuff - is it better to add markup for the  sphinx docs?

docs/tutorials/versioning-schemas.md
ssteinbach: Looks like I might have to do this as a second pass.  Unfortunately the check-links doc steps fails because the file isn't in the expected place... but the example won't exist until the PR is landed.

src/opentimelineio/typeRegistry.h
meshula: If we ever do a __declspec(dllexport) pass, this'll require special treatment. using a function accessor instead of extern const dodges the bullet. example of the complexity it introduces: https://github.com/AcademySoftwareFoundation/Imath/blob/main/src/Imath/ImathExport.h

rogernelson (commented):
I like the approach! A few c++ comments here and there, mostly minor.

docs/tutorials/versioning-schemas.md
JeanChristopheMorinPerso: Umm, effectively. At least we know the jobs I added are working!

To avoid this kind of problem, we would have to tell Sphinx that the examples are part of the documentation. Maybe https://linuxtv.org/downloads/sphinx-1.2.3/markup/code.html#directive-literalinclude or https://docutils.sourceforge.io/docs/ref/rst/directives.html#include?

src/py-opentimelineio/opentimelineio/core/__init__.py
```diff
@@ -54,28 +59,77 @@
...
 ]
 
 
-def serialize_json_to_string(root, indent=4):
-    return _serialize_json_to_string(_value_to_any(root), indent)
+def serialize_json_to_string(root, schema_version_targets=None, indent=4):
+    """Serialize root to a json string.  Optionally downgrade resulting schemas
+    to schema_version_targets.
 
+    :param SerializableObject root: root object to serialize
```
JeanChristopheMorinPerso, line 80: The type needs to be the FQDN, i.e. `opentimelineio.core.SerializableObject`. Also you might have to put the type on a separate `:type root: ...` line (sometimes Sphinx can get unhappy with "complex" types when inlned with the `:param:` instruction.

examples/io_perf_test.cpp
```diff
@@ -0,0 +1,225 @@
...
+        {"Clip", 1},
+        {"OtherThing", 12000}
+    };
+
+    if (RUN_STRUCT.CLONE_TEST)
+    {
+        otio::SerializableObject::Retainer<otio::Clip> cl = new otio::Clip("test");
+        cl->metadata()["example thing"] = "banana";
+        const auto intermediate = cl->clone(&err);
+        assert(intermediate != nullptr);
```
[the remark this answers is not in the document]
meshula: to that point, should be using an `assertNullptr(intermediate) here, and we really do need to swap out the use of assert in utils.h for something with a bit more reporting.

examples/io_perf_test.cpp
```diff
@@ -0,0 +1,225 @@
...
+        otio::SerializableObject::Retainer<otio::Clip> cl = new otio::Clip("test");
+        cl->metadata()["example thing"] = "banana";
+        chrono_time_point begin = std::chrono::steady_clock::now();
+        cl->to_json_file(
+                examples::normalize_path(tmp_dir_path + "/clip.otio"),
+                &err,
+                &downgrade_manifest
+        );
+        chrono_time_point end = std::chrono::steady_clock::now();
+        print_elapsed_time("downgrade clip", begin, end);
```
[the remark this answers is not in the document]
meshula: same note, should be using something in utils.h rather than the assert macro directly

src/opentimelineio/CORE_VERSION_MAP.cpp
```diff
@@ -0,0 +1,54 @@
...
+// This maps a "Label" to a map of Schema name to Schema version. The intent is
+// that these sets of schemas can be used for compatability with future
+// versions of OTIO, so that a newer version of OTIO can target a compatability
+// version of an older library.
+#include "opentimelineio/typeRegistry.h"
+
+namespace opentimelineio { namespace OPENTIMELINEIO_VERSION {
+
+const label_to_schema_version_map CORE_VERSION_MAP {
+    { "0.15.0.dev1",
```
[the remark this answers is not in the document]
meshula: if we did that, what would we do with the dev1?

src/opentimelineio/typeRegistry.cpp
```diff
@@ -113,6 +112,37 @@ TypeRegistry::TypeRegistry()
...
+                                &active_rkey
+                            )
+                    )
+                    {
+                        AnyDictionary active_ref;
+                        if (mrefs.get_if_set(active_rkey, &active_ref))
+                        {
+                            (*d)["media_reference"] = active_ref;
+                        }
+                    }
```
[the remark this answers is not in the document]
meshula: or an exception?

src/opentimelineio/CORE_VERSION_MAP.cpp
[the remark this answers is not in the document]
rogernelson: This would need to be defined, but one possibility is to map strings to numbers (we would need a predefined list)  "dev" = 0, "alpha" = 1, "beta" = 2, etc.

examples/io_perf_test.cpp
```diff
@@ -0,0 +1,225 @@
...
+/// utility function for printing std::chrono elapsed time
+void
+print_elapsed_time(
+        const std::string& message,
+        const chrono_time_point& begin,
+        const chrono_time_point& end
+)
+{
+    const auto dur = (
+            std::chrono::duration_cast<std::chrono::microseconds>(
```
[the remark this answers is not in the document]
ssteinbach: If I cast to seconds then I don't get sub-second resolution, but maybe I'm doing it wrong?

```
downgrade clip: 0 [s]
deserialize_json_from_file: 1 [s]
serialize_json_to_string: 1 [s]
serialize_json_to_string [no downgrade]: 0 [s]
serialize_json_to_file: 0 [s]
serialize_json_to_file [no downgrade]: 0 [s]
```

this is what I tried:

```c++
    const auto dur = (
            std::chrono::duration_cast<std::chrono::seconds>(
                end - begin
            ).count()
    );
    std::cout << message << ": " << dur << " [s]" << std::endl;
```

examples/io_perf_test.cpp
```diff
@@ -0,0 +1,225 @@
...
+            ).count()
+    );
+    std::cout << message << ": " << dur/1000000.0 << " [s]" << std::endl;
+}
+
+void
+print_version_map()
+{
+    std::cerr << "current version map: "  << std::endl;
+    for (auto kv_lbl: otio::CORE_VERSION_MAP)
```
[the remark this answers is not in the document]
ssteinbach: yup, thanks

examples/io_perf_test.cpp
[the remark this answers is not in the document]
ssteinbach: I'm not familiar with that assertion - where do I find it?

examples/io_perf_test.cpp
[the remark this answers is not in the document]
ssteinbach: well, I added the `assert()` call for now

src/opentimelineio/CORE_VERSION_MAP.cpp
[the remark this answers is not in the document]
ssteinbach: The dev1 doesn't show up anywhere outside of main branch development builds, that will disappear as part of the release process.  the released version should just be `0.15.0`.

I think its more direct to not have a key for the label if we haven't built a map for it.  We couldn't guarantee downgrading to that label if we don't have a spec or downgrade functions for it.  I'd vote against it.

src/opentimelineio/anyDictionary.h
```diff
@@ -122,6 +123,71 @@ class AnyDictionary : private std::map<std::string, any>
...
+
+        if (
+                 (it != this->end())
+                 && (
+                     it->second.type().hash_code() 
+                     == typeid(containedType).hash_code()
+                )
+        ) 
+        {
+            *result = any_cast<containedType>(it->second);
```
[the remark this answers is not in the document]
ssteinbach: I'm not sure how to repro this case, but if I try this, I get a compile error:
```c++
    otio::AnyDictionary test;
    test["danger"] = nullptr;
    decltype(nullptr) result;
    test.get_if_set("danger", result);
```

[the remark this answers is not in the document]
ssteinbach: Could that type be blocked by some other means?

```c++
&& (
                    typeid(decltype(nullptr)).hash_code() 
                    != typeid(containedType).hash_code()
                )
```

src/opentimelineio/typeRegistry.cpp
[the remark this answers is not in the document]
ssteinbach: The downgrading system doesn't have room for an error code but I could see adding that to the upgrade/downgrade system.  That would fit into the encoder but need some more retrofitting.  What about that?  Something like:

```c++
    std::function<ErrorStatus(AnyDictionary*)>
```

src/opentimelineio/typeRegistry.cpp
```diff
@@ -198,6 +247,28 @@ TypeRegistry::register_upgrade_function(
...
+    int                                 version_to_downgrade_from,
+    std::function<void(AnyDictionary*)> downgrade_function)
+{
+    std::lock_guard<std::mutex> lock(_registry_mutex);
+    if (auto r = _find_type_record(schema_name))
+    {
+        if (r->downgrade_functions.find(version_to_downgrade_from) ==
+            r->downgrade_functions.end())
+        {
+            r->downgrade_functions[version_to_downgrade_from] = (
```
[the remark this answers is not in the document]
ssteinbach: nice!  good catch.

src/py-opentimelineio/opentimelineio/core/__init__.py
ssteinbach: This type was linking up - I suspect because SerializableObject is in this module?  In any case, I can fully resolve it anyway, if it helps readability

JeanChristopheMorinPerso: Indeed! https://opentimelineio--1387.org.readthedocs.build/en/1387/api/python/opentimelineio.core.html#opentimelineio.core.serialize_json_to_file

Ignore this comment then.

examples/io_perf_test.cpp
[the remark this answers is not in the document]
rogernelson: I'm not 100% sure, but I think it might be because `end` and `begin` are integers so subtracting them gives std::chrono::duration<int>.  If you try `std::chrono::duration<float>( end - begin)` that might do the trick.

examples/io_perf_test.cpp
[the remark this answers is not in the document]
rogernelson: There's an `assertNotNull` in tests/utils.h.  I think that's likely what he is referring to.

src/opentimelineio/CORE_VERSION_MAP.cpp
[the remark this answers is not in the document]
rogernelson: If the string can be left off, I would definitely vote for that as well as it would simplify things.

For the version string to `uint64_t` conversion there are a lot of different techniques people use but one I think tends to work well is bit shifting.  Split around the '.' and shift each number.  Something like this for example:
```
uint64_t(a, b, c) = a << 32 | b << 16 | c
```
That would allow lexicographical comparison with a max version of ("65536"."65536"."65536"), and it can be converted back into a string as well by doing the reverse operation. 

There are other ways of doing it as well, so we could have a discussion about it if we want to take this approach.

src/opentimelineio/typeRegistry.cpp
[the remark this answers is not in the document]
rogernelson: I think is a pretty good idea, since the python bindings already have a mechanism for converting ErrorStatus to an exception.

src/opentimelineio/anyDictionary.h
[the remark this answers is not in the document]
rogernelson: Sorry, I meant to say `result` not `containedType` (nullptr is not a type in an of itself). You could block it in the condition by just `result != nullptr` if you want to avoid changing the return value.

src/opentimelineio/anyDictionary.h
```diff
@@ -122,6 +123,71 @@ class AnyDictionary : private std::map<std::string, any>
...
+                (d_it != this->end())
+                && (d_it->second.type() == typeid(containedType))
+        )
+        {
+            *result = any_cast<containedType>(d_it->second);
+            return true;
+        } 
+        else 
+        {
+            this->insert({key, *result});
```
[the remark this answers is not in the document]
ssteinbach: I think insert_or_assign is a C++17 feature, we're trying to stick to C++11 for now (because of VFX platform).

src/opentimelineio/typeRegistry.cpp
ssteinbach: We talked about this internally and are proposing that this becomes a follow up issue.  This would add further performance overhead and could probably use discussion, so I'll file an issue for it and leave it to a later date.

src/py-opentimelineio/opentimelineio-bindings/otio_bindings.cpp
```diff
@@ -51,8 +68,59 @@ static bool register_upgrade_function(std::string const& schema_name,
...
+    }
+
+    return true;
+}
+
+static bool 
+register_downgrade_function(
+        std::string const& schema_name,
+        int version_to_downgrade_from,
+        py::object const& downgrade_function_obj)
```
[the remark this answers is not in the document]
ssteinbach: I think I'm going to defer on this suggestion for right now - the pattern is working using the python objects directly and the right pybind11 incantation escapes me.  I would welcome a PR with the right magic though!  Good suggestion, sorry about punting on it for now.

Both the upgrade and downgrade functions would want to be converted as well.

docs/tutorials/versioning-schemas.md
ssteinbach: Good thought; lets look at this in a later pass.  Thanks!

src/py-opentimelineio/opentimelineio/adapters/otio_json.py
ssteinbach: I cleaned up the docs in core/__init__.py and from the python wrapped modules.  Doing a consistency pass on the adapter docs is probably a good idea but lets do it as a later thing.  Good catch though.  Thanks!

examples/io_perf_test.cpp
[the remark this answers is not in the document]
ssteinbach: Got this working, thanks.

examples/io_perf_test.cpp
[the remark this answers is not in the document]
ssteinbach: Ahh ok.  Lets tackle this when we expand the C++ unit testing.

src/opentimelineio/CORE_VERSION_MAP.cpp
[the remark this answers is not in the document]
ssteinbach: Lets do this as a follow up conversation around how to structure labels in the future.

src/opentimelineio/anyDictionary.h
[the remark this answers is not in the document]
ssteinbach: got it, added a check

src/opentimelineio/typeRegistry.cpp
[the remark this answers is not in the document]
ssteinbach: I'm going to file an issue for this and suggest this is done as a follow up overloaded optional thing.  So you can declare your up/downgrade functions as having an error state they can set.

src/py-opentimelineio/opentimelineio-bindings/otio_bindings.cpp
```diff
@@ -133,10 +254,35 @@ Return an instance of the schema from data in the data_dict.
...
+          "downgrade_function"_a);
+    m.def(
+            "release_to_schema_version_map",
+            [](){ return label_to_schema_version_map(CORE_VERSION_MAP);},
+            R"docstring(Fetch the compiled in CORE_VERSION_MAP.  
+
+The CORE_VERSION_MAP maps OTIO release versions to maps of schema name to schema version and is autogenerated by the OpenTimelineIO build and release system.  For example: `{"0.15.0": {"Clip": 2, ...}}`
+
+:returns: dictionary mapping core version label to schema_version_map
+:rtype: dict[str, dict[str, int]])docstring"
```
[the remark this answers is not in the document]
ssteinbach: I think in both of these cases I didn't see anything in the docs - but I might have been reading it incorrectly.

meshula (approved):

meshula (approved):

ssteinbach:
I added a performance scaling section to the pr text, just noting here in case folks following it are interested.

PR #1389, opened 2022-08-29
Load entrypoint plugins before builtin and contrib.

markreidvfx:
This pull request is meant to assist the moving contrib adapters to external repos.  
You can use `$OTIO_PLUGIN_MANIFEST_PATH` to override a contrib adapter, however this can be inflexible and difficult to test your adapter when you install it.
Being able to use a stock install of OTIO also means you don't have to manage pull requests for 2 repos at the same time when starting to move a contrib adapter.

This issue has been mentioned in https://github.com/AcademySoftwareFoundation/OpenTimelineIO/pull/1348

### DIFF
--- a/src/py-opentimelineio/opentimelineio/plugins/manifest.py
+++ b/src/py-opentimelineio/opentimelineio/plugins/manifest.py
@@ -202,10 +202,10 @@ def load_manifest():
 
     The order of loading (and precedence) is:
 
-       1. manifests specfied via the ``OTIO_PLUGIN_MANIFEST_PATH`` variable
-       2. builtin plugin manifest
-       3. contrib plugin manifest
-       4. ``setuptools.pkg_resources`` based plugin manifests
+       1. Manifests specified via the ``OTIO_PLUGIN_MANIFEST_PATH`` variable
+       2. Entrypoint based plugin manifests
+       3. Builtin plugin manifest
+       4. Contrib plugin manifest
     """
 
     result = Manifest()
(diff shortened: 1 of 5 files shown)


src/py-opentimelineio/opentimelineio/plugins/manifest.py
```diff
@@ -202,10 +202,10 @@ def load_manifest():
 
     The order of loading (and precedence) is:
 
-       1. manifests specfied via the ``OTIO_PLUGIN_MANIFEST_PATH`` variable
-       2. builtin plugin manifest
-       3. contrib plugin manifest
-       4. ``setuptools.pkg_resources`` based plugin manifests
+       1. manifests specified via the ``OTIO_PLUGIN_MANIFEST_PATH`` variable
+       2. ``setuptools.pkg_resources`` based plugin manifests
```
JeanChristopheMorinPerso, line 206: I know it was there before, but technically it's not "setuptools" based. It should say "Entrypoint based plugins".

markreidvfx: maybe this?
2. ``pkg_resources`` Entrypoint based plugin manifests

JeanChristopheMorinPerso: This is still inaccurate. `setuptools.pkg_resources` is just an API that can find and load entry points points. There is nothing setuptools specific when it comes to entry points. It just happens that they were introduced by setuptools long ago and is now a packaging specification (see https://packaging.python.org/en/latest/specifications/entry-points/).

Python 3 even has [a builtin API](https://docs.python.org/3/library/importlib.metadata.html) to support them, like mentioned in https://setuptools.pypa.io/en/latest/pkg_resources.html. `importlib.metadata` also has a backport package in PyPI, which we just happen to not be using.

markreidvfx: Okay how about I just change it to 
```
2. Entrypoint based plugin manifests
```

JeanChristopheMorinPerso: Seems good to me.

markreidvfx:
I'll take a look, I think I can use part of those existing tests and come up with something that isn't too hacky.

markreidvfx:
I added a unit test, please take a look when you have the time

jminor (approved):

jminor:
@ssteinbach does this conflict with the downgrader PR at all? Do you care if this lands before/after that one?

reinecke (approved):
Great! Thanks for adding the test!

ssteinbach:
> @ssteinbach does this conflict with the downgrader PR at all? Do you care if this lands before/after that one?

Not an issue.  The only file in common is in manifest.py and it isn't a conflicting change.  Go for it!

ssteinbach:
@jminor do you want to land this before or after the next release?  (IE do you want to land it at the beginning of the next cycle or at the end of this one)

markreidvfx:
It would be especially helpful for people working on extracting contrib adapters if it was in the next wheel release. 🙂 It would allow running unit test against a wheel, at least for me, instead of having to compile OTIO everytime i run CI. saves a lot time/cycles.

PR #1402, opened 2022-09-20
add .envrc to .gitignore

### LINKED ISSUE #1401: add .envrc to .gitignore to support using direnv when working with code locally
direnv (https://github.com/direnv/direnv) can be used for local developer workflow enhancements.  direnv uses .envrc file for configuration.  Because .envrc is usually specific to a developer workflow, it is generally not committed.

marksisson:
Signed-off-by: Mark Sisson <5761292+marksisson@users.noreply.github.com>

Fixes #1401

**Summarize your change.**

direnv (https://github.com/direnv/direnv) can be used for local developer workflow enhancements.  direnv uses .envrc file for configuration.  Because .envrc is usually specific to a developer workflow, it is generally not committed.

I think all tests provide coverage.  This change prevents .envrc file from being added to repository.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@ htmlcov
 xcuserdata/
 .venv/
 .cache
+.envrc
 
 # Pycharm metadata
 .idea/


marksisson:
Yes, but some projects do commit the .envrc file (especially in some internal projects where there are standardized project specific environments). It helps to ignore it in projects that do not use it. The global ignore breaks workflows where .envrc file is committed.

ssteinbach:
The global .gitignore solution covers that - it just ignores it for cases where it hasn't already been added.  You can still:

- add it via `git add -f` (it'll print a message about that)
- once you've added it, it'll appear as a normal un-ignored file

ssteinbach:
This direnv guide even advocates for this kind of an approach:
https://shivamarora.medium.com/a-guide-to-manage-your-environment-variables-in-a-better-way-using-direnv-2c1cd475c8e

marksisson:
I was not aware of the `git add -f`.  That solves both use cases nicely.  Thanks!

PR #1404, opened 2022-09-20
Fix for SerializableCollection::children_if

### LINKED ISSUE #1400: SerializableCollection::children_if doesn't recurse correctly
## C++ Repro

the assertion at the end of this program fails:

```c++
#include <iostream>

#include "opentimelineio/clip.h"
#include "opentimelineio/errorStatus.h"
#include "opentimelineio/serializableObject.h"
#include "opentimelineio/track.h"
#include "opentimelineio/timeline.h"
#include "opentimelineio/serializableCollection.h"

namespace otio = opentimelineio::OPENTIMELINEIO_VERSION;

int
main(void)
{
    opentimelineio::v1_0::ErrorStatus err;

    // make a clip
    otio::SerializableObject::Retainer<otio::Clip> cl = new otio::Clip();

    // put it in a track
    otio::SerializableObject::Retainer<otio::Track> tr = new otio::Track();
    tr->append_child(cl);
    
    // check track
    auto result = tr->children_if<otio::Clip>(&err);
    assert (result.size() == 1);
    result.clear();

    // Make a timeline and add the track
    otio::SerializableObject::Retainer<otio::Timeline> tl = new otio::Timeline();
    tl->tracks()->append_child(tr);

    result = tl->children_if<otio::Clip>(&err);
    assert (result.size() == 1);
    result.clear();

    // Make a serializableCollection and then add the timeline
    otio::SerializableObject::Retainer<otio::SerializableCollection> sc = new otio::SerializableCollection();
    result = sc->children_if<otio::Clip>(&err, {}, false);

    // fail
    assert (result.size() == 1);
    result.clear();

    std::cout << "success" << std::endl;
}
```

## Python Repro
```python
import opentimelineio as otio

cl = otio.schema.Clip()

tr = otio.schema.Track()
tr.append(cl)

assert len(tr.clip_if()) == 1

tl = otio.schema.Timeline()
tl.tracks.append(tr)

assert len(tl.clip_if()) == 1

sc = otio.schema.SerializableCollection()
sc.append(tl)

# fails
assert len(sc.clip_if()) == 1
```

## Parameters are not wrapped

Note that despite the comment/argument here:
https://github.com/AcademySoftwareFoundation/OpenTimelineIO/blob/8cdda64f5ef924855ce503822e1aeb125e6b011e/src/py-opentimelineio/opentimelineio/schema/track.py#L6

The `shallow_search` parameter is not passed through to the C++ function.

## Related
-  `shallow_search` parameter does not appear to be exposed to python
- there appear to be no python tests that call `clip_if` and the only test that calls `children_if` is the cmx_edl tests

## Items
- [ ] bug in `SerializableCollection::children_if` (should find the clip in the test above)
- [ ] exposing parameters of `clip_if` / `children_if`
- [ ] add python unit tests for `clip_if` and `children_if`

darbyjohnston:

Fixes #1400

Fixes an issue where SerializableCollection::children_if doesn't recurse into timelines. I'm marking this as a draft for now since it doesn't address all of the items in the issue (Python tests and missing arguments).

### DIFF
--- a/src/opentimelineio/serializableCollection.h
+++ b/src/opentimelineio/serializableCollection.h
@@ -99,8 +99,8 @@ SerializableCollection::children_if(
             out.push_back(valid_child);
         }
 
-        // if not a shallow_search, for children that are serialiable collections, compositions,
-        // or timelines, recurse into their children
+        // if not a shallow_search, for children that are serializable collections,
+        // compositions, or timelines, recurse into their children
         if (!shallow_search)
         {
             if (auto collection =
--- a/src/opentimelineio/serializableCollection.h
+++ b/src/opentimelineio/serializableCollection.h
@@ -5,6 +5,7 @@
 
 #include "opentimelineio/composition.h"
 #include "opentimelineio/serializableObjectWithMetadata.h"
+#include "opentimelineio/timeline.h"
 #include "opentimelineio/version.h"
 
 namespace opentimelineio { namespace OPENTIMELINEIO_VERSION {
@@ -98,8 +99,8 @@ SerializableCollection::children_if(
             out.push_back(valid_child);
         }
 
-        // if not a shallow_search, for children that are serialiable collections or compositions,
-        // recurse into their children
+        // if not a shallow_search, for children that are serialiable collections, compositions,
+        // or timelines, recurse into their children
         if (!shallow_search)
         {
             if (auto collection =
@@ -120,6 +121,19 @@ SerializableCollection::children_if(
             {
                 const auto valid_children =
                     composition->children_if<T>(error_status, search_range);
+                if (is_error(error_status))
+                {
+                    return out;
+                }
+                for (const auto& valid_child: valid_children)
+                {
+                    out.push_back(valid_child);
+                }
+            }
+            else if (auto timeline = dynamic_cast<Timeline*>(child.value))
+            {
+                const auto valid_children =
+                    timeline->children_if<T>(error_status, search_range);
                 if (is_error(error_status))
                 {
                     return out;
--- a/src/py-opentimelineio/opentimelineio/schema/serializable_collection.py
+++ b/src/py-opentimelineio/opentimelineio/schema/serializable_collection.py
@@ -32,7 +32,7 @@ def __repr__(self):
 
 @add_method(_otio.SerializableCollection)
 def each_child(self, search_range=None, descended_from_type=_otio.Composable,
-    shallow_search=False):
+               shallow_search=False):
     """ Generator that returns each child contained in the serializable
     collection in the order in which it is found.
 
--- a/src/py-opentimelineio/opentimelineio/schema/serializable_collection.py
+++ b/src/py-opentimelineio/opentimelineio/schema/serializable_collection.py
@@ -31,7 +31,7 @@ def __repr__(self):
 
 
 @add_method(_otio.SerializableCollection)
-def each_child(self, search_range=None, descended_from_type=_otio.Composable):
+def each_child(self, search_range=None, descended_from_type=_otio.Composable, shallow_search=False):
     """ Generator that returns each child contained in the serializable
     collection in the order in which it is found.
 
@@ -42,13 +42,15 @@ def each_child(self, search_range=None, descended_from_type=_otio.Composable):
                                    with the search range will be yielded.
     :param type descended_from_type: if specified, only children who are a descendent
                                      of the descended_from_type will be yielded.
+    :param bool shallow_search: if True, will only search children of self and not recurse
+                                into children of children.
     """
-    for child in self.children_if(descended_from_type, search_range):
+    for child in self.children_if(descended_from_type, search_range, shallow_search):
         yield child
 
 
 @add_method(_otio.SerializableCollection)
-def each_clip(self, search_range=None):
+def each_clip(self, search_range=None, shallow_search=False):
     """ Generator that returns each clip contained in the serializable
     collection in the order in which it is found.
 
@@ -57,6 +59,8 @@ def each_clip(self, search_range=None):
 
     :param TimeRange search_range: if specified, only children whose range overlaps
                                    with the search range will be yielded.
+    :param bool shallow_search: if True, will only search children of self and not recurse
+                                into children of children.
     """
-    for child in self.clip_if(search_range):
+    for child in self.clip_if(search_range, shallow_search):
         yield child
--- a/src/py-opentimelineio/opentimelineio/schema/serializable_collection.py
+++ b/src/py-opentimelineio/opentimelineio/schema/serializable_collection.py
@@ -31,7 +31,8 @@ def __repr__(self):
 
 
 @add_method(_otio.SerializableCollection)
-def each_child(self, search_range=None, descended_from_type=_otio.Composable, shallow_search=False):
+def each_child(self, search_range=None, descended_from_type=_otio.Composable,
+    shallow_search=False):
     """ Generator that returns each child contained in the serializable
     collection in the order in which it is found.
 
@@ -42,8 +43,8 @@ def each_child(self, search_range=None, descended_from_type=_otio.Composable, sh
                                    with the search range will be yielded.
     :param type descended_from_type: if specified, only children who are a descendent
                                      of the descended_from_type will be yielded.
-    :param bool shallow_search: if True, will only search children of self and not recurse
-                                into children of children.
+    :param bool shallow_search: if True, will only search children of self and not
+                                recurse into children of children.
     """
     for child in self.children_if(descended_from_type, search_range, shallow_search):
         yield child
@@ -59,8 +60,8 @@ def each_clip(self, search_range=None, shallow_search=False):
 
     :param TimeRange search_range: if specified, only children whose range overlaps
                                    with the search range will be yielded.
-    :param bool shallow_search: if True, will only search children of self and not recurse
-                                into children of children.
+    :param bool shallow_search: if True, will only search children of self and not
+                                recurse into children of children.
     """
     for child in self.clip_if(search_range, shallow_search):
         yield child
--- a/tests/test_serializable_collection.py
+++ b/tests/test_serializable_collection.py
@@ -71,6 +71,54 @@ class SerializableColTests(unittest.TestCase, otio_test_utils.OTIOAssertions):
             ")"
         )
 
+    def test_children_if(self):
+        cl = otio.schema.Clip()
+        tr = otio.schema.Track()
+        tr.append(cl)
+        tl = otio.schema.Timeline()
+        tl.tracks.append(tr)
+        sc = otio.schema.SerializableCollection()
+        sc.append(tl)
+        result = sc.children_if(otio.schema.Clip)
+        self.assertEqual(len(result), 1)
+        self.assertTrue(result[0], cl)
+        
+    def test_children_if_search_range(self):
+        range = otio.opentime.TimeRange(
+            otio.opentime.RationalTime(0.0, 24.0),
+            otio.opentime.RationalTime(24.0, 24.0))
+        cl0 = otio.schema.Clip()
+        cl0.source_range = range
+        cl1 = otio.schema.Clip()
+        cl1.source_range = range
+        cl2 = otio.schema.Clip()
+        cl2.source_range = range
+        tr = otio.schema.Track()
+        tr.append(cl0)
+        tr.append(cl1)
+        tr.append(cl2)
+        tl = otio.schema.Timeline()
+        tl.tracks.append(tr)
+        sc = otio.schema.SerializableCollection()
+        sc.append(tl)
+        result = sc.children_if(otio.schema.Clip, range)
+        self.assertEqual(len(result), 1)
+        self.assertTrue(result[0], cl0)
+    
+    def test_children_if_shallow_search(self):
+        cl = otio.schema.Clip()
+        tr = otio.schema.Track()
+        tr.append(cl)
+        tl = otio.schema.Timeline()
+        tl.tracks.append(tr)
+        sc = otio.schema.SerializableCollection()
+        sc.append(tl)
+        result = sc.children_if(otio.schema.Clip, shallow_search=True)
+        self.assertEqual(len(result), 0)
+        result = sc.children_if(otio.schema.Clip, shallow_search=False)
+        self.assertEqual(len(result), 1)
+        self.assertTrue(result[0], cl)
+
 
 if __name__ == '__main__':
     unittest.main()
--- a/tests/test_serializable_collection.py
+++ b/tests/test_serializable_collection.py
@@ -82,7 +82,7 @@ class SerializableColTests(unittest.TestCase, otio_test_utils.OTIOAssertions):
         result = sc.children_if(otio.schema.Clip)
         self.assertEqual(len(result), 1)
         self.assertTrue(result[0], cl)
-        
+
     def test_children_if_search_range(self):
         range = otio.opentime.TimeRange(
             otio.opentime.RationalTime(0.0, 24.0),
@@ -104,7 +104,7 @@ class SerializableColTests(unittest.TestCase, otio_test_utils.OTIOAssertions):
         result = sc.children_if(otio.schema.Clip, range)
         self.assertEqual(len(result), 1)
         self.assertTrue(result[0], cl0)
-    
+
     def test_children_if_shallow_search(self):
         cl = otio.schema.Clip()
         tr = otio.schema.Track()
--- a/tests/test_serializable_collection.py
+++ b/tests/test_serializable_collection.py
@@ -81,7 +81,7 @@ class SerializableColTests(unittest.TestCase, otio_test_utils.OTIOAssertions):
         sc.append(tl)
         result = sc.children_if(otio.schema.Clip)
         self.assertEqual(len(result), 1)
-        self.assertTrue(result[0], cl)
+        self.assertEqual(result[0], cl)
 
     def test_children_if_search_range(self):
         range = otio.opentime.TimeRange(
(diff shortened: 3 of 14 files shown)


src/opentimelineio/serializableCollection.h
```diff
@@ -129,6 +130,19 @@ SerializableCollection::children_if(
                     out.push_back(valid_child);
                 }
             }
+            else if (auto timeline = dynamic_cast<Timeline*>(child.value))
+            {
+                const auto valid_children =
+                    timeline->children_if<T>(error_status, search_range);
```
ssteinbach, line 136: Thanks for putting this together, @darbyjohnston!  This is pretty straightforward.  @meshula I think we'll need to return to this argument because it isn't specified what space the `search_range` is in.  In the case of a serializable collection, which doesn't have a temporal relationship between its children, I could see dropping this argument entirely.  For timelines, if its in the global space, then the offset should be subtracted first to get into the space of the `.tracks` stack.

I also sort of wonder if Serializable collection needs children_if/clip_if?  or needs one that does this kind of recursion?  I wonder if it makes more sense to drop it or make it always shallow search so that you can then operate on the results of this in a consistent way.  Any thoughts or opinions?  @jminor @reinecke 

We could leave it as is - or as it is after this PR or I could see an argument for dropping it.

src/opentimelineio/serializableCollection.h
```diff
@@ -5,6 +5,7 @@
 
 #include "opentimelineio/composition.h"
 #include "opentimelineio/serializableObjectWithMetadata.h"
+#include "opentimelineio/timeline.h"
```
ssteinbach, line 8: This might be a good time to break out test_serializableCollection.cpp

src/opentimelineio/serializableCollection.h
meshula: 1) I assume that in the case of a timeline, the range would be in the timeline's space, so wouldn't need a subtraction?

2) I do wonder how deep down the rabbit hole we should go in addressing the initial report. I'm inclined  to suggest solving for Timeline is sufficient for 0.15.0, but more fundamental thinking about *_if seems prime territory for exploration in 0.16.0.

src/opentimelineio/serializableCollection.h
```diff
@@ -129,6 +130,19 @@ SerializableCollection::children_if(
                     out.push_back(valid_child);
                 }
             }
+            else if (auto timeline = dynamic_cast<Timeline*>(child.value))
```
rogernelson, line 133: I'm not sure how far down the c++ template rabbit hole its worth going for this, but you could use SFINAE to check if `child.value` has the method `children_if<T>`, rather than casting to every type known to have it.  That would ensure the list would always remain exhaustive.

darbyjohnston: Interesting, I didn't know about the SFINAE, I'll check it out.

src/opentimelineio/serializableCollection.h
darbyjohnston: Good idea, I also broke out test_track.cpp and test_timeline.cpp since they also have ```children_if()```.

ssteinbach: awesome, thanks!

src/opentimelineio/serializableCollection.h
rogernelson: It stands for Substitution Failure Is Not An Error, which is essentially a technique in template programming to filter in the things you are interested in, and just ignore everything else that doesn't match.

Here is a relevant example from stack overflow:
https://stackoverflow.com/questions/87372/check-if-a-class-has-a-member-function-of-a-given-signature

rogernelson: I think adding something like this (maybe best suited for a utility file header):
```
template<typename, typename T>
struct has_children_if {
    static_assert(
        std::integral_constant<T, false>::value,
        "Second template parameter needs to be of function type.");
};

template<typename C, typename Ret, typename... Args>
struct has_children_if<C, Ret(Args...)> {
private:
    template<typename T>
    static constexpr auto check(T*) -> typename
        std::is_same<
            decltype( std::declval<T>().children_if( std::declval<Args>()... ) ),
            Ret 
        >::type;

    template<typename>
    static constexpr std::false_type check(...);

    typedef decltype(check<C>(0)) type;

public:
    static constexpr bool value = type::value;
};

bool has_children_if_method(SerializableObject const & so) {
   return has_children_if<so, std::vector<Retainer<T>>(ErrorStatus*, optional<TimeRange>, bool)>::value
}
```

And then replacing if/else chain with this single if
```
if( has_children_if_method(*(child.value)) ) {
```
_might_ do the trick (warning: I did not test any of that, there are probably errors).

Of course, putting a virtual has_children_if method in the SerializableObject class that returns false and overloading it the appropriate base classes to return true would also do the trick.  That is much easier to understand code, but at the same time I find once you go down that path, over time more and more of those start getting introduced and whenever you add a new class you have a bunch has_* methods that you may not know about and can effect the behaviour or your class is subtle, hard-to-debug ways.

darbyjohnston: I need to learn more about modern C++ templates before trying out that code, that's quite formidable looking. :) @ssteinbach brought up the question of whether SerializableCollection should support recursive searching, maybe we should come to a consensus on that before getting to far with the implementation.

darbyjohnston:
I think I have addressed all the items from the issue as well as adding more consistent use of the "shallow_search" parameter to the various functions. @rogernelson has some interesting thoughts on a more efficient implementation of the "children_if" function, but maybe we need to decide whether the recursive searching functionality is actually appropriate for SerializableCollection.

ssteinbach:
I think we should probably not do a major functionality switch on `SerializableCollection` right now, since we're in a release process.  That is a conversation we can have for beta 16.  Lets review this PR with its current functionality, and circle back to that other discussion later.

Thanks @darbyjohnston!  I'll take a look at the code.

src/opentimelineio/serializableCollection.h
ssteinbach: As I noted in the other comment, I think sticking with this for now is ok.  We can have a bigger discussion about adopting something more duck-type-y like the SFINAE that Roger proposes or even removing this function in the next beta.

Thanks for raising that @rogernelson!

src/py-opentimelineio/opentimelineio/schema/serializable_collection.py
```diff
@@ -31,7 +31,8 @@ def __repr__(self):
 
 
 @add_method(_otio.SerializableCollection)
-def each_child(self, search_range=None, descended_from_type=_otio.Composable):
+def each_child(self, search_range=None, descended_from_type=_otio.Composable,
```
ssteinbach, line 34: We marked these as deprecated in 0.14.  We should add an issue to remove these in 0.16.

tests/test_timeline.cpp
```diff
@@ -0,0 +1,83 @@
...
+        otio::SerializableObject::Retainer<otio::Track> tr =
+            new otio::Track();
+        tr->append_child(cl);
+        otio::SerializableObject::Retainer<otio::Timeline> tl =
+            new otio::Timeline();
+        tl->tracks()->append_child(tr);
+        opentimelineio::v1_0::ErrorStatus err;
+        auto result = tl->children_if<otio::Clip>(&err);
+        assertEqual(result.size(), 1);
+        assertTrue(result[0] = cl);
```
ssteinbach, line 34:
```suggestion
        assertEqual(result[0], cl)
```

Should this be `==` instead of `=`?  Or `assertEqual`?

tests/test_timeline.cpp
```diff
@@ -0,0 +1,83 @@
...
+        tr->append_child(cl0);
+        tr->append_child(cl1);
+        tr->append_child(cl2);
+        otio::SerializableObject::Retainer<otio::Timeline> tl =
+            new otio::Timeline();
+        tl->tracks()->append_child(tr);
+        opentimelineio::v1_0::ErrorStatus err;
+        auto result = tl->children_if<otio::Clip>(&err, range);
+        assertEqual(result.size(), 1);
+        assertTrue(result[0] = cl0);
```
ssteinbach, line 60:
```suggestion
        assertEqual(result[0], cl0);
```

Same question here about `assertTrue` vs `assertEqual`

tests/test_timeline.cpp
```diff
@@ -0,0 +1,83 @@
...
+        tr->append_child(cl);
+        otio::SerializableObject::Retainer<otio::Timeline> tl =
+            new otio::Timeline();
+        tl->tracks()->append_child(tr);
+        opentimelineio::v1_0::ErrorStatus err;
+        auto result = tl->children_if<otio::Clip>(&err, nullopt, true);
+        assertEqual(result.size(), 0);
+        result = tl->children_if<otio::Clip>(&err, nullopt, false);
+        assertEqual(result.size(), 1);
+        assertTrue(result[0] = cl);
```
ssteinbach, line 78: Same question here:

```suggestion
        assertEqual(result[0], cl);
```

tests/test_timeline.py
```diff
@@ -557,6 +557,48 @@ def test_tracks_set_null_tracks(self):
...
 
+    def test_children_if(self):
+        cl = otio.schema.Clip()
+        tr = otio.schema.Track()
+        tr.append(cl)
+        tl = otio.schema.Timeline()
+        tl.tracks.append(tr)
+        result = tl.children_if(otio.schema.Clip)
+        self.assertEqual(len(result), 1)
+        self.assertTrue(result[0], cl)
```
ssteinbach, line 568:
```suggestion
        self.assertEqual(result[0], cl)
```

tests/test_timeline.py
```diff
@@ -557,6 +557,48 @@ def test_tracks_set_null_tracks(self):
...
+        cl2.source_range = range
+        tr = otio.schema.Track()
+        tr.append(cl0)
+        tr.append(cl1)
+        tr.append(cl2)
+        tl = otio.schema.Timeline()
+        tl.tracks.append(tr)
+        result = tl.children_if(otio.schema.Clip, range)
+        self.assertEqual(len(result), 1)
+        self.assertTrue(result[0], cl0)
```
ssteinbach, line 588:
```suggestion
        self.assertEqual(result[0], cl0)
```

tests/test_timeline.py
```diff
@@ -557,6 +557,48 @@ def test_tracks_set_null_tracks(self):
...
+        cl = otio.schema.Clip()
+        tr = otio.schema.Track()
+        tr.append(cl)
+        tl = otio.schema.Timeline()
+        tl.tracks.append(tr)
+        result = tl.children_if(otio.schema.Clip, shallow_search=True)
+        self.assertEqual(len(result), 0)
+        result = tl.children_if(otio.schema.Clip, shallow_search=False)
+        self.assertEqual(len(result), 1)
+        self.assertTrue(result[0], cl)
```
ssteinbach, line 600:
```suggestion
        self.assertEqual(result[0], cl)
```

tests/test_track.cpp
```diff
@@ -0,0 +1,113 @@
...
+        using namespace otio;
+        otio::SerializableObject::Retainer<otio::Clip> cl =
+            new otio::Clip();
+        otio::SerializableObject::Retainer<otio::Track> tr =
+            new otio::Track();
+        tr->append_child(cl);
+        opentimelineio::v1_0::ErrorStatus err;
+        auto result = tr->children_if<otio::Clip>(&err);
+        assertEqual(result.size(), 1);
+        assertTrue(result[0] = cl);
```
ssteinbach, line 31:
```suggestion
        assertEqual(result[0], cl);
```

tests/test_track.cpp
```diff
@@ -0,0 +1,113 @@
...
+            new otio::Track();
+        tr->append_child(cl0);
+        tr->append_child(cl1);
+        tr->append_child(cl2);
+        opentimelineio::v1_0::ErrorStatus err;
+        auto result = tr->children_if<otio::Clip>(
+            &err,
+            TimeRange(RationalTime(0.0, 24.0), RationalTime(24.0, 24.0)));
+        assertEqual(result.size(), 1);
+        assertTrue(result[0] = cl0);
```
ssteinbach, line 56:
```suggestion
        assertEqual(result[0], cl0);
```

tests/test_track.cpp
```diff
@@ -0,0 +1,113 @@
...
+        result = tr->children_if<otio::Clip>(
+            &err,
+            TimeRange(RationalTime(24.0, 24.0), RationalTime(24.0, 24.0)));
+        assertEqual(result.size(), 1);
+        assertTrue(result[0] = cl1);
```
ssteinbach, line 61:
```suggestion
        assertEqual(result[0], cl1);
```

tests/test_track.cpp
```diff
@@ -0,0 +1,113 @@
...
+        result = tr->children_if<otio::Clip>(
+            &err,
+            TimeRange(RationalTime(48.0, 24.0), RationalTime(24.0, 24.0)));
+        assertEqual(result.size(), 1);
+        assertTrue(result[0] = cl1);
```
ssteinbach, line 66:
```suggestion
        assertEqual(result[0], cl1);
```

tests/test_track.py
```diff
@@ -0,0 +1,58 @@
...
+        tr = otio.schema.Track()
+        tr.append(cl0)
+        tr.append(st)
+        result = tr.children_if(otio.schema.Clip, shallow_search=True)
+        self.assertEqual(len(result), 1)
+        self.assertTrue(result[0], cl0)
+        result = tr.children_if(otio.schema.Clip, shallow_search=False)
+        self.assertEqual(len(result), 2)
+        self.assertTrue(result[0], cl0)
+        self.assertTrue(result[1], cl1)
```
ssteinbach, line 54:
```suggestion
        self.assertEqual(result[0], cl0)
        self.assertEqual(result[1], cl1)
```

ssteinbach (commented):
Looks good. Saw (maybe) a copy paste typo in the unit tests?  I probably missed some but I might just be misunderstanding the use of `assertTrue`, too.  Let me know what you think!

src/py-opentimelineio/opentimelineio/schema/serializable_collection.py
darbyjohnston: Thanks, good catch on the asserts! It's a bit embarrassing creating a bug and then copying and pasting it half a dozen times. :)

darbyjohnston: New issue to remove the deprecated functions: https://github.com/AcademySoftwareFoundation/OpenTimelineIO/issues/1409

tests/test_serializable_collection.py
```diff
@@ -71,6 +71,54 @@ def test_repr(self):
...
+        tr.append(cl0)
+        tr.append(cl1)
+        tr.append(cl2)
+        tl = otio.schema.Timeline()
+        tl.tracks.append(tr)
+        sc = otio.schema.SerializableCollection()
+        sc.append(tl)
+        result = sc.children_if(otio.schema.Clip, range)
+        self.assertEqual(len(result), 1)
+        self.assertTrue(result[0], cl0)
```
ssteinbach, line 106: Is this also meant to be `assertEqual`?
```suggestion
        self.assertEqual(result[0], cl0)
```

tests/test_serializable_collection.py
```diff
@@ -71,6 +71,54 @@ def test_repr(self):
...
+        tr.append(cl)
+        tl = otio.schema.Timeline()
+        tl.tracks.append(tr)
+        sc = otio.schema.SerializableCollection()
+        sc.append(tl)
+        result = sc.children_if(otio.schema.Clip, shallow_search=True)
+        self.assertEqual(len(result), 0)
+        result = sc.children_if(otio.schema.Clip, shallow_search=False)
+        self.assertEqual(len(result), 1)
+        self.assertTrue(result[0], cl)
```
ssteinbach, line 120:
```suggestion
        self.assertEqual(result[0], cl)
```

tests/test_serializable_collection.py
```diff
@@ -71,6 +71,54 @@ def test_repr(self):
...
+        cl = otio.schema.Clip()
+        tr = otio.schema.Track()
+        tr.append(cl)
+        tl = otio.schema.Timeline()
+        tl.tracks.append(tr)
+        sc = otio.schema.SerializableCollection()
+        sc.append(tl)
+        result = sc.children_if(otio.schema.Clip)
+        self.assertEqual(len(result), 1)
+        self.assertTrue(result[0], cl)
```
ssteinbach, line 84:
```suggestion
        self.assertEqual(result[0], cl)
```

tests/test_track.py
```diff
@@ -0,0 +1,58 @@
...
+
+class TrackTests(unittest.TestCase, otio_test_utils.OTIOAssertions):
+
+    def test_children_if(self):
+        cl = otio.schema.Clip()
+        tr = otio.schema.Track()
+        tr.append(cl)
+        result = tr.children_if(otio.schema.Clip)
+        self.assertEqual(len(result), 1)
+        self.assertTrue(result[0], cl)
```
ssteinbach, line 20:
```suggestion
        self.assertEqual(result[0], cl)
```

tests/test_track.py
```diff
@@ -0,0 +1,58 @@
...
+        cl1.source_range = range
+        cl2 = otio.schema.Clip()
+        cl2.source_range = range
+        tr = otio.schema.Track()
+        tr.append(cl0)
+        tr.append(cl1)
+        tr.append(cl2)
+        result = tr.children_if(otio.schema.Clip, range)
+        self.assertEqual(len(result), 1)
+        self.assertTrue(result[0], cl0)
```
ssteinbach, line 38:
```suggestion
        self.assertEqual(result[0], cl0)
```

ssteinbach (changes requested):
A few more `assertTrue` still in there

tests/test_serializable_collection.py
darbyjohnston: Oops, sorry, I forgot the Python tests. Should be fixed now.

ssteinbach (approved):
Alright, this looks great.  Thanks @darbyjohnston

PR #1412, opened 2022-09-23
Remove Travis traces

ssteinbach:
There were a few more places in the CMake/Manifest.in that were referencing travis, this should remove them.

**for landing in beta 16**

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,7 +3,7 @@
 cmake_minimum_required(VERSION 3.18.2)
 
 #------------------------------------------------------------------------------
-# Project Meta data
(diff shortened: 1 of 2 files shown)


meshula (approved):
I believe we do the `-latest` to `-22` change every time we do a release, then pop it back to `-latest` after the release? Is that how this works?

Assuming that's the case, and ubuntu 22.04 is right, lgtm.

meshula (approved):
thanks!

JeanChristopheMorinPerso (approved):

jminor (approved):

meshula (approved):

PR #1415, opened 2022-09-23
Added OTIO_CORE 0.14.0 version_manifest.

jminor:
In order to ensure that users can downgrade to 0.14.0 easily, this PR adds a version_manifest entry for "0.14.0" which specifies the schema versions prior to the new Clip.2 multiple media reference feature https://github.com/AcademySoftwareFoundation/OpenTimelineIO/pull/1241 

In addition, this PR updates the documentation to clearly show how to use the downgrade system at runtime (either via environment variable, or via otioconvert).


ssteinbach (approved):

ssteinbach (approved):

src/opentimelineio/CORE_VERSION_MAP.cpp
```diff
@@ -48,6 +48,39 @@ const label_to_schema_version_map CORE_VERSION_MAP {
...
+            { "MediaLinker", 1 },
+            { "MediaReference", 1 },
+            { "MissingReference", 1 },
+            { "PluginManifest", 1 },
+            { "SchemaDef", 1 },
+            { "SerializableCollection", 1 },
+            { "SerializableObject", 1 },
+            { "SerializableObjectWithMetadata", 1 },
+            { "Stack", 1 },
+            { "Test", 1 },
```
ssteinbach:
```suggestion
```

Test schema shouldn't be in there

ssteinbach (commented):
Once you run the make target for updating the file, it should be good to go.  A stray `Test` schema also slipped into the list

src/opentimelineio/CORE_VERSION_MAP.last.cpp
```diff
@@ -15,6 +15,39 @@
...
+            { "MediaLinker", 1 },
+            { "MediaReference", 1 },
+            { "MissingReference", 1 },
+            { "PluginManifest", 1 },
+            { "SchemaDef", 1 },
+            { "SerializableCollection", 1 },
+            { "SerializableObject", 1 },
+            { "SerializableObjectWithMetadata", 1 },
+            { "Stack", 1 },
+            { "Test", 1 },
```
ssteinbach, line 43:
```suggestion
```

src/opentimelineio/CORE_VERSION_MAP.cpp
```diff
@@ -15,6 +15,39 @@
...
```
ssteinbach, line 43:
```suggestion
```

ssteinbach (changes requested):

ssteinbach (approved):
LGTM

ssteinbach (approved):

src/opentimelineio/CORE_VERSION_MAP.cpp
```diff
@@ -15,6 +15,38 @@
 namespace opentimelineio { namespace OPENTIMELINEIO_VERSION {
 
 const label_to_schema_version_map CORE_VERSION_MAP {
+    { "0.14.0",
```
JeanChristopheMorinPerso, line 18: Should we also have 0.14.1?

jminor: Good question, @JeanChristopheMorinPerso. It would be identical to 0.14.0 right? Should we include every release from now on, or just add one when a schema changes?

JeanChristopheMorinPerso: I'm under the impression we don't need 0.14.1 since it only contains bug fixes. But I referred to ask just in case. Having all versions might make it easier for users to just say "write an OTIO file using that version" without knowing which version contained schema changes. But at the same time it's not like it's really difficult to get this map and figure out pretty quickly which version is required.

ssteinbach: Design notes:

- for this release, I think leaving it like this is fine, since we're backporting it anyway and don't have a strategy
- I think as a project we should have a discussion.  There are two strategies that I can think of off the top of my head:
  - we base everything off the library release version:
    - every release gets an entry
    - +: its straightforward, the only thing anyone needs to know is what version of the library you have
    - -: (potential) we'll end up with a lot of entries that are the same
    - -: (potential) we conflate software release with schema changes (less obvious when things are _actually_ incompatible schema-wise
  - we create a separate "schema version":
    - +: Clear when things actually change
    - +: might be more straightforward for nontechnical end users building pipelines
    - -: adds another version number to track for folks, need to add tools for getting it and possibly recording it, etc.

PR #1418, opened 2022-09-24
Remove python2.7 support

ssteinbach:
* Removes Python2.7 support from the library
* Removes shim code that was bridging differences between python3 and python2
* Follows from VFX Platform

TODO:
- [ ] Xges adapter


JeanChristopheMorinPerso:
I haven't r viewed it and it's still a draft, but I'm wondering if we could switch to https://docs.python.org/3/library/importlib.metadata.html in https://github.com/AcademySoftwareFoundation/OpenTimelineIO/blob/main/src/py-opentimelineio/opentimelineio/plugins/manifest.py#L255, like mentioned in https://github.com/AcademySoftwareFoundation/OpenTimelineIO/pull/1389#discussion_r957843801.

Also, should we run https://github.com/asottile/pyupgrade?

Issue #1395, opened 2022-09-15
Drop Python2.7 support


ssteinbach:
As noted by #1094, VFX platform 2019 is the last VFX platform that includes python 2.7.  This issue is to prepare for when we drop python 2.7 support.

- [ ] update check in setup.py: 
  - [ ] https://github.com/AcademySoftwareFoundation/OpenTimelineIO/blob/78941ddf12a16e6b3495290e7c920ce674b1f342/setup.py#L186
  - [ ] https://github.com/AcademySoftwareFoundation/OpenTimelineIO/blob/78941ddf12a16e6b3495290e7c920ce674b1f342/setup.py#L295
  - [ ] https://github.com/AcademySoftwareFoundation/OpenTimelineIO/blob/78941ddf12a16e6b3495290e7c920ce674b1f342/setup.py#L336
- [ ] update README.md
- [ ] update files that reference 2.7, especially places marked `# python2` or `# python 2.7`  (this is a rough list from grep)
```
tests/test_builtin_adapters.py
tests/test_console.py
tests/test_timeline.py
tests/test_cmx_3600_adapter.py
tests/test_otioz.py
tests/test_adapter_plugin.py
tests/baselines/plugin_module/otio_mockplugin/__init__.py
tests/test_hooks_plugins.py
tests/test_examples.py
contrib/opentimelineio_contrib/adapters/tests/test_hls_playlist_adapter.py
src/py-opentimelineio/opentimelineio/adapters/file_bundle_utils.py
src/py-opentimelineio/opentimelineio/adapters/otiod.py
src/py-opentimelineio/opentimelineio/adapters/fcp_xml.py
src/py-opentimelineio/opentimelineio/url_utils.py
```
- [ ] update `.github/workflows/python-package.yml` - [python-package.yml](https://github.com/AcademySoftwareFoundation/OpenTimelineIO/blob/main/.github/workflows/python-package.yml)
- [ ] update pybind11 to 2.10+
- [ ] try [pyupgrade](https://github.com/asottile/pyupgrade) to move to newer syntax where available

JeanChristopheMorinPerso:
Do we also want to update the code base to use newer syntax? If so then we can use https://github.com/asottile/pyupgrade to automatically upgrade code to use newer syntax. It also handles changing imports, and a bunch of different things. When running the tool, you basically tell it to target a specific Python version and it will take care of the rest.

JeanChristopheMorinPerso:
[...]

PR #1428, opened 2022-09-28
Remove Plugin schema `execution_scope` parameter

### LINKED ISSUE #728: remove execution_scope field
- This is confusing to new people
- it was for a feature we decided that we ultimately didn't need
- if we want to add it in later, we can do that

TODO:
- [ ] remove it from adapters
- [ ] remove it from hooks
- [ ] remove it from media linker
- [ ] remove it from the python_plugin system
- [ ] remove it from the docs

ssteinbach:
When the plugin system was originally designed for the adapter system, the idea was that adapters could be implemented in other languages by sub shelling out.  It was simpler to implement that directly via the `extern_*` pattern seen in the RV and Maya adapters, and now the library is emphasizing native support.  This feature was never implemented, so this PR removes the parameter from the library.

This PR follows #1395 

Closes #728

### DIFF
--- a/contrib/opentimelineio_contrib/adapters/contrib_adapters.plugin_manifest.json
+++ b/contrib/opentimelineio_contrib/adapters/contrib_adapters.plugin_manifest.json
@@ -4,63 +4,54 @@
         {
             "OTIO_SCHEMA": "Adapter.1",
             "name": "fcpx_xml",
-            "execution_scope": "in process",
             "filepath": "fcpx_xml.py",
             "suffixes": ["fcpxml"]
         },
         {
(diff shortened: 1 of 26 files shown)


meshula (approved):
This looks good to me. Are there knock-on affects to adapters outside of this repository? i.e. is it possible there are external callers to MediaLinker that might need to modify source with one less parameter?

ssteinbach:
I've tested against our proprietary plugin set and it looks good, trying to rustle up an additional tester.

ssteinbach:
Tested externally now, looks good there too (thanks @apetrynet and @reinecke).

PR #1433, opened 2022-09-29
initial commit of the _core_utils tests

ssteinbach:
* adds unit tests for testing `_core_utils` implementation details

Co-authored by: @JeanChristopheMorinPerso

PR #1436, opened 2022-09-30
Improve test coverage

JeanChristopheMorinPerso:
This PR adds test coverage for the AnyVector and AnyDictionary python bindings and the `opentimelineio.core._core_utils` module.

It supersedes https://github.com/AcademySoftwareFoundation/OpenTimelineIO/pull/1433.

I also change the builds a little bit so that the coverage is measured in debug mode to improve the accuracy of the C++ code coverage when running the python tests and I also added `--rc lcov_branch_coverage=1` to the `lcov` commands.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -11,7 +11,7 @@ env:
 
 on:
   push:
-    branches: [ main, 'improve_test_coverage' ]
+    branches: [ main ]
   pull_request:
     branches: [ main ]
 
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -126,6 +126,7 @@ jobs:
     - name: Install coverage dependency
       if: matrix.python-version == env.GH_COV_PY && matrix.os == env.GH_COV_OS && github.actor != env.GH_DEPENDABOT
       run: |
+        echo 'OTIO_CXX_DEBUG_BUILD=1' >> $GITHUB_ENV
         sudo apt-get install lcov
     - name: Install python build dependencies
       run: |
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -11,7 +11,7 @@ env:
 
 on:
   push:
-    branches: [ main ]
+    branches: [ main, 'improve_test_coverage' ]
   pull_request:
     branches: [ main ]
 
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 .PHONY: coverage test test_first_fail clean autopep8 lint doc-html \
-	python-version wheel manifest lcov
+	python-version wheel manifest lcov lcov-html lcov-reset
 
 # Special definition to handle Make from stripping newlines
 define newline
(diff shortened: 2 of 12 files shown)


src/py-opentimelineio/opentimelineio-bindings/otio_anyVector.h
```diff
@@ -46,7 +46,7 @@ struct AnyVectorProxy : public AnyVector::MutationStamp {
         AnyVector& v = fetch_any_vector();
         index = adjusted_vector_index(index, v);
         if (index < 0 || index >= int(v.size())) {
-            throw py::index_error();
+            throw py::index_error("list index out of range");
```
JeanChristopheMorinPerso, line 49: These should match the builtin `list` type exception messages.

.github/workflows/python-package.yml
```diff
@@ -126,6 +126,7 @@ jobs:
     - name: Install coverage dependency
       if: matrix.python-version == env.GH_COV_PY && matrix.os == env.GH_COV_OS && github.actor != env.GH_DEPENDABOT
       run: |
+        echo 'OTIO_CXX_DEBUG_BUILD=1' >> $GITHUB_ENV
```
ssteinbach, line 129: This would run a debug build only for the coverage cell in the matrix?

Makefile
```diff
@@ -99,17 +99,26 @@ ifndef OTIO_CXX_BUILD_TMP_DIR
...
 	lcov --list ${OTIO_CXX_BUILD_TMP_DIR}/coverage.filtered.info
 
+
+lcov-html: lcov
+	@echo -e "$(ccgreen)Generating C++ HTML coverage report...$(ccend)"
+	genhtml --quiet --branch-coverage --output-directory lcov_html_report ${OTIO_CXX_BUILD_TMP_DIR}/coverage.filtered.info
+
+lcov-reset:
+	@echo "$(tput setaf -Txterm 2)Resetting C++ coverage...$(tput sgr0)"
+	lcov -b . --directory ${OTIO_CXX_BUILD_TMP_DIR} --zerocounters
```
ssteinbach, line 120: For testing locally?  Good call

setup.py
```diff
@@ -41,6 +42,10 @@ def _debugInstance(x):
         print("{}:     {}".format(a, getattr(x, a)))
 
 
+def joinArgs(args):
```
ssteinbach, line 45: We're using snake case in python for functions:
```suggestion
def join_args(args):
```

setup.py
```diff
@@ -158,7 +163,10 @@ def cmake_preflight_check(self):
 
     def cmake_generate(self):
         self.announce('running cmake generation', level=2)
+
         cmake_args = ['cmake', SOURCE_DIR] + self.generate_cmake_arguments()
+        self.announce(joinArgs(cmake_args), level=2)
```
ssteinbach, line 168:
```suggestion
        self.announce(join_args(cmake_args), level=2)
```

src/py-opentimelineio/opentimelineio/core/_core_utils.py
```diff
@@ -337,7 +337,7 @@ def insert(self, index, item):
                 func = getattr(klass, name)
                 if (
                         isinstance(func, types.FunctionType)
-                        and name not in klass.__abstractmethods__
+                        and name not in klass.__abstractmethods__ # noqa and name != "__iter__"
```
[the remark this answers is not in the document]
ssteinbach: I like the hasattr line, want to make that change here?

ssteinbach (changes requested):
Small stuff, looks pretty good. Thanks @JeanChristopheMorinPerso!

.github/workflows/python-package.yml
JeanChristopheMorinPerso: Exactly. I was under the impression it wouldn't impact much. Is that a good assumption?

Makefile
JeanChristopheMorinPerso: Yes, it was pretty annoying to have to remember how to reset the coverage and so I figured a make target could help others.

.github/workflows/python-package.yml
meshula: I read that env var as SGI THUB, and puzzled for a second on what a THUB was

I guess OTIO_CXX_DEBUG_BUILD won't affect the choice of Microsoft CRT? A MSVC build uses the debug CRT, which isn't compatible with non-debug Python distros, which is why I ask

JeanChristopheMorinPerso: It will affect the choice of CRT, but the built DLLs will only be us d by the tests. The wheels will continue to be built in Release mode.

src/py-opentimelineio/opentimelineio/core/_core_utils.py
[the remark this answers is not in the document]
JeanChristopheMorinPerso: I'd like to, but we'd have to make sure we don't break anything by doing that. Unused code would start to be used and I have no idea if this code (the C++ code) was ever used in prod. At least my tests are covering that area.

.github/workflows/python-package.yml
meshula: ok, sounds good

src/py-opentimelineio/opentimelineio/core/_core_utils.py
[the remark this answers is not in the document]
JeanChristopheMorinPerso: I'll defer this after https://github.com/AcademySoftwareFoundation/OpenTimelineIO/pull/1432. Without the `_next__` methods, we get `TypeError: iter() returned non-iterator of type 'opentimelineio._otio.CompositionIterator'` exceptions everywhere.

ssteinbach (approved):

PR #1437, opened 2022-09-30
Remove deprecated each_child and each_clip functions

### LINKED ISSUE #1409: Deprecate each_child() and each_clip() Python functions
The deprecated functions are in composition.py, serializable_collection.py, stack.py, timeline.py, and track.py.

Calls to these functions can be replaced with the new "children_if()" and "clip_if()" functions.

darbyjohnston:

Fixes #1409

This PR removes the deprecated Python functions "each_child" and "each_clip", replacing them with "children_if" and "clip_if".

Python is not my expertise so here are a couple of notes about the changes I wasn't sure about:
* stack.py and track.py effectively became empty files after removing the functions. The linter didn't like Python files with just imports so I removed them.
* In a previous PR I had removed the function "each_clip" in clip.py which caused some issues and was later reverted. This time I renamed it to "clip_if", is that necessary?
* As far as I understand the "each_child" and "each_clip" functions were iterators? The new functions return lists, so the couple places that were calling "next" I replaced with a list index.


reinecke:
From TSC meeting: Should we consider adding something like `all_children` and `all_clips` for readability in the cases where there are no predicates?

JeanChristopheMorinPerso:
Maybe `get_children` and `get_clips` to make it obvious that they are not properties?

darbyjohnston:
I added the "all_children()" and "all_clips" functions, check out the changes and see if they look reasonable. I think in the meeting we had talked about them being Python functions, but I thought if they are useful why not add them to the C++ API as well? It might also be worth considering adding convenience functions for getting the first child or clip which can be useful.

I'm not sure why the "deps" directory is showing up in this PR, I think it got added when I merged the latest code from main.

darbyjohnston:
I think I fixed the "deps" commits but I'm not sure why they changed when I merged the main branch. Just for reference here are the steps I used:

Show the differences:
```
$ git diff main..remove_each_child -- src/deps
diff --git a/src/deps/optional-lite b/src/deps/optional-lite
...
-Subproject commit be720ebfd7add22abe60e49602735de9231105d0
+Subproject commit 7a66721b8da0affc30e43f9272a8635259562048
diff --git a/src/deps/pybind11 b/src/deps/pybind11
...
-Subproject commit 5bc0943ed96836f46489f53961f6c438d2935357
+Subproject commit 0176632e8cef72a4223f2df54d6dfb9e552ec71f
```
Change to the submodule directories and check out the previous commits:
```
$ cd src/deps/optional-lite
$ git checkout be720ebfd7add22abe60e49602735de9231105d0
$ cd ../pybind11
$ git checkout 5bc0943ed96836f46489f53961f6c438d2935357
```
Check in the changes:
```
cd ../../..
git add src/deps
git commit -m "Revert submodule" --signoff
```

meshula:
Does this PR need a rebase?

darbyjohnston:
I'm not sure, does it?

meshula:
Ok, no it doesn't need a rebase. When I checked last, the diff also showed all the CI modifications for Python 3.11.

meshula:
I re-ran all the checks just now.

meshula:
codecov failed straight up again. @JeanChristopheMorinPerso Do you know what the issue is?

JeanChristopheMorinPerso:
I unfortunately have no idea why it failed. Maybe just a glitch in the matrix... I'm guessing a retry might work.

darbyjohnston:
I think it may be an intermittent issue with CodeCov, I had a similar failure in one of my projects recently.

darbyjohnston:
@JeanChristopheMorinPerso You had suggested some alternate names:

> Maybe get_children and get_clips to make it obvious that they are not properties?

I'm not very familiar with Python, what is the issue with properties?

meshula (approved):

JeanChristopheMorinPerso:
> I'm not very familiar with Python, what is the issue with properties?

Oh I missed your message. Nothing wrong with properties. My comment was more about the fact that `all_children` sounds like it is a property while it is actually a method. Python being Python, it's quite easy to think that something is a property and not call it like a method/function and it will still work but won't do what we think it would do.

meshula:
I'd like to suggest we take the naming debate back to the issue.

https://github.com/AcademySoftwareFoundation/OpenTimelineIO/issues/1409

meshula (changes requested):
The form of the code is fine, let's agree on the naming before proceeding.

src/py-opentimelineio/opentimelineio/schema/__init__.py
```diff
@@ -44,15 +44,11 @@
...
     marker,
     serializable_collection,
-    stack,
     timeline,
-    track,
     transition,
     v2d,
 )
 
-track.TrackKind = TrackKind
```
jminor, line 54: Is this related to this change?

contrib/opentimelineio_contrib/adapters/tests/test_aaf_adapter.py
```diff
@@ -242,7 +242,7 @@ def test_aaf_read(self):
 
         self.assertEqual(len(timeline.audio_tracks()), 2)
 
-        clips = list(video_track.each_clip())
+        clips = list(video_track.find_clips())
```
jminor, line 245: I suspect the `list(...)` wrapper here is redundant now that `find_clips()` returns a list.

jminor (commented):
This looks great! I just had two questions/comments in-line.

contrib/opentimelineio_contrib/adapters/tests/test_aaf_adapter.py
darbyjohnston: Good catch, I also removed a couple of other redundant ```list()``` calls.

src/py-opentimelineio/opentimelineio/schema/__init__.py
jminor: Oh, I see. It was sort of odd before to have it in both places. This is cleaner.

jminor (approved):

jminor:
I think this is ready to go. @meshula can you take a last look? I see GitHub says you requested a change.

Thanks for persevering through our indecision with this @darbyjohnston - I think the `find_` naming really works great!

meshula (approved):
looks good to me!  thanks

PR #1442, opened 2022-10-04
Use glossary terms for environment variables references in the docs

JeanChristopheMorinPerso:
**Summarize your change.**

Simple PR that makes use of glossaries to define/document environment variables. This allows to reference environment variables from other places in the docs.

See https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-glossary.

The rendered docs can be seen at https://opentimelineio--1442.org.readthedocs.build/en/1442 and an example of how a reference looks like can be seen at https://opentimelineio--1442.org.readthedocs.build/en/1442/tutorials/write-an-adapter.html#custom-adapters.

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -21,6 +21,7 @@ help:
 	@echo "  htmlstrict to make standalone HTML files and fails if any warnings or errors are produced"
 	@echo "  dirhtml    to make HTML files named index.html in directories"
 	@echo "  singlehtml to make a single large HTML file"
+	@echo "  livehtml   to make standalone HTML files served by a web server that will automatically reload and rebuild when a file changes"
 	@echo "  pickle     to make pickle files"
 	@echo "  json       to make JSON files"
 	@echo "  htmlhelp   to make HTML files and a HTML help project"
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -73,6 +73,10 @@ singlehtml:
 	@echo
 	@echo "Build finished. The HTML page is in $(BUILDDIR)/singlehtml."
 
+.PHONY: livehtml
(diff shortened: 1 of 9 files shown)


docs/Makefile
```diff
@@ -73,6 +73,10 @@ singlehtml:
 	@echo
 	@echo "Build finished. The HTML page is in $(BUILDDIR)/singlehtml."
 
+.PHONY: livehtml
+livehtml:
```
JeanChristopheMorinPerso, line 77: Unrelated to this PR, but it's an handy target to have.

meshula (approved):
very nice!

docs/Makefile
ssteinbach: What is "livehtml" vs the `doc-html` target?

ssteinbach, line 77: How about we tie it together and add a comment?

```suggestion
# doc-html-live target starts a server mode for sphinx that watches the doc
# directory and will update if it sees changes. the doc-html target produces
# static html files.
.PHONY: doc-html-live
doc-html-live:
```

ssteinbach: got it! What if we change the name to `doc-html-live` (to tie it to the other target), and add a comment explaining the difference in the behavior?

JeanChristopheMorinPerso: `doc-html` is in the main Makefile. `livehtml` is just part of the docs Makefile.

JeanChristopheMorinPerso: I added the target into the help target. That way it will be listed when simply running `make` without any target.

ssteinbach (approved):

PR #1445, opened 2022-10-05
Improve type of function callbacks in register_upgrade_function and register_downgrade_function

### LINKED ISSUE #1425: Improve type hints for register_upgrade_function and register_downgrade_function
## Feature Request

`register_upgrade_function` and `register_downgrade_function` parameter type for the function is of type `py::object`. 

https://github.com/AcademySoftwareFoundation/OpenTimelineIO/blob/efc0aa676450d91df4fd93452cf451c39c81c2e6/src/py-opentimelineio/opentimelineio-bindings/otio_bindings.cpp#L115

This is not user friendly and forces us to document what argument the function takes, which duplicates this information since it's already specified in the code.

Pybind11 supports `std::function` (and see https://github.com/pybind/pybind11/blob/master/tests/test_callbacks.cpp#L23). From what I can see, when using `std::function`, any Python callable is supported.

For example, I made:
```c++
#include <pybind11/pybind11.h>
#include <pybind11/functional.h>

namespace py = pybind11;

PYBIND11_MODULE(pybind11_test, m) {
    m.def("add", [](const std::function<int(int)> &func, int &arg1) {
        return func(arg1);
    }, py::arg("func"), py::arg("number"));
}
```
Then in Python:
```python
import pybind11_test

def asd(arg):
    return arg + 2


class MyClass:
    def __call__(self, arg):
        return arg + 10


print(pybind11_test.add(asd, 1))

myclass = MyClass()
print(pybind11_test.add(myclass, 10))
```
which results in:
```
3
20
```

The advantage is that we get a really nice type hint when using `help` on the function and we also get that nice type hint in the Sphinx docs too:
```
add(...) method of builtins.PyCapsule instance
    add(func: Callable[[int], int], number: int) -> int
```

JeanChristopheMorinPerso:

Fixes #1425

**Summarize your change.**

Improve the advertised types of the callback functions in `register_upgrade_function` and `register_downgrade_function` functions.

See https://pybind11.readthedocs.io/en/stable/advanced/cast/functional.html for the Pybind11 documentation of `pybind11/functional.h`.

The help on both functions previously looked like:
```
register_downgrade_function(...) method of builtins.PyCapsule instance
    register_downgrade_function(schema_name: str, version_to_downgrade_from: int, downgrade_function: object) -> bool

register_upgrade_function(...) method of builtins.PyCapsule instance
    register_upgrade_function(schema_name: str, version_to_downgrade_from: int, upgrade_function: object) -> bool
```

while now it looks like:
```
register_downgrade_function(...) method of builtins.PyCapsule instance
    register_downgrade_function(schema_name: str, version_to_downgrade_from: int, downgrade_function: Callable[[opentimelineio._otio.AnyDictionary], None]) -> bool

register_upgrade_function(...) method of builtins.PyCapsule instance
    register_upgrade_function(schema_name: str, version_to_downgrade_from: int, upgrade_function: Callable[[opentimelineio._otio.AnyDictionary], None]) -> bool
```

### DIFF
--- a/src/py-opentimelineio/opentimelineio-bindings/otio_bindings.cpp
+++ b/src/py-opentimelineio/opentimelineio-bindings/otio_bindings.cpp
@@ -2,6 +2,7 @@
 // Copyright Contributors to the OpenTimelineIO project
 
 #include <pybind11/pybind11.h>
+#include <pybind11/functional.h>
 #include "otio_anyDictionary.h"
 #include "otio_anyVector.h"
 #include "otio_bindings.h"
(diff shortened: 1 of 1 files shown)


ssteinbach (approved):
Nice!

PR #1448, opened 2022-10-07
Improve types of childen, tracks, effects and markers parameters in function signatures

JeanChristopheMorinPerso:
**Summarize your change.**

Continuation of the work I've been doing to improve the parameter types in our bindings.

In this PR, I'm changing the types of the `children`, `tracks`, `effects` and `markers` parameters. A nice side effect is that we no more need the Python > C++ > Python > C++ > Python dance to convert the inputs from a `py::object` and `AnyVector`s.

Previously, there was an hand-coded bridging to `AnyVector` while now it's using facilities in pybind11 to accomplish the same thing. Part of the reason that's possible, is that `py::object` has been replaced in the interfaces with explicit types that pybind11 knows how to deal with.

 Here is an example of why it's important to expose the correct types in constructors (and functions/methods):

The previous behavior was something like this:
```python
>>> opentimelineio._otio.Track(name='asd', children=[1])
  File "/home/jcmorin/jcmenv/aswf/OpenTimelineIO/.venv/lib/python3.10/site-packages/opentimelineio/core/_core_utils.py", line 126, in _value_to_so_vector
    raise TypeError(
TypeError: Expected list/sequence of SerializableObjects; found element of type '<class 'int'>'
```
In this example there is two things to note. First, the error doesn't tell us which argument is of wrong type. Second, it doesn't tell us which specific types should be in the list/sequence. From the code, the list/sequence passed to the children parameter is not all subclasses of `SerializableObject`. It only accepts a list/sequence of type `Composable` and its sublcasses.

Compare this to the new behavior:
```python
>>> opentimelineio._otio.Track(name='asd', children=[1])
TypeError: __init__(): incompatible constructor arguments. The following argument types are supported:
    1. opentimelineio._otio.Track(name: object = '', children: Optional[List[opentimelineio._otio.Composable]] = None, source_range: Optional[opentimelineio._opentime.TimeRange] = None, kind: str = 'Video', metadata: object = None)

Invoked with: kwargs: name='asd', children=[1]
```

Now we can at least see all the arguments and the list of all accepted signatures. The signature is clear about that fact that it accepts a list of `Composable` and its subclasses. Pybind11 doesn't tell us which argument is wrong, but that could be contributed to Pybind11 if we feel like it would improve the experience.

---

There is one unanswered question though, at least for me, and it's what about the crashes mentioned in `otio_utils.cpp`? Andl also, what about `Every element has to be a SerializableObject::Retainer<>`? Was all this needed/casued by the fact that we were round tripping to Python?

### DIFF
--- a/src/py-opentimelineio/opentimelineio-bindings/otio_serializableObjects.cpp
+++ b/src/py-opentimelineio/opentimelineio-bindings/otio_serializableObjects.cpp
@@ -50,6 +50,15 @@ using SOWithMetadata = SerializableObjectWithMetadata;
 
 namespace {
 
+    template<typename T>
+    std::vector<T*> vector_or_default(optional<std::vector<T*>> item) {
+        if (item.has_value()) {
+            return item.value();
+        }
+
+        return std::vector<T*>();
+    }
+
     template<typename T, typename U>
     bool children_if(T* t, py::object descended_from_type, optional<TimeRange> const& search_range, bool shallow_search, std::vector<SerializableObject*>& l) {
         if (descended_from_type.is(py::type::handle_of<U>()))
@@ -248,10 +257,10 @@ a named collection.
 
 A :class:`~SerializableCollection` is useful for serializing multiple timelines, clips, or media references to a single file.
 )docstring")
-        .def(py::init([](std::string const& name, py::object children,
+        .def(py::init([](std::string const& name, optional<std::vector<SerializableObject*>> children,
                          py::object metadata) {
                           return new SerializableCollection(name,
-                                                py_to_vector<SerializableObject*>(children),
+                                                vector_or_default<SerializableObject>(children),
                                                 py_to_any_dictionary(metadata)); }),
              py::arg_v("name"_a = std::string()),
              "children"_a = py::none(),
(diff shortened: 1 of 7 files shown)


JeanChristopheMorinPerso:
Rebased to fix conflicts.

src/py-opentimelineio/opentimelineio-bindings/otio_serializableObjects.cpp
```diff
@@ -443,11 +452,11 @@ Base class for an :class:`~Item` that contains other :class:`~Item`\s.
 Should be subclassed (for example by :class:`.Track` and :class:`.Stack`), not used directly.
 )docstring")
         .def(py::init([](std::string name,
-                         py::object children,
+                         optional<std::vector<Composable*>> children,
```
meshula, line 455: Should the comment above, noting "Item", be revised to say "Composable"?

src/py-opentimelineio/opentimelineio-bindings/otio_utils.cpp
```diff
@@ -121,34 +120,6 @@ AnyDictionary py_to_any_dictionary(py::object const& o) {
...
-        _value_to_so_vector = core.attr("_value_to_so_vector");
-    }
-
-    std::vector<SerializableObject*> result;
-    if (o.is_none()) {
-        return result;
-    }
-
-    /*
-     * We're depending on _value_to_so_vector(), written in Python, to
```
meshula, line 136: I am not concerned about this comment. I believe it is referring to issues with bopping things back and forth to pyobject, and losing track of reference counts. As there are no unit tests to back up the note about crashing, (what was crashing? no way to tell), I think we have to move ahead assuming that the pybind11 explicit typing strategy is going to fundamentally change these assumptions. And if crashes do occur, we'll need unit tests to demonstrate a fix against.

src/py-opentimelineio/opentimelineio-bindings/otio_serializableObjects.cpp
JeanChristopheMorinPerso: Good idea. I will update that.

PR #1460, opened 2022-10-23
Add support for editable installs

### LINKED ISSUE #1286: Can't install locally with `pip install -e .`
Can't seem to install locally with `pip install -e .`

Main error is: `ModuleNotFound`

### Build Problem

```
git clone https://github.com/pedrolabonia/OpenTimelineIO.git --recursive           
Cloning into 'OpenTimelineIO'...
remote: Enumerating objects: 6444, done.
...
...
➜  otio cd OpenTimelineIO 
➜  OpenTimelineIO git:(main) python -m venv .venv
➜  OpenTimelineIO git:(main) source .venv/bin/activate
➜  OpenTimelineIO git:(main) (.venv) python -m pip install --upgrade pip
Collecting pip
  Using cached pip-22.0.4-py3-none-any.whl (2.1 MB)
Successfully installed pip-22.0.4
➜  OpenTimelineIO git:(main) (.venv) python -m pip install -e .
Obtaining file:///Users/pedrolabonia/dev/otio/OpenTimelineIO
  Installing build dependencies ... done
  Checking if build backend supports build_editable ... done
  Getting requirements to build wheel ... done
  Preparing metadata (pyproject.toml) ... done
Collecting pyaaf2~=1.4.0
  Using cached pyaaf2-1.4.0-py3-none-any.whl
Installing collected packages: pyaaf2, OpenTimelineIO
  Running setup.py develop for OpenTimelineIO
Successfully installed OpenTimelineIO-0.15.0.dev1 pyaaf2-1.4.0

➜  OpenTimelineIO git:(main) (.venv) python
Python 3.9.1 (default, Mar 21 2022, 17:23:00) 
[Clang 13.0.0 (clang-1300.0.29.30)] on darwin
>>> import opentimelineio
ModuleNotFoundError: No module named 'opentimelineio'

➜  OpenTimelineIO git:(main) (.venv) pip list
Package        Version     Editable project location
-------------- ----------- -------------------------------------------
OpenTimelineIO 0.15.0.dev1 /Users/pedrolabonia/dev/otio/OpenTimelineIO
pip            22.0.4
pyaaf2         1.4.0
setuptools     49.2.1
```

## To Reproduce

1. Mac OS 11.6
2. Python 3.9.1
4. OTIO https://github.com/PixarAnimationStudios/OpenTimelineIO/commit/ea64d92024ba4b521132f548df9ceee4585dbd46
5. Compiler information:
```
➜  OpenTimelineIO git:(main) (.venv) clang -v
Apple clang version 13.0.0 (clang-1300.0.29.30)
Target: arm64-apple-darwin20.6.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```

## Expected Behavior
Expected to be able to install locally with editable mode. I can install it normally via `pip install opentimelineio`. I'm trying to figure out what's going on. There's a `.egg-link` file in `site_packages` which seems to be correctly pointing to the root. Contents of the .egg-link:
```
/Users/pedrolabonia/dev/otio/OpenTimelineIO
.
```

JeanChristopheMorinPerso:

Fixes #1286.

**Summarize your change.**

`seetuptools` added back [editable installs](https://setuptools.pypa.io/en/latest/userguide/development_mode.html) not too long ago and it now supports extensions!

This PR adds support for installing OTIO in editable mode, using `pip install -e .`.

Note that due to the nature of OTIO, the binaries are not editable. There is not much we can do in regard to that since every time the C++ code changed, we need to re-compile the extension. But at least the Python files are editable.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -46,23 +46,25 @@ def join_args(args):
     return ' '.join(map(shlex.quote, args))
 
 
-class CMakeExtension(Extension):
-    def __init__(self, name):
-        Extension.__init__(self, name, sources=[])
-
-
 class OTIO_build_ext(setuptools.command.build_ext.build_ext):
     """
     def initialize_options(self):
         super(setuptools.command.build_ext.build_ext, self).initialize_options()
     """
 
+    built = False
+
     def run(self):
+        self.announce('running OTIO build_ext', level=2)
+        super().run()
+
+    def build_extension(self, _ext: Extension):
         # This works around the fact that we build _opentime and _otio
         # extensions as a one-shot cmake invocation. Usually we'd build each
         # separately using build_extension.
-        self.announce('running OTIO build_ext', level=2)
-        self.build()
+        if not self.built:
+            self.build()
+            self.built = True
 
     def build(self):
         self.build_temp_dir = (
@@ -256,20 +258,13 @@ class OTIO_build_py(setuptools.command.build_py.build_py):
     """Stamps PROJECT_METADATA into __init__ files."""
 
     def run(self):
-        setuptools.command.build_py.build_py.run(self)
+        super().run()
 
-        if not self.dry_run:
+        if not self.dry_run and not self.editable_mode:
+            # Only run when not in dry-mode (a dry run should not have any side effect)
+            # and in non-editable mode. We don't want to edit files when in editable
+            # mode because that could lead to modifications to the source files.
             _append_version_info_to_init_scripts(self.build_lib)
-
-
-def test_otio():
-    """Discovers and runs tests"""
-    try:
-        # Clear the environment of a preset media linker
-        del os.environ['OTIO_DEFAULT_MEDIA_LINKER']
-    except KeyError:
-        pass
-    return unittest.TestLoader().discover('tests')
 
 
 # copied from first paragraph of README.md
@@ -339,8 +334,8 @@ setup(
     ),
 
     ext_modules=[
-        CMakeExtension('_opentimelineio'),
-        CMakeExtension('_opentime'),
+        Extension('opentimelineio._otio', sources=[]),
+        Extension('opentimelineio._opentime', sources=[]),
     ],
 
     package_dir={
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -339,8 +333,8 @@ def test_otio():
     ),
 
     ext_modules=[
-        CMakeExtension('_opentimelineio'),
-        CMakeExtension('_opentime'),
+        Extension('opentimelineio._otio', sources=[]),
```
JeanChristopheMorinPerso, line 336: The full and correct module name is required here because setuptools needs to resolve the name to find the produced file and copy it into the source tree. (Because yes, editable install means that the install should point to the source tree, which means the `.so`s need to also be there alongside the python files).

setup.py
```diff
@@ -46,23 +45,25 @@ def join_args(args):
...
     """
     def initialize_options(self):
         super(setuptools.command.build_ext.build_ext, self).initialize_options()
     """
 
+    built = False
+
     def run(self):
+        self.announce('running OTIO build_ext', level=2)
+        super().run()
```
JeanChristopheMorinPerso, line 58: This is rather important. `build_ext.run` takes care of a couple of things, one of which is to copy the built files into the source tree (in `src/py-opentimelineio/opentimelineio`).

setup.py
```diff
@@ -256,22 +257,15 @@ class OTIO_build_py(setuptools.command.build_py.build_py):
     """Stamps PROJECT_METADATA into __init__ files."""
 
     def run(self):
-        setuptools.command.build_py.build_py.run(self)
+        super().run()
 
-        if not self.dry_run:
+        if not self.dry_run and not self.editable_mode:
```
JeanChristopheMorinPerso, line 262: `self.editable_mode` will be set to True when `-e` is given to the `pip install` command. See https://setuptools.pypa.io/en/latest/userguide/extension.html#setuptools.command.build.SubCommand.editable_mode.

meshula (changes requested):
This change, overall looks good to me, and seems like it will make the iteration process slightly less difficult to teach.

A general comment ~ you've left some general comments here in the PR of explanation for details that one can't know without special knowledge. All the comments, in my opinion, belong in the source as comments for reference by future maintainers.

meshula (approved):
Thanks :)

Issue #1464, opened 2022-10-26
Incorrect trimming of clips with LinearTimewarps


jminor:
When trimming a clip which contains a LinearTimewarp effect via `flatten_stack` or the underlying `track_trimmed_to_range` the resulting source_range on the trimmed clip is often incorrect.

To illustrate the problem see the screenshot below. When the tracks "Overlay B" and "Main B" are flattened to produce "Flattened B" the 2nd clip in the output (here named "Wrong Start") has the wrong start time. It should be "01:00:07:09" but instead it is "01:00:09:12".

![image](https://user-images.githubusercontent.com/169599/198154445-1c635924-e1f7-4419-b982-6643b9492775.png)

For comparison, when I slice the timewarped clip in Avid Media Composer, I get the correct result in this screenshot. Here we see that the clip "Sliced in Avid" has the correct start time "01:00:07:09", and so does the resulting "Correct Start" clip after flattening.

![image](https://user-images.githubusercontent.com/169599/198154625-264ac31b-e8a4-491e-b9dd-4586ab949d4a.png)

This bug is due to the fact that `track_trimmed_to_range` makes no attempt to honor the timewarp effect on the clip.

PR #1466, opened 2022-10-26
When flattening or trimming tracks, optionally honor linear timewarps.

jminor:
This PR is an attempt to fix https://github.com/AcademySoftwareFoundation/OpenTimelineIO/issues/1464

This change introduces an optional argument to both `flatten_tracks` and `track_trimmed_to_range` to indicate whether timewarps should be honored or not.

I'm not entirely happy with how this PR stands as-is, but I'm posting it to get feedback on it so it can be improved.


jminor:
Related also to this PR: https://github.com/AcademySoftwareFoundation/OpenTimelineIO/pull/842

PR #1469, opened 2022-10-27
Switch to using `importlib` from `pkg_resources`

### LINKED ISSUE #1426: switch to using importlib.metadata from pkgutil
We use pkgutil as one of our plugin mechanisms in python.  A more modern/robust solution is `importlib.metadata`, see the note here:

https://github.com/AcademySoftwareFoundation/OpenTimelineIO/pull/1423#issuecomment-1258732938

After #1423 lands.

mikemahony:
Fixes #1426

As requested in the issue, this replaces `pkg_resources` with `importlib` for plugin detection.

Quick and dirty benchmarks show an improvement in average runtime when accessing plugins:  
image

The `OTIO_DISABLE_PKG_RESOURCE_PLUGINS` env var and its corresponding test have been removed as they're no longer needed.

Less than ideal, `importlib.metadata` has been in flux since `3.7` so there's some code in here to work with old and new and to avoid a deprecation warning.

Existing tests in `test_plugin_detection` continue to provide coverage.

### DIFF
--- a/docs/tutorials/otio-env-variables.md
+++ b/docs/tutorials/otio-env-variables.md
@@ -18,10 +18,10 @@ OTIO_DEFAULT_MEDIA_LINKER
    The name of the default media linker to use after reading a file, if `""` then no
    media linker is automatically invoked.
 
-OTIO_DISABLE_PKG_RESOURCE_PLUGINS
-   By default, OTIO will use the `pkg_resource` entry_points mechanism to discover plugins
-   that have been installed into the current python environment. `pkg_resources`, however, can
-   be slow in certain cases, so for users who wish to disable this behavior, this variable can be set to 1.
+OTIO_DISABLE_ENTRYPOINTS_PLUGINS
+   By default, OTIO will use the `importlib.metadata` entry_points mechanism to discover plugins
+   that have been installed into the current python environment. For users who wish to disable this
+   behavior, this variable can be set to 1.
 
 OTIO_DEFAULT_TARGET_VERSION_FAMILY_LABEL
    If no downgrade arguments are passed to `write_to_file`/`write_to_string`, use the downgrade manifest
--- a/examples/sample_plugin/otio_counter/__init__.py
+++ b/examples/sample_plugin/otio_counter/__init__.py
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0
 # Copyright Contributors to the OpenTimelineIO project
-import pkg_resources
+
+import importlib.resources
 
 from opentimelineio.plugins import manifest
 
@@ -62,6 +63,7 @@ def plugin_manifest():
     # XXX: note, this doesn't get called.  For an example of this working,
     #      see the mockplugin unit test.
 
+    filepath = importlib.resources.files(__package__) / "plugin_manifest.json"
     return manifest.manifest_from_string(
-        pkg_resources.resource_string(__name__, 'plugin_manifest.json')
+        filepath.read_text()
     )
--- a/setup.py
+++ b/setup.py
@@ -361,8 +361,7 @@ setup(
 
     install_requires=[
         'pyaaf2>=1.4,<1.7',
-        # For python 3.7
-        'importlib_metadata==5.0.0',
+        'importlib_metadata==5.0.0; python_version < "3.8"',
     ],
     entry_points={
         'console_scripts': [
--- a/setup.py
+++ b/setup.py
@@ -361,7 +361,7 @@ setup(
 
     install_requires=[
         'pyaaf2>=1.4,<1.7',
-        'importlib_metadata==5.0.0; python_version < "3.8"',
+        'importlib_metadata>=1.4; python_version < "3.8"',
     ],
     entry_points={
         'console_scripts': [
--- a/setup.py
+++ b/setup.py
@@ -361,6 +361,8 @@ setup(
 
     install_requires=[
         'pyaaf2>=1.4,<1.7',
+        # For python 3.7
+        'importlib_metadata==5.0.0',
     ],
     entry_points={
         'console_scripts': [
--- a/src/py-opentimelineio/opentimelineio/plugins/manifest.py
+++ b/src/py-opentimelineio/opentimelineio/plugins/manifest.py
@@ -3,21 +3,17 @@
 
 """OTIO Python Plugin Manifest system: locates plugins to OTIO."""
 
+from importlib import resources
 import inspect
 import logging
 import os
+from pathlib import Path
 
-# In some circumstances pkg_resources has bad performance characteristics.
-# Using the envirionment variable: $OTIO_DISABLE_PKG_RESOURCE_PLUGINS disables
-# OpenTimelineIO's import and of use of the pkg_resources module.
-if os.environ.get("OTIO_DISABLE_PKG_RESOURCE_PLUGINS", False):
-    pkg_resources = None
-else:
-    try:
-        # on some python interpreters, pkg_resources is not available
-        import pkg_resources
-    except ImportError:
-        pkg_resources = None
+try:
+    from importlib import metadata
+except ImportError:
+    # For python 3.7
+    import importlib_metadata as metadata
 
 from .. import (
     core,
@@ -250,11 +246,14 @@ def load_manifest():
 
             result.extend(manifest_from_file(json_path))
 
-    # setuptools.pkg_resources based plugins
-    if pkg_resources:
-        for plugin in pkg_resources.iter_entry_points(
-                "opentimelineio.plugins"
-        ):
+    if not os.environ.get("OTIO_DISABLE_ENTRYPOINTS_PLUGINS"):
+        try:
+            entry_points = metadata.entry_points(group='opentimelineio.plugins')
+        except TypeError:
+            # For python <= 3.9
+            entry_points = metadata.entry_points().get('opentimelineio.plugins', [])
+
+        for plugin in entry_points:
             plugin_name = plugin.name
             try:
                 plugin_entry_point = plugin.load()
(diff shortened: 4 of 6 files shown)


docs/tutorials/otio-env-variables.md
```diff
@@ -18,11 +18,6 @@ OTIO_DEFAULT_MEDIA_LINKER
    The name of the default media linker to use after reading a file, if `""` then no
    media linker is automatically invoked.
 
-OTIO_DISABLE_PKG_RESOURCE_PLUGINS
-   By default, OTIO will use the `pkg_resource` entry_points mechanism to discover plugins
-   that have been installed into the current python environment. `pkg_resources`, however, can
-   be slow in certain cases, so for users who wish to disable this behavior, this variable can be set to 1.
-
```
ssteinbach, line 25: Let's keep the environment variable but update the name of it, just in case anyone in the future has problems with this feature (even via the new `importlib` path you're adding.

examples/sample_plugin/otio_counter/__init__.py
```diff
@@ -62,6 +63,6 @@ def plugin_manifest():
     # XXX: note, this doesn't get called.  For an example of this working,
     #      see the mockplugin unit test.
 
-    return manifest.manifest_from_string(
-        pkg_resources.resource_string(__name__, 'plugin_manifest.json')
-    )
+    with importlib.resources.path(__name__, "plugin_manifest.json") as p, \
+            open(p, 'rb') as f:
+        return manifest.manifest_from_string(f.read())
```
ssteinbach, line 68:
```suggestion
    with (
            importlib.resources.path(__name__, "plugin_manifest.json") as p,
            open(p, 'rb') as f
    ):
        return manifest.manifest_from_string(f.read())
```
Style preference for me, consider this optional - I'm not a fan of "\" for line continuation, would rather use ()

src/py-opentimelineio/opentimelineio/plugins/manifest.py
```diff
@@ -250,71 +245,59 @@ def load_manifest():
...
-            except Exception:
-                logging.exception(
-                    f"could not load plugin: {plugin_name}"
-                )
-                continue
+                    with resources.path(plugin.value, "plugin_manifest.json") as path, \
+                            open(path, 'rb') as manifest_stream:
+                        plugin_manifest = core.deserialize_json_from_string(
+                            manifest_stream.read().decode("utf-8")
+                        )
```
ssteinbach, line 289:
```suggestion
                    with (
                            resources.path(plugin.value, "plugin_manifest.json") as path, 
                            open(path, 'rb') as manifest_stream
                    ):
                        plugin_manifest = core.deserialize_json_from_string(
                            manifest_stream.read().decode("utf-8")
                        )
```
same nitpick

src/py-opentimelineio/opentimelineio/plugins/manifest.py
```diff
@@ -250,71 +245,59 @@ def load_manifest():
...
-    else:
-        # XXX: Should we print some kind of warning that pkg_resources isn't
-        #        available?
-        pass
+                        plugin_manifest._update_plugin_source(filepath)
+                        plugin_manifest.source_files.append(filepath)
+
+        except Exception:
+            logging.exception(
+                f"could not load plugin: {plugin_name}"
```
ssteinbach, line 296: Great idea! could we include the exception string in the log message too?

tests/baselines/plugin_module/otio_mockplugin/__init__.py
```diff
@@ -20,9 +20,6 @@
...
 def plugin_manifest():
-    return manifest.manifest_from_string(
-        pkg_resources.resource_string(
-            __name__,
-            'unusually_named_plugin_manifest.json'
-        )
-    )
+    with path(__name__, "unusually_named_plugin_manifest.json") as p, \
+            open(p, 'rb') as f:
+        return manifest.manifest_from_string(f.read())
```
ssteinbach, line 25:
```suggestion
    with (
            path(__name__, "unusually_named_plugin_manifest.json") as p,
            open(p, 'rb') as f
    ):
        return manifest.manifest_from_string(f.read())
```

ssteinbach (dismissed):
Looks good, some questions and suggestions!

src/py-opentimelineio/opentimelineio/plugins/manifest.py
```diff
@@ -3,21 +3,16 @@
...
-    try:
-        # on some python interpreters, pkg_resources is not available
-        import pkg_resources
-    except ImportError:
-        pkg_resources = None
+try:
+    from importlib import metadata
+except ImportError:
+    # For python 3.7
+    import importlib_metadata as metadata
```
JeanChristopheMorinPerso, line 15: This should be added to the project dependencies in the setup.py I think.

src/py-opentimelineio/opentimelineio/plugins/manifest.py
```diff
@@ -250,71 +245,59 @@ def load_manifest():
...
+                if not resources.is_resource(plugin.value, "plugin_manifest.json"):
+                    raise
 
-                    filepath = os.path.abspath(
-                        pkg_resources.resource_filename(
-                            plugin.module_name,
-                            'plugin_manifest.json'
-                        )
-                    )
+                with resources.path(plugin.value, "plugin_manifest.json") as filepath:
```
JeanChristopheMorinPerso, line 279: `path` is deprecated. [files](https://docs.python.org/3/library/importlib.resources.html#importlib.resources.files) should be used instead.

Also I believe this except block can simplified:

```
name = plugin_entry_point.__name__
filepath = importlib.resources.files(name) / "plugin_manifest.json"

if filepath.as_posix() in result.source_files:
    continue

try:
    plugin_manifest = core.deserialize_json_from_string(
        filepath.read_text()
    )
    plugin_manifest._update_plugin_source(filepath.as_posix())
    plugin_manifest.source_files.append(filepath.as_posix())
except IOError:
    raise
```

tests/test_plugin_detection.py
```diff
@@ -80,7 +78,8 @@ def test_detect_plugin(self):
         for linker in man.media_linkers:
```
JeanChristopheMorinPerso, line 78: We can't comment on non modified lines, so commenting here: You should modify the `setUp` method to remove the dependency on `pkg_resources`. You can see what I previously did in https://github.com/AcademySoftwareFoundation/OpenTimelineIO/compare/main...JeanChristopheMorinPerso:OpenTimelineIO:importlib?expand=1#diff-3d0c116209043c6578f8bb39f8d5fc32f70fad767498ef5a2e027af4396b5b71.

JeanChristopheMorinPerso (commented):
It's funny that you just created this PR. I was about to create one myself soon. I actually already had a [branch](https://github.com/AcademySoftwareFoundation/OpenTimelineIO/compare/main...JeanChristopheMorinPerso:OpenTimelineIO:importlib?expand=1#diff-3d0c116209043c6578f8bb39f8d5fc32f70fad767498ef5a2e027af4396b5b71) 😛

docs/tutorials/otio-env-variables.md
jminor: Yes, please. In some contexts where we know that plugins aren't in use, it will be good to have an option to turn this off & avoid the overhead of looking for them. A toggle is also hypothetically helpful in some security scenarios where you want to limit runtime discovery of untrusted code.

src/py-opentimelineio/opentimelineio/plugins/manifest.py
mikemahony: Thanks, @JeanChristopheMorinPerso - I love it.

Updating `otio_mockplugin` and `otio_counter` to match and adding support for `3.7`.

tests/test_plugin_detection.py
mikemahony: Good call.  Borrowing your code...

src/py-opentimelineio/opentimelineio/plugins/manifest.py
mikemahony: Thanks for catching that.

tests/test_plugin_detection.py
mikemahony: Thank you for providing your branch, @JeanChristopheMorinPerso - it was super helpful.

JeanChristopheMorinPerso: It's a pleasure!

src/py-opentimelineio/opentimelineio/plugins/manifest.py
```diff
@@ -3,21 +3,17 @@
...
-    try:
-        # on some python interpreters, pkg_resources is not available
-        import pkg_resources
-    except ImportError:
-        pkg_resources = None
+try:
+    from importlib import metadata
+except ImportError:
+    # For python 3.7
+    import importlib_metadata as metadata
```
JeanChristopheMorinPerso, line 16: We could theoretically always rely on `importlib_metadata` until `importlib.metadata` stabilizes and we drop 3.7.

jminor (approved):
This looks great. Are there any outstanding issues or comments to resolve before merging?

JeanChristopheMorinPerso:
Maybe https://github.com/AcademySoftwareFoundation/OpenTimelineIO/pull/1469#discussion_r1014530096.

src/py-opentimelineio/opentimelineio/plugins/manifest.py
mikemahony: I think, but can be convinced otherwise, that I'd like to leave it as is.  I like that it's only using the backport when necessary and will otherwise exercise the path moving forward (and gain any performance benefits therein).  Please unresolve if you feel strongly otherwise!

JeanChristopheMorinPerso: I find all the try/except cluttering. As for performance, `importlib_metadata` gets all the performance improvements before they land in CPython.

At least, if we move forward with the current state of the PR, we should limit the install of `importlib_metadata` or Pytohn 3.7 only, using a marker like `importlib_metadata; python_version < "3.8"` in `setup.py`.

jminor: That seems like a healthy pattern since it ensures that we have lots of proof that the newer module has been working well for a while before we drop the old one. If we rely on the old one, then the drop has more risk. However, @JeanChristopheMorinPerso you mentioned "until importlib.metadata stabilizes" - does that imply that it isn't yet stable? Is there a specific risk there?

jminor: Limiting the install to only when needed seems like a good change.

JeanChristopheMorinPerso: @jminor `importlib` is stable, so there is no risk. By stabilizing, I meant when they'll stop to change the interface. From what I understand, the interface is "stable" from 3.8 and onward. But I might be mistaken.

mikemahony: Good call on the marker, @JeanChristopheMorinPerso.  Added.

jminor:
This is all looking great. @JeanChristopheMorinPerso would you like to push the "Squash and merge" button?

setup.py
```diff
@@ -361,6 +361,7 @@ def run(self):
 
     install_requires=[
         'pyaaf2>=1.4,<1.7',
+        'importlib_metadata==5.0.0; python_version < "3.8"',
```
JeanChristopheMorinPerso, line 364: `importlib_metadata==5.0.0` could cause really painful to resolve conflicts to our users. What would be the lowest acceptable version? From https://pypi.org/project/importlib-metadata/#description I see hat 1.4 is equivalent to wwat's i CPython 3.8. SO I guess we could use `importlib_metadata>=1.4; python_version < "3.8"`?

JeanChristopheMorinPerso (changes requested):
One last comment before approving (sorry if I didn't spot that earlier).

setup.py
mikemahony: Another good call.  Updated.

JeanChristopheMorinPerso (approved):
Thanks a lot for creating that PR (and for following along with the multiple review passes)!

It's pretty exciting to get rid of the runtime dependency on setuptools! (You can't see my face right now, but I can guarantee you that I have a big smile on my face right now).

jminor:
Thanks @mikemahony and @JeanChristopheMorinPerso for bringing this one home. It was great to see y'all making this PR better at each step :)

Issue #1452, opened 2022-10-14
from_timecode doesn't treat rate properly


meshula:
@darbyjohnston reports

I was debugging an issue in my code and noticed something interesting about RationalTime::from_timecode(), however I'm not an expert in timecode so I don't know if this is expected behavior?
If I compare the result of RationalTime::from_timecode("01:00:00:00", 24.0) and RationalTime::from_timecode("01:00:00:00", 23.976), they both have the same value, 86400.0, but with the respective 24.0 and 23.976 rates. I guess I was expecting the second value to be 86,313.6 (23.976 * 3600.0)?
It looks like this is caused by RationalTime::from_timecode() running std::ceil() on the rate before calculating the result.

@meshula analyzes

I notice test_opentime.py, and the cpp, lack tests for equality between rates. For example, a test comparing two identical timecodes at different rates, converted to seconds (by dividing out the rate), should return the same result.
`nominal_fps` is just supposed to be used to check if the integer frame count in the time code string is greater than the number of frames that can possibly be valid indicated in the string.
The conversion to frames at the end of the function should be using `rate`, not `nominal_fps`.

Fix for this issue should include unit tests at the cpp level, optionally reported in the py bindings.

meshula:
When OpenTime implements time travel, I'm going back and explaining, emphatically, to the NTSC that their clever yet beknighted idea of wodging colorburst information between audio and the black and white in the transmission is going to screw up the computation of time, until the end of time. The last analog broadcast in North America was switched off in 2009, but we are still stuck with the analog hack for black and white/color compatibility!

darbyjohnston:
> The conversion to frames at the end of the function should be using rate, not nominal_fps.

Should a similar change also be made in RationalTime::to_timecode() so the values round trip? Also what about this test in test_opentime.py:
```
def test_timecode_23976_fps(self):
        # This should behave exactly like 24 fps
        timecode = "00:00:01:00"
        t = otio.opentime.RationalTime(value=24, rate=23.976)
        self.assertEqual(t, otio.opentime.from_timecode(timecode, 23.976))
```
Is this test irrelevant if 24 and 23.976 fps behave differently, or should the comment be removed and the test values updated to work?

meshula:
When is one second not a second? When NTSC has had a go. 

A more meaningful test to replace that one would be to generate rational times for "00:00:01:00" at 24 and 23.976, and then assert that that the difference between them is as expected due to framerate difference.

[...]

PR #1477, opened 2022-11-08
Improve handling of rates when converting to timecode strings

jminor:
This PR improves OTIO's `to_timecode`, `is_valid_timecode_rate`, and `nearest_valid_timecode_rate` functions so they handle rate values which are close, but not exactly matching the correct rates.

Specifically the previous strategy of keeping a table of commonly mis-typed non-integer rates (e.g. 29.97, 29.976, 23.98, etc.) is replaced by a heuristic which matches rates to the closest correct rate within some tolerance.

Rates checked by `is_valid_timecode_rate` and `nearest_valid_timecode_rate` now match each other, and adhere to [ST 12-1:2014 - SMPTE Standard - Time and Control Code](https://ieeexplore.ieee.org/document/7291029) (which is now freely available 👏🎉)

We also spot-checked some comparisons between Avid Media Composer and OTIO to make sure we got the drop frames in the right spot - that was already correct before this PR.

image

Note: there is 1 test failing which needs to be addressed, and my updates to the other tests should be scrutinized to make sure the changes are good.

For the failing test... something doesn't match up.
- Avid Media Composer's Calculator says: 1084319 frames = 05:01:11:29 @ 60 fps
- OpenTimelineIO says: 1084319 frames = 05:01:11:59 @ 60 fps
- https://robwomack.com/timecode-calculator/ says: 1084319 frames = 05:01:30:03 @ 60 fps

Test code: otio.opentime.from_frames(1084319, 60).to_timecode()


meshula:
To summarize, MC and OTIO  agree, but time-code-calculator significantly differs? Maybe we need to ping  the calculator's  author...

markreidvfx:
I would like to suggest `RationalTime::from_timecode`  explicitly take the nominal fps as a argument along with the frame rate

like what I proposed here: https://github.com/AcademySoftwareFoundation/OpenTimelineIO/issues/1452#issuecomment-1278486164

```cpp
RationalTime
RationalTime::from_timecode(std::string const& timecode, 
                            uint32_t           timecode_fps,
                            double             rate,
                            ErrorStatus*       error_status)
```

This would eliminate the need to do all the guessing from the playback frame rate.  
You might for example be using 24 frame timecode at playback rate of 30 fps, which isn't possible with the current method.

markreidvfx:
I tried that online [tc calculator](https://robwomack.com/timecode-calculator/). 
Its pretty finicky but you might have had it set to 59.97fps DF , I get
1084319 frames =  05:01:11:59 @ 60 fps.  
1084319 frames =  05:01:11:59 @ 59.97 fps.  **Non Drop Frame** (should be the same as 60)
and
1084319 frames =  05:01;30;03 @ 59.97 fps **Drop Frame**

That rate should really be **59.94**, not 59.97, but oh well.
I haven't figure out how to get 59.94 DF TC working on my avid yet to verify if that calculation is correct.

jminor:
MC and OTIO are off by 30 frames from each other.

markreidvfx:
The TC display in MC is might be showing 30fp TC where it duplicates every step.  My MC cannot display without using a burn in.
image

markreidvfx:
ok, in MC general settings there is option to change the TC display format from 30fps to 60 fps
image

image

Still haven't found a 60fps drop frame option.

jminor:
Here's the Media Composer features I used to check against:
- The "=" menu in the bottom left corner of the timeline window has "Show Track" which lets me turn on multiple timecode rulers along the timeline:
image

- The Timecode Window right-click menu lets me choose these:
image

- The Calculator Window lets you convert between "Total Frame Count" and various rates. This is where I got 1084319 frames = 05:01:11:29 @ 60 fps
image

jminor:
@markreidvfx can you explain more about using 24 frame timecode with 30 fps playback? That's not a scenario I've seen, so I'm confused about how it works.

With your proposed change, would these produce the same answer? 
```
a = otio.opentime.from_timecode(timecode_str, 24, 30)
b = otio.opentime.from_timecode(timecode_str, 24, 24).rescaled_to(30)
```

and the other direction:
```
a = otio.opentime.from_frames(frame_number, 30).rescaled_to(24).to_timecode()
b = otio.opentime.from_frames(frame_number, 30).to_timecode(rate=24)
```

markreidvfx:
No, those would produce different answers. maybe using 23.97 is a bit more clear.
It would be the equivalent of this

```
frame = otio.opentime.from_timecode(timecode_str, 24).value
a = otio.opentime.from_frames(frame, 23.97)
b = otio.opentime.from_timecode(timecode_str, 24).rescaled_to(23.97)

a.to_seconds() != b.to_seconds()
```
Timecode is converted to a frame number first.
Then that frame number is converted to a time at current frame rate.

Maybe adding another arg isn't the correct answer. The big thing I'm trying to get across is that timecode should be treated as frame number encoding and is not always reliable to be a timestamp.  It should really be called framecode in my opinion. :p

This is why I normally use a pattern like this
```
frames_to_seconds(timecode_to_frames(timecode_str, timecode_fps), frame_rate)
```
and the other direction:
```
frames_to_timecode(seconds_to_frames(seconds, frame_rate), timecode_fps, drop=False)
```

markreidvfx:
what is the the project format/edit rate?
image

some formats can't access certain timecodes.

jminor:
Aha! That explanation makes total sense now. The conversion to a frame count vs a timestamp really makes it clear - thanks @markreidvfx !

I'll check my MC project settings tomorrow.

I'd love to compare this to Resolve and/or Premiere as well. Are there other trusted timecode conversion tools or software libraries out there to compare to?

jminor:
This PR overlaps or conflicts with this older PR: https://github.com/AcademySoftwareFoundation/OpenTimelineIO/pull/1180

meshula:
@jminor I closed that one in favor of this one, since this one rectifies the treatments of documented SMPTE rates with the code. @splidje DCCs have various "tricks" for dealing with high frame rates, let's carry on the discussion of what to do with the high frame rates here.

markreidvfx:
I've verified those calculations in Resolve, it actually supports 59.94 TC with drop frame.

1084319 frames = 05:01:11:59 @ 59.94 fps without drop frame
1084319 frames = 05:01;30;03 @ 59.94 fps with **Drop Frame**

You can set the tc format when you create a timeline or in your project settings 
60fp_tc

You can toggle between frames and tc by right clicking on the timecode in the viewer
frame_to_tc

markreidvfx:
Resolve also additionally supports a 16, 18, 47.952 and 48 TC format
the 47.952 appears to just be 48 like 23.976 is just 24

image

meshula:
Thanks for doing the extra checks! So I think the theory is that MC is displaying time code at 30, even for a selected rate of 60 (as opposed to MC displaying a value that is 30 frames off of our calculation). A check for that would be to knock off 5 frames at 30, and check if MC then shows 24 instead of 29, or if it shows 19. 24 would demonstrate it is displaying a 30 rate, instead of 60; whereas 19 would show that the calculation is off by 30.

markreidvfx:
I'm quite sure @jminor had his avid set in 30 TC mode for 60. Its the default setting. The timecode ticks every 2 frames and the top dot of the last colon blinks on and off to tell you if your on even or odd frame.

https://user-images.githubusercontent.com/814966/203416726-a1cbd530-0c0e-4073-993f-79702d81e35e.mp4

This mode is somewhat described in the spec intro, page 5

> Progressive video systems with frame rates above 30 frames per second are described in this document, documenting what have become “de facto” implementations. Since the frame rate of these 50 and 60 frames-per-second progressive systems exceeds the frame count capacity of the time address, counting is done on frame pairs, which results in an edit resolution of two frames using traditional linear time code.

There are some further details in 12.1

markreidvfx:
I've also verified those timecode values in Unreal Engine 

60 fps NDF
image

59.94 DF

image

jminor:
Sorry for the delay in getting back to this. I should hopefully have time to revisit this in a couple of weeks.

In the meantime, I happened across this which is interesting to compare/contrast with OTIO's approach: https://github.com/orchetect/TimecodeKit

meshula:
Had a look at the link you provided, quite an interesting read ~ TimeCodeKit's major difference, I'd say, is that it introduces a strongly typed TimeCode object from which a string can be generated, and which functions as a mathematical object. It hasn't got a stronger ability than otio::RationalTime to represent a point in time, or a stronger ability to do math than otio::RationalTime, and would probably resist computations we are interested in, such as linear timewarps.

Another point of contrast, philosophically, is the view suggested by @markreidvfx earlier in the thread, that the timecode string should be considered, in a way, a "rendered" interpretation of a time, as opposed to a ground truth representation of time. 

TimeCodeKit does some interesting things that we don't support, at all. If you decrement time below zero, it will wrap around to 24 hours. We generate negative values. We could possibly do something similar in genarating time code strings. At the moment, negative time values are rejected. If the are negative and greater than -24 hours, we could wrap around.

I think it would be worth working through the math of TimeCodeKit to determine whether our conversion algorithm matches. Also, it would be worth working through the unit tests, and perhaps adding one to one corresponding unit tests to our own code.

meshula (approved):

reinecke (approved):

meshula:
I ran the tests in my own repo, the python unit tests are what failed, the tests themselves may need correction.

```
FAIL: test_invalid_rate_to_timecode_functions (test_opentime.TestTime.test_invalid_rate_to_timecode_functions)
  File "/home/runner/work/OpenTimelineIO/OpenTimelineIO/tests/test_opentime.py", line 400, in test_invalid_rate_to_timecode_functions
    with self.assertRaises(ValueError):
AssertionError: ValueError not raised

FAIL: test_timecode_infer_drop_frame (test_opentime.TestTime.test_timecode_infer_drop_frame)
---------------------------------------------------------------------
  File "/home/runner/work/OpenTimelineIO/OpenTimelineIO/tests/test_opentime.py", line 347, in test_timecode_infer_drop_frame
    self.assertEqual(t.to_timecode(rate, drop_frame=None), timecode)
AssertionError: '05:01:30;03' != '05:01:11;59'
- 05:01:30;03
+ 05:01:11;59
```

jminor:
Yes, @meshula that failing test is the one I mentioned in the initial post at the very top of this thread. My question is/was: what is the correct return value for `otio.opentime.from_frames(1084319, 60).to_timecode()` ?

meshula:
> `otio.opentime.from_frames(1084319, 60).to_timecode() # no dropframe rate requested`

referencing @markreidvfx's reported values from Resolve and Unreal

1084319 frames = 05:01:11:59 @ 59.94 fps without drop frame
1084319 frames = 05:01;30;03 @ 59.94 fps with Drop Frame

suggest that answer should be 05:01:11:59

jminor:
Following some illuminating conversation with @markreidvfx and Roger @ Avid, I have a better understanding of how the current OTIO timecode API is misleading. See for example this:
```
>>> otio.opentime.from_frames(1, 3).to_timecode(30)
'00:00:00:10'
>>> otio.opentime.from_frames(1, 30).to_timecode(30)
'00:00:00:01'
>>> otio.opentime.from_frames(1, 300).to_timecode(30)
'00:00:00:00'
>>> otio.opentime.from_frames(1, 3).to_timecode(30000/1001, True)
'00:00:00;09'
>>> otio.opentime.from_frames(1, 30).to_timecode(30000/1001, True)
'00:00:00;00'
>>> otio.opentime.from_frames(1, 300).to_timecode(30000/1001, True)
'00:00:00;00'
```

meshula:
Could you clarify why it's misleading? Maybe I've been looking at it too long and have got some assumptions that make me overlook something.

>>> otio.opentime.from_frames(1, 3).to_timecode(30)
'00:00:00:10'

1 frame 3fps, render to 30 fps, 10 frames.

>>> otio.opentime.from_frames(1, 30).to_timecode(30)
'00:00:00:01'

1 @ 30fps to 30 = 1

>>> otio.opentime.from_frames(1, 300).to_timecode(30)
'00:00:00:00'

1 @ 300fps to 30 is 0.1, round to zero

I don't have intuition about drop frame and haven't pulled out a calculator, but if I squint maybe these are fine?

>>> otio.opentime.from_frames(1, 3).to_timecode(30000/1001, True)
'00:00:00;09'
>>> otio.opentime.from_frames(1, 30).to_timecode(30000/1001, True)
'00:00:00;00'
>>> otio.opentime.from_frames(1, 300).to_timecode(30000/1001, True)
'00:00:00;00'

src/opentime/rationalTime.cpp
```diff
@@ -13,54 +13,33 @@ namespace opentime { namespace OPENTIME_VERSION {
...
-      60.0 }
+      60.0
+    }
 };
 
 bool
 RationalTime::is_valid_timecode_rate(double fps)
 {
-    auto b = valid_timecode_rates.begin(), e = valid_timecode_rates.end();
+    auto b = smpte_timecode_rates.begin(), e = smpte_timecode_rates.end();
```
rogernelson, line 42: Hopefully I am not opening a can of worms here... 

I see we were already using std::find (which uses operator==) before your changes, but might we want to put comparison with epsilon instead for floating point tolerance?

```
std::find_if(b, e,[fps](double v) { return std::fabs(fps - v) < std::numeric_limits<double>::epsilon(); });
```

The `nearest_valid_timecode_rate` function could be used as well for readability, although it's likely not as efficient:

```
return std::fabs(fps - nearest_valid_timecode_rate(fps)) < std::numeric_limits<double>::epsilon();
```

meshula: Agreed that exact comparisons are not as robust as tolerance based comparisons. `std::numeric_limits<double>::epsilon()` per se probably doesn't help because it is to small a value to catch inaccuracies from typical calculations (you lose a bit here and there after every math op, so imprecision is typically likely going to exceed epsilon by a factor on the order of the number of intervening calculations), although feel free to challenge my reasoning :) 

The highest framerate we test for is `60000.0 / 1001.0` so the choice I'd recommend for a tolerance is half that, e.g. `120000.0 / 1001.0` since that's double the max rate we check for.

rogernelson: I agree with you it makes sense to use a higher tolerance here, especially since we're supporting bindings to many different languages, and I've seen floating pointing numbers go awry just from crossing through language barriers.

So something like this should do it:
```
// Use half the highest frame rate as our max tolerance
static constexpr MAX_FPS_TOLERANCE = 120000.0 / 1001.0;
std::find_if(b, e,[fps](double v) { return std::fabs(fps - v) < MAX_FPS_TOLERANCE; });
```

meshula: I like that :)

reinecke:
Some relevant links from the TSC meeting:

- From @jhodges10 Some folks at Frame.io published [a paper ](https://ieeexplore.ieee.org/document/9780292)with SMPTE re: timecode handling that runs over 24 hours ([alternate link](https://atc21.smpte.org/home/session/728083/the-problem-with-timecode))
- From Ben Schofield: [Timecode in IMF](https://www.imfug.com/wp-content/uploads/2020/11/imfug-001-bp-timecode-in-imf.pdf) paper

src/opentime/rationalTime.cpp
rogernelson: I just noticed during the TSC meeting, I think we mean max tolerance = 1001.0 / 120000.0 right?

meshula:
Hi! One year later, are we stalemated? Today I am wondering if the work here fixes the very sad rounding bug reported yesterday.

How can we unstick this one? Have we bikeshedded it to death? Is this PR demonstrably an incremental improvement in what we have? If it is not perfect but incrementally better, and not incrementally worse, I want to merge it.

I am lost in the weeds of the many comment threads here. Please send help, SOS.

jminor:
Re-reading the thread, it looks like I need to do these things:
- Rebase this onto main
- Use "05:01:11:59" for the failing test.
- Rename functions to use "smpte" instead of the word "valid"

For the method rename, do we need to deprecate the old names & support both for a release?

apetrynet:
The use of the "valid_xx" methods have come up in quite a few discussions and issues.
I think we should supply both new and old method names with a deprecation warning in the old one for at least the upcoming release.

meshula:
@apetrynet I'm not following the comment "The use of the "valid_xx" functions have come up in quite a few discussions and issues." are you saying that supports merging, or that there some other work?

@jminor I concur with @apetrynet on the new/old+deprecation strategy,

Thanks for looking at this, folks :D

apetrynet:
@meshula 
What I meant was that in issues regarding timecodes and frame rates, people have referred `is_valid_timecode_rate` quite a few times. So they are in use and should be deprecated over time.

jminor:
Okay, I did these three things:
- Rebased on main
- Updated the test to use the correct timecode value (note I edited the comment above to correct the corrected value)
- Renamed both "valid_timecode" functions to "smpte_timecode" and deprecated the older functions.

Two questions:
- Is there a better way to tell pybind11 that those older functions are deprecated?
- Now there are failures in some adapter tests, where they should be using `nearest_smpte_timecode_rate`. These are probably legit issues that should be fixed. Should we fix them in this PR, or after the impending adapter split out?

jminor:
Hmm... that's odd. When I run the tests locally only the contrib ALE tests fail, but in the CI it fails earlier on `test_invalid_rate_to_timecode_functions`. I'll look further later in the week.

meshula:
@jminor Did https://github.com/AcademySoftwareFoundation/OpenTimelineIO/pull/1477/commits/28a8f5002f21bfb7783ef83a6e1fe98c0fdbd3d0 fix the issue

>When I run the tests locally only the contrib ALE tests fail, but in the CI it fails earlier on test_invalid_rate_to_timecode_functions. I'll look further later in the week.

? The commit seems to be after the comment :)

jminor:
I fixed the ALE rate problem in that commit, but this test still fails:
```
FAIL: test_invalid_rate_to_timecode_functions (test_opentime.TestTime)
  File "/home/runner/work/OpenTimelineIO/OpenTimelineIO/tests/test_opentime.py", line 424, in test_invalid_rate_to_timecode_functions
    with self.assertRaises(ValueError):
AssertionError: ValueError not raised
```

Let me try one other experiment...

jminor:
So I wondered if maybe a different exception was thrown? My experiment https://github.com/AcademySoftwareFoundation/OpenTimelineIO/pull/1477/commits/dc0051c0fc5f1c5aef5ace02d0f14fdf59c3eb45 shows that indeed no exception is thrown by `to_timecode` when given a bogus rate. However, when I run this locally, ValueError is thrown as desired.

I don't have an Ubuntu VM handy to try to repro this. I am using Python version (3.11) on my Mac. I'll see if I can tell the CI to run the Mac version of this to see if the failure is OS dependent.

If that doesn't help, maybe I can add some print statements in the C++ side of this function? I think the unit test system eats that output?

Any other debugging suggestions here are welcome.

meshula:
On the C++ side if the test framework eats the output, you can instead use an ofstream, and send your prints to say /var/tmp/test_log.txt or some such, and review the log after the fact?

jminor:
Progress... I was able to temporarily turn off `fail-fast` in the GitHub Action workflow to get the full matrix of CI jobs to run. Now we can see clearly which ones succeed and which fail.

The test fails on Ubuntu and Win-Mingw, but succeeds on macOS and Windows. It does not depend on Python version. (See screenshot below)

Let me make an Ubuntu VM to see if I can repro in there...

image

meshula:
@jminor At this point, it may come down to compiler settings, such as fast-math being enabled only on ubuntu. (fast-math, aka "non-reproducible sometimes broken math") should not be set at all for otio.

meshula:
@jminor This is looking good to me, aside from this: did you mean to add the ALE adaptor code to the commit?

jminor:
The ALE adapter fix was written before the adapters were moved, so those files are left in here just since I rebased this change. I can re-do those fixes in the new ALE adapter repo after this lands.

In the meantime, I was able to use the [tmate GitHub Action](https://mxschmitt.github.io/action-tmate/) to get a reverse shell on one of the runners in order to diagnose the Ubuntu-specific test failure. From poking at that in Python I can repro the issue, but I think I'll need to write a C++ test for `to_timecode` in order to find the true source of the problem.

src/opentime/rationalTime.cpp
jminor: I'm going to take the position here that any snapping or tolerance of slightly-off values should be handled via `nearest_smpte_timecode_rate` so that we can avoid propagating error into further calculations. Yes, those slightly-off cases will come up, but we want folks to use these functions to get the best value OTIO can provide, not to allow a slightly-off value to be used.

jminor:
Returning to this PR after a long hiatus... I rebased onto latest main, added some doc strings, fixed the bogus test case, and addressed some review notes from 2023.

src/opentime/rationalTime.cpp
rogernelson: That makes snese.  Perhaps then this method should return true/false based on whether or not `nearest_smpte_timecode_rate` is within the standard c++ tolerance for double comparison then so we don't run into the classic floating point comparison issues?

Simply:
```
return std::fabs(fps - nearest_smpte_timecode_rate(fps)) < std::numeric_limits<double>::epsilon();
```

The below is an additional thought, but I would not say it's required, I leave it up to you.

Note that there is already a hardcoded built in tolerance for  `nearest_smpte_timecode_rate` of 0.1 happening on line 461.

Perhaps to also address Nick's tolerance concerns for frame rates we could have a method like this that could be called there and used to replace that block of code:
```
double snap_to_smpte_rate(double rate, ErrorStatus* status)
{
   static constexpr double MAX_RATE_TOLERANCE = 1001.0 / 120000.0;
   const double nearest_rate = nearest_smpte_timecode_rate(rate);

   if (abs(nearest_rate) - rate) < MAX_RATE_TOLERANCE) return nearest_rate;
 
   if (error_status)
   {
      *error_status = ErrorStatus(ErrorStatus::INVALID_TIMECODE_RATE);
   }
   return 0.0;
}
```

PR #1481, opened 2022-11-18
Make opentimelineio headers public

michael-dm:
When linking to opentimelineio statically, cmake can't find imported headers :
```bash
fatal error: 'opentimelineio/clip.h' file not found
#include <opentimelineio/clip.h>
         ^~~~~~~~~~~~~~~~~~~~~~~
```

Making "target_include_directories" PUBLIC allows projects including opentimelineio in their CMakeLists.txt to correctly find the project headers.

### DIFF
--- a/src/opentimelineio/CMakeLists.txt
+++ b/src/opentimelineio/CMakeLists.txt
@@ -78,7 +78,7 @@ add_library(opentimelineio ${OTIO_SHARED_OR_STATIC_LIB}
 add_library(OTIO::opentimelineio ALIAS opentimelineio)
 
 target_include_directories(opentimelineio
-    PRIVATE       "${IMATH_INCLUDES}"
+    PUBLIC        "${IMATH_INCLUDES}"
                   "${PROJECT_SOURCE_DIR}/src"
                   "${PROJECT_SOURCE_DIR}/src/deps"
                   "${PROJECT_SOURCE_DIR}/src/deps/optional-lite/include"


src/opentimelineio/CMakeLists.txt
```diff
@@ -78,7 +78,7 @@ add_library(opentimelineio ${OTIO_SHARED_OR_STATIC_LIB}
 add_library(OTIO::opentimelineio ALIAS opentimelineio)
 
 target_include_directories(opentimelineio
-    PRIVATE       "${IMATH_INCLUDES}"
+    PUBLIC        "${IMATH_INCLUDES}"
```
meshula, line 81: Thanks for spotting this issue. This change makes sense in order to be able to use otio as a subproject. The reason it is PRIVATE currently is that the deps directory, and optional-lite directory are exposed, without checking first to see if the versions of the libraries in `deps` are used, or if the user has externally supplied their own dependent libraries. Some more work would be necessary to make the directory exports safe for general use.

meshula: In other words, the includes should be added via variables, such as `${OPTIONAL_LITE_INCLUDES}` or `${RAPIDJSON_INCLUDES}`. That would be better in any case...

meshula:
@michael-dm Any interest in landing this, or should I close it?

michael-dm:
> @michael-dm Any interest in landing this, or should I close it?

you can close, thanks 👍

Issue #1386, opened 2022-08-22
Rearrange PyPI projects, break out adapters


ssteinbach:
## Overview

The goal of this issue is to refactor the PyPI projects for OpenTimelineIO into two pypi projects from the current one:

* **OpenTimelineIO** (the current package) becomes **ONLY** the core python wrapped library
  * Versioned whenever a new release of the core comes out from GitHub (similar to the current release process)
* **OpenTimelineIO-Plugins** (a new package) is `OpenTimelineIO` + a curated set of plugins
  * Versioned whenever a plugin is added or removed from the list, or the `OpenTimelineIO` releases
  * A plugin may be removed from the list if it remains unsupported and falls behind the core, as determined by the OTIO TSC
  * PRs may be made to add or remove plugins from the list
  * The `Plugins` project will be in its own repository, with a README that describes briefly some criteria that the TSC is looking for when assessing whether to include a project or not
  * Plugins that are not present in `OpenTimelineIO-Plugins` may still be present on pip.  Users may install extra plugins through pip that aren't included in the `OpenTimelineIO-Plugins` project. 

## Outstanding tasks:

- [x] Remove the dependency on the contrib project from the top level setup.py
- [x] Add a note to the readme about the `OpenTimelineIO-Plugins`
- [x] Rename [umbrella repository](https://github.com/OpenTimelineIO/otio-umbrella) to `OpenTimeilneIO-Plugins`
- [x] Create pypi project for `OpenTimelineIO-Plugins`
- [x] add CI integration for pushing to pypi from `Plugins` repo
- [x] ~CI integration for checking new versions of `OpenTimelineIO` from the `Plugins` repo~ We'll leave this to adapter repo maintainers to do if they see fit
- [x] register pypi packages for the adapter plugin projects <- at least for the ones that we plan on having the umbrella project include
- [x] break out adapters
  - [x] as a TSC group we identify which adapters we want to include in the default installation
  - [x] each adapter gets its own package, which can be individually installed
  - [x] remove explicit references to the contrib package from the main library (see comment below) and documentation
  - [x] remove the contrib directory from the repository
  - [x] recommend a pinning strategy for how the adapters should handle their version dependencies + some CI for testing against new versions of the CORE 
- [x] ~add version freezing to release script for parent projects~ No Longer needed, we'll allow leaf adapter projects to set version constraints when needed
[...]

PR #1487, opened 2022-11-29
Extracted CMX3600 adapter and related sample data and tests

apetrynet:
This PR is related to issue #1386

**Summarize your change.**
This PR removes the `cmx_3600` adapter and related sample files and tests. 
There might still be some bits and pieces left, like in the documentation.


jminor:
This looks pretty good to me. If you want to be really clean about this change we _could_ separate the `imp` -> `importlib` change out and land that into `main` before the adapter switchero. Same for the tests that use `otio_json` instead of `cmx_3600`. That would make this PR just the removal of the EDL adapter.

tests/baselines/plugin_module/otio_override_adapter/plugin_manifest.json
```diff
@@ -3,9 +3,9 @@
     "adapters": [
         {
             "OTIO_SCHEMA" : "Adapter.1",
-            "name" : "cmx_3600",
+            "name" : "otio_json",
```
apetrynet, line 6: @jminor, is this what you meant regarding "otio_json"?
I believe this is a part of testing the plugin system and not directly related to the cmx adapter. Therefore I just updated it to something that will remain after the cmx extarction.

apetrynet:
What do we do about references to the cmx adapter in the documentation? Should we do that in this PR or do a proper doc overhaul when we're done with all the adapters?

tests/baselines/plugin_module/otio_override_adapter/plugin_manifest.json
jminor: Yes, I was just suggesting that this portion of the change could be landed into main if that helps make the extract_adapters changes easier to compare. It is fine as-is though :)

jminor:
Lets do the documentation change after we have the adapters split out, but before landing the extract_adapters branch into main.

tests/baselines/plugin_module/otio_override_adapter/plugin_manifest.json
apetrynet: On it! Thanks!

apetrynet:
Looks like we need to update some of the docs after all to pass the tests.

jminor (approved):

Issue #1474, opened 2022-11-01
Segmentation fault on pytest exit when creating a clip with metadata


ssteinbach:
A user reported that there is a segmentation fault when running pytest + objects with metadata + the "fs" (fake filesystem) capture system.

Repro:
```python
import opentimelineio as otio

# Does not segfault if you comment out this "test"
def test_empty_fs(fs):
    pass


def test_create_clip(capsys):
    # can be child classes of SerializableObjectWithMetadata (Clip, Gap, etc),
    # however other otio C++ classes appear to be ok (RationalTime, TimeRange,
    # TimeTransform)
    otio.core.SerializableObjectWithMetadata(
        # also does not segfault if you comment this out
        metadata={}
    )
```

Output:
```
❯ pytest otio_pytest_seg_fault.py --disable-warnings
============================================== test session starts ===============================================
platform darwin -- Python 3.8.13, pytest-7.2.0, pluggy-1.0.0
rootdir: /private/var/tmp
plugins: pyfakefs-5.0.0
collected 2 items

otio_pytest_seg_fault.py ..                                                                                [100%]

========================================== 2 passed, 1 warning in 0.09s ==========================================
zsh: segmentation fault  pytest otio_pytest_seg_fault.py --disable-warnings
```

(The warning is related to our use of `imp` and appears to be a red herring)

JeanChristopheMorinPerso:
When OTIO is built in debug mode, we get a little bit more details:

```
Fatal Python error: PyThreadState_Get: the function must be called with the GIL held, but the GIL is released (the current Python thread state is NULL)
Python runtime state: finalizing (tstate=0x000055649b90a060)

Aborted (core dumped)
```

JeanChristopheMorinPerso:
The fact that it crashes it metadata={} is given as an argument kind of suggests me it has to do with the round-trip in https://github.com/AcademySoftwareFoundation/OpenTimelineIO/blob/main/src/py-opentimelineio/opentimelineio-bindings/otio_utils.cpp#L99-L121. This part of the code is pretty painful to debug...

JeanChristopheMorinPerso:
```
(gdb) bt
#0  __pthread_kill_implementation (threadid=<optimized out>, signo=signo@entry=6, no_tid=no_tid@entry=0) at pthread_kill.c:44
[...]

PR #1490, opened 2022-12-04
Improve typing of the `metadata` parameter

JeanChristopheMorinPerso:
This PR is part of the series of PRs I have created to improve the types exposed by the Python bindings.

More particularly, this PR improves the `metadata` parameter type. But this is much easier said than done.

Note that I'm not fluent in C++, so it's entirely possible that I made some novice mistakes or that I did some bad things. I'll be more than happy to get feedback and learn new things.

# Changes

* Introduction of a new custom Pybind11 type caster that allows to expose an `AnyDictionaryProxy` that can both be an `AnyDictionaryProxy` and `py::dict` (this allows to keep compatibility with the previous implementation). It also handles recursively converting the types of the sub elements.

  This new caster is kind of special because like I said, it will handle both `AnyDictionaryProxy` and `py::dict`. It's also special because it will render the `AnyDictionaryProxy` type in Python docstrings as `Metadata`.
* Replaced `py_to_any_dictionary` with a pure C++ implementation, which removes the awkward Python > C++ > Python > C++ > Python round trip that it was doing to convert a `py::object` into an `AnyDictionary`.
* New `opentimelineio.core.Metadata` and `opentimelineio.core.MetadataValue` type aliases (defined in Python) that allows to have types clearly defined in Python. It's also nicely exposed in the Sphinx docs. This part is still pretty WIP. It kind of works but mypy isn't happy with it and it also requires a dirty hack for Sphinx.
* Added capability to construct `AnyDictionary` and `AnyVector` objects with values in Python. This means it's now possible to do `AnyDictionary({'asd': 1234})`. This is not strictly required for this PR, but since I was in the code for these and I always found it annoying that I couldn't do that, I thought I should just do it here.

# Side effects

I found no noticeable side effects, except that it fixes https://github.com/AcademySoftwareFoundation/OpenTimelineIO/issues/1474.

I'm planning on running https://github.com/bloomberg/memray to see if there is a difference in memory usage. It'd be nice if some people could try on big production OTIO files. That would actually also help to confirm that I didn't introduce a regression in functionality and behavior.

# Explanations

I have already written on the subject, see https://gist.github.com/JeanChristopheMorinPerso/0ad46432bf2c251b920c2bf648a86fe9#pyobject-as-parameter-type-in-class-signatures.

Previously, the `metadata` was exposed as `py::object` in constructors:
```cpp
py::class_<SOWithMetadata, SerializableObject,
               managing_ptr<SOWithMetadata>>(m, "SerializableObjectWithMetadata", py::dynamic_attr())

        // Notice the py::object
        .def(py::init([](std::string name, py::object metadata) {
                    return new SOWithMetadata(name, py_to_any_dictionary(metadata));
                }),
            py::arg_v("name"_a = std::string()),
            py::arg_v("metadata"_a = py::none()));
```

The main problem with this practice is that the signature becomes:
```
__init__(...)
    __init__(name: str = '', metadata: object = None) -> None
```

It's basically impossible to know what kind of type the metadata should be just by looking at the signature. But there is another thing to note about the previous approach. `py_to_any_dictionary` was doing a round-trip to python to do the type conversions! The round-trip looked like: Python (constructing the `SerializableObjectWithMetadata` object) > C++ (we are now in pybind11) > Python (`py_to_any_dictionary` calls `opentimelineio._core._core_utils._value_to_any`) > C++ (`_value_to_any` would create `PyAny` objects) > Python (PyAny returned to Python) > C++ (`py_to_any_dictionary` gets the return value from `_value_to_any`) > Python (Pybind11 does its thing and returns a SerializableObjectWithMetadata to Python).

That's pretty hacky and it was only to convert types. All this round tripping can be entirely replaced with a pure C++ approach.

With that, we now have:

```cpp
py::class_<SOWithMetadata, SerializableObject,
               managing_ptr<SOWithMetadata>>(m, "SerializableObjectWithMetadata", py::dynamic_attr())

        // Notice that we now accept an AnyDictionaryProxy
        .def(py::init([](std::string name, AnyDictionaryProxy* metadata) {
                    return new SOWithMetadata(name, metadata->fetch_any_dictionary());
                }),
            py::arg_v("name"_a = std::string()),
            py::arg_v("metadata"_a = py::dict())); // Note the default is now an empty dict
```

This now renders like this in Python:
```
__init__(...)
    __init__(name: str = '', metadata: opentimelineio.core.Metadata = {}) -> None
```
---

Even with this change, there is still one round-trip in the code base. And that remaining round-trip will be pretty hard to get rid of (at least for me. I guess it would be much easier for a C++ developer). The round-trip is basically https://github.com/AcademySoftwareFoundation/OpenTimelineIO/blob/main/src/py-opentimelineio/opentimelineio/core/_core_utils.py#L121. `AnyDictionaryProxy` C++ class (not the binding) wants `PyAny` objects, see https://github.com/AcademySoftwareFoundation/OpenTimelineIO/blob/main/src/py-opentimelineio/opentimelineio-bindings/otio_anyDictionary.h#L65. I tried to change that so that it takes Something else, but I couldn't come up with something that worked.


tests/test_serializable_object.py
```diff
@@ -36,6 +38,71 @@ def test_cons(self):
...
+                'AnyDictionary': d,
+                'RationalTime': RationalTime(
+                    value=10.0,
+                    rate=5.0
+                ),
+                'TimeRange': TimeRange(
+                    RationalTime(value=1.0),
+                    RationalTime(value=100.0)
+                ),
+                'TimeTransform': TimeTransform(
```
JeanChristopheMorinPerso, line 65: What about SerializableObject?

meshula:
This diff is pretty large, and I realize it's still a draft. I think it might make sense to schedule in cooperation with the TSC, an office hours zoom to have a look at it and help things a long.

JeanChristopheMorinPerso:
Sure! I'll be happy to have a call with TSC members to do a walk through of the changes.

Also, it's marked as a draft because there is still a thing or two to verify and tweak. For example the Metadata class should be renamed to something else (probably Dict, or something like that). And I need to make sure this Metadata class works with mypy and co. I also want to profile the changes and compare with the old code.

I'll be actively working on that in the next 4 weeks.

meshula:
splitting the PR as you suggest would be a good idea.

JeanChristopheMorinPerso:
I extracted some bits into https://github.com/AcademySoftwareFoundation/OpenTimelineIO/pull/1574. I'll update the PR once #1574 gets merged.

JeanChristopheMorinPerso:
I also created https://github.com/AcademySoftwareFoundation/OpenTimelineIO/pull/1575 to split another part out.

tests/test_serializable_object.py
JeanChristopheMorinPerso: I was starting to think that `SerializableObject` and co will cause issues, and it turns out that I was right. I'm not sure how I'll be able to make this work...

JeanChristopheMorinPerso: I have the same issue with the JS bindings, which doesn't come as a surprise since I'm using the same method to convert types...

JeanChristopheMorinPerso: I solved my problem. It took me almost two days of work, but we can now pass `SerializableObject`s (and its subclasses) like before.

src/py-opentimelineio/opentimelineio-bindings/otio_utils.cpp
```diff
@@ -94,30 +95,126 @@ void _build_any_to_py_dispatch_table() {
...
+        return any(py_to_cpp<TimeRange>(o));
+    }
+
+    if (py::isinstance<TimeTransform>(o)) {
+        return any(py_to_cpp<TimeTransform>(o));
+    }
+
+    if (py::isinstance<SerializableObject>(o)) {
+        SerializableObject::Retainer<> r(py::cast<SerializableObject*>(o));
+        return create_safely_typed_any(r.take_value());
```
JeanChristopheMorinPerso, line 167: I have to admit that I'm not too sure how the downcast works here, but it works. This is taken from https://github.com/AcademySoftwareFoundation/OpenTimelineIO/blob/main/src/py-opentimelineio/opentimelineio-bindings/otio_bindings.cpp#L31-L40.

We'll need more tests to make sure that everything works as expected.

src/py-opentimelineio/opentimelineio/core/__init__.py
```diff
@@ -32,6 +38,17 @@
     release_to_schema_version_map,
 )
 
+Metadata: TypeAlias = Union[Dict[str, 'MetadataValue'], AnyDictionary]
```
JeanChristopheMorinPerso, line 41: For these to work, we'll have to ove them into a [stub file](https://peps.python.org/pep-0484/#stub-files), more specifically `opentimelineio._otio.pyi`. We'll have to also create stubs for our bindings, which can be automated using [pybind11-stubgen](https://github.com/sizmailov/pybind11-stubgen).

PR #1514, opened 2023-01-07
Use duration() as a virtual function

### LINKED ISSUE #1513: Use duration() as a virtual function
I noticed a helper function ```_safe_duration()``` that was  dynamically casting a ```Composable``` just to call the function ```duration()```. This isn't necessary since ```duration()``` is a virtual function of ```Composable```. The function ```_safe_duration()``` could be removed and replaced with calls to ```duration()```.

darbyjohnston:
Fixes #1513

I noticed a helper function ```_safe_duration()``` that was  dynamically casting a ```Composable``` just to call the function ```duration()```. This isn't necessary since ```duration()``` is a virtual function of ```Composable```. The function ```_safe_duration()``` could be removed and replaced with calls to ```duration()```.

### DIFF
--- a/src/opentimelineio/composable.cpp
+++ b/src/opentimelineio/composable.cpp
@@ -66,7 +66,10 @@ Composable::duration(ErrorStatus* error_status) const
 {
     if (error_status)
     {
-        *error_status = ErrorStatus::NOT_IMPLEMENTED;
+        *error_status = ErrorStatus(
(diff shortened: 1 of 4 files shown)


meshula:
ltgm! This looks like some legacy that fell through the cracks. Thanks for the fix!

meshula (approved):

PR #1543, opened 2023-02-10
Use the C++ override keyword

### LINKED ISSUE #1542: Use the C++ override keyword
In C++11 the "override" keyword was added which checks at compile time whether a virtual function has been properly overridden. This can catch bugs from typos in derived classes and ensures that base class destructors are marked virtual. Some interesting discussion here, especially on whether destructors should use "override":
https://github.com/isocpp/CppCoreGuidelines/issues/721

darbyjohnston:

Fixes #1542

In C++11 the "override" keyword was added which checks at compile time whether a virtual function has been properly overridden. This can catch bugs from typos in derived classes and ensures that base class destructors are marked virtual. Some interesting discussion here, especially on whether destructors should use "override":
https://github.com/isocpp/CppCoreGuidelines/issues/721

Some compilers (e.g., the latest Xcode) now emit warnings for this:
```
/Users/darby/Dev/otio/OpenTimelineIO/src/opentimelineio/item.h:36:18: warning: 'visible' overrides a member function but is not marked 'override' [-Winconsistent-missing-override]
    virtual bool visible() const;
```

This PR changes:
* Use the "override" keyword on all functions that override a virtual function in a base class
* Make the virtual function SerializableObject::_schema_name_for_reference() protected instead of private
* Make the functions UnknownSchema::~ UnknownSchema() and UnknownSchema::_schema_name_for_reference() protected instead of private

### DIFF
--- a/src/opentimelineio/clip.h
+++ b/src/opentimelineio/clip.h
@@ -45,17 +45,17 @@ public:
                std::string const& new_active_key,
                ErrorStatus*       error_status = nullptr) noexcept;
 
-    virtual TimeRange
-    available_range(ErrorStatus* error_status = nullptr) const;
+    TimeRange
+    available_range(ErrorStatus* error_status = nullptr) const override;
 
-    virtual optional<Imath::Box2d>
-    available_image_bounds(ErrorStatus* error_status) const;
+    optional<Imath::Box2d>
+    available_image_bounds(ErrorStatus* error_status) const override;
 
 protected:
-    virtual ~Clip();
+    ~Clip() override;
 
-    virtual bool read_from(Reader&);
-    virtual void write_to(Writer&) const;
+    bool read_from(Reader&) override;
+    void write_to(Writer&) const override;
 
 private:
     template <typename MediaRefMap>
--- a/src/opentimelineio/clip.h
+++ b/src/opentimelineio/clip.h
@@ -52,7 +52,7 @@ public:
     available_image_bounds(ErrorStatus* error_status) const override;
 
 protected:
-    ~Clip() override;
+    virtual ~Clip();
 
     bool read_from(Reader&) override;
     void write_to(Writer&) const override;
--- a/src/opentimelineio/composable.h
+++ b/src/opentimelineio/composable.h
@@ -46,7 +46,7 @@ protected:
         return const_cast<Composable*>(this)->_highest_ancestor();
     }
 
-    ~Composable() override;
+    virtual ~Composable();
 
     bool read_from(Reader&) override;
     void write_to(Writer&) const override;
--- a/src/opentimelineio/composable.h
+++ b/src/opentimelineio/composable.h
@@ -46,10 +46,10 @@ protected:
         return const_cast<Composable*>(this)->_highest_ancestor();
     }
 
-    virtual ~Composable();
+    ~Composable() override;
 
-    virtual bool read_from(Reader&);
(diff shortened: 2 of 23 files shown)


src/opentimelineio/clip.h
```diff
@@ -45,17 +45,17 @@ class Clip : public Item
...
+    TimeRange
+    available_range(ErrorStatus* error_status = nullptr) const override;
 
-    virtual optional<Imath::Box2d>
-    available_image_bounds(ErrorStatus* error_status) const;
+    optional<Imath::Box2d>
+    available_image_bounds(ErrorStatus* error_status) const override;
 
 protected:
-    virtual ~Clip();
```
meshula, line 55: My reading of the iso cpp guidelines is that destructors must be virtual, not override.

http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c127-a-class-with-a-virtual-function-should-have-a-virtual-or-protected-destructor

http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rh-override

Do you read it the same as I?

meshula (commented):
This is a fantastic improvement, however my understanding is that destructors must be virtual (see inline comment).

meshula:
To clarify my comment, the outcome of the discussion on the iso issue you linked was the guidance in the document that I linked, and it seems, to my reading of the document, that they concluded destructors should be virtual.

darbyjohnston:
I think that is just for base classes though, once a function has been marked virtual doesn't that make the function virtual in any derived classes? I thought the last comment in the thread described it well "memory leaks (and other nasty problems), when deleting an object through a superclass pointer, because of forgetting to mark the superclass's destructor virtual". However I do think having "override" on the destructors looks a bit odd.

meshula:
Effective C++ says "If a class has any virtual function, it should have a virtual destructor, and that classes not designed to be base classes or not designed to be used polymorphically should not declare virtual destructors." Although it's true that if a destructor B is not virtual and it's super is A, that deleting a pointer to B will call the destructor on B then A, if you have a non-virtual C pointer, then only C's destructor will run. 

So the only case where override without virtual would be logical in our use, is if the class is also final, and the superclass is virtual. Since OTIO leverages polymorphism a lot, I think it is easier to reason about the code if we follow the "Effective" rule.

Effective C++ is older than the override keyword so if Meyers were to revise it he might publish a different opinion, but I think I am convincing myself that override is not going to be as maintainable and predictable as virtual, and that we know for sure we want the "virtual chained destructor" behavior in all cases. I hope this argument convinces you as well ;)

darbyjohnston:
I still think the comment makes a good case for catching non-virtual destructors in base classes, but seeing "override" on destructors is confusing enough maybe it's not worth it. Especially with OTIO since there is only a single base class? I updated the PR replacing override with virtual on the destructors.

Speaking of the single base class, SerializableObject, is that supposed to be instantiable? If not maybe it should have a pure virtual destructor?

darbyjohnston:
Ah, never mind, it looks like at least the type registry system needs to be able to create SerializableObjects.

meshula:
Thanks for the change! The semantics around override and virtual and destructor are slightly puzzling, that's for sure :)

meshula (approved):

meshula:
Do you have further work you want to do on this? If no, I say we are ready to land it :)

darbyjohnston:
I thinks it is good to go, thanks!

PR #1550, opened 2023-02-17
Fix exception when __doc__ is None with portable Python

### LINKED ISSUE #1548: opentimelineio won't import in an embedable Python interpreter

git commit d48163e4 introduced a change to omit external functions (ie functions from the python built-in `collections_abc` module) from the sphinx documentation.
The logic expects these functions to have \_\_doc\_\_ properties and performs a += operation to append the necessary tag for sphinx.

However embeddable python (eg. https://www.python.org/ftp/python/3.7.9/python-3.7.9-embed-amd64.zip) ships with only byte-code compiled versions of its built-in functions - including the collections module. pyc files do not contain docstrings, so as a result those functions will have \_\_doc\_\_ set to None and no string concatenation is possible.

I would like to submit a PR with the fix. Please let me know the next steps.

## To Reproduce

1. Operating System: Windows
2. Python version: Embeddable/portable 3.7.x and 3.9.x
3. Example snippet that demonstrates the issue - if it's a build issue,
    import opentimelineio

5. OpenTimelineIO release version or commit hash: 0.15.0
6. Compiler information: N/A, I've used prebuilt python wheel from pypi

## Expected Behavior

I expect that `import opentimelineio` succeeds.

## Screenshots

```
C:\Users\robsz\AppData\Local\Programs\al-editron-beta\resources\python>dist\python.exe -c "import opentimelineio"
  File "<string>", line 1, in <module>
  File "C:\Users\robsz\AppData\Local\Programs\al-editron-beta\resources\python\dist\dependencies\opentimelineio\__init__.py", line 14, in <module>
    from . import (
  File "C:\Users\robsz\AppData\Local\Programs\al-editron-beta\resources\python\dist\dependencies\opentimelineio\core\__init__.py", line 35, in <module>
    from . _core_utils import ( # noqa
  File "C:\Users\robsz\AppData\Local\Programs\al-editron-beta\resources\python\dist\dependencies\opentimelineio\core\_core_utils.py", line 407, in <module>
    _add_mutable_mapping_methods(AnyDictionary)
  File "C:\Users\robsz\AppData\Local\Programs\al-editron-beta\resources\python\dist\dependencies\opentimelineio\core\_core_utils.py", line 239, in _add_mutable_mapping_methods
    getattr(mapClass, name).__doc__ += '\n\n:meta private:'
TypeError: unsupported operand type(s) for +=: 'NoneType' and 'str'
```

robbiesz:

Fixes #1548

**Summarize your change.**

Embed-able Python distributions (at least python 3.7 and 3.9) ship with only byte-code compiled versions of their
built-in modules. As a result, the docstrings of classes, methods will not be set (ie. \_\_doc\_\_ will be set to None)

git commit  https://github.com/AcademySoftwareFoundation/OpenTimelineIO/commit/d48163e464cd37ae272d98ef5dca0c804f840475 introduced a change to exclude external functions (namely from `collections_abc`) from the OTIO sphinx documentation by appending a tag to the \_\_doc\_\_ string. However the code assumes that \_\_doc\_\_ is always set to a string.

On embedded python this will result in a crash with the following error:
```
C:\Users\robsz\AppData\Local\Programs\al-editron-beta\resources\python>dist\python.exe -c "import opentimelineio"
  File "<string>", line 1, in <module>
  File "C:\Users\robsz\AppData\Local\Programs\al-editron-beta\resources\python\dist\dependencies\opentimelineio\__init__.py", line 14, in <module>
    from . import (
  File "C:\Users\robsz\AppData\Local\Programs\al-editron-beta\resources\python\dist\dependencies\opentimelineio\core\__init__.py", line 35, in <module>
    from . _core_utils import ( # noqa
  File "C:\Users\robsz\AppData\Local\Programs\al-editron-beta\resources\python\dist\dependencies\opentimelineio\core\_core_utils.py", line 407, in <module>
    _add_mutable_mapping_methods(AnyDictionary)
  File "C:\Users\robsz\AppData\Local\Programs\al-editron-beta\resources\python\dist\dependencies\opentimelineio\core\_core_utils.py", line 239, in _add_mutable_mapping_methods
    getattr(mapClass, name).__doc__ += '\n\n:meta private:'
TypeError: unsupported operand type(s) for +=: 'NoneType' and 'str'
```

None

### DIFF
--- a/src/py-opentimelineio/opentimelineio/core/_core_utils.py
+++ b/src/py-opentimelineio/opentimelineio/core/_core_utils.py
@@ -177,7 +177,7 @@ def _add_mutable_mapping_methods(mapClass):
 
                 # Hide the method frm Sphinx doc.
                 # See https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#info-field-lists  # noqa
-                # __doc__ will be None on an embedded python interpreter where the built-in modules are provided  # noqa
+                # __doc__ will be None on a Windows embeddable package where the built-in modules are provided  # noqa
                 # as pyc files which do not include the docstrings.
                 if getattr(mapClass, name).__doc__ is not None:
                     getattr(mapClass, name).__doc__ += "\n\n:meta private:"
@@ -332,7 +332,7 @@ def _add_mutable_sequence_methods(
 
                     # Hide the method frm Sphinx doc.
                     # See https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#info-field-lists  # noqa
-                    # __doc__ will be None on an embedded python interpreter where the built-in modules are provided  # noqa
+                    # __doc__ will be None on a Windows embeddable package where the built-in modules are provided  # noqa
                     # as pyc files which do not include the docstrings.
                     if getattr(sequenceClass, name).__doc__ is not None:
                         getattr(sequenceClass, name).__doc__ += "\n\n:meta private:"
--- a/src/py-opentimelineio/opentimelineio/core/_core_utils.py
+++ b/src/py-opentimelineio/opentimelineio/core/_core_utils.py
@@ -177,9 +177,10 @@ def _add_mutable_mapping_methods(mapClass):
 
                 # Hide the method frm Sphinx doc.
                 # See https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#info-field-lists  # noqa
-                getattr(mapClass, name).__doc__ = "{}\n\n:meta private:".format(
-                    getattr(mapClass, name).__doc__ or ""
-                )
+                # __doc__ will be None on an embedded python interpreter where the built-in modules are provided  # noqa
+                # as pyc files which do not include the docstrings.
+                if getattr(mapClass, name).__doc__ is not None:
+                    getattr(mapClass, name).__doc__ += "\n\n:meta private:"
 
     mapClass.setdefault = setdefault
     mapClass.pop = pop
@@ -331,11 +332,10 @@ def _add_mutable_sequence_methods(
 
                     # Hide the method frm Sphinx doc.
                     # See https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#info-field-lists  # noqa
-                    getattr(sequenceClass, name).__doc__ = (
-                        "{}\n\n:meta private:".format(
-                            getattr(sequenceClass, name).__doc__ or ""
-                        )
-                    )
+                    # __doc__ will be None on an embedded python interpreter where the built-in modules are provided  # noqa
+                    # as pyc files which do not include the docstrings.
+                    if getattr(sequenceClass, name).__doc__ is not None:
+                        getattr(sequenceClass, name).__doc__ += "\n\n:meta private:"
 
     if not issubclass(sequenceClass, SerializableObject):
         def __copy__(self):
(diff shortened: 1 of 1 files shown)


JeanChristopheMorinPerso:
@robbiesz thanks! There is still one step left, which is DCO. See https://github.com/AcademySoftwareFoundation/OpenTimelineIO/pull/1550/checks?check_run_id=11427853276 for instructions on how to proceed.

src/py-opentimelineio/opentimelineio/core/_core_utils.py
```diff
@@ -177,7 +177,9 @@ def __deepcopy__(self, memo):
 
                 # Hide the method frm Sphinx doc.
                 # See https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#info-field-lists  # noqa
-                getattr(mapClass, name).__doc__ += '\n\n:meta private:'
+                getattr(mapClass, name).__doc__ = "{}\n\n:meta private:".format(
+                    getattr(mapClass, name).__doc__ or ""
+                )
```
JeanChristopheMorinPerso, line 182:
```suggestion
                # Add comment here to explain why we need to check for None
                if getattr(mapClass, name).__doc__ is not None:
                    getattr(mapClass, name).__doc__ += "\n\n:meta private:"
```
I think it would be clearer if we checked for `None` specifically. Also a little comment would be great so that we don't forget why sometimes it's None.

JeanChristopheMorinPerso (changes requested):

src/py-opentimelineio/opentimelineio/core/_core_utils.py
robbiesz: You're right! Sphinx won't ever be run to generate documentation on an embedded python, so there is no need to add the tag if \_\_doc\_\_ is None.

src/py-opentimelineio/opentimelineio/core/_core_utils.py
```diff
@@ -177,7 +177,10 @@ def __deepcopy__(self, memo):
 
                 # Hide the method frm Sphinx doc.
                 # See https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#info-field-lists  # noqa
-                getattr(mapClass, name).__doc__ += '\n\n:meta private:'
+                # __doc__ will be None on an embedded python interpreter where the built-in modules are provided  # noqa
```
JeanChristopheMorinPerso, line 180: Last nitpick I promise: Could you change "embedded python interpreter" with "Windows embeddable package" please? Embedded in this case has multiple meaning (for example it could be a Pybind11 embedded interpreter, etc). "Windows embeddable package" will make it clear that it's really the Windows "portable" install.

JeanChristopheMorinPerso (changes requested):

src/py-opentimelineio/opentimelineio/core/_core_utils.py
robbiesz: All good! :) I'll make the change

JeanChristopheMorinPerso (approved):
Thanks a lot!

robbiesz:
@JeanChristopheMorinPerso Did I open the PR against the wrong (main) branch? I can't seem to do the merge. "You’re not authorized to merge this pull request."

JeanChristopheMorinPerso:
@robbiesz You did everything right. Only maintainers and committers can commit code to the main branch. We'll take care of that for you.

PR #1574, opened 2023-04-07
Allow to construct AnyVector and AnyDictionary with values from Python

JeanChristopheMorinPerso:
Allow to construct `AnyVector` and `AnyDictionary` with values from Python.

Note that it only useful to OTIO developers who have to play with the Python bindings internals. For example I often have to create these objects when I work on the bindings, and it's quite annoying to have to do
```python
import opentimelineio as otio
d = otio._otio.AnyDictionary()
d['a'] = 'a'
```
while I could simply do `d = otio._otio.AnyDictionary({'a': 'a'})`.

This was extracted from https://github.com/AcademySoftwareFoundation/OpenTimelineIO/pull/1490.

### DIFF
--- a/src/py-opentimelineio/opentimelineio-bindings/otio_anyVector.cpp
+++ b/src/py-opentimelineio/opentimelineio-bindings/otio_anyVector.cpp
@@ -19,6 +19,14 @@ void otio_any_vector_bindings(py::module m) {
     
     py::class_<AnyVectorProxy>(m, "AnyVector")
         .def(py::init<>())
+        .def(py::init([](const py::iterable &it) {
+            any a;
+            py_to_any(it, &a);
+            AnyVector v = safely_cast_any_vector_any(a);
+            auto proxy = new AnyVectorProxy;
+            proxy->fetch_any_vector().swap(*v.get_or_create_mutation_stamp()->any_vector);
+            return proxy;
+        }))
         .def("__internal_getitem__", &AnyVectorProxy::get_item, "index"_a)
         .def("__internal_setitem__", &AnyVectorProxy::set_item, "index"_a, "item"_a)
         .def("__internal_delitem__", &AnyVectorProxy::del_item, "index"_a)
--- a/tests/test_core_utils.py
+++ b/tests/test_core_utils.py
@@ -2,6 +2,7 @@ import copy
 import unittest
 
 import opentimelineio._otio
+import opentimelineio.opentime
 import opentimelineio.core._core_utils
 
 
@@ -60,6 +61,38 @@ class AnyDictionaryTests(unittest.TestCase):
         with self.assertRaisesRegex(ValueError, "container mutated during iteration"):
             for key in d:
                 del d['b']
+
+    def test_construct_with_values(self):
+        d1 = opentimelineio.core._core_utils.AnyDictionary({'key1': 1234, 'key_2': {'asdasdasd': 5.6}})
+        v = opentimelineio.core._core_utils.AnyVector()
+        v.append(1)
+        v.append('inside any vector')
+
+        so = opentimelineio._otio.SerializableObject()
+        d2 = opentimelineio.core._core_utils.AnyDictionary({
+                'string': 'myvalue',
+                'int': -999999999999,
+                'list': [1, 2.5, 'asd'],
+                'dict': {'map1': [345]},
+                'AnyVector': v,
+                'AnyDictionary': d1,
+                'RationalTime': opentimelineio.opentime.RationalTime(
+                    value=10.0,
+                    rate=5.0
+                ),
+                'TimeRange': opentimelineio.opentime.TimeRange(
+                    opentimelineio.opentime.RationalTime(value=1.0),
+                    opentimelineio.opentime.RationalTime(value=100.0)
+                ),
+                'TimeTransform': opentimelineio.opentime.TimeTransform(
+                    offset=opentimelineio.opentime.RationalTime(value=55.0),
+                    scale=999
+                ),
+                'SerializableObject': so
+            })
+        self.assertEqual(d2['string'], 'myvalue')
+        self.assertEqual(d2['SerializableObject'], so)
+        self.assertEqual(d2['AnyDictionary'], d1)
 
     def test_raises_if_ref_destroyed(self):
         d1 = opentimelineio.core._core_utils.AnyDictionary()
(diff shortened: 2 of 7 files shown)


src/py-opentimelineio/opentimelineio-bindings/otio_anyVector.cpp
```diff
@@ -19,6 +19,14 @@ void otio_any_vector_bindings(py::module m) {
...
     py::class_<AnyVectorProxy>(m, "AnyVector")
         .def(py::init<>())
+        .def(py::init([](const py::iterable &it) {
+            any a;
+            py_to_any(it, &a);
+            AnyVector v = safely_cast_any_vector_any(a);
+            auto proxy = new AnyVectorProxy;
+            proxy->fetch_any_vector().swap(*v.get_or_create_mutation_stamp()->any_vector);
+            return proxy;
+        }))
```
ssteinbach, line 29: is this copying the data?

tests/test_core_utils.py
```diff
@@ -61,6 +62,42 @@ def test_raise_on_mutation_during_iter(self):
...
+
+        so = opentimelineio._otio.SerializableObject()
+        d2 = opentimelineio.core._core_utils.AnyDictionary(
+            {
+                'string': 'myvalue',
+                'int': -999999999999,
+                'list': [1, 2.5, 'asd'],
+                'dict': {'map1': [345]},
+                'AnyVector': v,
+                'AnyDictionary': d1,
```
ssteinbach, line 81: related to the question about copies, can you edit d1 in place and have d2["AnyDictionary"] change?

ssteinbach (commented):
This looks good to me. My only question is about copying and interacting with the reference system.. I wonder if there are any tests that could be added for that behavior.

ssteinbach:
C++ folks, @darbyjohnston, @davidbaraff, or @meshula -- can anyone else scan this?  Thanks @JeanChristopheMorinPerso!

tests/test_core_utils.py
JeanChristopheMorinPerso: Items removed or added in d1 in place won't be added/removed in d2, but values are shared. As far as I see, this is an existing behavior. For example, this passes:

```python
    def test_construct_with_values(self):
        # Contruct with values
        d1 = opentimelineio.core._core_utils.AnyDictionary(
            {'key1': 1234, 'key_2': {'asdasdasd': 5.6}}
        )

        # Construct without values and only add a value after instantiation.
        d3 = opentimelineio.core._core_utils.AnyDictionary()
        d3['key4'] = 'value4'

        # Create an SO and add it to d1 and d3.
        so = opentimelineio._otio.SerializableObjectWithMetadata('initial name')
        d1['so'] = so
        d3['so'] = so

        d2 = opentimelineio.core._core_utils.AnyDictionary(
            {
                'd1': d1,
                'd3': d3,
                'so': so
            }
        )
        self.assertEqual(d2['d1'], d1)
        self.assertEqual(d2['d3'], d3)

        # Edit d1 in place
        d1['key3'] = 'value3'
        d1['so'].name = 'new name set on d1["so"]'
        # Confirm that d2['d1'] is not updated
        self.assertEqual(d2['d1'], {'key1': 1234, 'key_2': {'asdasdasd': 5.6}, 'so': so})
        self.assertEqual(d1, {'key1': 1234, 'key_2': {'asdasdasd': 5.6}, 'key3': 'value3', 'so': so})
        # But values that were already there are shared.
        # We previously updated the name of so by doing d1['so'].name = '...'
        # and d2['so'] got updated.
        self.assertEqual(d2['so'].name, 'new name set on d1["so"]')

        # Same behavior here.
        d3['key5'] = 'value5'
        d3['so'].name = 'new name set on d3["so"]'
        self.assertEqual(d2['d3'], {'key4': 'value4', 'so': so})
        self.assertEqual(d3, {'key4': 'value4', 'key5': 'value5', 'so': so})
        self.assertEqual(d2['so'].name, 'new name set on d3["so"]')
```

JeanChristopheMorinPerso: Same result on the main branch (with slight modification in the code example to not construct with values):
```python
    def test_construct_with_values(self):
        d1 = opentimelineio.core._core_utils.AnyDictionary()
        d1['key1'] = 1234
        d1['key_2'] = {'asdasdasd': 5.6}

        # Construct without values and only add a value after instantiation.
        d3 = opentimelineio.core._core_utils.AnyDictionary()
        d3['key4'] = 'value4'

        # Create an SO and add it to d1 and d3.
        so = opentimelineio._otio.SerializableObjectWithMetadata('initial name')
        d1['so'] = so
        d3['so'] = so

        d2 = opentimelineio.core._core_utils.AnyDictionary()
        d2['d1'] = d1
        d2['d3'] = d3
        d2['so'] = so
        self.assertEqual(d2['d1'], d1)
        self.assertEqual(d2['d3'], d3)

        # Edit d1 in place
        d1['key3'] = 'value3'
        d1['so'].name = 'new name set on d1["so"]'
        # Confirm that d2['d1'] is not updated
        self.assertEqual(d2['d1'], {'key1': 1234, 'key_2': {'asdasdasd': 5.6}, 'so': so})
        self.assertEqual(d1, {'key1': 1234, 'key_2': {'asdasdasd': 5.6}, 'key3': 'value3', 'so': so})
        # But values that were already there are shared.
        # We previously updated the name of so by doing d1['so'].name = '...'
        # and d2['so'] got updated.
        self.assertEqual(d2['so'].name, 'new name set on d1["so"]')

        # Same behavior here.
        d3['key5'] = 'value5'
        d3['so'].name = 'new name set on d3["so"]'
        self.assertEqual(d2['d3'], {'key4': 'value4', 'so': so})
        self.assertEqual(d3, {'key4': 'value4', 'key5': 'value5', 'so': so})
        self.assertEqual(d2['so'].name, 'new name set on d3["so"]')
```

src/py-opentimelineio/opentimelineio-bindings/otio_anyVector.cpp
JeanChristopheMorinPerso: If by data you mean everything (like a deep copy in Python), no. If you mean the top level list/vector, yes. See https://github.com/AcademySoftwareFoundation/OpenTimelineIO/pull/1574/files#r1165693980.

What should be the behavior?

meshula:
My question is whether or not this opens a path to creating malformed constructions? (By opens a path, I mean, can you do something now to build something malformed, that could not be built before?)

davidbaraff:
An important invariant is not being able to build something in Python that causes C++ to crash. :) I’m fine with programming in C++ and doing illegal things in OTIO that cause a core dump. That’s the nature of C++.But the goal is that if you type something in in Python that can cause a core dump, it’s a bug on the implementors of the bridge (i.e., primary me) that should be fixed. The corollary is never to open up any new capabilities that permit this to happen. The bridging regarding dictionaries/vectors is extremely complicated because of things like modifying these structures during traversal, which could invalidate a C++ iterator. A lot of care was taken to get to where we are now (I don’t know that it’s perfect right now of course but I don’t know of any existing mechanisms that can cause a crash).So just saying that any new capabilities have to keep this front of mind. This stuff is NOT easy.Sent from my iPadOn Apr 13, 2023, at 9:37 AM, Nick Porcino ***@***.***> wrote:﻿
My question is whether or not this opens a path to creating malformed constructions? (By opens a path, I mean, can you do something now to build something malformed, that could not be built before?)

JeanChristopheMorinPerso:
> My question is whether or not this opens a path to creating malformed constructions? (By opens a path, I mean, can you do something now to build something malformed, that could not be built before?)

@meshula By that you mean will accept values/types that couldn't used before?

Basically, when we do `d['asd'] = 'value'` in python (where `d` is an `AnyDictionary` instance), it will call `d.__setitem__`, `d.__setitem__` will use `otio.core._core_utils._value_to_any` to convert `value` into a `PyAny` which will then be use to do `d.__internal_setitem__(key, PyAny_value)`. With the changes in this PR, the code path is almost the same when constructing with values. There is just an additional step which is that the Pybind11 class will call `py_to_any_dictionary(value)` to convert the value to an `AnyDictionary`. `py_to_any_dictionary` (on the C++ side) calls `py_to_any) (also C++), which roundtrips in Python to call `otio.core._core_utils._value_to_any`. The return value from that roundtrip is casted to an AnyDictionary. To me it's quite similar.

Also, I'm not aware of anyone actually using the `AnyDictionary` class in Python directly to construct metadata. As far as I know, in Python it's more an mplementation deal that behaves and looks like a normal dict to users than anything else. And users pass and use dicts with the Python API. For example, there is no easy way to instantiate `AnyVector` or `AnyDictionary`. They are only accessible from `opentimelineio._otio` or `opentimelineio.core._core_utils`, which are both "private".

Also, thanks @davidbaraff! There is one known crash with AnyDictionary, see https://github.com/AcademySoftwareFoundation/OpenTimelineIO/issues/1474. But that's existing behavior. And there is also a known exception:
```
>>> opentimelineio.core.SerializableObjectWithMetadata().metadata
  File "/home/jcmorin/jcmenv/aswf/OpenTimelineIO/build/lib.linux-x86_64-cpython-310/opentimelineio/core/_core_utils.py", line 127, in __repr__
    return repr(dict(self))
  File "/usr/lib/python3.10/_collections_abc.py", line 886, in __iter__
    yield from self._mapping
ValueError: Underlying C++ AnyDictionary has been destroyed
```

But that's all existing behaviour.

PR #1575, opened 2023-04-07
Add equality operators to AnyVector in Python

JeanChristopheMorinPerso:
Add equality operators to `AnyVector` in Python. This is mostly useful for OTIO development, but it might also be useful to anyone using OTIO.

Extracted from https://github.com/AcademySoftwareFoundation/OpenTimelineIO/pull/1490.


JeanChristopheMorinPerso:
The implementation is working, but only for `==`. I'll close and re-open when I have a better implementation.

PR #1578, opened 2023-04-11
Make serialize_json_to_string output compact when indent is less than 1

JeanChristopheMorinPerso:
When we use `opentimelineio.serialize_json_to_string` and we set `indent` to 0, it outputs using multiple lines with no indentation instead of outputting a compact, single line string:
```python
>>> opentimelineio.core.serialize_json_to_string(so, indent=0)
{
"OTIO_SCHEMA": "SerializableObjectWithMetadata.1",
"metadata": {
"a": "asd",
"b": "b"
},
"name": "sowithmet"
}
```

With this change, it will output a single line:
```python
>>> opentimelineio.core.serialize_json_to_string(so, indent=0)
{"OTIO_SCHEMA":"SerializableObjectWithMetadata.1","metadata":{"a":"asd","b":"b"},"name":"sowithmet"}
```

A compact string can be useful in unit tests.

Note that I'm still learning C++. The method I used adds some code duplication which I initially tried to avoid but couldn't find a way to avoid it. For example I tried to use a pointer or reference, but like mentioned in https://github.com/Tencent/rapidjson/issues/912#issuecomment-488550187, it doesn't work. Let me know if it could be improved!

### DIFF
--- a/src/opentimelineio/serialization.cpp
+++ b/src/opentimelineio/serialization.cpp
@@ -1232,10 +1232,7 @@ serialize_json_to_string_pretty(
         OTIO_rapidjson::kWriteNanAndInfFlag>
         json_writer(output_string_buffer);
 
-    if (indent >= 0)
-    {
-        json_writer.SetIndent(' ', indent);
-    }
+    json_writer.SetIndent(' ', indent);
 
     JSONEncoder<decltype(json_writer)> json_encoder(json_writer);
 
@@ -1256,8 +1253,7 @@ std::string
 serialize_json_to_string_compact(
     const any&                value,
     const schema_version_map* schema_version_targets,
-    ErrorStatus*              error_status,
-    int                       indent)
+    ErrorStatus*              error_status)
 {
     OTIO_rapidjson::StringBuffer output_string_buffer;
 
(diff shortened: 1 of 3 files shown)


src/opentimelineio/serialization.cpp
```diff
@@ -1251,6 +1251,61 @@ serialize_json_to_string(
     return std::string(output_string_buffer.GetString());
 }
 
+// to json_string
+std::string
+serialize_json_to_string_compact(
+    const any&                value,
+    const schema_version_map* schema_version_targets,
+    ErrorStatus*              error_status,
+    int                       indent)
```
ssteinbach, line 1260: in the compact form, is `indent` the number of spaces between things?  or is that unused?

ssteinbach (changes requested):

ssteinbach:
> ~I should add a test.~ (test added). Also, should we do the same for file writing? I wasn't sure if I should change that too since my main use case is just unit tests where it's annoying to have to deal with newline characters. It could also be useful to send over HTTP where it would save a couple of bytes... But for file writing, I can't see a situation where it would be useful to have a single line instead of multiple...

I would say leave the serialized to disk stuff as it is.  It can be zipped or there could be a binary representation if people want a compact form (like otioz).

src/opentimelineio/serialization.cpp
JeanChristopheMorinPerso: It's unused. I forgot to remove it.

ssteinbach (approved):
LGTM

PR #1583, opened 2023-04-15
Remove the remains of "contrib"

apetrynet:
As part of the reorganization effort listed in #1386 this PR removes the "contrib" directory tree and all references to the contrib adapters.
With exception of some of the documentation, all references to contrib should be removed. 

`make test` passes locally


ssteinbach (approved):
LGTM!  Thanks @apetrynet

PR #1598, opened 2023-05-13
Adjust license for package uploaded to PyPI

JeanChristopheMorinPerso:
I noticed that our package in PyPI was still advertising the old license (Modified Apache 2.0). So this PR fixes that to make sure. Our next release will now correctly show Apache 2.0.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -221,7 +221,7 @@ PROJECT_METADATA = {
     "version": "0.16.0.dev1",
     "author": 'Contributors to the OpenTimelineIO project',
     "author_email": 'otio-discussion@lists.aswf.io',
-    "license": 'Modified Apache 2.0 License',
+    "license": 'Apache 2.0 License',
 }
 
 METADATA_TEMPLATE = """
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -310,7 +310,7 @@ def run(self):
         'Topic :: Multimedia :: Video :: Display',
         'Topic :: Multimedia :: Video :: Non-Linear Editor',
         'Topic :: Software Development :: Libraries :: Python Modules',
-        'License :: Other/Proprietary License',
+        'License :: OSI Approved :: Apache Software License',
```
JeanChristopheMorinPerso, line 313: If anyone is curious, PyPI doesn't know about Apache 2.0 and this is what's used for Apache 2.0. See https://pypi.org/classifiers/.

setup.py
```diff
@@ -415,5 +415,6 @@ def run(self):
     },
 
     # expand the project metadata dictionary to fill in those values
-    **PROJECT_METADATA
+    **PROJECT_METADATA,
+    license_file='LICENSE.txt'
```
apetrynet, line 419: Is this correct? I tried finding this key in the setup.py docs, but only found it in setup.cfg

apetrynet (commented):
Only have one concern regarding the "license_file" key.

setup.py
JeanChristopheMorinPerso: It is based on https://setuptools.pypa.io/en/latest/references/keywords.html. But I just noticed that the docs says it's deprecated and we should instead use `license_files` which defaults to `LICENSE*`.

meshula (approved):
lgtm, given their classification system

JeanChristopheMorinPerso:
@meshula Note that the license will be reported correctly. The note about the classifier is really just for classifiers. They are used to filter packages on PyPI. But our Apache 2.0 license will be reported correctly in the package Metadata and in the UI.

jminor (approved):
Thanks for catching this!

PR #1606, opened 2023-05-20
Fix rendering of otioz file bundle structure in the docs

### LINKED ISSUE #1604: Documentation formatting issues
Some sections of the OTIO documentation are not formatted properly. For example, there is a list of a file bundle contents here: https://opentimelineio.readthedocs.io/en/stable/tutorials/otio-filebundles.html which should be formatted as a nested tree-like structure, but instead appears as a flat list which is confusing.

JeanChristopheMorinPerso:
Fixes #1604.

The rendered version can be seen at https://opentimelineio--1606.org.readthedocs.build/en/1606/tutorials/otio-filebundles.html#structure.

### DIFF
--- a/docs/tutorials/otio-filebundles.md
+++ b/docs/tutorials/otio-filebundles.md
@@ -27,13 +27,15 @@ The file bundle adapters expect the `target_url` field of the `media_reference` 
 
 File bundles, regardless of how they're encoded, have a consistent structure:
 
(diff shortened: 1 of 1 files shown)


meshula (approved):
lgtm, would be good to get a double-check from @ssteinbach

jminor (approved):
Thanks!

PR #1613, opened 2023-06-01
Use IMATH_NAMESPACE in place of Imath

### LINKED ISSUE #1499: Trouble with internal and external Imath
## Bug Report

The circumstances under which I ran into problems with the Imath dependency might be corner cases but I think there are some aspects to discuss and improve.

OTIO has Imath as a submodule. This internal Imath can cause non-obvious conflicts. The conditions are that `OTIO_IMATH_LIBS` is not set by the user, and `OTIO_FIND_IMATH` is `OFF` (default) or set to `ON` but `find_package()` fails to find Imath/IlmBase. In these cases, OTIO falls back to the internal Imath. This can result in linker errors if you have an application that uses a different version of Imath together with OpenEXR in the same version. It's not obvious that OTIO needs Imath and that installing a build may clobber (parts of) an Imath system installation, causing a mismatch between the OpenEXR and Imath versions.

I ran into a linker error because I wasn't aware of the options to use an external Imath and built OTIO in a container without my own Imath. While trying to correct this, I found code paths that are broken or could at least use better usage instructions.

### Build Problem

My setup is as follows:
- I compile OTIO in its own container (C++ only, no Python)
- I assemble a CI image from ASWF Docker images (including OpenEXR and Imath), plus my own OTIO image
- I try to compile my software, but the linker throws an error because it uses something from the bleeding edge Imath namespace (`Imath_3_2`) instead of the namespace of my actual Imath (`Imath_3_1`), which is caused by OTIO:
  `` undefined reference to `Imf_3_1::Header::Header(int, int, float, Imath_3_2::Vec2<float> const&, float, Imf_3_1::LineOrder, Imf_3_1::Compression)' ``

To make OTIO use my Imath, I need to add it to the OTIO build container (as well as the CI container) and tell it to use an external Imath.
- I tried the `OTIO_IMATH_LIBS` option. Is this supposed to point to one or multiple .so files, or the folder the shared libraries are in? I used the following:
  `-DOTIO_IMATH_LIBS="/usr/local/lib64/libImath-3_1.so.29.4.0"`
- The `OTIO_IMATH_LIBS` option does get picked up:
  `-- Using Imath from OTIO_MATH_LIBS: /usr/local/lib64/libImath-3_1.so.29.4.0`
  However, the compilation fails early because it cannot find the Imath header files. There is a mention of an `IMATH_INCLUDES` variable further down in the CMake code, which I tried to set to the Imath include dir:
  `-DIMATH_INCLUDES="/usr/local/include/Imath"`
  But this gave me an error:
  ```
  CMake Error at tests/CMakeLists.txt:24 (target_include_directories):
  target_include_directories called with invalid arguments
  ```
- Using `-DOTIO_FIND_IMATH=ON` instead finds the CMake config of my Imath in `/usr/local/lib64/cmake/Imath/`, which finds the necessary files (I use a workaround to ignore the Python support for which I miss a CMake target in my setup)

## To Reproduce

Test setup: https://github.com/Simran-B/otio_imath

To hit the different code paths (A through E), do the following:

- (A) Run `docker build -t otio_imath --progress=plain --no-cache .` to build the Docker image and get the linker error with a test application. This uses the internal Imath (`-- Using src/deps/Imath by default`)

- (B) Uncomment the indicate lines (1 line + 2 lines) in `Dockerfile` to add my Imath to the OTIO image (and apply a Python target workaround).
  Comment out code path A in `build_otio.sh` and uncomment B to set `OTIO_IMATH_LIBS` (`-- Using Imath from OTIO_MATH_LIBS: /usr/local/lib64/libImath-3_1.so.29.4.0`), and run `docker build -t otio_imath --progress=plain --no-cache .`, leading to a compile error because of missing includes

- (C) Comment out code path B in `build_otio.sh` and uncomment C / D to set `OTIO_IMATH_LIBS` as well as `IMATH_INCLUDES`, and run `docker build -t otio_imath --progress=plain --no-cache .`, leading to a CMake error for the test target

- (D) Uncomment the patch command (1 line) in `build_otio.sh` and run `docker build -t otio_imath --progress=plain --no-cache .`, leading to a successful build

- (E) Comment out code path C / D in `build_otio.sh` and uncomment E, to set `OTIO_FIND_IMATH=ON` (`-- Found Imath 3 at /usr/local/lib64/cmake/Imath/ImathConfig.cmake`), and run `docker build -t otio_imath --progress=plain --no-cache .`, leading to a successful build. If you revert `Dockerfile` to code path A, then you get the linker error again because of the fallback to the internal Imath (`-- Imath 3 and 2 were not found, using src/deps/Imath`)

Compiler information:

```
> gcc -v
Using built-in specs.
COLLECT_GCC=gcc
COLLECT_LTO_WRAPPER=/opt/rh/devtoolset-9/root/usr/libexec/gcc/x86_64-redhat-linux/9/lto-wrapper
Target: x86_64-redhat-linux
Configured with: ../configure --enable-bootstrap --enable-languages=c,c++,fortran,lto --prefix=/opt/rh/devtoolset-9/root/usr --mandir=/opt/rh/devtoolset-9/root/usr/share/man --infodir=/opt/rh/devtoolset-9/root/usr/share/info --with-bugurl=http://bugzilla.redhat.com/bugzilla --enable-shared --enable-threads=posix --enable-checking=release --enable-multilib --with-system-zlib --enable-__cxa_atexit --disable-libunwind-exceptions --enable-gnu-unique-object --enable-linker-build-id --with-gcc-major-version-only --with-linker-hash-style=gnu --with-default-libstdcxx-abi=gcc4-compatible --enable-plugin --enable-initfini-array --with-isl=/builddir/build/BUILD/gcc-9.3.1-20200408/obj-x86_64-redhat-linux/isl-install --disable-libmpx --enable-gnu-indirect-function --with-tune=generic --with-arch_32=x86-64 --build=x86_64-redhat-linux
Thread model: posix
gcc version 9.3.1 20200408 (Red Hat 9.3.1-2) (GCC)
```

## Expected Behavior

After some experimentation, I think the following changes should be made:

- Fix the `target_include_directories()` call for the test target by adding the mandatory visibility flag (PRIVATE seems to work here, validated with `cmake --build . --target test`):
  ```diff
  diff --git a/tests/CMakeLists.txt b/tests/CMakeLists.txt
  index 57a5231..e355553 100644
  --- a/tests/CMakeLists.txt
  +++ b/tests/CMakeLists.txt
  @@ -21,7 +21,7 @@ foreach(test ${tests_opentimelineio})
  
       target_link_libraries(${test} opentimelineio)
       if (NOT "${IMATH_INCLUDES}" STREQUAL "")
  -        target_include_directories(${test} "${IMATH_INCLUDES}")
  +        target_include_directories(${test} PRIVATE "${IMATH_INCLUDES}")
       endif()
       set_target_properties(${test} PROPERTIES FOLDER tests)
       add_test(NAME ${test}
  ```

- The description of the `OTIO_MATH_LIBS` option should clarify what it should be set to, e.g. by showing an example

- The description should also point out that `IMATH_INCLUDES` (which is not an option but only a variable!) needs to be set in addition - it could be checked if it's empty and the CMake configure phase aborted early to improve the UX

- Not set `${IMATH_INCLUDES}` [twice](https://github.com/AcademySoftwareFoundation/OpenTimelineIO/blob/main/src/opentimelineio/CMakeLists.txt#L80-L86) in the `target_include_directories()` call of the opentimelineio target - or does this serve a purpose?

- If `OTIO_FIND_IMATH` is set to `ON`, should there really be a fallback to the internal Imath? Seems like a possible pitfall because a failing find_package() may go unnoticed.

- Should the internal Imath be opt-in after all? The out-of-box experience would potentially be worse, but it would make other scenarios easier.

- Fix a [typo](https://github.com/AcademySoftwareFoundation/OpenTimelineIO/blob/main/CMakeLists.txt#L221) `"Using Imath from OTIO_MATH_LIBS: ${OTIO_IMATH_LIBS}")`, missing an I in OTIO_MATH_LIBS

Further, I would like to understand whether `OTIO_RESOLVED_IMATH_LIBRARIES` is picked up CMake somehow or if it's an unused variable.

peter-targett:
Fixes #1499

Use the Imath namespace preprocessor macro `IMATH_NAMESPACE` in place of `Imath`. This allows an external-use Imath with a custom namespace to be used. This PR addresses at least one of the issues from bug report #1499.

Changed all the source code references of `Imath::` with `IMATH_NAMESPACE::`.

### DIFF
--- a/src/opentimelineio/clip.cpp
+++ b/src/opentimelineio/clip.cpp
@@ -184,7 +184,7 @@ Clip::available_range(ErrorStatus* error_status) const
     return active_media->available_range().value();
 }
 
-optional<Imath::Box2d>
+optional<IMATH_NAMESPACE::Box2d>
 Clip::available_image_bounds(ErrorStatus* error_status) const
 {
     auto active_media = media_reference();
@@ -194,7 +194,7 @@ Clip::available_image_bounds(ErrorStatus* error_status) const
             ErrorStatus::CANNOT_COMPUTE_BOUNDS,
             "No image bounds set on clip",
             this);
-        return optional<Imath::Box2d>();
+        return optional<IMATH_NAMESPACE::Box2d>();
     }
 
     if (!active_media->available_image_bounds())
@@ -203,7 +203,7 @@ Clip::available_image_bounds(ErrorStatus* error_status) const
             ErrorStatus::CANNOT_COMPUTE_BOUNDS,
             "No image bounds set on media reference on clip",
             this);
-        return optional<Imath::Box2d>();
+        return optional<IMATH_NAMESPACE::Box2d>();
     }
 
     return active_media->available_image_bounds();
--- a/src/opentimelineio/clip.h
+++ b/src/opentimelineio/clip.h
@@ -48,7 +48,7 @@ public:
     TimeRange
     available_range(ErrorStatus* error_status = nullptr) const override;
 
-    optional<Imath::Box2d>
+    optional<IMATH_NAMESPACE::Box2d>
     available_image_bounds(ErrorStatus* error_status) const override;
 
 protected:
--- a/src/opentimelineio/composable.cpp
+++ b/src/opentimelineio/composable.cpp
@@ -74,11 +74,11 @@ Composable::duration(ErrorStatus* error_status) const
     return RationalTime();
 }
 
-optional<Imath::Box2d>
+optional<IMATH_NAMESPACE::Box2d>
 Composable::available_image_bounds(ErrorStatus* error_status) const
 {
     *error_status = ErrorStatus::NOT_IMPLEMENTED;
-    return optional<Imath::Box2d>();
+    return optional<IMATH_NAMESPACE::Box2d>();
 }
 
 }} // namespace opentimelineio::OPENTIMELINEIO_VERSION
--- a/src/opentimelineio/composable.h
+++ b/src/opentimelineio/composable.h
@@ -34,7 +34,7 @@ public:
 
     virtual RationalTime duration(ErrorStatus* error_status = nullptr) const;
 
-    virtual optional<Imath::Box2d>
+    virtual optional<IMATH_NAMESPACE::Box2d>
     available_image_bounds(ErrorStatus* error_status) const;
 
 protected:
--- a/src/opentimelineio/deserialization.cpp
+++ b/src/opentimelineio/deserialization.cpp
@@ -609,14 +609,14 @@ SerializableObject::Reader::_decode(_Resolver& resolver)
     else if (schema_name_and_version == "V2d.1")
     {
         double x, y;
-        return _fetch("x", &x) && _fetch("y", &y) ? any(Imath::V2d(x, y))
+        return _fetch("x", &x) && _fetch("y", &y) ? any(IMATH_NAMESPACE::V2d(x, y))
                                                   : any();
     }
     else if (schema_name_and_version == "Box2d.1")
     {
-        Imath::V2d min, max;
+        IMATH_NAMESPACE::V2d min, max;
         return _fetch("min", &min) && _fetch("max", &max)
-                   ? any(Imath::Box2d(std::move(min), std::move(max)))
+                   ? any(IMATH_NAMESPACE::Box2d(std::move(min), std::move(max)))
                    : any();
     }
     else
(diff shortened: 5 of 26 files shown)


meshula (approved):
That's great, thanks for doing this

PR #1623, opened 2023-06-11
Add utility functions for comparison and rounding

### LINKED ISSUE #1622: Add utility functions for comparison and rounding
It would be nice to have a couple extra utility functions for RationalTime:
* A function that compares whether two RationalTimes have the same value and rate. This is different from the "==" operator which scales the time before comparison (e.g., ```otime::RationalTime(30.2) == otime::RationalTime(60.4, 2.0)```).
* Rounding functions that return the ```floor()```, ```ceil()```, and ```round()``` of a RationalTime.

darbyjohnston:
Fixes #1622

This PR adds a few utility functions to RationalTime for comparison and rounding:
* A function that compares whether two RationalTimes have the same value and rate. This is different from the "==" operator which scales the time before comparison (e.g., otime::RationalTime(30.2) == otime::RationalTime(60.4, 2.0)).
* Rounding functions that return the floor(), ceil(), and round() of a RationalTime.

If these seem useful I can also add Python bindings for them.

### DIFF
--- a/src/opentime/rationalTime.h
+++ b/src/opentime/rationalTime.h
@@ -87,20 +87,20 @@ public:
 
     // Return a RationalTime with the largest integer value not greater than
     // this value.
-    RationalTime floor_value() const
+    RationalTime floor() const
     {
         return RationalTime{ std::floor(_value), _rate };
     }
 
     // Return a RationalTime with the smallest integer value not less than
     // this value.
-    RationalTime ceil_value() const
+    RationalTime ceil() const
     {
         return RationalTime{ std::ceil(_value), _rate };
     }
 
     // Return a RationalTime with the nearest integer value to this value.
-    RationalTime round_value() const
+    RationalTime round() const
     {
         return RationalTime{ std::round(_value), _rate };
     }
--- a/src/opentime/rationalTime.h
+++ b/src/opentime/rationalTime.h
@@ -80,27 +80,27 @@ public:
     // Return whether the value and rate are equal to another RationalTime.
     // This is different from the operator "==" that rescales the time before
     // comparison.
-    constexpr bool exactly_equal(RationalTime other) const noexcept
+    constexpr bool strictly_equal(RationalTime other) const noexcept
     {
         return _value == other._value && _rate == other._rate;
     }
 
     // Return a RationalTime with the largest integer value not greater than
     // this value.
-    RationalTime floor() const
+    RationalTime floor_value() const
     {
         return RationalTime{ std::floor(_value), _rate };
     }
 
     // Return a RationalTime with the smallest integer value not less than
     // this value.
-    RationalTime ceil() const
+    RationalTime ceil_value() const
     {
         return RationalTime{ std::ceil(_value), _rate };
     }
 
     // Return a RationalTime with the nearest integer value to this value.
-    RationalTime round() const
+    RationalTime round_value() const
     {
         return RationalTime{ std::round(_value), _rate };
     }
--- a/src/opentime/rationalTime.h
+++ b/src/opentime/rationalTime.h
@@ -77,6 +77,34 @@ public:
         return fabs(value_rescaled_to(other._rate) - other._value) <= delta;
     }
 
+    // Return whether the value and rate are equal to another RationalTime.
+    // This is different from the operator "==" that rescales the time before
+    // comparison.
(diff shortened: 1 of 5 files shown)


src/opentime/rationalTime.h
```diff
@@ -77,6 +77,34 @@ class RationalTime
...
+    // This is different from the operator "==" that rescales the time before
+    // comparison.
+    constexpr bool exactly_equal(RationalTime other) const noexcept
+    {
+        return _value == other._value && _rate == other._rate;
+    }
+
+    // Return a RationalTime with the largest integer value not greater than
+    // this value.
+    RationalTime floor() const
```
meshula, line 90: I almost wonder, to make the method names more self-describing, if they might be named `floor_to_rate`, `ceil_to_rate`, and `round_to_rate`.

src/opentime/rationalTime.h
```diff
@@ -77,6 +77,34 @@ class RationalTime
         return fabs(value_rescaled_to(other._rate) - other._value) <= delta;
     }
 
+    // Return whether the value and rate are equal to another RationalTime.
+    // This is different from the operator "==" that rescales the time before
+    // comparison.
+    constexpr bool exactly_equal(RationalTime other) const noexcept
```
meshula, line 83: I think `exactly_equal` harkens to javascript's `===`. 

If I see a function called `exactly_equal` my mind thinks that this is a strict equality  check, that is not rate conformed, which is correct, of course. I would prefer `strictly_equal` though since `exactly_equal` doesn't tell me whether or not the values are conformed before the comparison, whereas strict equality does. I believe `strict equality` is the modern term of art.

eg. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Strict_equality

darbyjohnston: Good suggestion, I'll update the name.

src/opentime/rationalTime.h
darbyjohnston: The ```_to_rate``` sounds a bit confusing to me, maybe ```floor_value```, ```ceil_value```, and ```round_value```?

jminor: I think `floor_value` is better, since it tells me which field is changing.

meshula: I can go for that :)

meshula (approved):

meshula:
Looking good to me, maybe get a second approval :)

src/opentime/rationalTime.h
```diff
@@ -263,11 +291,6 @@ class RationalTime
     static RationalTime     _invalid_time;
     static constexpr double _invalid_rate = -1;
 
-    RationalTime _floor() const noexcept
-    {
-        return RationalTime{ floor(_value), _rate };
-    }
-
```
apetrynet, line 270: Please excuse my lack of C++ foo. 
Is a method prefixed with an "_" meant to be "private" or not encouraged for use outside the class in C++?
I'm asking as replacing this could potentially cause some issues for people using it?

meshula: The underbar means "private" although it's not a general C++ convention. It's common in film, on linux/mac. @darbyjohnston I didn't think about the name when I approved, just the removal of the underbar. Is there a reason to add _value() to all the functions? floor() ceil() and round() are class methods so they won't conflict with the standard library.

meshula: Ah, I see the conversation. For the record, @jminor suggested them, and I said that sounded fine to me :) Josh's reasoning was that floor is taking the floor of the value. But now that time has passed, and I read it without context, value is such a generic term, that it is no longer intuitive to me that floor_value() means take the floor of the value member. I'll defer to Josh on this one ~ I'm sure I will be confused a couple more times until it sticks in my brain.

darbyjohnston: I think that's a point in favor of the shorter names, when you looked at it fresh that's what you expected to see. I personally think the "_value" is implicit and if we need additional rounding functions they can have more explicit names:
```
RationalTime floor() const;
RationalTime floor_rate() const;
RationalTime floor_value_and_rate() const;
```

darbyjohnston: I reverted back to the shorter names, hopefully it reads better and I think operating on the value is what people would expect.

meshula:
Thanks Darby!

PR #1627, opened 2023-06-12
Add support for ImageSequenceReference to otioz and otiod adapters

### LINKED ISSUE #1621: Add support for image sequences when creating .otioz files with otioconvert
It looks like image sequences are not supported when creating .otioz files with otioconvert? I tried converting an .otio file with an image sequence and it looks like the images were not bundled:
```
> zipinfo SingleClipSeq.otioz
Archive:  SingleClipSeq.otioz
Zip file size: 819 bytes, number of entries: 2
?rw-------  2.0 unx        5 b- defN 23-Jun-10 16:45 version.txt
?rw-------  2.0 unx     3185 b- defN 23-Jun-10 16:45 content.otio
2 files, 3190 bytes uncompressed, 599 bytes compressed:  81.2%
```

Looking at the code in ```file_bundile_utils.py``` there is this comment:
```
try:
    target_url = cl.media_reference.target_url
except AttributeError:
    # not an ExternalReference, ignoring it.
    continue
```

It would be nice to add support for image sequences or maybe add a warning that they are being ignored.

darbyjohnston:
Fixes #1621

This PR adds support for ImageSequenceReference media references to the otioz and otiod adapters. It should also add support for multiple media references but I haven't tested that yet.

The main changes are to the function ```_prepped_otio_for_bundle_and_manifest``` which is shared by both adapters. Previously the function returned a map of media paths and associated media references. The media paths were used to locate the media files and the media references were used to relink the resulting .otio file. Since image sequence references can produce multiple media paths (render.0001.exr, render.0002.exr, etc.) that mapping didn't seem to make sense any more so I changed the function to just return the list of media files. The relinking is now handled as a separate step that iterates through the media references on its own.

I have tested converting some simple .otio files with both movies and image sequences to .otioz and .otiod and have verified the results look OK. Most of the unit tests look like they are passing OK though I still need to add tests for the new functionality.

### DIFF
--- a/src/py-opentimelineio/opentimelineio/adapters/file_bundle_utils.py
+++ b/src/py-opentimelineio/opentimelineio/adapters/file_bundle_utils.py
@@ -77,10 +77,9 @@ def _prepped_otio_for_bundle_and_manifest(
 ):
     """ Create a new OTIO based on input_otio that has had media references
     replaced according to the media_policy.  Return that new OTIO and a
-    mapping of all the absolute file paths (not URLs) to be used in the bundle,
-    mapped to MediaReferences associated with those files.  Media references in
-    the OTIO will be relinked by the adapters to point to their output
-    locations.
+    list of all the absolute file paths (not URLs) to be used in the bundle.
+    Media references in the OTIO will be relinked by the adapters to point to
+    their output locations.
 
     The otio[dz] adapters use this function to do further relinking and build
     their bundles.
@@ -91,7 +90,7 @@ def _prepped_otio_for_bundle_and_manifest(
     # make sure the incoming OTIO isn't edited
     result_otio = copy.deepcopy(input_otio)
 
-    path_to_reference_map = {}
+    paths = set()
     invalid_files = set()
 
     # result_otio is manipulated in place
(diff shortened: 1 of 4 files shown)


src/py-opentimelineio/opentimelineio/adapters/file_bundle_utils.py
```diff
@@ -103,68 +102,85 @@ def _prepped_otio_for_bundle_and_manifest(
...
+            parsed_url = urlparse.urlparse(url)
+
+            # ensure that the urlscheme is either file or ""
+            # file means "absolute path"
+            # none is interpreted as a relative path, relative to cwd
+            if parsed_url.scheme not in ("file", ""):
+                if media_policy is MediaReferencePolicy.ErrorIfNotFile:
+                    raise NotAFileOnDisk(
+                        "The {} adapter only works with file URLs."
+                        " Got a URL of:  '{}'".format(adapter_name, url)
```
douglascomet, line 125: Minor note, update to `f-string`

```suggestion
                        f"The {adapter_name} adapter only works with file URLs."
                        f" Got a URL of:  '{url}'"
```

darbyjohnston:
Closing this PR, since there doesn't seem to be any interest (opened > 15 months ago).

Issue #1192, opened 2022-01-14
Q: CMake config files and interdependencies


foundry-markf:
Hi,

OTIO 0.14.0

I got approached with a question today, because in CMake, someone had done

```
find_package(OpenTimelineIO REQUIRED)
```

resulting in the error
```
CMake Error at CMakeLists.txt:326 (find_package):
  Found package configuration file:

    <some local path>/share/opentimelineio/OpenTimelineIOConfig.cmake

  but it set OpenTimelineIO_FOUND to FALSE so package "OpenTimelineIO" is
  considered to be NOT FOUND.  Reason given by package:

  The following imported targets are referenced, but are missing:
  OTIO::opentime
```
due to the `opentimelineio` import target requiring `opentime`.

I hadn't seen this in my own testing, because I had done 
```
find_package(OpenTime REQUIRED)
find_package(OpenTimelineIO REQUIRED)
```
(although I can't find that in documentation from a quick browse, so I must've just figured it out at the time.)

It's brought up the subsequent question of why are there _two_ different CMake config files exported during the build?

I'm not aware of any technical restrictions of exporting two libraries in a single config file, even if you only consume one.

I'm curious if there was a particular need for them to be separate?
Thanks

meshula:
The thought was that OpenTime might be more broadly interesting.

The OpenTime config file could stay as is, but it would make sense to add the OpenTime config information to the OpenTimelineIO config as well.

I think it's a reasonable user expectation that finding OpenTimelineIO itself is sufficient.

Are you able to prepare a PR for this?

foundry-markf:
> The OpenTime config file could stay as is, but it would make sense to add the OpenTime config information to the OpenTimelineIO config as well.

As far as I'm aware, in CMake, you can only export targets to a single config file.

> Are you able to prepare a PR for this?

Not at the present time. I asked management prior to Christmas about CLAs, but breaks/holidays/catching up has meant no progress. I can query again though since I've opened a few issue reports.


If I could though, I would recommend a single CMake config file, with all targets in it.
There is the [component ](https://cmake.org/cmake/help/git-stage/guide/importing-exporting/index.html#adding-components) concept in CMake packages, so users could request different parts, e.g.

```
find_package(OpenTimelineIO REQUIRED COMPONENTS OpenTime)
```
I'm spitballing this idea - I've not actually tried it, or have a good inkling for how it would feel to consumers.

jminor:
[...]

PR #1640, opened 2023-07-06
CMake config file improvements to find dependencies automatically for consumers

foundry-markf:
Proposed fixes for https://github.com/AcademySoftwareFoundation/OpenTimelineIO/issues/1192

**Summarize your change.**

In terms of CMake target resolution in link steps, just finding OpenTimelineIO was insufficient, as it depends on OpenTime, so consumers needed to know this detail. It's preferrable that consumers just find packages they are interested in, without knowledge of implementation details of build dependencies.

This change moves to using a canonical form of CMake config files, whereby the `*Config.cmake` is generated from a template allowing dependencies to be found, while the CMake generation of import targets is renamed to a `*Targets.cmake` file. The `*Config.cmake` file includes the `*Target.cmake` file.

See https://cmake.org/cmake/help/latest/manual/cmake-packages.7.html#creating-packages for an example.

There are a number of ways to solve the problem, but this was the least invasive.

This change adds 'consumer' C++ tests, which runs the CMake configuration step, which is sufficient to determine if all necessary CMake targets are present.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -39,6 +39,7 @@ jobs:
       OTIO_BUILD_CONFIG: Release
       OTIO_BUILD_DIR: ${{ github.workspace }}/build
       OTIO_INSTALL_DIR: ${{ github.workspace }}/install
+      OTIO_CONSUMER_TEST_BUILD_DIR: ${{ github.workspace }}/consumertest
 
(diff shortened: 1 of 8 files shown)


meshula (approved):
Thanks for the contribution, this looks good to me.

PR #1642, opened 2023-07-22
Added otiotool --remove-metadata-key option

jminor:
This PR adds a new `--remove-metadata-key` feature to `otiotool`. If specified, this option looks for the given metadata key(s) on all objects in the OTIO and removes it when found. This could be used to remove private, sensitive, or extraneous metadata from an OTIO.

```
  --remove-metadata-key KEY(s) [KEY(s) ...]
                        Remove one or more metadata keys from all objects.
```

This PR also improves the `otiotool` usage docs, adding some examples, clarifying some arguments, and making the short summary more readable by using the `--long-options` instead of single-letter options.

```
usage: otiotool [-h] [--input PATHs) [PATH(s) ...]] [--video-only] [--audio-only] [--only-tracks-with-name NAME(s) [NAME(s) ...]]
                [--only-tracks-with-index INDEX(es) [INDEX(es) ...]] [--only-clips-with-name NAME(s) [NAME(s) ...]]
                [--only-clips-with-name-regex REGEX(es) [REGEX(es) ...]] [--remove-transitions] [--trim START END] [--flatten TYPE] [--keep-flattened-tracks]
                [--stack] [--concat] [--relink-by-name FOLDER(s) [FOLDER(s) ...]] [--copy-media-to-folder FOLDER] [--remove-metadata-key KEY(s) [KEY(s) ...]]
                [--redact] [--stats] [--list-clips] [--list-tracks] [--list-media] [--verify-media] [--list-markers] [--inspect NAME(s) [NAME(s ...]]
                [--downgrade FAMILY:VERSION] [--list-versions] [--output PATH]
```

### DIFF
--- a/src/py-opentimelineio/opentimelineio/console/otiotool.py
+++ b/src/py-opentimelineio/opentimelineio/console/otiotool.py
@@ -112,7 +112,12 @@ def main():
         for timeline in timelines:
             copy_media_to_folder(timeline, args.copy_media_to_folder)
 
-    # Phase 6: Redaction
+    # Phase 6: Remove/Redaction
+
+    if args.remove_metadata_key:
+        for timeline in timelines:
+            for key in args.remove_metadata_key:
+                remove_metadata_key(timeline, key)
 
     if args.redact:
         for timeline in timelines:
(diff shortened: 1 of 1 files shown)


apetrynet (commented):
If I read this correctly it will only remove a key from the "top level" of a metadata dict. 
I believe this will suffice in most cases, but we should perhaps mention this in the docs?

jminor:
@apetrynet I addressed your note about "top level" in the usage statement. Does that look good to you?

apetrynet (approved):
Looks good to me @jminor
I also like the more verbose help strings in the other arguments.

meshula (approved):

PR #1539, opened 2023-02-05
Support for C++17

darbyjohnston:
Fixes #1526

Changes:

* Set the C++ version to 17 and CMAKE_OSX_DEPLOYMENT_TARGET to 10.14 in the top-level CMakelist.txt file
* Replaced any and optional dependencies with versions from the standard library
* Removed the any and optional dependencies

(This PR replaces https://github.com/AcademySoftwareFoundation/OpenTimelineIO/pull/1530)

PR #1646, opened 2023-08-09
Add C++17 guard to casters.h

ssteinbach:
When building with C++17 turned ON, our `optional` and `any` libraries, `nonstd::optional` and `nonstd::any`, are replaced with the ones present in `std` in C++17 on.  Because Pybind11 already defines casters for the `std::optional`, this causes a build error, duplicate definition, with the definition found in casters.h.  This PR guards the definition in casters.h and adds to the comment describing why its guarded and when casters.h can be removed.

### DIFF
--- a/src/py-opentimelineio/bindings-common/casters.h
+++ b/src/py-opentimelineio/bindings-common/casters.h
@@ -15,6 +15,8 @@ namespace pybind11 { namespace detail {
     // becomes std::optional starting in C++17, and Pybind11 already defines
     // casters for std::optional, this becomes a duplicate (and therefore build
     // error).
+    // The casters can be removed when C++ versions less than 17 are no longer
+    // supported by OTIO.
     template<typename T> struct type_caster<optional<T>>
         : public optional_caster<optional<T>> {};
 
--- a/src/py-opentimelineio/bindings-common/casters.h
+++ b/src/py-opentimelineio/bindings-common/casters.h
@@ -5,14 +5,20 @@
 using nonstd::optional;
 using nonstd::nullopt_t;
 
+#if __cplusplus < 201703L
 namespace pybind11 { namespace detail {
     // Caster for converting to/from nonstd::optional.
     // Pybind11 supports optional types when C++17 is used.
-    // So for now we have to create the casters manually.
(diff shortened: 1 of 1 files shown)


JeanChristopheMorinPerso (approved):
LGTM. I'm surprised we didn't catch that before!

ssteinbach:
I don't think we have any C++17 builds in our CI?  which we should also add... I can file an issue for that but I won't have time to do that today

darbyjohnston:
How far off is the next release? We were hoping to merge the "c++17" branch shortly after which will enable C++17 support and removes "casters.h":
https://github.com/AcademySoftwareFoundation/OpenTimelineIO/pull/1539

ssteinbach:
Ah yeah.  So @darbyjohnston does this still apply in that case?  Or should we skip this entirely.

darbyjohnston:
I'm not sure, I was hoping the new release would be out by now and we could merge the "c++17" branch which would take care of "casters.h" and enable C++17 for the CI. But if you need C++17 support now then this seems like a reasonable change.

> When building with C++17 turned ON, our optional and any libraries, nonstd::optional and nonstd::any, are replaced with the ones present in std 

How does this work, don't the includes and namespaces need to be updated as well to use the std:: versions?

ssteinbach:
I think the trick is that the any and optional libraries get swapped out.  And since internally we hide them behind otio::optional and otio::any, if you're referring those types, they'll become std too.  

IE: 
https://github.com/martinmoene/optional-lite/blob/be720ebfd7add22abe60e49602735de9231105d0/include/nonstd/optional.hpp#L201

And yeah, this enables C++17 builds with pre landing the C++17 build.

darbyjohnston:
Ah, that's confusing since it looks like they are still being used. Hopefully at some point we can fully migrate to C++17 and remove the dependencies.

It looks like older versions of Visual Studio on Windows didn't support the `__cplusplus` macro, though the CI builds passed so it seems to be fixed in Visual Studio 2022 at least:
https://devblogs.microsoft.com/cppblog/msvc-now-correctly-reports-__cplusplus/
https://gitlab.kitware.com/cmake/cmake/-/issues/18837

Issue #1652, opened 2023-08-20
find_children is broken with two tracks.


ggarra13:
### Incorrect Functionality and General Questions

C++'s timeline->find_children<otio::Item> is broken when there are two tracks and a search range is involved.

## To Reproduce

I created a PR for a test for it to show both the correct and incorrect functionality.  I am also attaching the test here as .txt file.

[test_find_children.cpp.txt](https://github.com/AcademySoftwareFoundation/OpenTimelineIO/files/12388154/test_find_children.cpp.txt)

## Expected Behavior

timeline->find_children<otio::Item>() should return all otio::Items in the range.

darbyjohnston:
Thanks for creating this test! I created a new PR (#1766) that has a potential fix and and the test, so I'll close this one.

ggarra13:
Darby, would you be a darling and test Issue #1750? I've been wanting to attend the OTIO meetings and for one reason or another I cannot make it.
Issue #1750 is an incorrect functionality between M1 and M3 machines, which seems like a macOS compiler issue.  I don't have any M1/M3 machine but I would like to know if the test works fine on an M1 machine at least.

PR #1653, opened 2023-08-20
Find children broken

ggarra13:

This is a test for issue #1652.  It does not fix it.  It just adds a test for it.

**Summarize your change.**

Added a test_find_children.cpp that shows the problem.

test_find_children.cpp.


jminor:
It looks like this was covered by https://github.com/AcademySoftwareFoundation/OpenTimelineIO/pull/1766
Feel free to reopen if I've misunderstood.

PR #1651, opened 2023-08-18
Editing commands

ggarra13:
This commit is for:

Implements https://github.com/AcademySoftwareFoundation/OpenTimelineIO/issues/1649.
Makes PR https://github.com/AcademySoftwareFoundation/OpenTimelineIO/pull/1518 obsolete.

It implements edit commands in the otio::algo:: namespace. The commands implemented are those discussed in https://github.com/AcademySoftwareFoundation/OpenTimelineIO/issues/711.

They are further explained more clearly in:
https://github.com/mccartnm/OpenTimelineIO/blob/openedit_design/docs/design/editorial_design.md

Summarize your change.

It implements the needed base commands for editing with otio, building on top of @darbyjohnston's PR.

@darbyjohnston 's PR https://github.com/AcademySoftwareFoundation/OpenTimelineIO/pull/1518 was the base for it, but I debugged two subtle precision bugs in his code and moved it all to the algo:: namespace.

Implementation also modifies errorStatus.h/cpp to add NOT_A_GAP enum.

Implementation, as @darbyjohnston's, makes Composition's protected _index_of_child a now public member of the API as index_of_child and updates track.cpp accordingly.

Reference associated tests.

test_editAlgorithm.cpp is added as in @darbyjohnston, but with tests for all commands.

PR #1650, opened 2023-08-17
Edit commands

ggarra13:
Signed-off-by: Gonzalo Garramuño [ggarra13@gmail.com](mailto:ggarra13@gmail.com)

This commit is for:

Implements #1649.
Makes PR #1518 obsolete.

It implements edit commands in the otio::algo:: namespace.  The commands implemented are those discussed in #711.

They are further explained more clearly in:
https://github.com/mccartnm/OpenTimelineIO/blob/openedit_design/docs/design/editorial_design.md

**Summarize your change.**

It implements the needed base commands for editing with otio, building on top of @darbyjohnston's PR.

@darbyjohnston 's PR #1518 was the base for it, but I debugged two subtle precision bugs in his code and moved it all to the algo:: namespace.

Implementation also modifies errorStatus.h/cpp to add NOT_A_GAP enum.

Implementation, as @darbyjohnston's, makes Composition's protected _index_of_child a now public member of the API as index_of_child and updates track.cpp accordingly.

**Reference associated tests.**

test_editAlgorithm.cpp is added as in @darbyjohnston, but with tests for all commands.

Issue #1649, opened 2023-08-16
OTIO edit commands


ggarra13:
## Feature Request

This is a feature request to add editing commands based on #711 and experimental PR #1518.

## Description

I am currently working on adding all the commands described in #711, without Undo/Redo capabilities and, as a first step, without taking into account transitions.  I am currently adding them to a new namespace (edit), instead of algo as it was discussed.

## Context

I currently have (with unit tests):  slice, overwrite, insert, slip, slide and (without checking available_range) trim.
I am still missing: finishing trim, and work on ripple, roll and fill.

timlehr:
Seems like this one is fully done now @reinecke.

Issue #711, opened 2020-05-19
Common Edit Command Library


mccartnm:
### Overview
The industry is full of timeline management software but nearly all of them include a fundamental set of tools for manipulation. The most well known being:

- Overwrite
- Trim
- Cut
- Slip
- Slide
- Ripple
- Roll
- Fill (3/4 Point Edit)

OpenTimelineIO, while not seeking to become a fully realized NLE, could benefit from having a simple toolkit to help automate building/modifying timelines based on time rather than index alone.

### Proposal
For editing commands, we should strive to do all validation and assert that everything works before committing anything on the timeline. I'm proposing the `EditEvent` which might something like:
```cpp
class EditEvent {
public:
    EditEventKind kind; // e.g. Insert, Append, Remove, Modify
    Retainer<Composition> parent;
    Retainer<Composable> composable;

    // ... Additional fields to execute the above as required

    bool run(ErrorStatus* error_status);
    bool revert();
};
```
An event is atomic in nature and does "one thing" to a `Composable`. Each edit maneuver (e.g. `otio.edit.place(...)`) would generate a vector of these events that can be played forward and, possibly, backward. The result of them collectively is the commands result.

```cpp
for (EditEvent &event: events) {
    event.run(error_status);
    if (error_status) {
        for (auto it = completed_events.rbegin(); /*...*/)
            completed.revert();
        break;
    }
    completed_events.push_back(event);
}
```
Alternatively, we could build a more robust transaction system since all items can be cloned but there are memory considerations for that and beyond the scope of this initial proposal.

### API
The API might look like:
```
otio.edit.place(item, into=track, at=rat_time, placement_kind=otio.edit.PlacementKind)
otio.edit.trim(item, edge=otio.edit.ItemEdge)
etc.
```
Additional commands can be added as the editing foundation is built.

### Pros:
- Atomic events make reuse simple and should scale well with the complexity of edit commands
- Hooks up nicely to an undo/redo framework on third party software

### Cons:
- Additional code to handle event care

### Questions/Considerations:
- Could we handle `preview` flags for edit commands? This would let third party tools inspect the outcome of events before committing them.
- Should `EditEvent` be a class system (e.g. `InsertEvent`, `ModifyEvent`) or something else entirely?

mccartnm:
[...]

PR #1654, opened 2023-08-20
Experimental Editing Commands

ggarra13:
This commit is for:

Implements https://github.com/AcademySoftwareFoundation/OpenTimelineIO/issues/1649.
Makes PR https://github.com/AcademySoftwareFoundation/OpenTimelineIO/pull/1518,  PR #1650 and PR #1651 obsolete.

It implements edit commands in the otio::algo:: namespace. The commands implemented are those discussed in https://github.com/AcademySoftwareFoundation/OpenTimelineIO/issues/711.

They are further explained more clearly in:
https://github.com/mccartnm/OpenTimelineIO/blob/openedit_design/docs/design/editorial_design.md

Summarize your change.

It implements the needed base commands for editing with otio, building on top of @darbyjohnston's PR.

@darbyjohnston 's PR https://github.com/AcademySoftwareFoundation/OpenTimelineIO/pull/1518 was the base for it, but I debugged two subtle precision bugs in his code and moved it all to the algo:: namespace.

Implementation also modifies errorStatus.h/cpp to add NOT_A_GAP enum.

Implementation, as @darbyjohnston's, makes Composition's protected _index_of_child a now public member of the API as index_of_child and updates track.cpp accordingly.

Reference associated tests.

test_editAlgorithm.cpp is added as in @darbyjohnston, but with tests for all commands.    

I have now tested the commands on an actual application.  The commands overwrite, insert, slice have been tested on mrv2.

Signed-off-by: Gonzalo Garramuño <ggarra13@gmail.com>


src/opentimelineio/algo/editAlgorithm.h
```diff
@@ -0,0 +1,188 @@
...
+    Fit
+};
+
+// Overwrite an item or items.
+//
+// | A | B |  ->  |A| C |B|
+//   ^   ^
+//   | C |
+//
+//          item = item to overwrite (usually a clip)
```
meshula, line 24: I think this is C in the diagram, could you make a note?

Could you write some notes about the corner cases?

C starts or ends before A
C starts or ends after B

src/opentimelineio/algo/editAlgorithm.h
```diff
@@ -0,0 +1,188 @@
...
+    bool             remove_transitions = true,
+    Item*            fill_template      = nullptr,
+    ErrorStatus*     error_status       = nullptr);
+
+// Insert an item.
+// |     A     | B |  ->  | A | C | A | B |
+//       ^
+//     | C |
+//
+//          item = item to insert (usually a clip)
```
meshula, line 43: I assume that if length of AB is L1, and length of C is L2, then length of the result is L1+L2? In other words A is split?

src/opentimelineio/algo/editAlgorithm.h
```diff
@@ -0,0 +1,188 @@
...
+    bool const          remove_transitions = true,
+    Item*               fill_template      = nullptr,
+    ErrorStatus*        error_status       = nullptr);
+
+//
+// Adjust a single item's start time or duration.
+//
+// |    A    | B | C |  ->  |  A  |FILL| B | C |
+//        <--*
+//
```
meshula, line 63: If FILL is nullptr, then is the result of the trim

| A | B | C |

in other words,  B and C slip to the left?

src/opentimelineio/algo/editAlgorithm.h
```diff
@@ -0,0 +1,188 @@
...
+    RationalTime const& time,
+    bool const          remove_transitions = true,
+    ErrorStatus*        error_status       = nullptr);
+
+//
+// Slip an item start_time by + or -, clamping to available_range if available.
+//
+// |   A   |
+//  <----->
+//
```
meshula, line 95: how is the slip accomplished? what happens if the delta results in negative values?

src/opentimelineio/algo/editAlgorithm.h
```diff
@@ -0,0 +1,188 @@
...
+//          item = item to slip (usually a clip)
+//         delta = +/- rational time to slip the item by.
+void slip(Item* item, RationalTime const& delta);
+
+//
+// Slide an item start_time by + or -, moving the previous item's duration.
```
meshula, line 101:
```suggestion
// Slide an item start_time by + or -, adjusting the previous item's duration.
```

src/opentimelineio/algo/editAlgorithm.h
```diff
@@ -0,0 +1,188 @@
...
+// Clamps previous item's duration to available_range if available.
+//
```
meshula, line 103: what if there is no previous item?

src/opentimelineio/algo/editAlgorithm.h
```diff
@@ -0,0 +1,188 @@
...
+// Adjust a source_range without affecting any other items.
+//
+// |   A   |   B   |  ->  | A |  B  |FILL|
+//      <--*
+//            
+//      item = Item to apply ripple to (usually a clip)
+//  delta_in = RationalTime that the item's source_range().start_time()
+//             will be adjusted by
+// delta_out = RationalTime that the item's
+//             source_range().end_time_exclusive() will be adjusted by
```
meshula, line 121: what happens if the adjustment results in an element with zero or negative duration?

src/opentimelineio/algo/editAlgorithm.h
```diff
@@ -0,0 +1,188 @@
...
+// End time in parent of Item after input item will never change
+//
+// |   A   |   B   |  ->  | A |  B      |
+//      <--*
+//            
+//      item = Item to apply roll to (usually a clip)
+//  delta_in = RationalTime that the item's source_range().start_time()
+//             will be adjusted by
+// delta_out = RationalTime that the item's
+//             source_range().end_time_exclusive() will be adjusted by
```
meshula, line 143: same question if result is a zero or negative length item?

src/opentimelineio/algo/editAlgorithm.h
```diff
@@ -0,0 +1,188 @@
...
+//            
+// | A | C | B |  ->  | A |GAP| B |
+//       ^
+//       |
+//
+//           track = track to remove item from.
+//            time = RationalTime
+//            fill = whether to fill the hole with fill_template.
+//   fill_template = if nullptr, use a gap to fill the hole.
+//
```
meshula, line 180: If GAP is not supplied, does this become a ripple?

meshula (commented):
This is fantastic. I have some questions that I am hoping you can answer as new comment documentation, where I have asked the questions? 

Someone can discover the answers to the questions by reading the code, but I think it would be good to have the explanation in the header for easy reference.

src/opentimelineio/algo/editAlgorithm.h
ggarra13: No.  It always creates a gap.  Clarified in new comments.

src/opentimelineio/algo/editAlgorithm.h
ggarra13: They are accepted as long as there's no available range or the available range supports the negative values.

src/opentimelineio/algo/editAlgorithm.h
ggarra13: Changed.

src/opentimelineio/algo/editAlgorithm.h
ggarra13: It is clamped.

meshula (approved):
thanks for updating the docs

PR #1664, opened 2023-09-24
Adds support to `filepath_from_url` to support non-encoded URLs

douglascomet:
`filepath_from_url` works for encoded URLs but not for URLs that are not encoded. I encountered this when working with URLs from both `.xml` and `.aaf` files. 

The `.xml` file came from Premiere and encoded the URL while the `.aaf` came from Avid and the URL was not encoded.

Examples:

- `"file://localhost/S%3a/path/file.ext"` -> `"S:/path/file.ext"`
- `"file://S:/path/file.ext"` -> `"/path/file.ext"`

My PR updates the code to return the same result with both of the example URLs:

- `"file://localhost/S%3a/path/file.ext"` -> `"S:/path/file.ext"`
- `"file://S:/path/file.ext"` -> `"S:/path/file.ext"`

I tested and dev-ed this on Windows with OTIO `0.15.0` and `0.16.0.dev1`.

### DIFF
--- a/src/py-opentimelineio/opentimelineio/url_utils.py
+++ b/src/py-opentimelineio/opentimelineio/url_utils.py
@@ -9,8 +9,10 @@ from urllib import (
     parse as urlparse,
     request
 )
-import pathlib
-from pathlib import Path, PureWindowsPath
+from pathlib import (
+    Path,
+    PureWindowsPath
+)
 
 
 def url_from_filepath(fpath):
@@ -19,7 +21,7 @@ def url_from_filepath(fpath):
     try:
         # appears to handle absolute windows paths better, which are absolute
         # and start with a drive letter.
-        return urlparse.unquote(pathlib.Path(fpath).as_uri())
+        return urlparse.unquote(Path(fpath).as_uri())
     except ValueError:
         # scheme is "file" for absolute paths, else ""
         scheme = "file" if os.path.isabs(fpath) else ""
--- a/src/py-opentimelineio/opentimelineio/url_utils.py
+++ b/src/py-opentimelineio/opentimelineio/url_utils.py
@@ -4,7 +4,6 @@
 """Utilities for conversion between urls and file paths"""
 
 import os
-import re
 
 from urllib import (
     parse as urlparse,
--- a/src/py-opentimelineio/opentimelineio/url_utils.py
+++ b/src/py-opentimelineio/opentimelineio/url_utils.py
@@ -4,6 +4,7 @@
 """Utilities for conversion between urls and file paths"""
 
 import os
+import re
 
 from urllib import (
     parse as urlparse,
@@ -52,7 +53,7 @@ def filepath_from_url(urlstr):
     filepath = filepath.replace("\\", "/")
 
     # If on Windows, remove the first leading slash left by urlparse
-    if os.name == 'nt' and filepath.startswith('/'):
+    if filepath.startswith('/') and re.match(r"/[a-zA-Z]:/.*", filepath):
         filepath = filepath[1:]
 
     return filepath
--- a/src/py-opentimelineio/opentimelineio/url_utils.py
+++ b/src/py-opentimelineio/opentimelineio/url_utils.py
@@ -49,6 +49,8 @@ def filepath_from_url(urlstr):
     else:
         filepath = parsed_result.netloc + parsed_result.path
 
+    filepath = filepath.replace("\\", "/")
+
     # If on Windows, remove the first leading slash left by urlparse
     if os.name == 'nt' and filepath.startswith('/'):
         filepath = filepath[1:]
--- a/src/py-opentimelineio/opentimelineio/url_utils.py
+++ b/src/py-opentimelineio/opentimelineio/url_utils.py
@@ -62,8 +62,9 @@ def filepath_from_url(urlstr):
 
     filepath = filepath.replace("\\", "/")
 
-    # If on Windows, remove the first leading slash left by urlparse
-    if filepath.startswith('/') and re.match(r"/[a-zA-Z]:/.*", filepath):
+    # If on Windows and using a drive letter,
+    # remove the first leading slash left by urlparse
+    if re.match(r"/[a-zA-Z]:/.*", filepath):
         filepath = filepath[1:]
 
     return filepath
--- a/src/py-opentimelineio/opentimelineio/url_utils.py
+++ b/src/py-opentimelineio/opentimelineio/url_utils.py
@@ -11,6 +11,7 @@ from urllib import (
     request
 )
 import pathlib
+from pathlib import Path, PureWindowsPath
 
 
 def url_from_filepath(fpath):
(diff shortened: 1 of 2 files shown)


meshula (approved):

meshula:
Looking at the documentation (https://docs.python.org/3/library/urllib.request.html) , and searching error reports, it seems pretty clear that the urllib has not had a lot of attention around aspects such as drive letters and windows specific encodings. It's disappointing that we need to do this, but I think it's necessary.

Have you filed any issues with urllib around this, as the problem does seem to lie at their end. We should not gate merging this PR on such resolution, but it seems like something needs to happen upstream for the benefit of all.

tests/test_url_conversions.py
```diff
@@ -51,6 +52,17 @@ def test_roundtrip_rel(self):
         # should have reconstructed it by this point
         self.assertEqual(os.path.normpath(result), MEDIA_EXAMPLE_PATH_REL)
 
+    def test_premiere_xml_urls(self):
```
apetrynet, line 55: I was just wondering if it would be possible to test this without introducing the xml file?
Could we pass strings instead?
The reason I'm asking is that we'll be removing all those sample files as a part of the adapter extraction.

apetrynet:
Now that we've left python 2 behind I think we could start using pathlib for some of this as well.

douglascomet:
> Have you filed any issues with urllib around this, as the problem does seem to lie at their end. We should not gate merging this PR on such resolution, but it seems like something needs to happen upstream for the benefit of all.

@meshula, now that you mention it you're right this is a `urllib` issue and not OTIO's`. The motivation for this PR was to offload the responsibility of maintaining this code in a custom hook I made and help future developers looking to use OTIO on WIndows. Incorporating this into OTIO seems like the quicker route for now.

I looked through the [cpython](https://github.com/python/cpython) repo and found these open tickets that seem to relate to this very issue:

- https://github.com/python/cpython/issues/90812, which led to this proposal: https://discuss.python.org/t/file-uris-in-python/15600
- https://github.com/python/cpython/issues/85168

Since this issue is known on the `cypthon` that is still being debated I'm not sure if there is any benefit opening a new issue that is similar to existing ones.

douglascomet:
> Now that we've left python 2 behind I think we could start using pathlib for some of this as well.

@apetrynet, what were you thinking about using `pathlib` here? I also found this closed PR about generating `Path` from a URL, https://github.com/python/cpython/issues/91504.

tests/test_url_conversions.py
douglascomet: Sure, could totally switch to using strings for unit tests! That's how I tested this code anyway, by copying examples from `.xml` and `.aaf` files to figure out how the differences in approach.

src/py-opentimelineio/opentimelineio/url_utils.py
```diff
@@ -40,4 +40,19 @@ def filepath_from_url(urlstr):
...
+
+    # Otherwise, combine the netloc and path
+    else:
+        filepath = parsed_result.netloc + parsed_result.path
+
+    filepath = filepath.replace("\\", "/")
+
+    # If on Windows, remove the first leading slash left by urlparse
+    if os.name == 'nt' and filepath.startswith('/'):
+        filepath = filepath[1:]
```
douglascomet, line 56: @apetrynet, it occurs to me that since this is an OS dependent check it will cause the unit tests I made to fail. Any suggestions on how to test Windows paths properly? I could make a regex that looks for a something like `/S:/` before omitting the leading slash but I'm not sure if a colon is ever used in a Linux or Mac path since I do not typically work in those operating systems.

douglascomet: Nvm, I looked into it further and colons are not used in Unix paths so a regex seemed like the best approach for testing locally can and in CI/CD

apetrynet:
> > Now that we've left python 2 behind I think we could start using pathlib for some of this as well.
> 
> @apetrynet, what were you thinking about using `pathlib` here? I also found this closed PR about generating `Path` from a URL, [python/cpython#91504](https://github.com/python/cpython/issues/91504).

I was thinking about some [windows path handling](https://docs.python.org/3/library/pathlib.html#pathlib.PureWindowsPath), but looking at it I'm not sure it'll help in this case.

apetrynet (commented):
I've provided a suggestion for an alternative way of dealing with this. I'm not sure if it's any cleaner, but I try to rely on the builtin modules to resolve this correctly. Should hopefully adapt to the underlying OS

src/py-opentimelineio/opentimelineio/url_utils.py
```diff
@@ -37,7 +38,33 @@ def url_from_filepath(fpath):
...
+        filepath = parsed_result.netloc + parsed_result.path
+
+    filepath = filepath.replace("\\", "/")
+
+    # If on Windows and using a drive letter,
+    # remove the first leading slash left by urlparse
+    if re.match(r"/[a-zA-Z]:/.*", filepath):
+        filepath = filepath[1:]
+
+    return filepath
```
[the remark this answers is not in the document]
douglascomet: Oh wow, I really like this suggestion! I don't know the `pathlib` library well but I had to look a few things up to understand what is happening. I added a few comments to your suggestion to explain to future devs.

tests/test_url_conversions.py
```diff
@@ -51,6 +55,14 @@ def test_roundtrip_rel(self):
         # should have reconstructed it by this point
         self.assertEqual(os.path.normpath(result), MEDIA_EXAMPLE_PATH_REL)
 
+    def test_windows_urls(self):
+        for url in (ENCODED_WINDOWS_URL, WINDOWS_URL):
+            print(f"Original URL: {url}")
+            self.assertTrue(url.startswith("file://"))
+            processed_url = otio.url_utils.filepath_from_url(url)
+            print(f"Processed URL Path: {processed_url}")
```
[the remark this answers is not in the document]
douglascomet: I'll remove the prints and I added `self.assertTrue(url.startswith("file://"))` because the other unit tests were doing something similar. I can remove it as well though.

apetrynet:
@douglascomet Could you please look at the failed CI task? It looks like a simple fix removing the unused "re" import

src/py-opentimelineio/opentimelineio/url_utils.py
```diff
@@ -4,12 +4,14 @@
 """Utilities for conversion between urls and file paths"""
 
 import os
+import re
```
apetrynet, line 7:
```suggestion
```

The unused "re" import stopped the CI.

src/py-opentimelineio/opentimelineio/url_utils.py
```diff
@@ -4,12 +4,14 @@
...
 
 from urllib import (
     parse as urlparse,
     request
 )
 import pathlib
```
apetrynet, line 13: Super nitpick request.
Could you please remove this import as well and use "Path" in the `url_from_filepath()` function?

apetrynet (changes requested):
Good call on the comments!
I have a couple  of minor requests.

src/py-opentimelineio/opentimelineio/url_utils.py
douglascomet: Sure can do! Had no idea the CI was so thorough!

src/py-opentimelineio/opentimelineio/url_utils.py
douglascomet: Sure, can do!

apetrynet (approved):
Thanks @douglascomet !
LGTM

Issue #1668, opened 2023-10-09
Python 3.12 support / imp module deleted


EvanBldy:
## Bug Report

OpenTimeLineIO plugin for python installs correctly with pip but it cannot import be imported after. It's due to an import of imp module which is deleted and deprecated since python 3.4.

```
Python 3.12.0 (main, Oct  9 2023, 17:10:54) [GCC 11.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import opentimelineio
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/mnt/DATA/DEV/virtualenv/linuxmint/zou3.12/lib/python3.12/site-packages/opentimelineio/__init__.py", line 14, in <module>
    from . import (
  File "/mnt/DATA/DEV/virtualenv/linuxmint/zou3.12/lib/python3.12/site-packages/opentimelineio/schema/__init__.py", line 34, in <module>
    from . schemadef import (
  File "/mnt/DATA/DEV/virtualenv/linuxmint/zou3.12/lib/python3.12/site-packages/opentimelineio/schema/schemadef.py", line 7, in <module>
    from .. import (
  File "/mnt/DATA/DEV/virtualenv/linuxmint/zou3.12/lib/python3.12/site-packages/opentimelineio/plugins/__init__.py", line 8, in <module>
    from .python_plugin import (
  File "/mnt/DATA/DEV/virtualenv/linuxmint/zou3.12/lib/python3.12/site-packages/opentimelineio/plugins/python_plugin.py", line 7, in <module>
    import imp
ModuleNotFoundError: No module named 'imp'
```

I will make a PR to fix that. 

## To Reproduce

1. Operating System :  Linux Mint
2. Python version : 3.12

## Expected Behavior

To work on python 3.12.

EvanBldy:
I close because my PR is validated :)

PR #1669, opened 2023-10-09
Python 3.12 support / imp module deleted

EvanBldy:

https://github.com/AcademySoftwareFoundation/OpenTimelineIO/issues/1668

**Summarize your change.**

The imp module in Python is deprecated since Python 3.4 and is removed in Python 3.12. So the package cannot be imported in Python 3.12 because the imp module doesn't exist anymore. See my issue.

By reading the docs it have to be changed by importlib. So I have made these changes and this works for importing adapters for example :) 

Some documentation :
https://docs.python.org/3.12/whatsnew/3.12.html#imp
https://docs.python.org/3.11/library/imp.html

I have added Python 3.12 in the workflow for Python package. 

I use some linter (black) in my vscode so my changes can be more than the real modifications. If you want I can revert these changes. 

Don't have launched the tests, they will run via GitHub Actions, no ?

I'm sorry it's my first PR to OpenTimeLineIO :)

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -93,7 +93,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
-        python-version: ['3.7', '3.8', '3.9', '3.10', '3.11']
+        python-version: ['3.7', '3.8', '3.9', '3.10', '3.11', '3.12']
         include:
           - { os: ubuntu-latest, shell: bash }
           - { os: macos-latest, shell: bash }
@@ -163,7 +163,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
-        python-build: ['cp37*', 'cp38*', 'cp39*', 'cp310*', 'cp311*']
+        python-build: ['cp37*', 'cp38*', 'cp39*', 'cp310*', 'cp311*', 'cp312*']
     steps:
       - uses: actions/checkout@v3
 
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -168,7 +168,7 @@ jobs:
       - uses: actions/checkout@v3
 
       - name: Build wheels (Python 3)
-        uses: pypa/cibuildwheel@v2.11.1
+        uses: pypa/cibuildwheel@v2.16.2
         with:
           output-dir: wheelhouse
         env:
--- a/src/py-opentimelineio/opentimelineio/console/autogen_serialized_datamodel.py
+++ b/src/py-opentimelineio/opentimelineio/console/autogen_serialized_datamodel.py
@@ -200,6 +200,7 @@ def _generate_model_for_module(mod, classes, modules):
             if (
                 inspect.ismodule(thing)
                 and thing not in modules
+                and "opentimelineio" in thing.__name__
                 and all(not thing.__name__.startswith(t) for t in SKIP_MODULES)
             )
         ),
--- a/src/py-opentimelineio/opentimelineio/plugins/python_plugin.py
+++ b/src/py-opentimelineio/opentimelineio/plugins/python_plugin.py
@@ -4,10 +4,10 @@
 """Base class for OTIO plugins that are exposed by manifests."""
 
 import os
-import imp
 import inspect
 import collections
 import copy
+import importlib.util
 
 from .. import (
     core,
@@ -107,21 +107,23 @@ class PythonPlugin(core.SerializableObject):
     def _imported_module(self, namespace):
         """Load the module this plugin points at."""
 
-        pyname = os.path.splitext(os.path.basename(self.module_abs_path()))[0]
-        pydir = os.path.dirname(self.module_abs_path())
-
-        (file_obj, pathname, description) = imp.find_module(pyname, [pydir])
-
-        with file_obj:
-            # this will reload the module if it has already been loaded.
-            mod = imp.load_module(
-                f"opentimelineio.{namespace}.{self.name}",
-                file_obj,
-                pathname,
-                description
+        module_name = f"opentimelineio.{namespace}.{self.name}"
+        try:
+            # Attempt to import the module using importlib first
+            mod = importlib.import_module(module_name)
+        except ImportError:
+            # If the module couldn't be imported, import it manually
+            pyname = os.path.splitext(os.path.basename(self.module_abs_path()))[0]
+            pydir = os.path.dirname(self.module_abs_path())
+            spec = importlib.util.spec_from_file_location(
+                module_name,
+                os.path.join(pydir, f"{pyname}.py"),
             )
 
-            return mod
+            mod = importlib.util.module_from_spec(spec)
+            spec.loader.exec_module(mod)
+
+        return mod
 
     def module(self):
         """Return the module object for this adapter. """
--- a/src/py-opentimelineio/opentimelineio/url_utils.py
+++ b/src/py-opentimelineio/opentimelineio/url_utils.py
@@ -10,8 +10,9 @@ from urllib import (
     request
 )
 from pathlib import (
-    Path,
-    PureWindowsPath
+    PurePath,
+    PureWindowsPath,
+    PurePosixPath
 )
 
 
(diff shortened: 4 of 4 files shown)


JeanChristopheMorinPerso:
Hi @EvanBldy , it would be preferable if you remove any whitespace changes and formatting changes if possible.

EvanBldy:
> Hi @EvanBldy , it would be preferable if you remove any whitespace changes and formatting changes if possible.

It should be better :)

EvanBldy:
The tests are passing in Python 3.8 but not in 3.12, I don't understand the error neither what this code is intended to (https://github.com/AcademySoftwareFoundation/OpenTimelineIO/actions/runs/6462945630/job/17545363452?pr=1669)

reinecke:
Hi @EvanBldy, thank you for working on this!

If you want to try running the tests locally, the usual process is to use these commands to setup and activate a development virtual environment:
```zsh
python3 -m venv .venv
source .venv/bin/activate
pip install -e '.[dev]'  # This builds and installs OTIO and the development dependencies in your virtualenv
```
Depending on your setup, you may want to do `python3.12 -m venv .venv` to ensure the environment is built with the exact python version you want.

Then, each time you make a change, you can run:
```zsh
python setup.py install  # This rebuilds OTIO with the changes and installs it into the env - goes quicker after the first time
make ci-prebuild  # This is what CI uses to lint and make sure all the extra data files needed are included
make test  # A shortcut for running the test suite
make ci-postbuild  # This is where the CI error was coming from
```

That should help you see what will happen in CI locally.

In terms of what autogen_serialized_datamodel.py does, that script autogenerates the markdown for [this file](https://github.com/AcademySoftwareFoundation/OpenTimelineIO/blob/main/docs/tutorials/otio-serialized-schema.md) and [this file](https://github.com/AcademySoftwareFoundation/OpenTimelineIO/blob/main/docs/tutorials/otio-serialized-schema-only-fields.md).

Digging into the error a bit more, I think the issue had to do with some internals of how a `module` is converted into a string. The important part of the trace is here:
```python
  File "/opt/hostedtoolcache/Python/3.12.0/x64/lib/python3.12/site-packages/opentimelineio/console/autogen_serialized_datamodel.py", line 206, in <lambda>
    key=lambda mod: str(mod)
                    ^^^^^^^^
  File "<frozen importlib._bootstrap>", line 545, in _module_repr
  File "<frozen importlib._bootstrap>", line 830, in _module_repr_from_spec
AttributeError: 'VendorImporter' object has no attribute '_path'
```

[This block of code](https://github.com/AcademySoftwareFoundation/OpenTimelineIO/blob/bd7ae178921e4e35fef79980b76f94cdfd2871a0/src/py-opentimelineio/opentimelineio/console/autogen_serialized_datamodel.py#L197-L207) is meant to traverse down into all the submodules. It's hitting some hiccup when it encounters `pkg_resources.extern.VendorImporter`. Upon further inspection, the block of code will actually dive into modules that OpenTimelineIO imports, not just OpenTimelineIO submodules.

To address this, I did a small rewrite of that block of code:
```python
new_mods = sorted(
    (
        thing for _, thing in inspect.getmembers(mod)
        if (
            inspect.ismodule(thing)
            and thing not in modules
            and "opentimelineio" in thing.__name__
            and all(not thing.__name__.startswith(t) for t in SKIP_MODULES)
        )
    ),
    key=lambda mod: str(mod)
)
```

The main part is that I added an extra filter condition in there to only get modules that are part of opentimelineio. Testing locally, this seems to work for me. Feel free to give that a whirl and update the PR - I think that should fix it up!

EvanBldy:
Hi @reinecke,
Thanks for all the informations :) 
And yeah the extra filter condition seems to do the trick. 
I've updated my PR with this code and some other to resolve some "DeprecationWarning: Testing an element's truth value will raise an exception in future versions.".

.github/workflows/python-package.yml
```diff
@@ -93,7 +93,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
-        python-version: ['3.7', '3.8', '3.9', '3.10', '3.11']
+        python-version: ['3.7', '3.8', '3.9', '3.10', '3.11', '3.12']
```
JeanChristopheMorinPerso, line 96: I think this will require a bump of Pybind11: https://github.com/pybind/pybind11/releases/tag/v2.11.0. From what I see in the test it seems to work with the version we currently use, but it would be preferable to use a version that is known to work and officially works with 3.12.

JeanChristopheMorinPerso (commented):
One test is failing on Windows with 3.12 (https://github.com/AcademySoftwareFoundation/OpenTimelineIO/actions/runs/6677197744/job/18146955106?pr=1669#step:9:24). Also I left a note about Pybind11.

.github/workflows/python-package.yml
EvanBldy: Hi ! 
pybind11 seems to be installed via a submodule who point to the repo of pybind11, so I think it's always taking the last commit of the master branch. 
See here : https://github.com/AcademySoftwareFoundation/OpenTimelineIO/blob/main/.gitmodules

EvanBldy:
> One test is failing on Windows with 3.12 (https://github.com/AcademySoftwareFoundation/OpenTimelineIO/actions/runs/6677197744/job/18146955106?pr=1669#step:9:24). Also I left a note about Pybind11.

Yeah, I'm trying to know what's the problem :) The test seems to be buggy only on Windows / Python 3.12, probably a new change in Python 3.12

.github/workflows/python-package.yml
JeanChristopheMorinPerso: That's not how submodules work. Submodules are pinned to a specific commit.

EvanBldy: Yeah I didn't remember ! 
I have updated the submodule for pybind11 to the latest tag (v2.11.1)

EvanBldy:
It should be fixed :)

EvanBldy:
Hi @JeanChristopheMorinPerso @reinecke, are you okay with my PR ? 
We use OTIO in our project and want to support Python 3.12, when do you plan to do a new release?

EvanBldy:
Hi ! Do you have news about this ?

reinecke:
Hi @EvanBldy -
Sorry it took us a while to get back to you.
The change looks pretty good, but I saw the following files were modified with fixes unrelated to this PR:

- [src/py-opentimelineio/opentimelineio/url_utils.py](https://github.com/AcademySoftwareFoundation/OpenTimelineIO/pull/1669/files#diff-3916e9ff9ad21f9ef264e2e1bb38fbfe509f37274565681839d3cf362c8a42c9)
- [contrib/opentimelineio_contrib/adapters/fcpx_xml.py](https://github.com/AcademySoftwareFoundation/OpenTimelineIO/pull/1669/files#diff-fdffca9979ebdfb153c25dad4337e2ab97d87a4f38b5539dc3e33763c8fb8ca7)

Are those intentional? We prefer to keep PRs targeted at one specific thing. Would it be possible to revert those files and submit the changes to them as separate PRs?

Otherwise this is a great update and I'd love to get it in before the 0.16.0 release.

EvanBldy:
Hi @reinecke, thanks for your answer :) 

For the changes :
* [src/py-opentimelineio/opentimelineio/url_utils.py](https://github.com/AcademySoftwareFoundation/OpenTimelineIO/pull/1669/files#diff-3916e9ff9ad21f9ef264e2e1bb38fbfe509f37274565681839d3cf362c8a42c9) : they are intended to fix tests in Python 3.12 (due to a change in the API, and they are preferred to manipulate only path string). This test : https://github.com/AcademySoftwareFoundation/OpenTimelineIO/actions/runs/6677197744/job/18146955106?pr=1669#step:9:24
* [contrib/opentimelineio_contrib/adapters/fcpx_xml.py](https://github.com/AcademySoftwareFoundation/OpenTimelineIO/pull/1669/files#diff-fdffca9979ebdfb153c25dad4337e2ab97d87a4f38b5539dc3e33763c8fb8ca7) : I will remove these changes and make a new PR, it was intended to remove some warnings. 

Cool if it can be in the 0.16.0 release :)

EvanBldy:
I have removed the changes about DeprecationWarning and made a separate PR : https://github.com/AcademySoftwareFoundation/OpenTimelineIO/pull/1688

reinecke (approved):
Thank you for all your work on this!

PR #1670, opened 2023-10-11
0.17.0 dev1 release preparation

reinecke:
This PR bumps the adapter extraction to 0.17.0 in accordance with the new release roadmap and gets the branch prepped for a preview release.


.github/workflows/python-package.yml
```diff
@@ -24,13 +24,13 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-latest, windows-latest, macos-latest]
+        os: [ubuntu-latest`, `windows-latest`, and `macos-latest`, and `macos-latest`, `windows-latest`, and `macos-latest`, and `macos-latest`, and `macos-latest, windows-latest`, and `macos-latest`, and `macos-latest, macos-latest]
```
JeanChristopheMorinPerso, line 27: This doesn't look right to me...

jhodges10:
There may be a way to do the matrix strategies differently so you don't have to repeat them as much.

Check out - https://discuss.circleci.com/t/dynamically-set-parameter-values-of-matrix/47172

.github/workflows/python-package.yml
reinecke: No, it does not, does it.
This was done via [this Makefile target](https://github.com/AcademySoftwareFoundation/OpenTimelineIO/blob/745e61438d3742e9a11fed3f8987e9a22f67bded/Makefile#L238-L240) in [this script](https://github.com/AcademySoftwareFoundation/OpenTimelineIO/blob/main/maintainers/freeze_ci_versions.py).

Hrm, [this](https://github.com/AcademySoftwareFoundation/OpenTimelineIO/blob/745e61438d3742e9a11fed3f8987e9a22f67bded/maintainers/freeze_ci_versions.py#L58-L59) may not be the best.

I'll dig into those and see what's going on.

reinecke: There was a formatting change on the [upstream github README](https://github.com/AcademySoftwareFoundation/OpenTimelineIO/blob/745e61438d3742e9a11fed3f8987e9a22f67bded/maintainers/freeze_ci_versions.py#L16-L18) the script was parsing - that broke [this regex](https://github.com/AcademySoftwareFoundation/OpenTimelineIO/blob/745e61438d3742e9a11fed3f8987e9a22f67bded/maintainers/freeze_ci_versions.py#L68).

I made a code update to that script locally and have a new version of the parsing code to handle the updated format. I'll create a separate PR to main for that particular change soon.

JeanChristopheMorinPerso (approved):

PR #1673, opened 2023-10-24
Updated cibuildwheel settings to generate apple silicon wheels

### LINKED ISSUE #1672: Add Apple Silicon/ARM wheels
## Feature Request

We'd like to distribute Apple Silicon/ARM wheels.

## Description

Right now, we're still building only classic x86_64 wheels for macOS and Apple Silicon users are forced to build from the source tarball. Ideally, we'd offer a range of prebuilt wheels that includes Apple Silicon where applicable.

## Context

We've resisted this in the past because we didn't have a mechanism to test Apple Silicon artifacts to make sure they worked correctly, this looks like it may be [solved now](https://github.blog/2023-10-02-introducing-the-new-apple-silicon-powered-m1-macos-larger-runner-for-github-actions/).

Checking the 0.17.0.dev1 release, we're already cross-compiling a universal binary for the python bindings:

```
% lipo -archs _otio.cpython-37m-darwin.so
x86_64 arm64
```

We also may look at reshuffling `py_build_test` and `package_wheels` - we could consider building all the wheels in one step (using x86_64 to cross-build an Apple Silicon wheel). And have a separate action that pulls down the wheels and runs the test suite against them in a matrix fashion. This way, only the test suite would run on an expensive Apple Silicon runner and the rest can happen on free runners.

References:

- [cibuildwheel docs about Apple Silicon](https://cibuildwheel.readthedocs.io/en/stable/faq/#apple-silicon)
- [Github has released Apple Silicon runners as a paid option](https://github.blog/2023-10-02-introducing-the-new-apple-silicon-powered-m1-macos-larger-runner-for-github-actions/)
- OpenImageIO has done [some experiements](https://academysoftwarefdn.slack.com/archives/C0169RX7MMK/p1696264947232289) with enabling these

reinecke:
Fixes #1672 

This is a WIP PR for enabling Apple Silicon Support.

Currently, this PR simply enables cibuildwheel for x86_64 and arm64.

TODO:

- [ ] Decide which wheels we feel like we need (is there a demand for `universal2` or are the optimized wheels fine?)
- [ ] Figure out how to do testing on Apple Silicon since [arm64 wheels can't be tested on x86_64](https://cibuildwheel.readthedocs.io/en/stable/faq/#how-to-cross-compile)

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -168,7 +168,7 @@ jobs:
       - uses: actions/checkout@v3
 
       - name: Build wheels (Python 3)
-        uses: pypa/cibuildwheel@v2.11.1
+        uses: pypa/cibuildwheel@v2.16.2
         with:
           output-dir: wheelhouse
         env:
@@ -176,6 +176,7 @@ jobs:
           CIBW_SKIP: '*musllinux*'
           CIBW_MANYLINUX_X86_64_IMAGE: manylinux2014
           CIBW_MANYLINUX_I686_IMAGE: manylinux2014
+          CIBW_ARCHS_MACOS: universal2 arm64
 
       - uses: actions/upload-artifact@v3
         with:
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -176,7 +176,7 @@ jobs:
           CIBW_SKIP: '*musllinux*'
           CIBW_MANYLINUX_X86_64_IMAGE: manylinux2014
           CIBW_MANYLINUX_I686_IMAGE: manylinux2014
-          CIBW_ARCHS_MACOS: universal2 arm64
+          CIBW_ARCHS_MACOS: x86_64 arm64
 
       - uses: actions/upload-artifact@v3
         with:


JeanChristopheMorinPerso:
>Decide which wheels we feel like we need (is there a demand for universal2 or are the optimized wheels fine?)

Maybe look at https://cibuildwheel.readthedocs.io/en/stable/faq/#what-to-provide?

>Figure out how to do testing on Apple Silicon since [arm64 wheels can't be tested on x86_64](https://cibuildwheel.readthedocs.io/en/stable/faq/#how-to-cross-compile)

I hear through the branches that it's coming. See https://github.blog/2023-10-02-introducing-the-new-apple-silicon-powered-m1-macos-larger-runner-for-github-actions/#how-to-use-the-runner and https://github.com/AcademySoftwareFoundation/wg-ci/blob/main/meetings/2023-10-11.md?plain=1#L89. I'll check to see if we could have access.

reinecke:
I spot checked the Apple Silicon python 3.11 wheel [from the first build](https://github.com/AcademySoftwareFoundation/OpenTimelineIO/actions/runs/6633380411?pr=1673).

I unzipped that dir into `/tmp/otiowheels` and installed from it using the following command:

```shell
python -m pip install --find-links /tmp/otiowheels opentimelineio==0.16.0.dev1
```

I was then able to use `make ci-postbuild` to run tests against the installed wheel.

Our CI currently has a `py_build_test` job and a package_wheels job.
These redundantly build OTIO on all the platforms but `py_build_test` just [does a](https://github.com/AcademySoftwareFoundation/OpenTimelineIO/blob/745e61438d3742e9a11fed3f8987e9a22f67bded/.github/workflows/python-package.yml#L142-L145) `pip install .[dev] -v` whereas `package_wheels` uses [cibuildwheel](https://github.com/pypa/cibuildwheel) to make the final artifacts.

I think a good refactor might be to use have one job that just runs the lint and check-manifest in a single platform (like we do with `package_sdist`), then a dependent job that runs `package_wheels` to generate the artifacts, and finally make a dependent job that pulls down the artifacts, installs the right one for the platform, and runs tests (basically use [these steps](https://github.com/AcademySoftwareFoundation/OpenTimelineIO/blob/745e61438d3742e9a11fed3f8987e9a22f67bded/.github/workflows/python-package.yml#L115-L136), put a step in to pull down the built wheels and install from them, and then run [these steps](https://github.com/AcademySoftwareFoundation/OpenTimelineIO/blob/745e61438d3742e9a11fed3f8987e9a22f67bded/.github/workflows/python-package.yml#L146-L158)).

This would have some specific benefits:
- We would only do the OTIO libaray builds once
- We would be testing the exact artifact that will deliver to PyPI and might be more likely to catch things like dynamic linker  pathing issues that could come up
- We can limit what is run on Apple Silicon to the bare minimum

The one downside I see is that cibuildwheel seems to take ~5m to build whereas `pip install .` is taking closer to ~2m in many instances. I suspect this is because cibuildwheel downloads and installs it’s own python. My personal feeling is that the benefits outweigh the extra few minutes.

reinecke:
Another consideration - universal binaries are a bit slower to import possibly - what's the tax of this?

From TSC meeting: let's go with x86_64 + universal2 archs

reinecke:
This seems to be addressed by #1214 - we can revisit this if we decide we need to explicitly signal through cibuildwheel as well.

PR #1674, opened 2023-10-25
Added support to filepath_from_url for UNC paths and tests for UNC and posix paths

douglascomet:
This PR is a follow up to https://github.com/AcademySoftwareFoundation/OpenTimelineIO/pull/1664 to add support to `filepath_from_url` and tests for UNC paths.

### DIFF
--- a/src/py-opentimelineio/opentimelineio/url_utils.py
+++ b/src/py-opentimelineio/opentimelineio/url_utils.py
@@ -82,8 +82,9 @@ def filepath_from_url(urlstr):
         # Paths of type: "file://host/share/path/to/file.ext" provide "host" as netloc
         filepath = PurePath('//', parsed_result.netloc + decoded_parsed_path)
 
-        # When running `as_posix` the resulting path will have only 1 leading `/`,
-        # so we insert another `/` at the front of the string path and return it.
+        # Executing `as_posix` on Windows seems to generate a path with only
+        # 1 leading `/`, so we insert another `/` at the front of the string path
+        # to match Linux and Windows UNC conventions and return it.
         conformed_filepath = filepath.as_posix()
         if not conformed_filepath.startswith('//'):
             conformed_filepath = '/' + conformed_filepath
--- a/src/py-opentimelineio/opentimelineio/url_utils.py
+++ b/src/py-opentimelineio/opentimelineio/url_utils.py
@@ -66,8 +66,8 @@ def filepath_from_url(urlstr):
     if PureWindowsPath(parsed_result.netloc).drive:
         filepath = PurePath(parsed_result.netloc + parsed_result.path)
 
-    # Check if the specified index is a windows drive, if it is then do nothing
-    elif PureWindowsPath(filepath.parts[0]).drive:
+    # Check if the specified index has a specified `drive`, if so then do nothing
+    elif filepath.drive:
         filepath = filepath
 
     # Check if the specified index is a windows drive, then offset the path
@@ -75,8 +75,8 @@ def filepath_from_url(urlstr):
         # Remove leading "/" if/when `request.url2pathname` yields "/S:/path/file.ext"
         filepath = PurePosixPath(*filepath.parts[1:])
 
-    # Last resort, if using a "file" schema, then strip the "file:" prefix
-    elif parsed_result.scheme == 'file':
+    # Last resort, strip the "file:" prefix
+    else:
         filepath = Path(decoded_url_str.strip('file:'))
 
     # Convert "\" to "/" if needed
--- a/src/py-opentimelineio/opentimelineio/url_utils.py
+++ b/src/py-opentimelineio/opentimelineio/url_utils.py
@@ -40,7 +40,7 @@ def url_from_filepath(fpath):
         )
 
 
-def filepath_from_url(urlstr):
+def filepath_from_url(urlstr, as_posix=True):
     """
     Take an url and return a filepath.
 
@@ -81,6 +81,13 @@ def filepath_from_url(urlstr):
     elif parsed_result.netloc and parsed_result.netloc != 'localhost':
         # Paths of type: "file://host/share/path/to/file.ext" provide "host" as netloc
         filepath = PurePath('//', parsed_result.netloc + decoded_parsed_path)
+
+        # When running `as_posix` the resulting path will have only 1 leading `/`,
+        # so we insert another `/` at the front of the string path and return it.
+        conformed_filepath = filepath.as_posix()
+        if conformed_filepath.startswith('/'):
+            conformed_filepath = '/' + conformed_filepath
+        return conformed_filepath
 
     # Convert "\" to "/" if needed
     return filepath.as_posix()
--- a/src/py-opentimelineio/opentimelineio/url_utils.py
+++ b/src/py-opentimelineio/opentimelineio/url_utils.py
@@ -4,6 +4,7 @@
 """Utilities for conversion between urls and file paths"""
 
 import os
+
 from urllib import (
     parse as urlparse,
     request
--- a/src/py-opentimelineio/opentimelineio/url_utils.py
+++ b/src/py-opentimelineio/opentimelineio/url_utils.py
@@ -76,7 +76,7 @@ def filepath_from_url(urlstr):
     # as parsing "file:///some/path/to/file.ext" doesn't provide a netloc
     elif parsed_result.netloc and parsed_result.netloc != 'localhost':
         # Paths of type: "file://host/share/path/to/file.ext" provide "host" as netloc
-        filepath = Path('//', parsed_result.netloc + decoded_parsed_path)
+        filepath = PurePath('//', parsed_result.netloc + decoded_parsed_path)
 
     # Convert "\" to "/" if needed
     path = filepath.as_posix()
--- a/src/py-opentimelineio/opentimelineio/url_utils.py
+++ b/src/py-opentimelineio/opentimelineio/url_utils.py
@@ -4,7 +4,6 @@
 """Utilities for conversion between urls and file paths"""
 
 import os
-import sys
 from urllib import (
     parse as urlparse,
     request
--- a/src/py-opentimelineio/opentimelineio/url_utils.py
+++ b/src/py-opentimelineio/opentimelineio/url_utils.py
@@ -4,6 +4,7 @@
 """Utilities for conversion between urls and file paths"""
 
 import os
+import sys
 from urllib import (
     parse as urlparse,
     request
@@ -11,7 +12,8 @@ from urllib import (
 from pathlib import (
     PurePath,
     PureWindowsPath,
-    PurePosixPath
+    PurePosixPath,
+    Path
 )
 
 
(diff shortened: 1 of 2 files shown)


douglascomet:
Just occurred to me, should `filepath_from_url` raise or do nothing if a `urlstr` does not have the `file://` prefix? What about a different scheme?

src/py-opentimelineio/opentimelineio/url_utils.py
```diff
@@ -62,10 +65,18 @@ def filepath_from_url(urlstr):
     if PureWindowsPath(parsed_result.netloc).drive:
         filepath = Path(parsed_result.netloc + parsed_result.path)
 
-    # Otherwise check if the specified index is a windows drive, then offset the path
+    # Check if the specified index has a specified `drive`, if so then do nothing
+    elif filepath.drive:
+        filepath = filepath
+
+    # Check if the specified index is a windows drive, then offset the path
```
apetrynet, line 72: I'm a bit concerned that this test only will work in a windows environment and that the "else" block below will manipulate almost all URLs passing through this function.

src/py-opentimelineio/opentimelineio/url_utils.py
```diff
@@ -62,10 +65,18 @@ def filepath_from_url(urlstr):
...
     elif PureWindowsPath(filepath.parts[1]).drive:
         # Remove leading "/" if/when `request.url2pathname` yields "/S:/path/file.ext"
         filepath = filepath.relative_to(filepath.root)
 
+    # Last resort, strip the "file:" prefix
+    else:
+        filepath = Path(decoded_url_str.strip('file:'))
+
```
apetrynet, line 80: In other environments than windows or URLs without windows drives/shares , this will manipulate all URLs.

apetrynet (commented):
Thanks @douglascomet for working on covering more windows URL cases.

Here is a suggestion for an alternative way of achieving this. It only manipulates the path if a windows drive is found or an URL only contains two leading slashes like so: "file://host/share/path/to/file.ext\"

Please note that I've also done some minor adjustments to code untouched by your PR.

``` python
def filepath_from_url(urlstr):
    """
    Take an url and return a filepath.

    URLs can either be encoded according to the `RFC 3986`_ standard or not.
    Additionally, Windows mapped drive letter and UNC paths need to be accounted for
    when processing URL(s); however, there are `ongoing discussions`_ about how to best
    handle this within Python developer community. This function is meant to cover
    these scenarios in the interim.

    .. _RFC 3986: https://tools.ietf.org/html/rfc3986#section-2.1
    .. _ongoing discussions: https://discuss.python.org/t/file-uris-in-python/15600
    """

    # Parse provided URL
    parsed_result = urlparse.urlparse(urlstr)

    # De-encode the parsed path
    decoded_parsed_path = urlparse.unquote(parsed_result.path)

    # Convert the parsed URL to a path
    filepath = Path(request.url2pathname(decoded_parsed_path))

    # If the network location is a window drive, reassemble the path
    if PureWindowsPath(parsed_result.netloc).drive:
        filepath = Path(parsed_result.netloc + decoded_parsed_path)

    # Check if the specified index is a windows drive, then offset the path
    elif PureWindowsPath(filepath.parts[1]).drive:
        # Remove leading "/" if/when `request.url2pathname` yields "/S:/path/file.ext"
        filepath = filepath.relative_to(filepath.root)

    # Should catch UNC paths, as parsing "file:///some/path/to/file.ext" doesn't provide a netloc
    elif parsed_result.netloc:
        # Paths of type: "file://host/share/path/to/file.ext" provide "host" as netloc
        filepath = Path('//', parsed_result.netloc + decoded_parsed_path)

    # Convert "\" to "/" if needed
    return filepath.as_posix()

```

src/py-opentimelineio/opentimelineio/url_utils.py
douglascomet: This seems to work with the example paths for the Posix paths. Did i use invalid paths?

src/py-opentimelineio/opentimelineio/url_utils.py
douglascomet: Unfortunately, since I'm dev-ing on Windows I'm getting behavior that the CI can't replicate. Even the coverage results highlighted this section but it passes and is used on my machine.

src/py-opentimelineio/opentimelineio/url_utils.py
douglascomet: Also, from what I can tell, when AVID references a UNC path on Windows it only has `//` and `///`, that's why I just stripped `file:`

apetrynet: >This seems to work with the example paths for the Posix paths. Did i use invalid paths?

With the exception of possibly leaving in `localhost` in the path (more on this below), I don't think it produced invalid paths. My main point was that in a non-Windows environment this will run on all paths which seems unnecessary.
 
>Also, from what I can tell, when AVID references a UNC path on Windows it only has `//` and `///`, that's why I just stripped `file`:

Would you be able to provide example paths we can look at that show the two variations so we can dissect the paths?

From what I gather in [this](https://en.wikipedia.org/wiki/File_URI_scheme) wikipedia page there should "always" be three slashes. 
`file:///path/to/file.ext` is more or less the same as `file://localhost/path/to/file.ext` once resolved. Difference being that `localhost` is implied in the first one and defined in the second one. 
If we replace `localhost` with `SERVERNAME` we're getting closer to the UNC paths provided in your tests.

In both your and my approach will produce a path with `localhost` in it given the URL: `file://localhost/path/to/file.ext` -> `//localhost/path/to/file.ext` which I think would produce an error down the line.
Since `localhost` is "reserved" it shouldn't make it in a file path.

So whether we go for stripping the scheme or joining parts I think we should adjust the logic to:
`elif parsed_result.netloc and parsed_result.netloc != 'localhost':` (updated in comment above)
which should produce: `//WINDOWSSHARE/path/to/file.ext` but not `//localhost/path/to/file.ext`

We should perhaps add another path to test
`POSIX_LOCALHOST_URL = "file://localhost/path/sub dir/file.ext"`

meshula:
@douglascomet Does your latest commit resolve either of the conversations?

douglascomet:
> @douglascomet Does your latest commit resolve either of the conversations?

Yes, it will. Apologies for the delay with pushing this PR forward.

src/py-opentimelineio/opentimelineio/url_utils.py
douglascomet: > Would you be able to provide example paths we can look at that show the two variations so we can dissect the paths?

@apetrynet, I debugged the parsing an `.aaf` and came across this section of code in `advanced_authoring_format.py` 
image

Seems the `.aaf` I happened to be testing had a clip with a url like this: `file://unc/path/to/file.ext`, with only 2 `/` and not 3. I couldn't tell you why that is the case but that's what led me to make this PR to fix that.


> So whether we go for stripping the scheme or joining parts I think we should adjust the logic to:
elif parsed_result.netloc and parsed_result.netloc != 'localhost': (updated in comment above)
which should produce: //WINDOWSSHARE/path/to/file.ext but not //localhost/path/to/file.ext

I like your suggestion way better than mine and switched to it! Unfortunately, I'm getting failed unit tests with `tests.test_url_conversions.TestConversions.test_windows_unc_urls`. I'm not sure how we resolve this.

image

> We should perhaps add another path to test POSIX_LOCALHOST_URL = "file://localhost/path/sub dir/file.ext"

Added!

src/py-opentimelineio/opentimelineio/url_utils.py
douglascomet: I removed this because your other suggestion from https://github.com/AcademySoftwareFoundation/OpenTimelineIO/pull/1674#discussion_r1375120236 is better.

src/py-opentimelineio/opentimelineio/url_utils.py
douglascomet: @apetrynet, I think I've reached a solution that gets all the unit tests to pass. Let me know what you think!

src/py-opentimelineio/opentimelineio/url_utils.py
```diff
@@ -56,17 +57,37 @@ def filepath_from_url(urlstr):
...
+    # as parsing "file:///some/path/to/file.ext" doesn't provide a netloc
+    elif parsed_result.netloc and parsed_result.netloc != 'localhost':
+        # Paths of type: "file://host/share/path/to/file.ext" provide "host" as netloc
+        filepath = PurePath('//', parsed_result.netloc + decoded_parsed_path)
+
+        # When running `as_posix` the resulting path will have only 1 leading `/`,
+        # so we insert another `/` at the front of the string path and return it.
+        conformed_filepath = filepath.as_posix()
+        if conformed_filepath.startswith('/'):
+            conformed_filepath = '/' + conformed_filepath
```
apetrynet, line 89: I think this test will always be `True`.
On Linux the `filepath` already has two leading slashes from the join above. Since `startwith('/')` will be true for 1+ slashes you get three.
Does Windows remove one of the slashes when joining or at the `as_posix()` call?

src/py-opentimelineio/opentimelineio/url_utils.py
```diff
@@ -40,14 +40,15 @@ def url_from_filepath(fpath):
         )
 
 
-def filepath_from_url(urlstr):
+def filepath_from_url(urlstr, as_posix=True):
```
apetrynet, line 43: Is the `as_posix` argument used anywhere?

meshula:
@douglascomet No need to apologize, I was just checking status, no urgency.

src/py-opentimelineio/opentimelineio/url_utils.py
douglascomet: Ah, this can be removed I was playing around with ideas and forgot to remove this.

src/py-opentimelineio/opentimelineio/url_utils.py
douglascomet: Yes, that's what appears to be happening on my Windows machine even though the `filepath` was defined to start with `//`.

douglascomet: Thanks for the context, I decided to re-work this a bit and it seems to be passing the tests now
```python
        conformed_filepath = filepath.as_posix()
        if not conformed_filepath.startswith('//'):
            conformed_filepath = '/' + conformed_filepath
```

reinecke:
@douglascomet in response to:
> Just occurred to me, should `filepath_from_url` raise or do nothing if a urlstr does not have the file:// prefix? Same question if a url with a different scheme is provided?

I would say it should not raise - the implication is that that is a relative URL. That does, however, imply that `filepath_from_url` should maybe accepts something like a `base_path` with is the filepath that paths can be relative to (usually this would be the path of parent dir the loaded otio file is in).

But I think this is out of scope for this PR.

reinecke (approved):

src/py-opentimelineio/opentimelineio/url_utils.py
apetrynet: Sorry I disappeared, must of zoned off while attempting to telepathically  connect with @meshula who also had some comments earlier. I think we should go ahead as the tests pass and we cover the old behavior.

PR #1675, opened 2023-10-27
Replaced usages of deprecated pkg_resources module

reinecke:
**Summarize your change.**

`pkg_resources` is now deprecated. This PR removes references to that module.

Refactored `plugin.manifest` module to factor the pip installed package discovery into a separate function. Made `otiopluginfo` and `otioconvert` use that function rather than doing their own plugin discovery.

I did not add new tests, but I validated the output of `otiopluginfo --version` and `otioconvert --version` to validate they're nearly identical. The only difference is that the string `Plugins from pkg_resources:` is now `Plugins from installed packages:`.


JeanChristopheMorinPerso (approved):
LGTM

PR #1679, opened 2023-11-16
add rendering notes

meshula:
This PR adds notes on rendering since rendering behavior is unclear from the specification itself.

### DIFF
--- a/docs/tutorials/otio-timeline-structure.md
+++ b/docs/tutorials/otio-timeline-structure.md
@@ -10,7 +10,7 @@ audio/video timeline.
 Rendering of the image tracks in a timeline is done in painter order. The layers
 in a stack are iterated from the bottom (the first entry in the stack) towards
 the top (the final entry in the stack). Images in a stack overlay lower images
-with no blending.
+using an alpha composite operation respecting any alpha in the source materials.
 
 Within a track, clips may overlap via a transition. In that case, the 
 contribution of track is the linear blend of the elements joined by the
--- a/docs/tutorials/otio-timeline-structure.md
+++ b/docs/tutorials/otio-timeline-structure.md
@@ -5,6 +5,26 @@ clips, gaps, and transitions. This document is meant to clarify how these
 objects nest within each other, and how they work together to represent an
 audio/video timeline.
 
+## Rendering
+
+Rendering of the image tracks in a timeline is done in painter order. The layers
+in a stack are iterated from the bottom (the first entry in the stack) towards
+the top (the final entry in the stack). Images in a stack overlay lower images
+with no blending.
+
+Within a track, clips may overlap via a transition. In that case, the 
+contribution of track is the linear blend of the elements joined by the
+transition.
+
+If there are effects on a clip, OpenTimelineIO does not say anything about the
+impact of the effect and deviation from the base behavior is application
+specific.
+
+Rendering of the audio tracks is additive. It is strongly advised, but not
+required, that the summed audio is summed as floating point, and that it is 
+processed through a compression filter in order to prevent clipping and 
+distortion. 
+
 ## Simple Cut List
 
 Let’s start with a simple cut list of a few clips. This is stored as a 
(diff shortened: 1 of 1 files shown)


docs/tutorials/otio-timeline-structure.md
```diff
@@ -5,6 +5,26 @@ clips, gaps, and transitions. This document is meant to clarify how these
 objects nest within each other, and how they work together to represent an
 audio/video timeline.
 
+## Rendering
+
+Rendering of the image tracks in a timeline is done in painter order. The layers
+in a stack are iterated from the bottom (the first entry in the stack) towards
+the top (the final entry in the stack). Images in a stack overlay lower images
+with no blending.
```
darbyjohnston, line 13: As @reinecke mentioned in the Slack thread, the current documentation recommends the tracks be alpha composited:
https://github.com/AcademySoftwareFoundation/OpenTimelineIO/wiki/OpenTimelineIO-Application-Integrator's-Guide#stack

That seems like it would be preferable for things like information overlays (logos, shot info, etc.)?

darbyjohnston:
What would also be really useful is the equivalent of the "project settings" in Final Cut Pro for establishing the output resolution, audio format, etc:
https://support.apple.com/guide/final-cut-pro/final-cut-pro-project-settings-ver1b946a4ff/10.7/mac/13.5

Currently in tlRender I make a guess at the settings by using the video and audio from the first clip in the timeline. This works mostly OK but fails in some situations and also requires opening the media.

Properties_01
Properties_04

reinecke:
I spun the project settings suggestion @darbyjohnston made into #1683 - I know it'd been talked about before but we'd failed to capture it as an issue. Thanks for highlighting it again!

docs/tutorials/otio-timeline-structure.md
meshula, line 13:
```suggestion
using an alpha composite operation respecting any alpha in the source materials.
```

darbyjohnston:
Looks good to me, thanks. For transparent clips that allow the background to show through, we assume the background color is black?

docs/tutorials/otio-timeline-structure.md
```diff
@@ -5,6 +5,26 @@ clips, gaps, and transitions. This document is meant to clarify how these
 objects nest within each other, and how they work together to represent an
 audio/video timeline.
 
+## Rendering
+
+Rendering of the image tracks in a timeline is done in painter order. The layers
+in a stack are iterated from the bottom (the first entry in the stack) towards
+the top (the final entry in the stack). Images in a stack overlay lower images
+using an alpha composite operation respecting any alpha in the source materials.
+
```
meshula, line 14:
```suggestion
All compositing is assumed to occur over a background of zero values in color components, and 100% values for alpha components.
```

meshula:
@darbyjohnston yes, I added a bit about zero values, and 100% values for alpha. Note that I specified zero, not black, because if we say "black" then it invites discussion about black levels and what we expect to happen for non-zero blacks. I also said 100% so that no one wonders if we mean "1", "1.0", or "255" or ...

darbyjohnston:
Thanks, I think the note about "zero" is good so we don't get into color issues as you mentioned. Why the 100% alpha? For viewers it is nice to preserve the alpha channels for checking renders, though that could also be achieved by viewing the individual elements instead of the final composite.

reinecke:
This update looks really good to me. On the background color, I think 100% opaque zero color values is a good baseline to match most NLEs. That said, if an application wants to allow people to override that behavior, there’s nothing stopping them.

meshula:
If someone wants to go ahead and approve, we can merge.

darbyjohnston:
It looks like the DCO check is failing...

meshula:
right, if we accept changes in the online form, as we I did, the dco is not attached. heavy sigh.

meshula:
Does anyone know a trick for repairing that? If not, I'm inclined to just merge, as there are not enough hours in the day to recreate a PR and go through the whole rigamarole again

reinecke (approved):

reinecke:
I forced DCO pass and merged - thanks for updating the docs @meshula

PR #1691, opened 2024-02-01
Fix non latin symbols decoding in adapters

### LINKED ISSUE #1321: FCP7 XML adapter issue: Py 3.9 changes default encoding behavior on 'open'
Justin Webster reports (https://academysoftwarefdn.slack.com/archives/CMQ9J4BQC/p1653968371353599) ~

Since switching from Python 3.7 to 3.9, parsing FCP7 XML fails.
Python seems to be defaulting to ascii mode and requires that utf-8 encoding be explicitly set in the call to open()
```
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 90: ordinal not in range(128)
```

Apparently recent Python releases behave differently and rely on the system specific environment to determine the default encoding when using open(). https://stackoverflow.com/questions/27401474/pydev-3-8-or-3-9-requires-utf-8-encoding-to-be-manually-set-in-python-3-code-un

Changing line 137 in adapter.py to the following solves it on my test system:

```py
 with open(filepath, 'r', encoding="UTF-8") as fo:
```
But I’m not sure if this is 100% safe and won’t affect any other adapter other that the XML and AAF ones which I’m testing.

Nick reports:

As best as I can tell, the encoding parameter on open exists in any version of Python3. It doesn't exist in Python 2.
https://docs.python.org/2/library/functions.html#open
https://docs.python.org/3/library/functions.html#open
So we need a 2/3 check?

```
Python 2.7.16 (default, Mar  5 2021, 01:03:47)
[GCC 6.3.1 20170216 (Red Hat 6.3.1-3)] on linux2
>>> f = open("/var/tmp/foo", 'r', encoding="UTF-8")
TypeError: 'encoding' is an invalid keyword argument for this function
```

he XML standard says "If no encoding declaration exists in a document's XML declaration, that XML document is required to use either UTF-8 or UTF-16 encoding."
Do we have to read an XML file to find out what the encoding is in order to be able to open the file for reading?

The specification of FCP7 XML is silent on the encoding, and refers us to the XML specifications. [https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/FinalCutPro_XML/AboutThisDoc/AboutThisDoc.html#//appl[…]REF101](https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/FinalCutPro_XML/AboutThisDoc/AboutThisDoc.html#//apple_ref/doc/uid/TP30001152-TPXREF101)

A general solution seems out of the question, so it would have to be a binary peek at the initial chunk looking for explicit encoding attributes....  https://stackoverflow.com/questions/436220/how-to-determine-the-encoding-of-text

@reinecke reports

Yeah, it looks like there is a little chicken/egg dance that’s meant to happen here.
Which, unfortunately, is how my ALE/EDL handler works. It basically tries a couple common encodings until it lands on one that works. At least with XML it has a way to tell you what the encoding should be.
We’ve talked previously about having a read_from_bytes function in adapters so that they can manage some of this dance themselves.

my feeling is that delegating this to individual adapters may be the best way forward. If we see some reusable patterns emerge we could look into factoring those into some shared code somewhere

kalemas:
 
Fixes #1321

**Summarize your change.**

As Py2 support is dropped we should be free to commit utf-8 encoding in Python Adapters to handle UTF in input data.

### DIFF
--- a/src/py-opentimelineio/opentimelineio/adapters/adapter.py
+++ b/src/py-opentimelineio/opentimelineio/adapters/adapter.py
@@ -105,7 +105,7 @@ class Adapter(plugins.PythonPlugin):
             not self.has_feature("read_from_file") and
             self.has_feature("read_from_string")
         ):
-            with open(filepath) as fo:
+            with open(filepath, encoding="utf-8") as fo:
                 contents = fo.read()
             result = self._execute_function(
                 "read_from_string",
@@ -181,7 +181,7 @@ class Adapter(plugins.PythonPlugin):
             self.has_feature("write_to_string")
         ):
             result = self.write_to_string(input_otio, **adapter_argument_map)
-            with open(filepath, 'w') as fo:
+            with open(filepath, 'w', encoding="utf-8") as fo:
                 fo.write(result)
             result = filepath
 


meshula (approved):
lgtm, thanks for the contribution.

reinecke:
@kalemas We've addressed the windows build issues in the HEAD of the current main branch, would you mind rebasing on top of that so we can get a clean build?
Sorry about the confusion!

kalemas:
> rebasing on top of that so we can get a clean build?

@reinecke Done.

PR #1697, opened 2024-02-14
Add marker to timeline when we can't find valid track

### LINKED ISSUE #1696: OpenTimelineIO raises exception when trying to load AAF exported from Avid with "Use Selected Tracks" selected
## Bug Report
I’ve found a bug in Avid Media Composer AAF file generation:
- Make 3 video tracks
- On the third track, place a clip with a marker
- Deselect track 2
- Export an AAF with “selected tracks only” setting
Since track 2 is excluded from the AAF, track 3 gets assigned PhysicalTrackNumber = 2  in the AAF (reasonable!)
BUT - the marker doesn’t get updated! The Marker in the AAF still points at track 3! If you try to import this AAF back into Avid, that marker just disappears.
If you import that AAF into DaVinci Resolve, the marker gets added to the timeline itself instead of a clip.

Currently OTIO will raise an exception when trying to load this AAF:
```python
raise AAFAdapterError(
    "Marker '{}' cannot be attached to an item. SlotID: '{}', "
    "PhysicalTrackNumber: '{}'".format(
        marker.name, slot_id, track_number
    )
)
```
Since this is a "valid" aaf generated by Avid Media Composer, OpenTimelineIO shouldn't explode when reading in the file.

## To Reproduce

Here's an AAF file generated by Avid Media Composer 2021.12.2.55649.0, using the instructions above:
[bad_marker_track_from_avid.aaf.zip](https://github.com/AcademySoftwareFoundation/OpenTimelineIO/files/14285416/bad_marker_track_from_avid.aaf.zip)

## Logs

```text
  File "/Users/awitherspoon/src/studio/laika-editorial/venv/bin/otioview", line 8, in <module>
    sys.exit(main())
  File "/Users/awitherspoon/src/studio/laika-editorial/venv/lib/python3.9/site-packages/opentimelineview/console.py", line 300, in main
    window.load(args.input)
  File "/Users/awitherspoon/src/studio/laika-editorial/venv/lib/python3.9/site-packages/opentimelineview/console.py", line 191, in load
    file_contents = otio.adapters.read_from_file(
  File "/Users/awitherspoon/src/studio/laika-editorial/venv/lib/python3.9/site-packages/opentimelineio/adapters/__init__.py", line 137, in read_from_file
    return adapter.read_from_file(
  File "/Users/awitherspoon/src/studio/laika-editorial/venv/lib/python3.9/site-packages/opentimelineio/adapters/adapter.py", line 125, in read_from_file
    result = self._execute_function(
  File "/Users/awitherspoon/src/studio/laika-editorial/venv/lib/python3.9/site-packages/opentimelineio/plugins/python_plugin.py", line 153, in _execute_function
    return (getattr(self.module(), func_name)(**kwargs))
  File "/Users/awitherspoon/src/studio/laika-editorial/venv/lib/python3.9/site-packages/opentimelineio_contrib/adapters/advanced_authoring_format.py", line 1590, in read_from_file
    result = _attach_markers(result)
  File "/Users/awitherspoon/src/studio/laika-editorial/venv/lib/python3.9/site-packages/opentimelineio_contrib/adapters/advanced_authoring_format.py", line 1245, in _attach_markers
    raise AAFAdapterError(
opentimelineio.adapters.AAF.AAFAdapterError: Marker 'Marker on Track 3! When exported with Track 2 deselected and exported with "Selected Tracks Only", this marker will incorrectly point at track 2 in the AAF even though it doesn't exist (since the original track 2 was ommitted, and track 3 was renamed to 2 by avid!)' cannot be attached to an item. SlotID: '9', PhysicalTrackNumber: '3'
```

austinwitherspoon:
```Fixes #1696```

See issue #1696 , certain AAF files generated by Avid Media Composer with "Use Selected Tracks" may contain markers with invalid PhysicalTrackNumber properties.

The current behavior is to raise an exception when this occurs, but it seems preferable to handle this situation gracefully.
Avid currently handles these AAFs by dropping the marker entirely.
DaVinci Resolve handles these AAFs by adding the marker to the timeline, rather than a clip.

I personally prefer the Resolve approach, since we don't delete the data entirely. It means we always have _somewhere_ where the marker data will end up.

I have update the AAF parser to do this approach - instead of raising an exception when OTIO can't find a matching track, we apply the marker to the timeline itself (or rather, the `tracks` property on the timeline. This is how OTIO files generated by Resolve represent timeline markers.)

Here is a screenshot of `otioview` running on this AAF, where the marker is applied to the timeline itself rather than the clip:
image

I've added a test for this case in `contrib/opentimelineio_contrib/adapters/tests/test_aaf_adapter.py` called `test_aaf_marker_with_bad_track()`
This test currently only ensures that we don't raise an exception when parsing a file like this.


markreidvfx:
@austinwitherspoon thanks for the PR! we are in the process of moving all the adapters to separate repositories. Are you able to port your change to this repo?
https://github.com/OpenTimelineIO/otio-aaf-adapter

Your change might conflict with some recent changes made there related to markers. Maybe this is already fixed, 
https://github.com/OpenTimelineIO/otio-aaf-adapter/pull/20
but it does sound like a different issue.

austinwitherspoon:
> @austinwitherspoon thanks for the PR! we are in the process of moving all the adapters to separate repositories. Are you able to port your change to this repo? https://github.com/OpenTimelineIO/otio-aaf-adapter
> 
> Your change might conflict with some recent changes made there related to markers. Maybe this is already fixed, [OpenTimelineIO/otio-aaf-adapter#20](https://github.com/OpenTimelineIO/otio-aaf-adapter/pull/20) but it does sound like a different issue.

Ah okay, I'll take a look and see if I can move my changes over there! Thanks.

PR #1698, opened 2024-02-15
Add optional comment field to Marker

### LINKED ISSUE #744: Marker: Add comment to schema
In addition to color and/or name, some NLE packages support a comment in their markers. The marker schema could add a top-level `comment` property to hold this data.

There are some format-specific differences.

FCP 7 XML seems to line up pretty nicely with this using the `comment` tag: https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/FinalCutPro_XML/Elements/Elements.html#//apple_ref/doc/uid/TP30001156-TPXREF106

AAF provides some interesting discussion points as it does not have a discrete name and comment field. Per @stefanschulze:
> there is not really a name in AAF for a marker, so it probably would work to use the name as the comment-content. But a marker comment in AAF could be large and is really different than what I’d understand to be a name. I think it would be better not to conflate these two notions.

nickblt:

Fixes #744

**Summarize your change.**

This adds an optional comment field to the Marker object as requested in #744 

Added new tests that specifically test to make sure the comment is being added and serialized/deserialized successfully

### DIFF
--- a/docs/tutorials/otio-serialized-schema-only-fields.md
+++ b/docs/tutorials/otio-serialized-schema-only-fields.md
@@ -212,6 +212,7 @@ parameters:
 
 parameters:
 - *color*
+- *comment*
 - *marked_range*
 - *metadata*
 - *name*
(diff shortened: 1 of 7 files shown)


meshula (approved):
Looks good to me.

PR #1703, opened 2024-02-18
Normalize track lengths with gaps before flatten stacks

### LINKED ISSUE #1430: Flatten stack produces undesirable results if track does not end with gap or clip
## Repro test

C++:
```c++
// SPDX-License-Identifier: Apache-2.0
// Copyright Contributors to the OpenTimelineIO project

#include "opentime/timeRange.h"
#include "utils.h"

#include <opentimelineio/clip.h>
#include <opentimelineio/stack.h>
#include <opentimelineio/track.h>
#include <opentimelineio/stackAlgorithm.h>

#include <iostream>

namespace otime = opentime::OPENTIME_VERSION;
namespace otio  = opentimelineio::OPENTIMELINEIO_VERSION;

int
main(int argc, char** argv)
{
    Tests tests;

    tests.add_test(
        "test_flatten_stack", [] {
        using namespace otio;

        otio::RationalTime rt_0_24{0, 24};
        otio::RationalTime rt_150_24{150, 24};
        otio::TimeRange tr_0_150_24{rt_0_24, rt_150_24};

        // all three clips are identical, but placed such that A is over B and
        // has no gap or end over C
        // 0         150          300
        // [    A     ]
        // [    B     |     C     ]
        // 
        // should flatten to:
        // [    A     |     C     ]

        otio::SerializableObject::Retainer<otio::Clip> cl_A =
            new otio::Clip("track1_A", nullptr, tr_0_150_24);
        otio::SerializableObject::Retainer<otio::Clip> cl_B =
            new otio::Clip("track1_B", nullptr, tr_0_150_24);
        otio::SerializableObject::Retainer<otio::Clip> cl_C =
            new otio::Clip("track1_C", nullptr, tr_0_150_24);

        otio::SerializableObject::Retainer<otio::Track> tr_over =
            new otio::Track();
        tr_over->append_child(cl_A);

        otio::SerializableObject::Retainer<otio::Track> tr_under =
            new otio::Track();
        tr_under->append_child(cl_B);
        tr_under->append_child(cl_C);

        otio::SerializableObject::Retainer<otio::Stack> st =
            new otio::Stack();
        st->append_child(tr_under);
        st->append_child(tr_over);

        auto result = flatten_stack(st);

        assertEqual(result->children()[0]->name(), std::string("track1_A"));
        assertEqual(result->children().size(), 2);
        assertEqual(result->duration().value(), 300);
    });

    tests.run(argc, argv);
    return 0;
}
```

Python:

```python
import opentimelineio as otio

rt_0_24 = otio.opentime.RationalTime(0, 24)
rt_150_24 = otio.opentime.RationalTime(150, 24)
tr_0_150_24 = otio.opentime.TimeRange(
    start_time=rt_0_24,
    duration=rt_150_24
)

# all three clips are identical, but placed such that A is over B and has no
# gap or end over C
# 0         150          300
# [    A     ]
# [    B     |     C     ]
# 
# should flatten to:
# [    A     |     C     ]

clip_A = otio.schema.Clip(name="track1_A", source_range=tr_0_150_24)
clip_B = otio.schema.Clip(name="track2_B", source_range=tr_0_150_24)
clip_C = otio.schema.Clip(name="track2_C", source_range=tr_0_150_24)

tr_over = otio.schema.Track(children=[clip_A])
tr_under = otio.schema.Track(children=[clip_B, clip_C])

st = otio.schema.Stack(children=[tr_under, tr_over])

result = otio.algorithms.flatten_stack(st)

# does select A on top for the first clip
assert(result[0].name == 'track1_A')

# ...but trims C instead of including it
# the resulting track is just this:
# [    A     ]
assert(result[1].name == 'track2_C')

# these assertions also fail (see below)

# should have two clips
assert(len(result) == 2)

# with duration 300
assert(result.duration().value == 300)
```

If the stack is flipped (such that the track with two clips is above the track with one, then it functions as expected.

(original bug report follows below)

When using `flatten_stack` if a video track does not end with a clip or gap above another video track that does the child from the bottom video track will be omitted from the returned timeline.

## To Reproduce

1. Win 10
2. Python 3.9
3. OpenTimelineIO 0.14.1

## Expected Behavior

`flatten_stack` should insert a gap prior to flattening or account for this scenario when flattening so the clip on the bottom track is accounted for and included in the returned timeline

## Screenshots

Initial results, the top image from otioview is the original timeline and the bottom image from otioview is the result of `flatten_stack`
image

Tested this by inserting a new clip into the above track and re-ran `flatten_stack` and got the same result where the end of the bottom clip is being omitted from the returned timeline.

image

markreidvfx:
This PR fixes #1430

I added a `_normalize_tracks_lengths` function that adds gap items to the end of tracks that are shorter before the flattening happens.

### DIFF
--- a/src/opentimelineio/stackAlgorithm.cpp
+++ b/src/opentimelineio/stackAlgorithm.cpp
@@ -5,10 +5,12 @@
 #include "opentimelineio/track.h"
 #include "opentimelineio/trackAlgorithm.h"
 #include "opentimelineio/transition.h"
+#include "opentimelineio/gap.h"
 
 namespace opentimelineio { namespace OPENTIMELINEIO_VERSION {
 
 typedef std::map<Track*, std::map<Composable*, TimeRange>> RangeTrackMap;
+typedef std::vector<SerializableObject::Retainer<Track>> TrackRetainerVector;
 
 static void
 _flatten_next_item(
@@ -123,10 +125,49 @@ _flatten_next_item(
     }
 }
 
+// make all tracks the same length by adding a gap to the end if they are shorter
+static void
+_normalize_tracks_lengths(std::vector<Track*>& tracks,
+                          TrackRetainerVector& tracks_retainer,
+                          ErrorStatus*         error_status)
+{
+    RationalTime duration;
+    for (auto track: tracks)
+    {
+        duration = std::max(duration, track->duration(error_status));
+        if (is_error(error_status))
+        {
+            return;
+        }
+    }
+
+    for(int i = 0; i < tracks.size(); i++)
+    {
+        Track *track = tracks[i];
+        RationalTime track_duration = track->duration(error_status);
+        if (track_duration < duration)
+        {
+            track = static_cast<Track*>(track->clone(error_status));
+            if (is_error(error_status))
+            {
+                return;
+            }
+            tracks_retainer.push_back(SerializableObject::Retainer<Track>(track));
+            track->append_child(new Gap(duration - track_duration), error_status);
+            if (is_error(error_status))
+            {
+                return;
+            }
+            tracks[i] = track;
+        }
+    }
+}
+
 Track*
 flatten_stack(Stack* in_stack, ErrorStatus* error_status)
 {
     std::vector<Track*> tracks;
+    TrackRetainerVector tracks_retainer;
     tracks.reserve(in_stack->children().size());
 
     for (auto c: in_stack->children())
@@ -151,6 +192,12 @@ flatten_stack(Stack* in_stack, ErrorStatus* error_status)
         }
     }
 
+    _normalize_tracks_lengths(tracks, tracks_retainer, error_status);
+    if (is_error(error_status))
+    {
+        return nullptr;
+    }
+
     Track* flat_track = new Track;
     flat_track->set_name("Flattened");
 
@@ -168,6 +215,20 @@ flatten_stack(Stack* in_stack, ErrorStatus* error_status)
 Track*
 flatten_stack(std::vector<Track*> const& tracks, ErrorStatus* error_status)
 {
+    std::vector<Track*> flat_tracks;
+    TrackRetainerVector tracks_retainer;
+    flat_tracks.reserve(tracks.size());
+    for (auto track : tracks)
+    {
+        flat_tracks.push_back(track);
+    }
+
+    _normalize_tracks_lengths(flat_tracks, tracks_retainer, error_status);
+    if (is_error(error_status))
+    {
+        return nullptr;
+    }
+
     Track* flat_track = new Track;
     flat_track->set_name("Flattened");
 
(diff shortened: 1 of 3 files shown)


src/opentimelineio/stackAlgorithm.cpp
```diff
@@ -123,10 +125,49 @@ _flatten_next_item(
...
+        }
+    }
+
+    for(int i = 0; i < tracks.size(); i++)
+    {
+        Track *track = tracks[i];
+        RationalTime track_duration = track->duration(error_status);
+        if (track_duration < duration)
+        {
+            track = static_cast<Track*>(track->clone(error_status));
```
jminor, line 150: In reading this I got confused about the reuse of `track` as both the old one and the new one. Maybe `old_track` and `new_track` could make that more clear?

src/opentimelineio/stackAlgorithm.cpp
```diff
@@ -168,14 +215,28 @@ flatten_stack(Stack* in_stack, ErrorStatus* error_status)
 Track*
 flatten_stack(std::vector<Track*> const& tracks, ErrorStatus* error_status)
 {
+    std::vector<Track*> flat_tracks;
+    TrackRetainerVector tracks_retainer;
+    flat_tracks.reserve(tracks.size());
+    for (auto track : tracks)
+    {
+        flat_tracks.push_back(track);
+    }
```
jminor, line 224: Could you put a comment above this tracks_retainer section with a clue to future readers about why the retainer is needed? I _think_ I understand, but I had to stare at it for a while :)

src/opentimelineio/stackAlgorithm.cpp
```diff
@@ -123,10 +125,49 @@ _flatten_next_item(
     }
 }
 
+// make all tracks the same length by adding a gap to the end if they are shorter
```
jminor, line 128: Could you add a comment clarifying that this updates the `tracks` vector with a mix of old unchanged tracks and new padded tracks? At first read I was confused by that.

jminor: Also, do you think anyone else might want to call this function for other use cases? I could imagine that some adapters might want to have equal-duration padded tracks. We could leave that for a future PR.

jminor (changes requested):
Looks good! I left some requests for comments in a few places. Overall it looks correct to me.

src/opentimelineio/stackAlgorithm.cpp
markreidvfx: It could be useful, not sure. I can't think of anywhere atm in the AAF adapter this would be needed. I'm not sure if any other adapters.

markreidvfx:
I think I addressed your notes. please take a look when you get a chance :)

jminor (approved):
Looks great! Thanks Mark!

PR #1705, opened 2024-02-29
Add to_dict methods

JeanChristopheMorinPerso:
Still WIP. I will add tests later.

This is the result from https://academysoftwarefdn.slack.com/archives/CMQ9J4BQC/p1708976729390279.

### DIFF
--- a/src/py-opentimelineio/opentimelineio/core/_core_utils.py
+++ b/src/py-opentimelineio/opentimelineio/core/_core_utils.py
@@ -4,6 +4,7 @@
 import types
 import collections.abc
 import copy
+import json
 
 from .. import (
     _otio,
@@ -13,6 +14,11 @@ from .. _otio import (
     AnyDictionary,
     AnyVector,
     PyAny
+)
+from .. _opentime import (
+    RationalTime,
+    TimeRange,
+    TimeTransform
 )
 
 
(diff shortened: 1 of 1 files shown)


src/py-opentimelineio/opentimelineio/core/_core_utils.py
```diff
@@ -388,3 +394,57 @@ def __deepcopy__(self, *args, **kwargs):
 @add_method(SerializableObject)
 def __copy__(self, *args, **kwargs):
     raise ValueError("SerializableObjects may not be shallow copied.")
+
+
+@add_method(AnyDictionary)
+def to_dict(self):  # noqa: F811
```
JeanChristopheMorinPerso, line 400: I'm not convinced that this is the best name. Should we instead call it `to_py`? It would be more generic. I don't like that we have `to_dict` on everything except `AnyVector` which has `to_list`.

JeanChristopheMorinPerso:
This is ready to review. I would need help with the model regeneration. I'm not too sure what's going on with that.

src/py-opentimelineio/opentimelineio/core/_core_utils.py
ksallee: `to_json`?

JeanChristopheMorinPerso: There is no type named json in Python. `to_json` to me means "to JSON formatted string".

ksallee: OK I see what you mean. `to_py` is indeed more generic then, even if it's a bit less intuitive?

apetrynet: I'd prefer `to_dict` and `to_list` over `to_py`.
That way you know what you get. 
Of course the nature of the object will indicate what `to_py` returns, but I still prefer the explicitness of `to_dict` and `to_list`

douglascomet: I wouldn't mind using `to_dict` because I'm used to seeing and using this to make a complex structure a simple dictionary. To me it makes sense.

JeanChristopheMorinPerso: `to_py` would allow us to iterate through a list of OTIO objects (of any type) and convert them to python types without using `isinstance` to pick the right method.

reinecke:
I think we're thumbsup on adding this. The test failure is a bit of a head-scratcher - we'll need to look in [this vicinity](https://github.com/AcademySoftwareFoundation/OpenTimelineIO/blob/afcb55214b6ef8a625fe04e48766b784b888ff6e/src/py-opentimelineio/opentimelineio/console/autogen_serialized_datamodel.py#L318) to determine why the module path is coming out wrong in the docs.

apetrynet:
If this is doc related it could you'll need to run `make doc-plugins-update` locally and push again.
Ran into a similar issue on the adapter breakouts.

src/py-opentimelineio/opentimelineio/core/_core_utils.py
douglascomet: Follow up to this, I ended up having to make a helper function to do exactly this at work. I ended up calling the function `pythonify` to communicate its returning Python objects.

`to_py` seems in line with other conversion method names on OTIO objects and is clear of what to expect. Another reason to use a `py`/`python` name is because `AnyDictionary` and `AnyVector` can nest each other so you're not just getting a `dict` or `list` but rather the entire struct is `pythonify`-ed.

meshula (approved):
this looks good to me assuming the CI failures are unrelated to the implementation itself.

src/py-opentimelineio/opentimelineio/core/_core_utils.py
jminor: Can we take some naming inspiration from `@dataclass` or some other plain-old-data terminology?

PR #1712, opened 2024-03-26
Redact media reference names

### LINKED ISSUE #1678: otiotool --redact doesn't redact media reference names
### Incorrect Functionality

otiotool's `--redact` feature is supposed to remove all potentially sensitive information from an OTIO file, however it leaves the "name" attribute of MediaReference objects as-is. This could potentially reveal information about project names, talent, locations, dates, etc. which might be sensitive.

## To Reproduce

[redacted_media_out_of_bounds.otio.zip](https://github.com/AcademySoftwareFoundation/OpenTimelineIO/files/13245861/redacted_media_out_of_bounds.otio.zip)

1. Download attached OTIO file, which is already redacted, but still has media reference names.
2. Run `otiotool -i input.otio --redact -o output.otio
3. Notice that output.otio still has names like "i345_13c" in it.

## Expected Behavior

Media reference names should be replaced with anonymous names like "Media Reference #7".

peter-targett:
**Pull Request is related to.**

```Fixes #1678```

**Summarize your change.**

For `otiotool --redact` make sure to redact media reference names.

### CODE REVIEW

**Comment by jminor** [comment]
This looks great! The test failure(s) seem unrelated to this change.

Something about ```
ModuleNotFoundError: No module named 'importlib_metadata'
  error: subprocess-exited-with-error
```

I'll press the Retry button to see if that helps.

jminor (approved):

jminor:
Retry worked.

@apetrynet you're right that using a feature branch is a good practice, but I don't think we need @peter-targett to redo anything. The branch name will be irrelevant once this lands.

apetrynet:
Thank you for your contribution @peter-targett !

peter-targett:
I agree it could be on a branch but as 1 line change I did wonder if it could be pushed on main.Thanks, Peter On 26 Mar 2024, at 18:42, Joshua Minor ***@***.***> wrote:﻿
Retry worked.
@apetrynet you're right that using a feature branch is a good practice, but I don't think we need @peter-targett to redo anything. The branch name will be irrelevant once this lands.

PR #1713, opened 2024-03-27
Support for C++17 Part 2

### LINKED ISSUE #1526: Support for C++17
## Support for C++ 17

VfxPlatform CY2020 specifies C++14, but CY2021 and greater specify C++17.
We currently use c++17 shim libraries to get the functionality of optional and any.
Adopting C++17 would eliminate two dependencies from our build and thus simplify our build scripts, and simplify our build configuration.

## Description

When the next OpenTimelineIO release is tagged, we will note that it is the last release supporting CY2020.
We will go ahead at that point with removing the optional and any shims.
Should bugs arise that need to be addressed in that release for the CY2020 platform, we will create a maintenance branch and perform corrections as required in that branch. Subsequent point releases in support of CY2020 will be made against that branch.
Support for CY2020 will be dropped at the end of calendar year 2023.

## Context

Participants representing studios and vendors queried in the 2023 Jan 19 TSC meeting indicate they already building with C++17.

darbyjohnston:
Fixes #1526

Changes:
* Set the C++ version to 17 and the minimum macOS version to 10.13 in the top-level CMakelist.txt file
* Replace the use of `any` and `optional` classes with `std::any` and `std::optional`
* Remove the `any` and `optional` dependencies
* Removed the file `py-opentimelineio\bindings-common\casters.h` which is not necessary with C++17

This PR is a replacement for #1539 which is proving difficult to merge.


meshula (approved):
That's a lotta small touches! Thanks
Looks good.

PR #1715, opened 2024-03-30
Fixed AttributeError: editable_mode in setup.py

reinecke:
The `setup.py` build_py action uses the `editable_mode` attribute, however there are some invocation paths where this isn't set.

This adds defaulting behavior to avoid an `AttributeError` in those cases.

I validated this through the various `make`, `pip`, and `python setup.py...` commands.


apetrynet (approved):
LGTM

Issue #830, opened 2020-11-02
timecode problem with FCP XML converter


msheby:
### Incorrect Functionality and General Questions

We used `otioconvert` on a FCP file with content described as:
```
<rate>
  <timebase>24</timebase>
  <ntsc>TRUE</ntsc>
</rate>
```
and got an output OTIO file with a bunch of
```
"rate": 23.976023976023979
```
entries. So far so good. However, `otiostat` is unhappy with the resulting file:
```
parsed: True
top level object: Timeline.1
number of tracks: 3
Tracks are the same length: True
deepest nesting: 4
number of clips: 6
total duration: RationalTime(33820, 23.976)
There was a system error: invalid timecode rate
top level rate: 23.97602397602398
clips with cdl data: 0
Tracks with non standard types: 0
```

## To Reproduce

1. CentOS Linux release 7.2.1511 (Core)
2. Python 3.6.8
3. n/a
4. OpenTimelineIO 0.13.0 (via pip install)
5.
```
Using built-in specs.
COLLECT_GCC=gcc
COLLECT_LTO_WRAPPER=/usr/libexec/gcc/x86_64-redhat-linux/4.8.5/lto-wrapper
Target: x86_64-redhat-linux
Configured with: ../configure --prefix=/usr --mandir=/usr/share/man --infodir=/usr/share/info --with-bugurl=http://bugzilla.redhat.com/bugzilla --enable-bootstrap --enable-shared --enable-threads=posix --enable-checking=release --with-system-zlib --enable-__cxa_atexit --disable-libunwind-exceptions --enable-gnu-unique-object --enable-linker-build-id --with-linker-hash-style=gnu --enable-languages=c,c++,objc,obj-c++,java,fortran,ada,go,lto --enable-plugin --enable-initfini-array --disable-libgcj --with-isl=/builddir/build/BUILD/gcc-4.8.5-20150702/obj-x86_64-redhat-linux/isl-install --with-cloog=/builddir/build/BUILD/gcc-4.8.5-20150702/obj-x86_64-redhat-linux/cloog-install --enable-gnu-indirect-function --with-tune=generic --with-arch_32=x86-64 --build=x86_64-redhat-linux
Thread model: posix
gcc version 4.8.5 20150623 (Red Hat 4.8.5-4) (GCC) 
```

## Expected Behavior

[...]

PR #1717, opened 2024-04-04
Add implementation of to_nearest_timecode

vade:
Add implementation of `to_nearest_timecode` which makes approximate conversion of timecode from a rate using the closest valid time code rate.

This is intended to fix various downstream adaptor hiccups where source material might be variable frame rate, or close but not quite an exact SMPTE timecode.

https://github.com/AcademySoftwareFoundation/OpenTimelineIO/issues/830

**Summarize your change.**

This PR does not directly fix the FCP_XML adaptor, but rather isolates changes to OTIO core and exposes a `to_nearest_timecode` with the same method signature as `to_timecode`. 

This is intended to act as a entry point for changes to other adaptors if deemed a valid approach

No new tests just yet.

### DIFF
--- a/src/opentime/rationalTime.cpp
+++ b/src/opentime/rationalTime.cpp
@@ -571,6 +571,26 @@ RationalTime::to_timecode(
 }
 
 std::string
+RationalTime::to_nearest_timecode(
+    double          rate,
+    IsDropFrameRate drop_frame,
+    ErrorStatus*    error_status) const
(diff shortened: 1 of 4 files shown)


meshula (approved):
lgtm

vade:
Thanks for your patience. DCO Sign off passed. I'll try to keep that in mind in the future!

PR #1726, opened 2024-04-12
Fixed freeze script github yaml parsing and unfroze the ci versions

reinecke:
The freeze CI versions script had broken with an update to the GitHub documentation formatting. This update uses a slightly more markdown-aware approach to extract the needed info.

Also addressed minor escaping issue in Makefile I'd discovered while cutting the 0.16.0 release.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -251,7 +251,7 @@ update-contributors: check-github-token
 	@echo "Updating CONTRIBUTORS.md..."
 	@python maintainers/fetch_contributors.py \
 		--repo AcademySoftwareFoundation/OpenTimelineIO \
-		--token $(OTIO_RELEASE_GITHUB_TOKEN)
+		--token "${OTIO_RELEASE_GITHUB_TOKEN}"
 
 dev-python-install:
 	@python setup.py install
--- a/Makefile
+++ b/Makefile
@@ -251,7 +251,7 @@ update-contributors: check-github-token
 	@echo "Updating CONTRIBUTORS.md..."
 	@python maintainers/fetch_contributors.py \
 		--repo AcademySoftwareFoundation/OpenTimelineIO \
-		--token "${OTIO_RELEASE_GITHUB_TOKEN}"
+		--token $(OTIO_RELEASE_GITHUB_TOKEN)
 
 dev-python-install:
 	@python setup.py install
--- a/maintainers/freeze_ci_versions.py
+++ b/maintainers/freeze_ci_versions.py
@@ -49,7 +49,7 @@ def _parsed_args():
     return parser.parse_args()
 
 
-def fetch_plat_map():
+def _fetch_plat_map():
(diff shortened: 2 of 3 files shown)


maintainers/freeze_ci_versions.py
```diff
@@ -49,24 +49,69 @@ def _parsed_args():
...
+                label for label in available_labels
+                if label.split("-")[-1] == "latest"
             )
+        except StopIteration:
+            # not a latest label
+            continue
+
+        label = next(
+            label for label in available_labels
+            if label.split("-")[-1] not in {"xl", "latest"}
```
apetrynet, line 101: There doesn't seem to be any "xl" only "xlarge"
```suggestion
            if label.split("-")[-1] not in {"xl", "xlarge", "latest"}
```
Keeping "xl" just in case

Makefile
```diff
@@ -259,7 +259,7 @@ update-contributors: check-github-token
 	@echo "Updating CONTRIBUTORS.md..."
 	@python maintainers/fetch_contributors.py \
 		--repo AcademySoftwareFoundation/OpenTimelineIO \
-		--token $(OTIO_RELEASE_GITHUB_TOKEN)
+		--token "${OTIO_RELEASE_GITHUB_TOKEN}"
```
ssteinbach, line 262: are you sure about this change?  The `$(OTIO_RELEASE_GITHUB_TOKEN)` means the variable is being subsituted in make, but the extra quoting and braces mean its going to be evaluated in the subshell...I'm not sure if there is a difference.  Is that because you don't want the token to appear in the shell output?

maintainers/freeze_ci_versions.py
```diff
@@ -49,24 +49,69 @@ def _parsed_args():
...
+            continue
+        entries[platform] = label
+
+    return entries
+
+
+def main():
+    args = _parsed_args()
+
+    plat_map = fetch_plat_map()
```
ssteinbach, line 114:
```suggestion
    plat_map = _fetch_plat_map()
```

maintainers/freeze_ci_versions.py
```diff
@@ -49,24 +49,69 @@ def _parsed_args():
     return parser.parse_args()
 
 
-def main():
-    args = _parsed_args()
-
+def fetch_plat_map():
```
ssteinbach, line 52:
```suggestion
def _fetch_plat_map():
```

ssteinbach (commented):
Seems fine, this rearranges how the non-latest versions are fetched AFAICT?  I had one question in there on the Makefile, otherwise seems fine.

Makefile
reinecke: I'll admit, I'm not sure about this. It seemed like it'd fixed an issue with the token not coming through properly. I'm going to revert this change though and I'll see if everything goes through okay next release without it.

apetrynet (approved):
lgtm

PR #1727, opened 2024-04-12
Update codecov to run on python 3.10 builds

ssteinbach:
Codecov uploader is no longer compatible w/ python 3.7, which was causing action failures against the python 3.7 target.  This switches it to use the 3.10 version of python instead.  It also updates to the v4 codecov uploader.


reinecke (approved):
LGTM!

PR #1741, opened 2024-04-15
do not break the build if codecov fails

ssteinbach:
Make sure that problems uploading to codecov do not indicate broken PRs.


meshula (approved):
Approving in order to unblock all other automation

JeanChristopheMorinPerso (approved):
That makes sense to me.

PR #1746, opened 2024-05-09
Add Doxygen comments to RationalTime

### LINKED ISSUE #708: Doxygen documentation for the C++ code

### LINKED ISSUE #1753: Add Doxygen comments to the C++ API
# Required:
------------
[ ] I believe this isn't a duplicate topic  
[x ] This report is not related to an adapter  

## Select One:

[ ] Build problem  
[ ] Incorrect Functionality or bug  
[x ] New feature or functionality  

## Description
As a follow up to #708, this issue covers adding Doxygen style comments to the C++ API. Specifically:

* All public C++ API should be documented (e.g., classes, structs, enumerations, member functions, etc.)
* C++ style comments are used:
```
/// @brief Returns the time value converted to a new rate.
constexpr double value_rescaled_to(double new_rate) const noexcept;
```
* Comments should always have an explicit `brief` section; detailed descriptions, parameter strings, and return strings are optional for shorter functions:
```
/// @brief Returns the time value converted to a new rate.
constexpr double value_rescaled_to(double new_rate) const noexcept;
```
```
/// @brief Parse a string in the form "hours:minutes:seconds".
///
/// The string may have a leading negative sign. Seconds may have up to microsecond precision.
///
/// @param time_string The time string.
/// @param rate The time rate.
/// @param error_status Optional error status.
static RationalTime from_timecode(
        std::string const& timecode,
        double rate,
        ErrorStatus* error_status = nullptr);
```
* Given the amount of work involved, it should probably be broken into multiple PRs.

Example of the Doxygen HTML documentation:

OTIO_HTML

Example Doxygen tooltip in Visual Studio:

OTIO_VS_Tooltip

Example Doxygen tooltip in Xcode:

OTIO_Xcode_Tooltip

darbyjohnston:

Fixes #1753

**Summarize your change.**

Adds Doxygen comments to the RationalTime class. These comments are based on the Python doc strings with some editing for consistency. A couple of notes:

* I kept the Doxygen markup as simple as possible to keep it easy to write, and to allow it to be parsed by IDEs like Visual Studio
* I made a couple of modifications to the Doxygen configuration file to support the simpler markup
* This is just a first pass at the comments, it would be nice to get some feedback to help improve them

Here is a screenshot of the Doxygen comments providing help in Visual Studio:

OTIO Doxygen VS Tooltip

Here is a screenshot of the HTML generated for RationalTime:

OTIO Doxygen HTML


meshula:
Overall this is a great improvement.

I'm not familiar with ABBREVIATE_BRIEF. Does it preclude documenting the parameters formally?

darbyjohnston:
The ABBREVIATE_BRIEF option is a bit odd, it filters the text to try and make it more concise. So if your comment looks like this:

`
/// The Addition class provides functionality for adding numbers.
`

It will be changed to: "Functionality for adding numbers". I always find it a bit surprising so I like to turn it off.

If these changes look OK how should we go about documenting the rest of the code? It would be nice to do the work in chunks to keep it manageable, maybe one PR per class?

meshula:
My first question about ABBREVIATE_BRIEF was really poking at whether that was supposed to do more, like infer parameters and returns.

I'd prefer we go heavy-handed with documentation, ie with full adornment. I assume ABBREVIATE_BRIEF means that we can skip the `@brief` tag. Sometimes this feels overboard, but I think it's better to be thorough and consistent in order that all the tooling around doxygen syntax can come into play.

```
/**
 * @brief Converts a XYZ color to RGB color space.
 * 
 * Converts a XYZ color to RGB color space using the provided color space.
 * 
 * @param cs Pointer to the color space object.
 * @param xyz The XYZ color to convert.
 * @return The RGB color.
 */
NCAPI NcRGB NcXYZToRGB(const NcColorSpace* cs, NcXYZ xyz);
```

darbyjohnston:
The ABBREVIATE_BRIEF option doesn't do any inference, it's just for filtering text. To automate brief descriptions I was trying the options MULTILINE_CPP_IS_BRIEF=YES and REPEAT_BRIEF=YES, so a comment like this:

```
/// Converts a XYZ color to RGB color space. The conversion is performed
/// using the provided color space.
```

Will have the first sentence used for the brief description, and the entire text used for the full description.

I like the @param and @return keywords, though I wonder if we can omit them for simple functions, especially class member get methods?

meshula:
Sure, why don't we have a look at that, it sounds fine to me.

darbyjohnston:
After more testing the MULTILINE_CPP_IS_BRIEF option didn't work very well so I turned it off and added explicit `@brief` markup as you suggested.

I think the detailed description, `@param` markup, and `@return` markup should be optional since it adds a lot of boiler plate for simple functions like member get functions. They should just be used where necessary to provide extra information that isn't obvious from the function name.

Some other notes:
* I disabled the Latex and XML output since they are probably not necessary?
* I used `///` for comments since it's a bit easier to type than the C style comments, but can change that if there is already a precedent for using the C style comments.

meshula (approved):
For more complex functions we could use @return, but I don't spot anything here that really warrant's it. I think this is a good pattern to start with.

darbyjohnston:
I created a new issue since the original Doxygen issue was closed awhile back. Given the amount of work maybe we can use that issue as an umbrella for multiple PRs.

Any ideas on how to fix the CI build, the error doesn't seem related to this change?

meshula:
The problem in the CI is that pip on windows under msys is failing because distlib can't be found. @reinecke does that ring any bells? I feel like we struggled through something like that last year but the details escape me.

darbyjohnston:
Strange, it looks like distlib is being installed as part of the msys setup:
```
installing mingw-w64-x86_64-python-distlib...
```

darbyjohnston:
Sorry, was trying to follow the GitHub rebasing steps and seemed to have created a mess again... I may just create a new PR instead of trying to fix this one.

PR #1749, opened 2024-05-16
Fix sign-compare warning when compiling with GCC 14

JeanChristopheMorinPerso:
Simple pull request to remove a new warning under GCC 14.

The warning is
```
/home/jcmorin/jcmenv/aswf/OpenTimelineIO/src/opentimelineio/stackAlgorithm.cpp: In function ‘void opentimelineio::v1_0::_normalize_tracks_lengths(std::vector<Track*>&, TrackRetainerVector&, ErrorStatus*)’:
/home/jcmorin/jcmenv/aswf/OpenTimelineIO/src/opentimelineio/stackAlgorithm.cpp:147:23: warning: comparison of integer expressions of different signedness: ‘int’ and ‘std::vector<opentimelineio::v1_0::Track*>::size_type’ {aka ‘long unsigned int’} [-Wsign-compare]
  147 |     for (int i = 0; i < tracks.size(); i++)
      |                     ~~^~~~~~~~~~~~~~~
```

### DIFF
--- a/src/opentimelineio/stackAlgorithm.cpp
+++ b/src/opentimelineio/stackAlgorithm.cpp
@@ -7,6 +7,8 @@
 #include "opentimelineio/trackAlgorithm.h"
 #include "opentimelineio/transition.h"
 
+#include <cstddef>
+
 namespace opentimelineio { namespace OPENTIMELINEIO_VERSION {
 
 typedef std::map<Track*, std::map<Composable*, TimeRange>> RangeTrackMap;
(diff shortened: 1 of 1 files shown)


src/opentimelineio/stackAlgorithm.cpp
```diff
@@ -144,7 +146,7 @@ _normalize_tracks_lengths(
         }
     }
 
-    for (int i = 0; i < tracks.size(); i++)
+    for (std::size_t i = 0; i < tracks.size(); i++)
```
meshula, line 149: tracks.size() returns a size_t; if you just use size_t, you don't need to include the c header file.

meshula (changes requested):

meshula (approved):
awesome, thanks

PR #1755, opened 2024-06-04
Fix for find_children() with stacks

### LINKED ISSUE #1652: find_children is broken with two tracks.

darbyjohnston:
Fixes #1652

After some investigation, it looks like the `find_children()` function was not working with stacks when given a search range. To fix this, an override is added for the function `children_in_range()` that works with "stacks" of children.

### DIFF
--- a/src/opentimelineio/composition.h
+++ b/src/opentimelineio/composition.h
@@ -101,7 +101,7 @@ public:
         bool                shallow_search = false) const;
 
     // Return all objects within the given search_range.
-    std::vector<Retainer<Composable>> children_in_range(
+    virtual std::vector<Retainer<Composable>> children_in_range(
         TimeRange const& search_range,
         ErrorStatus*     error_status = nullptr) const;
 
--- a/src/opentimelineio/stack.cpp
+++ b/src/opentimelineio/stack.cpp
@@ -79,6 +79,26 @@ Stack::range_of_all_children(ErrorStatus* error_status) const
     return result;
 }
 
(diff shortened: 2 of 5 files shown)


jminor (approved):
This is a great find & a clear fix.
Skimming the history of the codebase, I think this bug has existed since before the port from Python to C++. The implementation of children_in_range in Composition only works correctly for Tracks. This bug was likely introduced by the bisection optimization, but I didn't look back that far in the history to confirm.
I wonder if there's a general implementation that could go into Composition, and only the optimized one could be in Track, but since there are only 2 subclasses of Composition anyways, I think your fix is great and should be long lived.

jminor:
Oh, I thought of one other edge case... Could you try this with a Track that has a source_range trim on it? I suspect that the new function you made might need to use `trimmed_range_in_parent` instead of `trimmed_range`.

darbyjohnston:
I didn't realize tracks could have a source range, how does that affect the clips? If the track source range duration is shorter than the clips, are they trimmed?

darbyjohnston:
Replaced by #1766.

PR #1766, opened 2024-06-14
Fix for find_children() with stacks #2

### LINKED ISSUE #1652: find_children is broken with two tracks.

darbyjohnston:
Fixes #1652

This PR is a replacement for #1755 which has some trouble with the git history.

After some investigation, it looks like the find_children() function was not working with stacks when given a search range. To fix this, an override is added for the function children_in_range() that works with "stacks" of children.

### DIFF
--- a/src/opentimelineio/composition.h
+++ b/src/opentimelineio/composition.h
@@ -101,7 +101,7 @@ public:
         bool                shallow_search = false) const;
 
     // Return all objects within the given search_range.
-    std::vector<Retainer<Composable>> children_in_range(
+    virtual std::vector<Retainer<Composable>> children_in_range(
         TimeRange const& search_range,
         ErrorStatus*     error_status = nullptr) const;
 
--- a/src/opentimelineio/stack.cpp
+++ b/src/opentimelineio/stack.cpp
@@ -85,12 +85,12 @@ Stack::children_in_range(
     ErrorStatus* error_status) const
 {
     std::vector<SerializableObject::Retainer<Composable>> children;
-    for (auto child : this->children())
+    for (const auto& child : this->children())
     {
-        if (auto item = dynamic_retainer_cast<Item>(child))
+        if (const auto& item = dynamic_retainer_cast<Item>(child))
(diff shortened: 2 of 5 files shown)


darbyjohnston:
@jminor Sorry, I managed to mess up the git history on the previous PR while trying to update with the latest changes from main. I created this new PR as a replacement.

I believe your last comment was about using `trimmed_range_in_parent()` and testing with a source range on the track? I didn't realize that tracks could have source ranges, what would that test look like?

jminor:
There's an example of a track with a source_range in Figure 3 here: 
https://opentimelineio.readthedocs.io/en/stable/tutorials/otio-timeline-structure.html
In that diagram, Track-003 is offset by using a source_range with a negative start time. A simpler use case would be to use a track's source range to just trim a nested track to a sub-range of the whole track.

There's also an example of a stack with a source_range in Figure 4 on that same page. In that case the source_range is used to trim the nested stack. Although rare, the source_range could be set on the top level stack of a timeline also.

darbyjohnston:
Thanks, those figures were very helpful; I guess I should have checked there first. :)

I changed the code to use `trimmed_range_in_parent()` and added extra C++ tests that put various source ranges on the track. If the C++ tests look OK I will update the Python tests to match.

darbyjohnston:
@jminor Just checking if you have time to review the latest changes?

jminor (approved):
Looks great. The use of trimmed_range_in_parent looks correct to me :)

darbyjohnston:
Thanks!

PR #1779, opened 2024-07-10
#1677 otiotool verify ranges

### LINKED ISSUE #1677: otiotool --verify-ranges
## Description

It would be really helpful if `otiotool` had a feature to scan all clips in a timeline and report if any have source_range which exceeds the available_range of media.

## Context

OTIO allows clips to use media outside the bounds of available media because some tools and workflows rely on this to serve as a request for new media (e.g. a new take, or new render) that is longer. However some tools don't support this case, and produce errors or incorrect playback, so it would be great if there was a quick way to check an OTIO for this situation.

See attached OTIO which has one clip (named "Clip #5") with an out-of-bounds source_range.
[redacted_media_out_of_bounds.otio.zip](https://github.com/AcademySoftwareFoundation/OpenTimelineIO/files/13245821/redacted_media_out_of_bounds.otio.zip)

semagnum:
Fixes #1677

**Summary**

Adds `--verify-ranges` option for otiotool's inspection feature, verifies that the clip's source range doesn't start before the available start range or ends beyond it (if the available range exists).

### DIFF
--- a/src/py-opentimelineio/opentimelineio/console/otiotool.py
+++ b/src/py-opentimelineio/opentimelineio/console/otiotool.py
@@ -870,8 +870,8 @@ def summarize_timeline(list_tracks, list_clips, list_media, verify_media,
                 print(f"TRACK: {child.name} ({child.kind})")
         if isinstance(child, otio.schema.Clip):
             if list_clips or verify_ranges:
-                range_msg = ""
                 if verify_ranges:
+                    range_msg = ""
                     try:
                         source = child.source_range
                         available = child.available_range()
@@ -889,7 +889,9 @@ def summarize_timeline(list_tracks, list_clips, list_media, verify_media,
                             range_msg = "IN BOUNDS"
                     except Exception:  # available range is, well, unavailable
                         pass
-                print("  CLIP:", child.name, range_msg)
+                    print("  CLIP:", child.name, range_msg)
+                else:
+                    print("  CLIP:", child.name)
             if list_media or verify_media:
                 try:
                     url = child.media_reference.target_url
--- a/src/py-opentimelineio/opentimelineio/console/otiotool.py
+++ b/src/py-opentimelineio/opentimelineio/console/otiotool.py
@@ -137,7 +137,8 @@ def main():
                         args.list_media or
                         args.verify_media or
                         args.list_tracks or
-                        args.list_markers)
+                        args.list_markers or
+                        args.verify_ranges)
     if should_summarize:
         for timeline in timelines:
             summarize_timeline(
@@ -146,6 +147,7 @@ def main():
                 args.list_media,
                 args.verify_media,
                 args.list_markers,
+                args.verify_ranges,
                 timeline)
 
     # Final Phase: Output
@@ -207,7 +209,7 @@ This tool works in phases, as follows:
 
 6. Inspect
     Options such as --stats, --list-clips, --list-tracks, --list-media,
-    --verify-media, --list-markers, and --inspect will examine the OTIO and
+    --verify-media, --list-markers, --verify-ranges, and --inspect will examine the OTIO and
     print information to standard output.
 
 7. Output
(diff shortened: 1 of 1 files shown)


meshula (approved):
lgtm

jminor:
This change looks good to me. One thing we should clarify though is that there are plenty of cases where it is valid for an OTIO Clip to reference media outside the available range of its media. Having this tool to check for the cases where that isn't okay is helpful for sure, but that isn't always an error case.

semagnum:
> One thing we should clarify though is that there are plenty of cases where it is valid for an OTIO Clip to reference media outside the available range of its media. Having this tool to check for the cases where that isn't okay is helpful for sure, but that isn't always an error case.

Agreed. Are you wanting that clarified in the verify message, or in the argument help?

jminor:
I think something brief in the usage message for that option would be wonderful. Maybe it could say something like this?

"Verify that each clip in a timeline has a source range within the available range of media (acceptable in some use cases, not in others)"

That would be just enough of a hint to users that it depends on the use case.

Thanks :)

src/py-opentimelineio/opentimelineio/console/otiotool.py
```diff
@@ -861,8 +870,29 @@ def summarize_timeline(list_tracks, list_clips, list_media, verify_media,
...
+                        avail_dur = available.duration.value
+
+                        starts_early = src_start < avail_start
+                        ends_late = (src_dur - src_start) > (avail_dur - avail_start)
+                        if starts_early or ends_late:
+                            range_msg = "SOURCE MEDIA OUT OF BOUNDS"
+                        else:
+                            range_msg = "IN BOUNDS"
+                    except Exception:  # available range is, well, unavailable
+                        pass
```
ssteinbach, line 892: So the error condition is that both available and source ranges are present but the source range is not within the available range?

src/py-opentimelineio/opentimelineio/console/otiotool.py
```diff
@@ -861,8 +870,29 @@ def summarize_timeline(list_tracks, list_clips, list_media, verify_media,
...
+                        source = child.source_range
+                        available = child.available_range()
+
+                        src_start = source.start_time.value
+                        src_dur = source.duration.value
+                        avail_start = available.start_time.value
+                        avail_dur = available.duration.value
+
+                        starts_early = src_start < avail_start
+                        ends_late = (src_dur - src_start) > (avail_dur - avail_start)
```
ssteinbach, line 886: I think you can use the time algebra to answer this question... see:

https://github.com/AcademySoftwareFoundation/OpenTimelineIO/blob/42158ebd33eb4ddf5dc2843a63998b4d8523be54/src/opentime/timeRange.h#L145

I think what you want is available_range contains source_range

ssteinbach (changes requested):

src/py-opentimelineio/opentimelineio/console/otiotool.py
semagnum: Correct, if there's no available range defined than it won't do the bounds check.

src/py-opentimelineio/opentimelineio/console/otiotool.py
semagnum: Good call, the `contains()` didn't seem to work for me initially when the source range matched available range exactly. But that'd be an easy check

semagnum: So it turns out `contains()` uses `end_time_exclusive()`, which means it fails if the source and available ranges end at exactly the same time (such as using the entire available range). But using `end_time_inclusive()` drastically simplifies the time algebra. Pushed the change.

jminor: @ssteinbach this is the same issue that @reinecke and I talked about a while back (on Slack I think?). There is/was a suggestion to add a convenience for this case where `contains()` doesn't behave as people expect. I think we can leave the code in this PR as-is and address that API change some other time.

jminor (approved):

ssteinbach (approved):

PR #1787, opened 2024-07-29
Offer alternative for cxxabi on Windows (Swift+Clang).

furby-tm:
**Summarize your change.**

This fixes the clang compilation of **OpenTimelineIO** on **Microsoft Windows**, when compiling with Swift. Previously, the `cxxabi.h` header was erroneously getting included in the Windows clang compilation of OpenTimelineIO's `stringUtils.cpp` file because it was being conditionally compiled for clang across all platforms, and this header does not exist on the Windows platform.

- The include for `cxxabi.h` is now guarded via `__has_include(<cxxabi.h>)`.
- A new preprocessor definition for `OTIO_HAVE_DEMANGLER` is set to **`1`** if `cxxabi.h` exists, and **`0`** if this header does not exist.
- Previous checks against `defined(__GNUC__) || defined(__clang__)` are replaced with conditional compilation checks against the new `OTIO_HAVE_DEMANGLER` preprocessor define.

### DIFF
--- a/src/opentimelineio/stringUtils.cpp
+++ b/src/opentimelineio/stringUtils.cpp
@@ -4,14 +4,20 @@
 #include "opentimelineio/serializableObject.h"
 #if defined(__GNUC__) || defined(__clang__)
 #    include <cstdlib>
-#    include <cxxabi.h>
+#    if __has_include(<cxxabi.h>)
+#       define OTIO_HAVE_DEMANGLER 1
+#       include <cxxabi.h>
+#    else // !__has_include(<cxxabi.h>)
+#       define OTIO_HAVE_DEMANGLER 0
+#    endif // __has_include(<cxxabi.h>)
 #    include <memory>
 #else
+#    define OTIO_HAVE_DEMANGLER 0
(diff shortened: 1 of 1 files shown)


meshula (approved):
Thanks for this contribution, this looks good to me. 

I do have a question for the team, perhaps the cxxabi functions are rather overkill for this application. I'd prefer that we simplify this code to not do this demangling, I don't honestly think it's adding value.

furby-tm:
If the team prefers to remove the cxxabi demangling, I can update this PR to deprecate it instead. Please feel free to ping me in this issue once the team has reached a consensus.

jminor:
From some offline discussion, it seems there is no objection to removing the demangling. @furby-tm we would gladly accept this PR as-is, or with the demangling removed. Please let us know which you'd like.

furby-tm:
I'll go ahead and remove the de-mangling @jminor

furby-tm:
Closing in favor of https://github.com/AcademySoftwareFoundation/OpenTimelineIO/pull/1800

PR #1789, opened 2024-08-26
Disable 32-bit Python builds

darbyjohnston:
Fixes #1788 

This PR disables x86 builds to help reduce CI time and troubleshooting.

Marking this as WIP to test the changes.

PR #1795, opened 2024-09-06
Add Python build for Linux on ARM architecture

tomviner:
**Summarize your change.**

Add support for building wheels for Linux on ARM architecture.

This covers both ARM64 hardware running Linux directly, as well as Docker containers running under other operating systems, such as Docker on MacOS.

Our use case is to speed up installs of the package in Docker running on Apple M1 MacBooks.

See docs https://cibuildwheel.pypa.io/en/stable/faq/#emulation
Note there is a time penalty to building this wheel under emulation.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -193,11 +193,8 @@ jobs:
         env:
           CIBW_BUILD: ${{ matrix.python-build }}
           CIBW_SKIP: '*musllinux*'
-          # configure cibuildwheel to build native archs ('auto'), and some
-          # emulated ones
-          CIBW_ARCHS_LINUX: auto aarch64
+          CIBW_ARCHS_LINUX: x86_64 aarch64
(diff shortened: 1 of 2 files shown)


mmcardle:
👍

jminor:
This sounds like a good idea. @tomviner can you just double check that this PR doesn't conflict/overlap with this which just landed? https://github.com/AcademySoftwareFoundation/OpenTimelineIO/pull/1789

tomviner:
> This sounds like a good idea. @tomviner can you just double check that this PR doesn't conflict/overlap with this which just landed? #1789

Good point. I've added some changes to take account of this.

Now we'll see just `x86_64` and `aarch64` (and not `i686`) getting built under the Ubuntu wheels.

JeanChristopheMorinPerso (approved):
LGTM, thanks for creating this PR!

PR #1798, opened 2024-09-27
Updated Effect schema with enabled flag

### LINKED ISSUE #1187: Add enabled Flag to Effects
New Feature

## Description

In #1175 the `enabled` flag was added to the `Item` schema. This would also be useful to add to effects as well.

#1186 addresses adding this flag for Transitions.

## References

- [FCP 7 XML models this using the enabled element](https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/FinalCutPro_XML/Elements/Elements.html#//apple_ref/doc/uid/TP30001156-TPXREF104)

fschleich:
**This PR addresses #1187 - Add enabled Flag to Effects**

```Fixes #1187```

**Summarize your change.**

- Added enabled flag to Effect schema
~~- Updated Effect schema to version 2~~
~~- Updated core version to 0.18.0~~

- Updated unit tests in `test_effect.py` to account for enabled flag
- added `empty_effect.json` and updated `test_json_backend.py` accordingly

### CODE REVIEW

**Comment by fschleich** [comment]
I need some guidance on how to correctly update `CORE_VERSION_MAP.cpp`, as running `make version-map-update` results in the below error. Manually setting the version to `0.18.0` results in the failed tests, e.g. [here](https://github.com/AcademySoftwareFoundation/OpenTimelineIO/actions/runs/11075330036/job/30776039254?pr=1798). Would appreciate any pointers or guidance!

```
(.venv) fschleich@Florians-MacBook-Pro OpenTimelineIO % make version-map-update
updating the CORE_VERSION_MAP...
  File "/Users/fschleich/Developer/github/fschleich/OpenTimelineIO/src/py-opentimelineio/opentimelineio/console/autogen_version_map.py", line 116, in <module>
    main()
  File "/Users/fschleich/Developer/github/fschleich/OpenTimelineIO/src/py-opentimelineio/opentimelineio/console/autogen_version_map.py", line 86, in main
    args = _parsed_args()
           ^^^^^^^^^^^^^^
  File "/Users/fschleich/Developer/github/fschleich/OpenTimelineIO/src/py-opentimelineio/opentimelineio/console/autogen_version_map.py", line 38, in _parsed_args
    default=otio.__version__,
            ^^^^^^^^^^^^^^^^
AttributeError: module 'opentimelineio' has no attribute '__version__'
make: *** [version-map-update] Error 1
```

jminor:
A while back when we added `enabled` to Clip, we were able to do so without increasing the schema version because the new property has a clear fallback value (true). I think you can do the same here for Effect. Here's the PR for Clip for reference: https://github.com/AcademySoftwareFoundation/OpenTimelineIO/pull/1175

reinecke (approved):
Looks great - thanks for the update!

PR #1799, opened 2024-09-27
Bindings: UnknownSchema added data property

### LINKED ISSUE #1234: UnknownSchema objects should allow access to data fields as a dictionary
## Feature Request

To allow developers to access data on objects of `UnknownSchema`, a dictionary-type interface should be exposed for the data fields.

## Context

When the deserializer doesn't recognize the schema of an object, it deserializes to an instance of `UnknownSchema`. Unfortunately, the `UnknownSchema` object only exposes `original_schema_name` and `original_schema_version` - this means the majority data in the object is opaque and inaccessible.

See #1230 for more discussion on this.

natchar:
Fixes #1234 

Python bindings return dictionary-type python object for UnknownSchema data.

### DIFF
--- a/src/opentimelineio/unknownSchema.h
+++ b/src/opentimelineio/unknownSchema.h
@@ -39,6 +39,11 @@ public:
         return _original_schema_version;
     }
 
+    AnyDictionary data() const noexcept
+    {
+        return _data;
+    }
+
     bool read_from(Reader&) override;
     void write_to(Writer&) const override;
 
--- a/src/opentimelineio/unknownSchema.h
+++ b/src/opentimelineio/unknownSchema.h
@@ -39,6 +39,7 @@ public:
         return _original_schema_version;
     }
 
+    /// @brief Return a copy of the data.
     AnyDictionary data() const noexcept
     {
         return _data;
--- a/src/py-opentimelineio/opentimelineio-bindings/otio_serializableObjects.cpp
+++ b/src/py-opentimelineio/opentimelineio-bindings/otio_serializableObjects.cpp
@@ -301,6 +301,9 @@ define_bases1(py::module m)
     py::class_<UnknownSchema, SerializableObject, managing_ptr<UnknownSchema>>(
         m,
         "UnknownSchema")
+        .def_property_readonly("data", [](UnknownSchema* schema) {
+                auto ptr = schema->data().get_or_create_mutation_stamp();
+            return (AnyDictionaryProxy*)(ptr); }, py::return_value_policy::take_ownership)
         .def_property_readonly(
             "original_schema_name",
             &UnknownSchema::original_schema_name)
--- a/src/py-opentimelineio/opentimelineio-bindings/otio_serializableObjects.cpp
+++ b/src/py-opentimelineio/opentimelineio-bindings/otio_serializableObjects.cpp
@@ -302,8 +302,9 @@ define_bases1(py::module m)
         m,
         "UnknownSchema")
         .def_property_readonly("data", [](UnknownSchema* schema) {
-                auto ptr = schema->data().get_or_create_mutation_stamp();
-            return (AnyDictionaryProxy*)(ptr); }, py::return_value_policy::take_ownership)
+                AnyDictionary copy = schema->data();
+                return any_to_py(copy, true /*top_level*/);
+            })
         .def_property_readonly(
             "original_schema_name",
             &UnknownSchema::original_schema_name)
--- a/tests/test_unknown_schema.py
+++ b/tests/test_unknown_schema.py
@@ -63,6 +63,32 @@ class UnknownSchemaTests(unittest.TestCase, otio_test_utils.OTIOAssertions):
         unknown = self.orig.media_reference.metadata["stuff"]
         self.assertTrue(unknown.is_unknown_schema)
 
+    def test_unknown_to_dict(self):
+        unknown = self.orig.media_reference.metadata["stuff"]
+        self.assertTrue(unknown.is_unknown_schema)
+        unknown_data = unknown.data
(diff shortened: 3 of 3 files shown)


jminor:
This looks good to me. Once this lands, then this feature in Raven should be easy to implement: https://github.com/OpenTimelineIO/raven/issues/26

@ssteinbach do you have any concerns with this?

src/opentimelineio/unknownSchema.h
```diff
@@ -31,6 +31,11 @@ class UnknownSchema : public SerializableObject
         return _original_schema_version;
     }
 
+    AnyDictionary& data() noexcept
```
meshula, line 34: is the intention to provide mutable access? Reading the original issue it talks about accessing dictionaries presumably for reading. A reference opens the door to lifetime issues where the reference to the AnyDictionary can outlive the UnknownSchema object.

I think it would be better to return a copy of the AnyDictionary, even though that incurs overhead, and if we want mutability provide a method to supply a new dictionary.

Alternatively, we could change _data to be a shared pointer, but treating the AnyDictionary as a value type would give safer concurrency, if we also added a mutex during the copy and replace.

We don't have much consideration for concurrency in otio at the moment, but it would make sense to me to think about it moving forward.

Does anyone (@jminor ?) have a sense of whether these dictionaries might tend to be massive (dozens or hundreds of entries), or are they more on the order of a few entries?

If the former, I think I'd bias to a shared_ptr solution and ignore concurrency inside the object; if the latter, return/set by value pattern with a mutex in anticipation of concurrency.

ssteinbach: @meshula Would a `const &` do the trick?  Or you want a copy?  And if its a copy should we call it  `data_copy()` or something to disambiguate that you're not getting direct access to the child dictionary?

jminor: Regarding size: a typical OTIO object should be small, with possibly dozens of JSON key/values in metadata. However, given that this class, UnknownSchema, is meant for user defined schema types, they really could be anything. This is the place where we're encouraging developers to add interesting new stuff, so we should be open-ended about what might be in there.

That leads me to support your suggestion that the API be defensive about UnknownSchema by providing read-only access. That will address the desired workflow (making UnknownSchema introspectable) without opening the door to meddling with the contents.

meshula: Cool :) @ssteinbach `const&` doesn't protect against object lifetime issues, or concurrent modification, which I'm worried about. It would work if we had a notion of a holder on the container that also blocked writing while the container is held, but we haven't got that :]

jminor: I don't think there's a way to modify the contents of an UnknownSchema object via our API, except via `read_from` which fully replaces the contents, so I suspect concurrency is not an issue. If a developer wants to modify these things, then they can register a schema and get a real object instead of UnknownSchema.

Would lifetime issues be handled by returning a copy as suggested earlier in this thread?

meshula: Yes, a copy solves lifetime issues :)

jminor: Ok, cool. @natchar could you alter `data()` to return a copy of `_data`?

jminor (changes requested):
Summarizing discussion thread: could you alter `data()` to return a copy of `_data`?

src/opentimelineio/unknownSchema.h
natchar: If it is returning a copy of data, perhaps instead of it being a property, it should be a function definition so that the call is `unknownSchema.data()`.

Maybe I'm missing something, but if I do change the function signature to `AnyDictionary data() const noexcept` that should return a copy of `_data`. However, in the python bindings we use `AnyDictionaryProxy` for AnyDictionary and running the test will yield `Underlying C++ AnyDictionary has been destroyed`. Is there another change I need to make to AnyDictionaryProxy?

meshula: It's reporting that the copy has been destroyed, right? So I guess we need to look at the binding, and make sure that's it's doing what we expect, and then remove the warning... Any other thoughts?

natchar: Updated as part of Dev Day! Apologies I wasn't watching the issue threads.

timlehr:
@jminor this should be good to re-review. 
@apetrynet can you kick off the Doc build again? Seems like it died randomly.

apetrynet:
@timlehr @natchar 
I see this error message on RTD. 

> Config file validation error Config validation error in build.os. Expected one of (ubuntu-22.04, ubuntu-24.04, ubuntu-26.04, ubuntu-lts-latest), got type str (ubuntu-20.04). Double check the type of the value. A string may be required (e.g. "3.10" instead of 3.10)

Output from log:

> git clone --depth 1 https://github.com/AcademySoftwareFoundation/OpenTimelineIO.git . 
git fetch origin --force --prune --prune-tags --depth 50 pull/1799/head:external-1799 
git checkout --force 412312291a9e9744fec98f269094fccc2b0b5f59 
cat .readthedocs.yml 

Looks like we need to update the rtd config file. Perhaps in a separate PR?

apetrynet:
I've submitted a [PR](https://github.com/AcademySoftwareFoundation/OpenTimelineIO/pull/2027) to unblock this. Sorry about the delay.

apetrynet:
The main branch should be at a good state to continue with this PR.

timlehr:
@jminor seems like your comment has been adressed. Is this good to be approved?

meshula (approved):

jminor (approved):

PR #1808, opened 2024-11-07
Add effect and marker initialization to Clip

peter-targett:
**Summarize your change.**

Extend `Clip` to allow effects and markers to be set, matching how `Gap`'s can be initialized.

Added basic C++ and Python tests.

### DIFF
--- a/src/py-opentimelineio/opentimelineio/schema/clip.py
+++ b/src/py-opentimelineio/opentimelineio/schema/clip.py
@@ -7,11 +7,13 @@ from .. import _otio
 
 @add_method(_otio.Clip)
 def __str__(self):
-    return 'Clip("{}", {}, {}, {})'.format(
+    return 'Clip("{}", {}, {}, {} {} {})'.format(
         self.name,
         self.media_reference,
         self.source_range,
-        self.metadata
+        self.metadata,
+        self.effects,
+        self.markers
     )
 
 
@@ -23,11 +25,15 @@ def __repr__(self):
         'media_reference={}, '
         'source_range={}, '
         'metadata={}'
+        'effects={}'
+        'markers={}'
         ')'.format(
             repr(self.name),
             repr(self.media_reference),
             repr(self.source_range),
             repr(self.metadata),
+            repr(self.effects),
+            repr(self.markers)
         )
     )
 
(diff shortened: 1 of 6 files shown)


src/py-opentimelineio/opentimelineio/schema/clip.py
```diff
@@ -7,11 +7,13 @@
 
 @add_method(_otio.Clip)
 def __str__(self):
-    return 'Clip("{}", {}, {}, {})'.format(
+    return 'Clip("{}", {}, {}, {} {} {})'.format(
```
jminor, line 10: Should there be commas between these?

src/py-opentimelineio/opentimelineio/schema/clip.py
```diff
@@ -23,11 +25,15 @@ def __repr__(self):
         'media_reference={}, '
         'source_range={}, '
         'metadata={}'
+        'effects={}'
+        'markers={}'
```
jminor, line 29: Missing ", " separators like the pattern above.

jminor (changes requested):

src/py-opentimelineio/opentimelineio/schema/clip.py
peter-targett: Apologies, bungled the changes, and committed too early.

jminor:
Thanks for the update. Your change looks good to me. There's a build failure which I think is unrelated to your change. Let me see if we can find someone to help with that.

```
    File "/tmp/pip-build-env-hksa4rl4/overlay/lib/python3.9/site-packages/setuptools/_distutils/command/build_ext.py", line 502, in _build_extensions_serial
      self.build_extension(ext)
    File "<string>", line 70, in build_extension
    File "<string>", line 85, in build
    File "<string>", line 166, in cmake_preflight_check
  RuntimeError: CMake Error: Could not find CMAKE_ROOT !!!
  CMake has most likely not been installed correctly.
  Modules directory not found in
  /tmp/pip-build-env-hksa4rl4/overlay/share/cmake-3.31
  CMake Error: Error executing cmake::LoadCache(). Aborting.
  error: subprocess-exited-with-error
```

jminor (approved):

peter-targett:
This could be a separate pull request, I note the order of the arguments between gaps and clips could be more aligned.

PR #1816, opened 2024-11-23
Switch python build to scikit-build-core

JeanChristopheMorinPerso:
**Summarize your change.**

The current python builds use setuptools which is really complicated and hard to maintain. This PR replaces setuptools with a more recent and modern build backend that was specifically designed to support CMake+Python. That backend is [scikit-build-core](https://scikit-build-core.readthedocs.io/en/stable/index.html).

Note that this is not the old `scikit-build`. It's from the same people, but it's not the same thing. `scikit-buil` was building on top of setuptools. `scikit-build-core` is a complete rewrite and doesn't rely on setuptools.

This makes the build, debugging and packaging experience much better.

`scikit-build-core` is now widely used and has active maintainers. It's also used by other projects in the ASWF (https://github.com/AcademySoftwareFoundation/MaterialX, https://github.com/AcademySoftwareFoundation/openexr, https://github.com/AcademySoftwareFoundation/OpenImageIO/pull/4428 for example).

I'm leaving this PR as a draft because there is some failures on Python 3.7 and I'm not too sure why.


.github/workflows/python-package.yml
```diff
@@ -11,7 +11,7 @@ env:
 
 on:
   push:
-    branches: [ main ]
+    branches: [ main, "switch_to_scikit_build_core" ]
```
JeanChristopheMorinPerso, line 14:
```suggestion
    branches: [ main ]
```

.github/workflows/python-package.yml
```diff
@@ -93,6 +93,7 @@ jobs:
   py_build_test:
     runs-on: ${{ matrix.os }}
     strategy:
+      fail-fast: false
```
JeanChristopheMorinPerso, line 96:
```suggestion
```

pyproject.toml
```diff
@@ -1,10 +1,97 @@
...
+
+logging.level = "DEBUG"
+build.verbose = true
+
+sdist.include = ["src/opentimelineio/adapters/builtin_adapters.plugin_manifest.json"]
+
+wheel.packages = ["src/py-opentimelineio/opentimelineio", "src/opentimelineview"]
+wheel.install-dir = "opentimelineio"
+
+build-dir = "build"
```
JeanChristopheMorinPerso, line 79:
```suggestion
```

pyproject.toml
```diff
@@ -1,10 +1,97 @@
...
+    "Topic :: Multimedia :: Video :: Display",
+    "Topic :: Multimedia :: Video :: Non-Linear Editor",
+    "Topic :: Software Development :: Libraries :: Python Modules",
+    "License :: OSI Approved :: Apache Software License",
+    "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: 3.7",
+    "Programming Language :: Python :: 3.8",
+    "Programming Language :: Python :: 3.9",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
```
JeanChristopheMorinPerso, line 27:
```suggestion
    "Programming Language :: Python :: 3.11",
    "Programming Language :: Python :: 3.12",
```

Issue #1635, opened 2023-06-26
Flattening tracks sometimes splits contiguous clips into segments


jminor:
### Incorrect Functionality

If a track has adjacent Gaps, then items below the boundary between those Gaps are sliced in an unexpected way.

## To Reproduce

Using this test file "bug_before_flatten_redacted.otio"...
[flatten_bug_repro.zip](https://github.com/AcademySoftwareFoundation/OpenTimelineIO/files/11872212/flatten_bug_repro.zip)

Run this command:
```
% otiotool -i bug_before_flatten_redacted.otio --flatten video -o out.otio
```

## Expected Behavior

Clips below adjacent Gaps should not be segmented by the flatten operation. In the provided example, "Clip #1" should be left as-is.

## Screenshots

Here is a screenshot before flattening (note the highlighted Gap on the top track)
![Screenshot 2023-06-26 at 10 45 25 AM](https://github.com/AcademySoftwareFoundation/OpenTimelineIO/assets/169599/832d50ac-0dc5-446d-bc49-3e01963477a8)

Here is the result after flattening (note that "Clip #1" has been split in two)
![Screenshot 2023-06-26 at 10 45 38 AM](https://github.com/AcademySoftwareFoundation/OpenTimelineIO/assets/169599/5576d951-fa80-469d-ad7b-f8fda50e4d03)

douglascomet:
Seems related to the behavior outlines in https://github.com/AcademySoftwareFoundation/OpenTimelineIO/issues/1430

praerie:
hi, I'd like to work on this for DevDays!

jminor:
That would be great @praerie - let us know if you want any guidance.

markreidvfx:
Hi @praerie, someone on my team has been looking into this issue (and was kinda hoping to work on it for dev days too) but since you claimed it first, happy to share our findings. 

Our initial thought is to combine the time ranges of contiguous non-visibile items before the recursively trimming occurs.

praerie:
thank you, @markreidvfx! I would be happy to collaborate if your teammate still wants to work on this.

rameshbabuvfxtd:
Hi @praerie, I'm the one looking into this issue. I have an idea for this issue happy to collaborate with you. Thanks for accepting for collaboration.

talk2mybot:
Hi all, I'm also interested into collaborating on dev days with you on this project, if thats o'right

jminor:
[...]

PR #1818, opened 2024-12-09
Fix flattening tracks sometimes splits contiguous clips into segments issue

rameshbabuvfxtd:
This pull request fixes the flattening tracks issue. When an invisible item like Gap is on an upper track, it cops the clip below when flattening. 

Fixes #[1635](https://github.com/AcademySoftwareFoundation/OpenTimelineIO/issues/1635)

### Example of flatten stack:

The source edit file contains two tracks: the top track has two gaps, and the bottom track has one clip.

source

### Before the fix
When the flattening process was performed, it incorrectly split the clip into two segments, which was not the intended behavior.

before

### After the fix
The flattening process now correctly maintains the clip as a single segment, ensuring that the track contains one continuous clip.

after

The `test_stack_algo.cpp` file has been updated to include new tests for the flatten_stack function.

* These tests cover scenarios where gaps are placed over clips, ensuring that the clips remain uncut and are correctly flattened.


jminor:
FYI, the Python 3.7 test failure in this PR should be fixed by https://github.com/AcademySoftwareFoundation/OpenTimelineIO/pull/1820 as soon as that lands.

PR #1820, opened 2025-01-13
Fix CI issues with readthedocs, python 3.7, and broken clip test

reinecke:
**Summarize your change.**

A grab-bag of fixes for issues causing failing CI, including:

1. Updated the CI to run linux python 3.7 checks on ubuntu 22.04 - the last version that includes python 3.7 in github runners- (see [this version manifest](https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json) for reference.
1. Added an explicit path for the sphinx configuration in accordance with [this blog post](https://about.readthedocs.com/blog/2024/12/deprecate-config-files-without-sphinx-or-mkdocs-config/).
1. Updated some clip unittests to account to effects and markers being included in the `__str__` and `__repr__` methods as of #1808

PR #1823, opened 2025-01-16
Fix CI issue with gcov

darbyjohnston:
**Summarize your change.**

It looks like the gcc version in the CI may have changed which is causing issues with gcov/geninfo:
```
geninfo: ERROR: "/home/runner/work/OpenTimelineIO/OpenTimelineIO/src/deps/rapidjson/include/rapidjson/reader.h":271: mismatched exception tag for id 1, 1: '0' -> '1'
```

This PR fixes the issue by removing rapidjson from the coverage. It also includes other build fixes from this PR:
https://github.com/AcademySoftwareFoundation/OpenTimelineIO/pull/1820

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -98,6 +98,7 @@ jobs:
         python-version: ['3.7', '3.8', '3.9', '3.10', '3.11', '3.12']
         include:
           - { os: ubuntu-latest, shell: bash }
+          - { os: ubuntu-22.04, shell: bash, python-version: 3.7 }
           - { os: macos-latest, shell: bash }
           - { os: macos-13, shell: bash }
           - { os: windows-latest, shell: pwsh }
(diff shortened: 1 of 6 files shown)


darbyjohnston:
@reinecke It looks like this change fixes the gcov error, but now there is also a different error for the Windows/msys build:
```
Run curl -sS https://bootstrap.pypa.io/get-pip.py | python
error: externally-managed-environment

This environment is externally managed

To install Python packages system-wide, try 'pacman -S
$MINGW_PACKAGE_PREFIX-python-xyz', where xyz is the package you
are trying to install.

If you wish to install a non-MSYS2-packaged Python package,
create a virtual environment using 'python -m venv path/to/venv'.
Then use path/to/venv/bin/python and path/to/venv/bin/pip.

If you wish to install a non-MSYS2 packaged Python application,
it may be easiest to use 'pipx install xyz', which will manage a
virtual environment for you. Make sure you have $MINGW_PACKAGE_PREFIX-python-pipx
installed via pacman.

note: If you believe this is a mistake, please contact your Python installation or OS distribution provider. You can override this, at the risk of breaking your Python installation or OS, by passing --break-system-packages.
hint: See PEP 668 for the detailed specification.
Error: Process completed with exit code 1.
```

JeanChristopheMorinPerso (approved):

PR #1825, opened 2025-01-24
Add convenience functions

darbyjohnston:
**Summarize your change.**

Add a few convenience functions:
* Add a TimeRange constructor that takes three doubles; the start time, duration, and rate. This allows you to construct a TimeRange like: `TimeRange(0.0, 48.0, 24.0)`, which is a bit more convenient than: `TimeRange(RationalTime(0.0, 24.0), RationalTime(48.0, 24.0))`.
* Add `RationalTime::is_valid_time()` to complement `RationalTime::is_invalid_time()`. This can be a bit easier to read: `if (time.is_valid_time())`, instead of: `if (!time.is_invalid_time())`.
* Add `TimeRange::is_invalid_range()` and `TimeRange::is_valid_range()` to match RationalTime.

Small fix:
* Remove explicit from TimeRange constructors. The explicit keyword only needs to be used with constructors that take a single argument.


src/opentime/timeRange.h
```diff
@@ -40,13 +40,39 @@ class TimeRange
...
+        , _duration{ duration, rate }
+    {}
+
+    /// @brief Returns true if the time range is invalid.
+    ///
+    /// The time range is considered invalid if either the start time or
+    /// duration is invalid.
+    bool is_invalid_range() const noexcept
+    {
+        return _start_time.is_invalid_time() || _duration.is_invalid_time();
```
meshula, line 64: Although a negative start time is valid, a negative duration isn't, so I think duration needs an additional test.

src/opentime/timeRange.h
```diff
@@ -40,13 +40,39 @@ class TimeRange
...
+    }
+
+    /// @brief Returns true if the time range is valid.
+    ///
+    /// The time range is considered valid if both the start time and
+    /// duration are valid.
+    bool is_valid_range() const noexcept
+    {
+        return _start_time.is_valid_time() && _duration.is_valid_time();
```
meshula, line 73: ditto

tests/test_opentime.cpp
```diff
@@ -22,6 +23,15 @@ main(int argc, char** argv)
         assertEqual(t.rate(), 1.0);
     });
 
+    tests.add_test("test_valid", [] {
```
meshula, line 26: should add tests for negative duration please

meshula (changes requested):
good additions, some requests

src/opentime/timeRange.h
darbyjohnston: Good point, I added the checks for a negative duration.

From a coding style point of view, what do you think of having `is_invalid_range()` look like this instead:
```
bool is_invalid_range() const noexcept
{
    return !is_valid_range();
}
```
That way the logic only needs to be in a single function.

Also what do you think about having a static variable that represents an invalid time and invalid range? Something like:
```
    static const RationalTime invalidTime;
```
That way you could construct times that are invalid, without needing to remember the negative rate:
```
    RationalTime time = RationalTime::invalidTime;
```
There is something like this as a private member variable in RationalTime, but I think it would be handy to make it public and also have an equivalent for TimeRange.

meshula (approved):
thanks for the update!

meshula (approved):

src/py-opentimelineio/opentime-bindings/opentime_timeRange.cpp
```diff
@@ -42,6 +42,15 @@ meaning that :meth:`end_time_inclusive` (last portion of a sample in the time ra
                         return TimeRange(*start_time, *duration);
                     }
         }), "start_time"_a=nullptr, "duration"_a=nullptr)
+        .def(py::init<double, double, double>(), "start_time"_a, "duration"_a, "rate"_a)
+        .def("is_invalid_range", &TimeRange::is_invalid_range, R"docstring(
+Returns true if the time range is invalid. The time range is considered invalid if either the start time or
+duration is invalid.
```
jminor, line 48: Could you also document the `duration >= 0` check here?

jminor:
I had to think for a moment about the duration<0 check... To be clear, it should be fine for a RationalTime to have a negative value (e.g. when subtracting two times `a-b` you might get positive or negative result) but I think it is good to constrain a TimeRange to have a non-negative duration, so that any code using ranges can assume the start is before the end.

darbyjohnston:
What do you think about having invalid time and range member variables? That way you can use them to initialize values like this:
```
RationalTime t = RationalTime::invalid_time;
...
if (t.is_valid_time())
{
    ...
}
```
Which I think is a bit more readable and consistent than constructing invalid times manually.

There is an `_invalid_time` member in RationalTime, but it's currently private.

src/py-opentimelineio/opentime-bindings/opentime_timeRange.cpp
darbyjohnston: I added comments to the C++ header and Python bindings.

jminor:
I'm a little hesitant about that since people might think `if (myTime == RationalTime::invalid_time) { ... }` might be equivalent to `if (myTime.is_invalid()) { ... }` but it won't be the same since the specific value/rate values might differ in how they are invalid.

jminor (approved):

darbyjohnston:
That's a good point. I find the `==` operator for RationalTime is already a bit tricky since it rescales the rates, which is not always what I expect.

meshula:
Equality versus strict equality, both are useful...

PR #1827, opened 2025-01-30
Bumped upload-artifact action from v3 to v4

reinecke:
Supersedes #1740 

Bump upload-artifact github action to v4 in line with [deprecation](https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/).

The V4 update no longer supports implicit merge of matrixed build products into a single uploaded artifact (in v3 all wheel platform/python version variants were implicitly merged into a `wheels` artifact). This update makes a separate artifact for each matrix variant ant then merges them on download when creating a release to push to pypi.

PR #1828, opened 2025-01-30
Update GH actions to fix CI builds

darbyjohnston:
**Summarize your change.**

This PR includes the changes from #1827 plus some additional changes to fix the CI:
* Update dawidd6/action-download-artifact to v8
* Update actions/setup-python to v5.4.0
* Use `ubuntu-24.04-arm` for native Linux ARM builds (replacing the QEMU emulator builds)
* Update pypa/cibuildwheel to v2.22.0

### DIFF
--- a/.github/workflows/pypi-publish.yml
+++ b/.github/workflows/pypi-publish.yml
@@ -22,7 +22,8 @@ jobs:
         workflow: python-package.yml
         workflow_conclusion: success
         commit: ${{ github.sha }}
-        name: wheels
+        name_is_regexp: true
+        name: wheel-.*
         path: dist
 
     - name: Download sdist from commit ${{ github.sha }}
--- a/.github/workflows/pypi-publish.yml
+++ b/.github/workflows/pypi-publish.yml
@@ -23,7 +23,7 @@ jobs:
         workflow_conclusion: success
         commit: ${{ github.sha }}
         name_is_regexp: true
-        name: wheel-.*
+        name: `wheel-.*`
         path: dist
 
     - name: Download sdist from commit ${{ github.sha }}
--- a/.github/workflows/pypi-publish.yml
+++ b/.github/workflows/pypi-publish.yml
@@ -17,7 +17,7 @@ jobs:
     # happen for this actual commit (the commit that the tag points to).
     # It also restores the files timestamps.
     - name: Download wheels from commit ${{ github.sha }}
-      uses: dawidd6/action-download-artifact@v2
+      uses: dawidd6/action-download-artifact@v8
       with:
         workflow: python-package.yml
         workflow_conclusion: success
@@ -27,7 +27,7 @@ jobs:
         path: dist
 
     - name: Download sdist from commit ${{ github.sha }}
-      uses: dawidd6/action-download-artifact@v2
+      uses: dawidd6/action-download-artifact@v8
       with:
         workflow: python-package.yml
         workflow_conclusion: success
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -133,7 +133,7 @@ jobs:
           git
     - name: Set up Python ${{ matrix.python-version }}
       if: matrix.python-version != 'mingw64'
-      uses: actions/setup-python@v4.3.0
+      uses: actions/setup-python@v5.4.0
       with:
         python-version: ${{ matrix.python-version }}
     - name: Install coverage dependency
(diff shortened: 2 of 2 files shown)


.github/workflows/python-package.yml
```diff
@@ -171,31 +171,31 @@ jobs:
...
         with:
           output-dir: wheelhouse
         env:
-          CIBW_BUILD: ${{ matrix.python-build }}
+          CIBW_BUILD: ${{ matrix.python-build }}*
           CIBW_SKIP: '*musllinux*'
-          CIBW_ARCHS_LINUX: x86_64 aarch64
-          CIBW_MANYLINUX_X86_64_IMAGE: manylinux2014
-          CIBW_MANYLINUX_AARCH64_IMAGE: manylinux2014
+          CIBW_ARCHS_LINUX: ${{ matrix.arch }}
```
JeanChristopheMorinPerso, line 193: I think you can drop that now.

.github/workflows/python-package.yml
```diff
@@ -171,31 +171,31 @@ jobs:
...
-        uses: pypa/cibuildwheel@v2.16.5
+        uses: pypa/cibuildwheel@v2.22.0
         with:
           output-dir: wheelhouse
         env:
-          CIBW_BUILD: ${{ matrix.python-build }}
+          CIBW_BUILD: ${{ matrix.python-build }}*
           CIBW_SKIP: '*musllinux*'
-          CIBW_ARCHS_LINUX: x86_64 aarch64
-          CIBW_MANYLINUX_X86_64_IMAGE: manylinux2014
```
JeanChristopheMorinPerso, line 193: We need to keep these to make sure we compile with the right glibc.

.github/workflows/pypi-publish.yml
```diff
@@ -17,16 +17,17 @@ jobs:
...
     - name: Download wheels from commit ${{ github.sha }}
-      uses: dawidd6/action-download-artifact@v2
+      uses: dawidd6/action-download-artifact@v8
       with:
         workflow: python-package.yml
         workflow_conclusion: success
         commit: ${{ github.sha }}
-        name: wheels
+        name_is_regexp: true
+        name: wheel-.*
```
JeanChristopheMorinPerso, line 26:
```suggestion
        name: `wheel-.*`
```
Just in case the YAML parser decides to do funny things

.github/workflows/python-package.yml
darbyjohnston: I needed to set this explicitly because the default was trying to build both `x86_64` and `aarch64`.

.github/workflows/python-package.yml
darbyjohnston: Do we need to set it explicitly? It looks like the default value for `CIBW_MANYLINUX_X86_64_IMAGE` is `maylinux2014`:
https://cibuildwheel.pypa.io/en/stable/options/

.github/workflows/python-package.yml
JeanChristopheMorinPerso: It's safer to be explicit, just in case the next version of the action bumps the images.

.github/workflows/python-package.yml
JeanChristopheMorinPerso: It's safer to be explicit, just in case the next version of the action bumps the images.

.github/workflows/python-package.yml
JeanChristopheMorinPerso: Umm, the default is supposed to be x86_64 and i686 according to https://cibuildwheel.pypa.io/en/stable/options/#archs. We could maybe force it to `native`?

.github/workflows/python-package.yml
darbyjohnston: I added it back.

.github/workflows/python-package.yml
darbyjohnston: Hmm, I'll double check, I just tried removing it...

darbyjohnston: Yeah, for some reason the default is trying to build `x86_64` and `aarch64`; from the log:
```
Build options:
  platform: linux
  allow_empty: False
  architectures: aarch64, x86_64
```
Let me try setting it to native...

darbyjohnston: Cool, looks like `native` works.

JeanChristopheMorinPerso (approved):
LGTM

PR #1852, opened 2025-03-09
Drop Imath2 support and modernize Imath includes

jlskuz:
**Summarize your change.**

As quickly discussed in the Slack channel...

OpenEXR developers [recommend to use full namespace includes](https://github.com/AcademySoftwareFoundation/Imath/issues/136#issuecomment-812671628) like `#include <Imath/ImathBox.h>`, but we still use `#include <ImathBox.h>`.

The support of OpenEXR/Imath 2 blocks changing the includes to full namespace. However Imath 3 has be released 4 years ago and OpenEXR/Imath 2 is no longer on the vfxplatform horizon so it seems appropriated to just remove support for it.

### DIFF
--- a/src/opentimelineio/CMakeLists.txt
+++ b/src/opentimelineio/CMakeLists.txt
@@ -78,14 +78,12 @@ add_library(opentimelineio ${OTIO_SHARED_OR_STATIC_LIB}
 add_library(OTIO::opentimelineio ALIAS opentimelineio)
 
 target_include_directories(opentimelineio
-    PRIVATE       "${IMATH_INCLUDES}"
-                  "${PROJECT_SOURCE_DIR}/src"
+    PRIVATE       "${PROJECT_SOURCE_DIR}/src"
                   "${PROJECT_SOURCE_DIR}/src/deps"
-                  "${PROJECT_SOURCE_DIR}/src/deps/rapidjson/include"
-                  "${IMATH_INCLUDES}")
+                  "${PROJECT_SOURCE_DIR}/src/deps/rapidjson/include")
 
 target_link_libraries(opentimelineio 
-    PUBLIC opentime ${OTIO_IMATH_TARGETS})
+    PUBLIC opentime Imath::Imath)
 
 set_target_properties(opentimelineio PROPERTIES
     DEBUG_POSTFIX "${OTIO_DEBUG_POSTFIX}"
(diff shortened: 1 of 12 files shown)


src/opentimelineio/CMakeLists.txt
```diff
@@ -78,14 +78,16 @@ add_library(opentimelineio ${OTIO_SHARED_OR_STATIC_LIB}
...
-    PRIVATE       "${IMATH_INCLUDES}"
-                  "${PROJECT_SOURCE_DIR}/src"
+    PRIVATE       "${PROJECT_SOURCE_DIR}/src"
                   "${PROJECT_SOURCE_DIR}/src/deps"
-                  "${PROJECT_SOURCE_DIR}/src/deps/rapidjson/include"
-                  "${IMATH_INCLUDES}")
+                  "${PROJECT_SOURCE_DIR}/src/deps/rapidjson/include")
+
+if(USE_DEPS_IMATH)
+    target_include_directories(opentimelineio PRIVATE "${PROJECT_SOURCE_DIR}/src/deps/Imath/src")
```
darbyjohnston, line 86: I know `PRIVATE` was from the original code, but should it be removed since Imath is part of the public API? Would that allow other targets using the `opentimelineio` target to pick up the Imath dependency? Like the Python bindings, tests, or if OTIO is being used as a Git sub-module.

meshula: you are correct, it needs to be public since the imath types appear in interfaces

jlskuz: I am happy with whatever you prefer

meshula: @darbyjohnston you are saying this pattern avoids the prefixing problem? (I think that it would) `include_directories("${PROJECT_SOURCE_DIR}/src/deps/Imath/src")`

darbyjohnston: I think so, it would also simplify the changes a bit by removing the various calls to "target_include_directories".

I'm not sure why the "Imath::Imath" "INTERFACE_INCLUDE_DIRECTORIES" don't handle the prefix correctly, is that something that needs to be fixed on the Imath side? Or something about how OTIO is using Imath as a sub-module?

@jlskuz Do you mind trying this change at the top-level CMakeLists.txt, and removing the "target_include_directories" calls? Otherwise the changes look good, thanks for the help updating this.
```
if(USE_DEPS_IMATH)
    include_directories("${PROJECT_SOURCE_DIR}/src/deps/Imath/src")
endif()
```

jlskuz: Done

darbyjohnston: Thanks for the changes!

@meshula Just confirming that we are OK with dropping support for Imath2? Or does that need to be a larger conversation?

PR #1859, opened 2025-03-27
Allow to use rapidjson from the system

jlskuz:
**Summarize your change.**

For (Linux) distro packaging it is common to use system packages instead of submodules and co.

With this option one can ask to use the system rapidjson with CMake flag instead of maintaining custom patches such as https://build.opensuse.org/projects/KDE:Applications/packages/opentimelineio/files/0001-Use-system-rapidjson.patch?expand=1

Note: if `OTIO_FIND_RAPIDJSON` is `ON` but rapidjson is not found cmake fails. This behavior is different to `OTIO_FIND_IMATH` where it will just continue with `src/deps` as a fallback. However I think if the user explicitly request the system package it makes more sense to fail if not found. In my opinion this should be changed for Imath as well, but that is offtopic for this PR


jlskuz:
The readthedocs.org failure seems unrelated?

darbyjohnston:
Unrelated; sometimes readthedocs has these failures, possibly a network error. I'm not sure how to re-run the job, it's separate from the Github actions. Maybe wait for a bit and try pushing another change to try again.

ssteinbach:
I restarted the RTD build - sometimes that does flake out, sorry

darbyjohnston:
Thanks for the changes; I agree about the fallback, the Imath behavior could be changed in a separate PR.

Issue #1182, opened 2021-12-27
Proper shared library handling for the C++ bindings


darix:
## Feature Request

Proposed Change to existing Behavior

## Description

1. right now libopentime.so and libopentimelineio.so are lacking an soversion. it would be nice to assign a proper soversion for those.
2. allow the python extensions to use the shared library.

## Context

This would make it easier to package OpenTimelineIO and use it with other programs olive.

meshula:
This would be a welcome PR!

It's a more complex request that it appears on first blush though, so I'd like to lay out what I'd hope to see in a PR:

For maintainers for whom SO version policy is not their day job, it is extraordinarily difficult to make sense of the SO versioning documentation. On other projects, SO versioning has been a bit of a challenge, as we've frequently got, sometimes urgent, "you messed it up" issues, but very few "here's how to do it right" PRs or comments leaving the maintainers in the frustrating position of attempting to iterate fixes and wait for validation.

I am aware of pages like this https://tldp.org/HOWTO/Program-Library-HOWTO/shared-libraries.html, or https://www.debian.org/doc/debian-policy/ch-sharedlibs.html, but these pages are filled with caveats, like "some" or "most" and "normally." I'd like to see something specific to OTIO with OTIO's policy explicitly laid out.

On that note, I request that any PR introducing SO version should please also include 

- additions to the documentation that describe clearly a policy on when and how SO version is to be updated; something along the lines of "when this condition is met, modify the SO version as follows" for whatever unique combinations of conditions and actions exist.

- required naming conventions for Python per platform,

- required naming conventions for Linux, and also differences for Linux per distro, if differences exist (I'm not sure about that).

- expectations about symlinking versions, e.g. if it's expected on Linux that otio 1.1.5 is linked by a 1.1 symlink and that by a 1 symlink, that should be laid out as well, and accomplished within the Cmake script.

- mods as necessary to the Cmake config file installation.

Thanks

darix:
For the question "when do we need to change the soversion" ... there are tools now and you do not have to do that manually anymore. Those can even be integrated into the testsuite or so. 

- https://youtu.be/b1fO-RLtDME
- https://media.ccc.de/v/1234-libabigail-how-semantic-analysis-of-c-and-c-elf-binaries-can-be-used-to-analyze-abi-changes
- https://github.com/ansasaki/abimap

meshula:
Thanks for the pointers!

[...]

Issue #872, opened 2021-01-21
include semantic versions in the built C++ library names


ssteinbach:
currently they include the python info but not the version info:  `_opentime.cpython-38-darwin.so`

meshula:
Notes here on "doing it right." https://github.com/pvanhoof/dir-examples

PR #1860, opened 2025-03-28
Add SOVERSION for the C++ libraries

darbyjohnston:
Related to #872 and #1182, but this is just for the C++ libraries.

This change sets the SOVERSION for the C++ libraries. This version should be incremented anytime the ABI changes. OTIO is currently using the minor version number for breaking changes, e.g. v0.15, v0.16.0, v0.17.0, so SOVERSION is set as a combination of OTIO_VERSION_MAJOR and OTIO_VERSION_MINOR. When OTIO reaches version 1.x.x then SOVERSION could be changed to track just OTIO_VERSION_MAJOR.


CMakeLists.txt
```diff
@@ -11,6 +11,15 @@ set(OTIO_VERSION_MINOR "18")
...
 set(OTIO_VERSION ${OTIO_VERSION_MAJOR}.${OTIO_VERSION_MINOR}.${OTIO_VERSION_PATCH})
 
+# Set the SOVERSION. This version should be incremented anytime the ABI
+# changes. OTIO is currently using the minor version number for breaking
+# changes, e.g. v0.15, v0.16.0, v0.17.0, so the SOVERSION should be incremented
+# to match. If the next OTIO version is 0.19.0 and has ABI changes, the
+# SOVERSION should be incremented to 0.19. When OTIO reaches 1.x.x the
+# SOVERSION can be changed to match the major version which is a more
+# common convention.
+set(OTIO_SOVERSION 0.18)
```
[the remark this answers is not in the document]
darbyjohnston: Good point, and when OTIO reaches version 1.x.x it could be changed to just OTIO_VERSION_MAJOR.

darix:
TBH: using the version number for the SO version is a good starting point and bumping with the version is an easy way to avoid problems.

but there is actual tooling now for understand when you need to bump your soversion. e.g. libabigail. 

https://media.ccc.de/v/1234-libabigail-how-semantic-analysis-of-c-and-c-elf-binaries-can-be-used-to-analyze-abi-changes

meshula:
I propose the following policy to make the policy concrete



The SO version must be incremented every time a change occurs to the ABI that causes a backward incompatibility. Such changes include, exhaustively:

 - a change to struct or class layout
 - enum changes that would cause a renumbering of previously published enums
 - a removal of a struct, class, enumeration, or function
 - a change in parameters to a free standing function
 - a removal of a global variable

OTIO currently designates the minor version number for breaking changes, e.g. v0.15, v0.16.0, v0.17.0, accordingly the SOVERSION will be incremented to match. SO version must be monotically increasing, so the ABI version should be computed as major*100 + revision*1. The third digit will never implicate an ABI version change.

Distros that wish to minimize version changes are encouraged to pick the version of OpenTimelineIO that was released most recently before the last cut off date of the most recent vfxplatform year. Distros that wish to always be on head are advised that ABI version is considered carefully before a bump and subject to OpenTimelineIO project policy.

JeanChristopheMorinPerso:
@meshula I find the following confusing:

>Distros that wish to minimize version changes are encouraged to pick the version of OpenTimelineIO that was released most recently before the last cut off date of the most recent vfxplatform year. Distros that wish to always be on head are advised that ABI version is considered carefully before a bump and subject to OpenTimelineIO project policy.

I'm curious, why mention the vfxplaform? Also, why does picking a version released before the latest VFX platform year minimize changes?

As for the last statement, it might be just because I'm not a native English speaker, but I find it confusing. Is the intent to say that the head isn't yet stabilized and can't be relied on, ABI wise?

darix:
target platform shouldnt have any impact on when to bump the soversion. the important bit is when you break the ABI.

that's why I suggest using tools like abigail over guessing. :)

meshula:
>I'm curious, why mention the vfxplaform? Also, why does picking a version released before the latest VFX platform year minimize changes?

> Distros that wish to minimize version changes are encouraged to pick the version of OpenTimelineIO that was released most recently before the last cut off date of the most recent vfxplatform year. Distros that wish to always be on head are advised that ABI version is considered carefully before a bump and subject to OpenTimelineIO project policy.

That's PTSD from OpenEXR/Imath and distros trying to slice and dice an old OpenEXR (2.x) with a new Imath (3.x) . Maybe we don't need the advice; if people run into problems with the Imath dependency they should use the interred version, and maybe we don't need to say anything about that.

> As for the last statement, it might be just because I'm not a native English speaker, but I find it confusing. Is the intent to say that the head isn't yet stabilized and can't be relied on, ABI wise?

No,  now that you've asked me to think more carefully about my comment, it's meant to say if you are building an old vfxplatform and you want to use system Imath, you should pick an OTIO from the vfxplatform year you care about, otherwise you might run into Imath2/3 problems.

Again, maybe Imath2 is gone by now, so maybe not worth caring about.

darbyjohnston:
@meshula I updated the comments with your feedback, and also added some examples of release versions and their corresponding ABI versions. I left out the parts about the distros, hopefully moving to Imath3 will sort that out. When you have a chance take a look and see what you think.

CMakeLists.txt
```diff
@@ -118,6 +118,29 @@ else()
...
+# to match. So for example, the following OTIO versions would map to these
+# ABI versions:
+#
+# * v0.18.0 - 0.18
+# * v0.19.0 - 0.19
+# * v0.19.1 - 0.19 # No ABI changes with minor version changes
+# * v1.0.0 - 1.0
+# * v1.1.0 - 1.1
+#
+set(OTIO_SOVERSION ${OTIO_VERSION_MAJOR}.${OTIO_VERSION_MINOR})
```
meshula, line 142: Although you can set a dotted version for SOVERSION it will cause issues in  tooling that thinks SOVERSION is an integer. 

That's why I was suggesting SOVERSION should be major*10 + minor.

```suggestion
math(EXPR OTIO_SOVERSION "${OTIO_VERSION_MAJOR} * 100 + ${OTIO_VERSION_MINOR}")
```

darbyjohnston: Ah OK, that makes sense.

meshula (approved):
lgtm

PR #1864, opened 2025-04-01
Don't fallback to Imath from src/deps

jlskuz:
If the user explicitely requests to use find_package to use the system Imath it seems wrong to fallback to the delivered Imath

As discussed in #1859


darbyjohnston:
Looks good, thanks for following up on this.

PR #1865, opened 2025-04-01
Work/deps ci

jlskuz:

```Fixes #123```

**Summarize your change.**

### CODE REVIEW

File: .github/workflows/python-package.yml [hunk 1]
```diff
@@ -88,6 +88,63 @@ jobs:
         cd ${{ env.OTIO_CONSUMER_TEST_BUILD_DIR }}
         cmake ${{ github.workspace }}/tests/consumer -DCMAKE_PREFIX_PATH=${{ env.OTIO_INSTALL_DIR }}
 
+  cpp_build_external_deps:
```
ssteinbach, line 91: A comment describing this job and what distinguishes this from the simple `cpp_build` job would be nice. 
 With three jobs, it feels like we would want comments on the other two as well, but that is outside the scope of thisPR.

ssteinbach (commented):
I appreciate you making this PR.  My only concern is complicating the build/CI system with a job that is very similar to the `cpp_build`.   The matrix is narrow on that job, so it wouldn't fan out in the way the other jobs do.

PR #1866, opened 2025-04-01
Add Doxygen comments to opentime

darbyjohnston:
This change completes adding Doxygen style comments to the opentime API. The changes are just comments, there are no functional changes to the API.

Most of the comments were straightforward, but any feedback to help improve them is welcome.

I added a Doxygen group for the TimeRange "relations" and `<pre>` tags to keep the ASCII diagrams intact in the Doxygen HTML.

I also made a couple of tweaks to the Doxygen configuration to remove non-public info and keep the sorting order the same as the header files.

### DIFF
--- a/doxygen/config/dox_config
+++ b/doxygen/config/dox_config
@@ -494,7 +494,7 @@ EXTRACT_ALL            = NO
 # be included in the documentation.
 # The default value is: NO.
 
-EXTRACT_PRIVATE        = YES
+EXTRACT_PRIVATE        = NO
 
 # If the EXTRACT_PRIV_VIRTUAL tag is set to YES, documented private virtual
 # methods of a class will be included in the documentation.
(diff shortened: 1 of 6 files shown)


src/opentime/timeRange.h
```diff
@@ -10,43 +10,46 @@
...
- * This default epsilon_s value is used in comparison between floating numbers.
- * It is computed to be twice 192khz, the fastest commonly used audio rate.
- * It can be changed in the future if necessary due to higher sampling rates
- * or some other kind of numeric tolerance detected in the library.
- */
+/// @brief This default epsilon_s value is used in comparison between floating numbers.
+///
+/// It is computed to be twice 192khz, the fastest commonly used audio rate. It
+/// can be changed in the future if necessary due to higher sampling rates or
+/// some other kind of numeric tolerance detected in the library.
```
ssteinbach, line 17: Might be worth noting that twice 192khz implies 1/2 a 192khz frame

doxygen/config/dox_config
```diff
@@ -643,7 +643,7 @@ INLINE_INFO            = YES
 # name. If set to NO, the members will appear in declaration order.
 # The default value is: YES.
 
-SORT_MEMBER_DOCS       = YES
+SORT_MEMBER_DOCS       = NO
```
ssteinbach, line 646: Might add a comment that this is off so that sort order matches the headers

src/opentime/timeRange.h
darbyjohnston: Sounds good, I updated the comment.

ssteinbach (approved):

PR #1874, opened 2025-04-18
Drop dead OTIO_IMATH_LIBS cmake option

jlskuz:
It seems to be a relic from Imath2 where it would have made sense as it required a custom find OpenEXR script. However the option is dead (and appears to have ever been) and is not usefull for Imath3 anyways.

As discussed with @meshula on Slack


meshula (approved):

meshula:
lgtm!

Issue #1094, opened 2021-10-01
VFX Platform compatibility


JeanChristopheMorinPerso:
In the last TSC meeting, we discussed about which VFX Platform years we support.

Meeting notes: https://docs.google.com/document/d/14PmEo9ukURmU9OKh4TAo02-e6cSqgnfHFgsURvlqFK8

Excerpt from the meeting notes:

> OTIO release 0.14 will support:
> * VFX platform 2021, 2020, 2019, 2018
> * Python 2.7 - 3.9
> * Notice that Python 2.7 is deprecated & we plan to drop it in OTIO release 0.15
> * We can still patch things into 0.14.x for folks who need to stay on Python 2.7 VFX 2019, etc.

I thought it would be nice to have a checklist to verify our compliance with the VFX Platform.

- [x] Linux (`glibc` 2.17): Our Python wheels are built using the `manylinux2010` docker images, which use glibc 2.17.
- [ ] Linux: GCC 6.3.1 and 9.3.1. Right now we are testing with 9.3.x that comes with Ubuntu 20.04 (https://wiki.ubuntu.com/FocalFossa/ReleaseNotes#Toolchain_Upgrades_.2B2D3e4P4P-). The `manylinux2010` docker images use GCC 8.3.1 which means we also test this version, at least for compilation). We would need to add a job to test against 6.3.1.
- [ ] macOS: We are currently building against 10.15 (and soon 11.0 because GitHub are rolling it out). We need to target 10.13.
- [ ] Windows: Visual Studio 2017 for the VFX Platform 2021. Undefined for previous years. We are currently building with Visual Studio 2019.
- [ ] C++: Minimum version is C++14 and maximum is C++17. Are we compliant?
- [ ] PySide2: 2.0.x to 5.15. We currently require 5.11 and higher (https://github.com/PixarAnimationStudios/OpenTimelineIO/blob/master/setup.py#L352)... Do we want to be compliant here?

# Notes on Python wheels
While writing this, I had some thoughts about what it means for our Python wheels.

For example, how should we treat our wheels? Do we want to build them inside VFX Platform compliant docker images for our Linux wheels? What about other platforms?

* Linux: I think we should continue to build our Linux Python wheels with `cibuildwheel`. It uses the `manylinux2010` docker image by default, which means our wheels are compatible with glibc 2.17 and higher. In other words I don't think we need to use VFX Platform compliant docker images to build them.
[...]

PR #1883, opened 2025-05-03
Remove vfx platform 2020 support from README.md, add 2024 and 2025, note that 3.12 is also supported

### LINKED ISSUE #1804: Update to VFX 2024 and 2025
Our policy is to support current year VFX Platform + 3 prior releases. Our README and status badges currently show "VFX platform 2023, 2022, 2021, 2020" which is one year behind that.

We should update that badge, README, and the CI config to include the right versions of Python.

One important detail noted in https://github.com/AcademySoftwareFoundation/OpenTimelineIO/issues/1802 is that we will need to pin some of our older Python CI jobs to specific versions of OS rather than, for example, `ubuntu-latest`) so that we aren't vulnerable to CI failures when "latest" updates automatically. Having some jobs run against "latest" is good so we can discover issues with recent/new versions, but old ones should pin to matching versions.

### LINKED ISSUE #1876: Update VFX Platform Badge to remove support for v2020

ssteinbach:
Fixes #1876 
Fixes #1804 

* Removes VFX Platform 2020 from the badge, adds 2024 and 2025
* stripped out code blocks that indicated they were present for python versions before 3.9.
* Adds notes that Pyhon 3.12 is supported (CI is already running on 3.12, binaries are already built)

As noted in the comments below by @JeanChristopheMorinPerso, this PR doesn't fully cover #1094, it doesn't add any of the VFX platform container related testing stuff.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@ OpenTimelineIO
 ==============
 
 [![Supported VFX Platform Versions](https://img.shields.io/badge/vfx%20platform-2022--2025-lightgrey.svg)](http://www.vfxplatform.com/)
-![Supported Versions](https://img.shields.io/badge/python-3.9%2C%203.10%2C%203.11-blue)
+![Supported Versions](https://img.shields.io/badge/python-3.9%2C%203.10%2C%203.11-%2C%203.12-blue)
 [![Build Status](https://github.com/AcademySoftwareFoundation/OpenTimelineIO/actions/workflows/python-package.yml/badge.svg)](https://github.com/AcademySoftwareFoundation/OpenTimelineIO/actions/workflows/python-package.yml)
 [![codecov](https://codecov.io/gh/AcademySoftwareFoundation/OpenTimelineIO/branch/main/graph/badge.svg)](https://codecov.io/gh/AcademySoftwareFoundation/OpenTimelineIO)
 [![docs](https://readthedocs.org/projects/opentimelineio/badge/?version=latest)](https://opentimelineio.readthedocs.io/en/latest/index.html)
@@ -56,7 +56,7 @@ Supported VFX Platforms
 -----------------
 The current release supports:
 - VFX platform 2025, 2024, 2023, 2022
-- Python 3.9 - 3.11
+- Python 3.9 - 3.12
 
 For more information on our vfxplatform support policy: [Contribution Guidelines Documentation Page](https://opentimelineio.readthedocs.io/en/latest/tutorials/contributing.html)
 For more information on the vfxplatform: [VFX Platform Homepage](https://vfxplatform.com)
@@ -154,7 +154,7 @@ You can also install the PySide2 dependency with `python -m pip install .[view]`
 
 You may need to escape the `[` depending on your shell, `\[view\]` .
 
-Currently the code base is written against python 3.9, 3.10 and 3.11,
+Currently the code base is written against python 3.9-3.12,
 in keeping with the pep8 style.  We ask that before developers submit pull
 request, they:
 
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@ OpenTimelineIO
 ==============
 
 [![Supported VFX Platform Versions](https://img.shields.io/badge/vfx%20platform-2021--2023-lightgrey.svg)](http://www.vfxplatform.com/)
-![Supported Versions](https://img.shields.io/badge/python-3.7%2C%203.8%2C%203.9%2C%203.10%2C%203.11-blue)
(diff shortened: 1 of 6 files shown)


meshula (changes requested):
Hi Stephan, current vfxplatform year is 2025, we currently claim to support 2023,22,21. Actually it should be cy+3 which means 25,24,23,22. So we should remove 2021 as well. Accordingly this also gives us a bump to python 3.9 so we can at long last drop python 3.7 support.

ssteinbach:
Added changes in to remove 3.7 and 3.8

README.md
```diff
@@ -154,7 +154,7 @@ You can also install the PySide2 dependency with `python -m pip install .[view]`
 
 You may need to escape the `[` depending on your shell, `\[view\]` .
 
-Currently the code base is written against python 3.7, 3.8, 3.9, 3.10 and 3.11,
+Currently the code base is written against python 3.9, 3.10 and 3.11,
```
JeanChristopheMorinPerso, line 157: I feel like we should say that we support 3.9 through 3.12, because we already ship wheels for 3.12.

setup.py
```diff
@@ -357,10 +355,9 @@ def run(self):
     },
 
     # Disallow 3.9.0 because of https://github.com/python/cpython/pull/22670
-    python_requires='>=3.7, !=3.9.0',  # noqa: E501
+    python_requires='>=3.9, !=3.9.0',  # noqa: E501
```
JeanChristopheMorinPerso, line 358: This expression can be simplified to 
```suggestion
    python_requires='>3.9.0',  # noqa: E501
```

```python
>>> from packaging.version import Version
>>> Version('3.9.1') > Version('3.9.0')
True
>>> Version('3.9.0') > Version('3.9.0')
False
```

src/py-opentimelineio/opentimelineio/plugins/manifest.py
```diff
@@ -317,18 +308,10 @@ def load_manifest():
         )
 
     # the builtin plugin manifest
-    try:
-        builtin_manifest_path = (
-            resources.files("opentimelineio.adapters")
-            / "builtin_adapters.plugin_manifest.json"
-        ).as_posix()
-    except AttributeError:
-        # For python <= 3.7
```
JeanChristopheMorinPerso, line 326: Removing this feels so goood!!!

JeanChristopheMorinPerso (commented):
This looks good to me. The only thing that I would add is a mention that we also support 3.12 since we already ship wheels for 3.12 and according to https://clickpy.clickhouse.com/dashboard/opentimelineio?min_date=2024-03-03&max_date=2025-04-20, users effectively depend on 3.12.

Also, a nice side effect of this PR is that it will enable https://github.com/AcademySoftwareFoundation/OpenTimelineIO/pull/1816.

ssteinbach:
@JeanChristopheMorinPerso to be clear, does this PR also address #1094 ?

README.md
ssteinbach: Gotcha.  I held off on that because if isn't in the VFX platform but you're right, we should just put it out there.

JeanChristopheMorinPerso:
@ssteinbach I don't think it does. We aren't really testing against the VFX platform targets like we should, at least on Linux. Doing so would require us to run the tests in docker containers that match the platform years.

ssteinbach:
Ok got it. I'll leave that issue off this in that car, that's a different project.

JeanChristopheMorinPerso:
@ssteinbach I might be interpreting your comment wrong, but I don't see a note about python 3.12, is this expected?

ssteinbach:
@JeanChristopheMorinPerso yup!  Forgot to push a commit.  That is there now too.

src/py-opentimelineio/opentimelineio/plugins/manifest.py
```diff
@@ -7,13 +7,9 @@
...
 import logging
 import os
-from pathlib import Path
 
-try:
-    from importlib import metadata
-except ImportError:
-    # For python 3.7
-    import importlib_metadata as metadata
+# for python <= 3.9
```
meshula, line 11: does this comment tell us we will delete it when we drop 3.9?

meshula (approved):
lgtm

src/py-opentimelineio/opentimelineio/plugins/manifest.py
ssteinbach: Yeah, exactly

JeanChristopheMorinPerso (approved):

PR #1896, opened 2025-06-06
fixed image bounds error status null bug + test case

cato-o:
**Summarize your change.**

Added null pointer check for the error status pointer in the available image bounds function.

Added two new test cases to test_clip.cpp called "test_error_ptr_null" and "test_error_ptr_null_no_media". One test has a media reference, the other test has no media reference.

### DIFF
--- a/tests/test_clip.cpp
+++ b/tests/test_clip.cpp
@@ -289,17 +289,7 @@ main(int argc, char** argv)
     tests.add_test("test_error_ptr_null_no_media", [] {
         using namespace otio;
 
-        // tests for no image bounds and no media reference on clip
-        otio::ErrorStatus              status;
-        SerializableObject::Retainer<> so =
-            SerializableObject::from_json_string(
-                R"(
-            {
-                "OTIO_SCHEMA": "Clip.1"
-            })",
-                &status);
-
-        Clip* clip = dynamic_cast<Clip*>(so.value);
+        SerializableObject::Retainer<Clip> clip(new Clip);
 
         // set media reference to empty
         Clip::MediaReferences empty_mrs;


tests/test_clip.cpp
```diff
@@ -260,6 +260,107 @@ main(int argc, char** argv)
...
+
+        // check that if null ptr, nothing happens
+        otio::ErrorStatus* null_test = nullptr;
+
+        assertEqual(
+            mr_clip->available_image_bounds(null_test), std::optional<IMATH_NAMESPACE::Box2d>()
+        );
+
+
+
```
jminor, line 321: Can you remove these blank lines?

darbyjohnston:
Hi, and thanks for the fix!

It would also be nice to fix the `available_image_bounds()` function declarations, it looks like they are missing the default `nullptr` value. We generally default the `error_status` argument to `nullptr` so there is a bit less code to write if the developer does not want to check the error status.

It looks like the declarations need to be fixed in `clip.h` and `composable.h`, just by changing:
```
std::optional<IMATH_NAMESPACE::Box2d>
    available_image_bounds(ErrorStatus* error_status) const;
```
To:
```
std::optional<IMATH_NAMESPACE::Box2d>
    available_image_bounds(ErrorStatus* error_status = nullptr) const;
```

tests/test_clip.cpp
```diff
@@ -260,6 +260,104 @@ main(int argc, char** argv)
...
+        );
+    });
+
+    tests.add_test("test_error_ptr_null_no_media", [] {
+        using namespace otio;
+
+        // tests for no image bounds and no media reference on clip
+
+        otio::ErrorStatus              status;
+        SerializableObject::Retainer<> so =
```
darbyjohnston, line 327: I think it is OK to construct the `Clip` from JSON here, since it looks like it creates one with an empty media reference. I'm not sure the C++ API allows you to set an empty `media_reference`.

Out of curiosity, was there a bug caused by having a clip without a media reference?

cato-o: There wasn't a specific bug but I wrote this to try and get into and test the `if (!active_media)` code path. I'm not certain if it's ever possible to get there though since I think if there's an empty `media_reference`, it gets checked for and set to DEFAULT_MEDIA instead.

cato-o:
Hi Darby! I made the changes you mentioned and it's ready for re-review! Thank you :)

darbyjohnston:
Thanks for the changes! Just one more small change and I think it is ready; in the second test, you could use `new` to create the clip since you are setting the media references to empty:
```
SerializableObject::Retainer<Clip> clip(new Clip);
```
Or keep the JSON and remove the part about setting the media references to empty, they both do the same thing and only one is needed.

darbyjohnston:
Looks good, thanks again!

PR #1901, opened 2025-06-11
Add core C++ support for otioz and otiod

### LINKED ISSUE #1869: Add core C++ support for otioz and otiod
After some discussion in Slack:
https://academysoftwarefdn.slack.com/archives/CMQ9J4BQC/p1729878665473649

I wanted to create a proposal for adding core C++ support for otioz and otiod. I think adding them to the C++ API would have a couple of benefits:
 
* Broader adoption by users of the C++ API and other language bindings.
* Provide a common interface to ensure bundles are created according to the specification.

I propose adding minizip-ng as a new dependency to support otioz. It has a permissive license (zlib), and is also used by OpenColorIO for their .ocioz files. The use of minizip-ng would be internal and not part of the OTIO public API.

Outline of the required work:

#### Add an enum for the media reference policy
(https://opentimelineio.readthedocs.io/en/stable/tutorials/otio-filebundles.html#mediareferencepolicy)
```
enum class MediaReferencePolicy
{
    ErrorIfNotFile,
    MissingIfNotFile,
    AllMissing
 };
```

#### Add constants for the file bundle versions
```
static std::string const otiozVersion = "1.0.0";
static std::string const otiodVersion = "1.0.0";
```

#### Add I/O functions to SerializableObject
```
bool to_otioz_file(
    std::string const&        file_name,
    MediaReferencePolicy      media_reference_policy   = MediaReferencePolicy::ErrorIfNotFile,
    ErrorStatus*              error_status             = nullptr,
    const schema_version_map* target_family_label_spec = nullptr,
    int                       indent                   = 4) const;
```
* Create a new ZIP file.
* Resolve external references based on the media reference policy and add them to the ZIP file.
* Create the `version` file and add it to the ZIP file.
* Clone the `SerializableObject` and modify it based upon the media reference policy, and to change the external reference paths.
* Call `to_json_string()` to create `content.otio` and add it to the ZIP file.

```
static SerializableObject* from_otioz_file(
    std::string const& file_name,
    std::string const& temp_dir,
    ErrorStatus*       error_status = nullptr);
```
* Unzip the file to the temporary directory.
* Call `from_json_file()` on the `content.otio` file in the temporary directory.

```
bool to_otiod_bundle(
    std::string const&        dir_name,
    MediaReferencePolicy      media_reference_policy   = MediaReferencePolicy::ErrorIfNotFile,
    ErrorStatus*              error_status             = nullptr,
    const schema_version_map* target_family_label_spec = nullptr,
    int                       indent                   = 4) const;
```
* Create the given directory.
* Resolve external references based on the media reference policy and copy them to the directory.
* Create the `version` file in the directory.
* Clone the `SerializableObject` and modify it based upon the media reference policy, and to change the external reference paths.
* Call `to_json_string()` to create the `content.otio` file in the directory.

Is a `from_otiod_bundle()` function necessary? The `content.otio` file could just be opened inside the directory.

#### Python bindings
Add Python bindings for the new C++ functions and remove the Python adapters. Modify `otioconvert` and any other Python utilities to use the new functions.

darbyjohnston:
Fixes #1869

This PR adds support for otioz and otiod bundles to the C++ API. (Updated 12/17/25)

The code follows the original Python as close as possible, so the logic and error handling should be similar.

The python otioz and otioz adapters still exist and work the same, but now they call into the C++ code for the bundle functionality.

Support for image sequences has been added, previously the Python adapters only supported media references.

Other changes:
* Added dependencies for working with ZIP files (minizip) and file compression (zlib). The minizip library is part of zlib.
 As an alternative, minizip-ng and zlib-ng could be used since they provide the same APIs in compatibility mode.
* Added a new C++ namespace "bundle" for reading and writing bundles.
* Added a C++ example to demonstate reading and writing bundles.
* Added miscellaneous C++ utilities for working with URLs and files.
* Added the define OTIO_TESTS_DIR for C++ tests to locate the sample data directory.
* Replaced most of the Python bundle tests with equivalent C++ tests. Some Python bundle tests remain to test that the adapters work.

Questions:
* Should bundle support be optional with a CMake flag?
* Should the otioz/otiod file versions be bumped?
* Should we use a third party library for URL handling? We only use `file://` URLs, and do lossy conversions between file paths, so for now I implemented just the necessary functionality. All of the tests pass, but I'm not sure how comprehensive they are.

Todo:
* Remove the minizip-ng and zlib-ng submodules.
* Test 64-bit support (files > 4GB).
* Add file information/attributes when adding files to the ZIP archive.


darbyjohnston:
I wanted to try a different approach to this, and created a new PR with some help from Claude #2021

PR #1912, opened 2025-08-01
Added --remove-effects feature to otiotool.

jminor:
Sometimes you have effects in your timeline that you want to remove. This PR adds a `--remove-effects` option to otiotool to make that easy and convenient.

Example:
```
otiotool -i effects.otio --remove-effects -o no-effects.otio
```


jminor:
@ssteinbach yes, those `-O otio_json` options were redundant. That is automatically inferred from the file extension. I just cleaned that up while I was in there so folks wouldn't think that's required.

PR #1918, opened 2025-08-12
Update submodule deps for cmake 4

HO-COOH:
**Summarize your change.**
Now that the latest cmake version goes to 4.x, and cmake <3.5 compatibility has been removed, which the `pybind11` submodules uses. It's best to update all those submodules to latest.

CI are all passed.


darbyjohnston:
Hi, and sorry for the delay; we've updated the submodule dependencies to more recent versions: Imath v3.2.1, pybind11 v2.13.6, and the latest rapidjson. I believe these will work with CMake 4, but if not feel free to reopen this PR or create a new issue.

Issue #26, opened 2016-11-08
Comparing Timelines


ssteinbach:
There have been requests of the ability to have a tool or library for diffing OTIO data. This includes:
- When things change on items that are the same
- A way to programmatically specify how to hash an item (for example, by a particular metadata key or by name)
- A way to detect when an item has moved to another location in the hierarchy

meshula:
Flagging for TSC, should probably close

PR #1922, opened 2025-08-15
Adding diff feature (otiodiff) to otiotool

cato-o:
Added a diff feature that compares the clips of two OTIO timelines. Clips are categorized based on changes from timeline A to timeline B. Outputs an annotated and color coded OTIO with tracks from both timelines visualizing differences as well as a text summary.

Addresses #26 

Demo-ed at 2025 ASWF OTIO Virtual Town Hall

otiodiff


src/py-opentimelineio/opentimelineio/console/otiodiff/clipData.py
```diff
@@ -0,0 +1,162 @@
...
+    track_num = None    # not originally stored in otio.schema.Clip
+    source_clip = otio.schema.Clip()
+    # everything below holds comparison result info
+    note = ""
+    matched_clipData = None
+
+    def __init__(self, source_clip, track_num, note=None):
+        self.full_name = source_clip.name
+        self.name = self.splitFullName(source_clip)[0]
+        self.version = self.splitFullName(source_clip)[1]
```
jminor, line 33: This could avoid calling the function twice:
```python
self.name, self.version = self.splitFullName(source_clip)
```

jminor (changes requested):
This is getting close!

Could you remove some of the blocks of commented-out code that you don't think would be useful to another person working on this later.

Let's walk through the remaining TODOs & either address them if they're quick, or turn them into "suggestion for future work".

tacheconfuse:
Can we expect this feature anytime soon? 👀

PR #1929, opened 2025-08-25
Hide our custom fabs() function

darbyjohnston:
This PR hides our custom `fabs()` function. We added the function as part of the constexpr work, but it really should not be part of the public API. When we upgrade to C++23 we can remove the function entirely and replace it with the version from the std library.

Note that there are two copies of the `fabs()` function, in both RationalTime and TimeRange. The code is short enough it seemed OK to duplicate, especially since we can remove it eventually.

### DIFF
--- a/src/opentime/rationalTime.h
+++ b/src/opentime/rationalTime.h
@@ -20,21 +20,6 @@ enum IsDropFrameRate : int
     ForceYes      = 1,
 };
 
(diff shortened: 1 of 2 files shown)


meshula (approved):

meshula:
Awesome, thank you very much :D

PR #1936, opened 2025-09-07
Fix various typos

luzpaz:

n/a

Found via `codespell -L bui,fo,thirdparty`

Describe the reason for the change: uniformity

n/a


jminor:
@luzpaz I merged in the latest from main to resolve the build issues you hit before. The only outstanding issue to fix is the DCO check. You can see how to resolve that here: https://github.com/AcademySoftwareFoundation/OpenTimelineIO/pull/1936/checks?check_run_id=49903144636

luzpaz:
Oh man, I fubar'd this PR up :disappointed:

jminor:
Haha. Yeah, it's "fun with git" day today :)

We can still make this work. When we do land this PR we will squash it like we do to all the other commits into main. If you can do an interactive rebase of your branch onto the latest main, keeping just your one commit that you want to keep, that should resolve these issues. If you haven't done that before, lmk and I can help find a tutorial.

luzpaz:
Ok, if the CIs complete, then this should be good to go. Sorry for all the noise :grimacing:

jminor:
Thanks!

PR #1953, opened 2025-10-08
Export C++ symbols part 2

### LINKED ISSUE #1919: Exception when using CMAKE_CXX_VISIBILITY_PRESET=hidden
When using the option `CMAKE_CXX_VISIBILITY_PRESET=hidden`, an exception can be thrown when using metadata in an application. It seems that classes created in the application can have different type information than the classes in OTIO, causing an exception when serializing the OTIO file:
```
libc++abi: terminating due to uncaught exception of type std::bad_any_cast: bad any cast
```

This only seems to happen with some compilers (clang), and when OTIO is linked as a shared library.

Example:
```
#include <opentimelineio/marker.h>
#include <opentimelineio/timeline.h>

int main(int argc, char** argv)
{
    OTIO_NS::SerializableObject::Retainer<OTIO_NS::Timeline> timeline(new OTIO_NS::Timeline);
    OTIO_NS::SerializableObject::Retainer<OTIO_NS::Track> track(new OTIO_NS::Track);
    timeline->tracks()->append_child(track);
    OTIO_NS::SerializableObject::Retainer<OTIO_NS::Marker> marker(new OTIO_NS::Marker);

    //
    // This AnyDictionary will have different type information than OTIO.
    //
    OTIO_NS::AnyDictionary md;
    md["type"] = static_cast<int64_t>(23);
    marker->metadata()["kdenlive"] = md;
    track->markers().push_back(marker);

    //
    // std::bad_any_cast will be thrown when serializing the file with -DCMAKE_CXX_VISIBILITY_PRESET=hidden
    //
    if (!timeline->to_json_file("test.otio"))
    {
        return 1;
    }
    return 0;
}
```

Example CMake file:
```
cmake_minimum_required(VERSION 3.3)
project(test)

set(CMAKE_CXX_STANDARD 17)

find_package(OpenTimelineIO REQUIRED)

set(CMAKE_CXX_VISIBILITY_PRESET hidden)

add_executable(test test.cpp)
target_link_libraries(test OTIO::opentimelineio)
```

Adding `__attribute__((__visibility__("default")))` to the `AnyDictionary` class seems to fix the example.

Related:
https://developers.redhat.com/articles/2021/10/27/compiler-option-hidden-visibility-and-weak-symbol-walk-bar#the_llvm_package_build
https://cmake.org/cmake/help/latest/module/GenerateExportHeader.html

Thanks to @jlskuz for troubleshooting the issue.

darbyjohnston:
Fixes #1919

This is a replacement for PR #1920

These changes are based on the recommendations from @meshula for exporting symbols without using the CMake GenerateExportHeader functionality.

This PR is a WIP, I've just done the minimum amount of exporting to try out the process. With these changes my test app works OK, without them I get this exception as mentioned in the issue:
```
libc++abi: terminating due to uncaught exception of type std::bad_any_cast: bad any cast
```

To do / questions:
* Should the export.h file be shared by the opentime and opentimelineio libraries, or should each library get it's own export.h file? (EDIT: I added a separate export.h file for opentime and opetimelineio, with some shared functionality in the opentime export.h)
* The AnyDictionary class exposes some symbols with the `using` keyword; can these be exported or do they need to be replaced with function wrappers? (EDIT: It looks like this is not necessary since everything is inline)
* For the opentimelineio defines I used OTIO_API for brevity (instead of OPENTIMELINEIO_API)

My test app for reference:
```
#include <opentimelineio/marker.h>
#include <opentimelineio/timeline.h>

int main(int argc, char** argv)
{
    OTIO_NS::SerializableObject::Retainer<OTIO_NS::Timeline> timeline(new OTIO_NS::Timeline);
    OTIO_NS::SerializableObject::Retainer<OTIO_NS::Track> track(new OTIO_NS::Track);
    timeline->tracks()->append_child(track);
    OTIO_NS::SerializableObject::Retainer<OTIO_NS::Marker> marker(new OTIO_NS::Marker);
    OTIO_NS::AnyDictionary md;
    md["type"] = static_cast<int64_t>(23);
    marker->metadata()["kdenlive"] = md;
    track->markers().push_back(marker);
    if (!timeline->to_json_file("test.otio"))
    {
        return 1;
    }
    return 0;
}
```

The CMake file:
```
cmake_minimum_required(VERSION 3.3)
project(test)

set(CMAKE_CXX_STANDARD 17)

find_package(OpenTimelineIO REQUIRED)

set(CMAKE_CXX_VISIBILITY_PRESET hidden)

add_executable(test test.cpp)
target_link_libraries(test OTIO::opentimelineio)
```

### DIFF
--- a/src/opentime/export.h
+++ b/src/opentime/export.h
@@ -3,35 +3,38 @@
 
 #pragma once
 
+// For an explanation of how these export defines work, see:
+// https://github.com/PixarAnimationStudios/OpenUSD/blob/dev/pxr/base/arch/export.h
 #if defined(_WINDOWS)
 #   if defined(__GNUC__) && __GNUC__ >= 4 || defined(__clang__)
-#       define ARCH_EXPORT __attribute__((dllexport))
-#       define ARCH_IMPORT __attribute__((dllimport))
-#       define ARCH_HIDDEN
-#       define ARCH_EXPORT_TYPE
+#       define OPENTIMELINIO_EXPORT __attribute__((dllexport))
+#       define OPENTIMELINIO_IMPORT __attribute__((dllimport))
+#       define OPENTIMELINIO_HIDDEN
+#       define OPENTIMELINIO_EXPORT_TYPE
 #   else
-#       define ARCH_EXPORT __declspec(dllexport)
-#       define ARCH_IMPORT __declspec(dllimport)
-#       define ARCH_HIDDEN
-#       define ARCH_EXPORT_TYPE
+#       define OPENTIMELINIO_EXPORT __declspec(dllexport)
+#       define OPENTIMELINIO_IMPORT __declspec(dllimport)
+#       define OPENTIMELINIO_HIDDEN
+#       define OPENTIMELINIO_EXPORT_TYPE
 #   endif
 #elif defined(__GNUC__) && __GNUC__ >= 4 || defined(__clang__)
-#   define ARCH_EXPORT __attribute__((visibility("default")))
-#   define ARCH_IMPORT
-#   define ARCH_HIDDEN __attribute__((visibility("hidden")))
+#   define OPENTIMELINIO_EXPORT __attribute__((visibility("default")))
+#   define OPENTIMELINIO_IMPORT
+#   define OPENTIMELINIO_HIDDEN __attribute__((visibility("hidden")))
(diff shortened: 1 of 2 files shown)


jlskuz:
Looks good as far as I can tell

PR #1955, opened 2025-10-09
Initialize test_v2d.py variables

darbyjohnston:
I have been seeing occasional CI failures with this error:
```
FAIL: test_serialization_round_trip (test_v2d.V2dTests.test_serialization_round_trip)
  File "D:/a/OpenTimelineIO/OpenTimelineIO/tests/test_v2d.py", line 112, in test_serialization_round_trip
    self.assertEqual(v2d, deserialized)
AssertionError: otio.schema.V2d(x=1.2375583497411374e+214, y=3.503154865959304e+151) != otio.schema.V2d(x=1.2375583497411374e+214, y=3.5031548659593044e+151)

Ran 489 tests in 13.798s

FAILED (failures=1, skipped=2)
make: *** [Makefile:79: coverage-core] Error 1
Error: Process completed with exit code 2.
```

It looks like the variables being used are not initialized, causing them to have random values that can trigger the failure. I'm not sure why the serialization is failing but we should probably initialize the values so the tests are consistent.

### DIFF
--- a/tests/test_v2d.py
+++ b/tests/test_v2d.py
@@ -101,12 +101,12 @@ class V2dTests(unittest.TestCase, otio_test_utils.OTIOAssertions):
         self.assertEqual(otio.schema.V2d.baseTypeEpsilon(), sys.float_info.epsilon)
 
     def test_json_serialization(self):
-        serialized = otio.adapters.otio_json.write_to_string(otio.schema.V2d())
+        serialized = otio.adapters.otio_json.write_to_string(otio.schema.V2d(0.1, 0.2))
         json_v2d = json.loads(serialized)
         self.assertEqual(json_v2d["OTIO_SCHEMA"], "V2d.1")
 
     def test_serialization_round_trip(self):
-        v2d = otio.schema.V2d()
+        v2d = otio.schema.V2d(0.3, 0.4)
         serialized = otio.adapters.otio_json.write_to_string(v2d)
         deserialized = otio.adapters.otio_json.read_from_string(serialized)
         self.assertEqual(v2d, deserialized)


apetrynet:
Good catch!
Sneaky bug indeed.

However, should we look into why an uninitialized `v2d` gets a random value? I would think the serialized value of an "empty" v2d should be `null`, `[]` or even `[0,0]`?
Perhaps we should look at the otio implementation of `v2d`? I assume the random values stem from old data in memory at some location, but I'm in no way a C++ expert

darbyjohnston:
The default `v2d` constructor in C++ doesn't initialize the values, so it can be any random memory value as you point out. This is done on purpose, I believe as an optimization. For example if you create a vector of ten thousand `v2d` values, you might not want the overhead of assigning zero to all of them.

I'm not a Python expert, does Python have the concept of uninitialized values? Maybe the Python bindings should initialize the values unlike the C++ code.

It's also interesting that the serialization fails, but maybe crazy numbers like `3.503154865959304e+151` aren't handled.

jminor:
Python doesn't have that notion, and some languages are moving away from uninitialized variables as a guard against unexpected bugs or security issues (e.g. reading passwords from someone else's freed memory). 
I think we should make that constructor initialize with zeros if possible. Maybe that's possible in the bindings layer?
With that said, I would happily accept this PR as-is and handle that task in a future PR.

darbyjohnston:
It should be fairly simple to initialize the values in the bindings, I can create a separate PR for that. I think we should merge this PR so there aren't any more random CI failures. Plus using values like `v2d(0.1, 0.2)` is probably better for serialization tests than `v2d(0, 0)`.

jminor (approved):

PR #1956, opened 2025-10-10
Initialize Python Imath values

darbyjohnston:
This PR initializes Imath values created in Python to 0.0. In C++ the default constructor does not initialize the values which means they could be anything. This is done as an optimization for C++ but is probably not necessary and also unexpected in Python.

For example this was recently causing CI failures: #1955

### DIFF
--- a/src/opentimelineio/color.cpp
+++ b/src/opentimelineio/color.cpp
@@ -14,17 +14,17 @@ Color::Color(
     double const         b,
     double const         a,
     std::string const&   name)
-    : _name(name),
-    _r(r),
+    : _r(r),
     _g(g),
     _b(b),
-    _a(a) {}
+    _a(a),
+    _name(name) {}
 
-Color::Color(Color const& other) : _name(other.name()),
-                                   _r(other.r()),
+Color::Color(Color const& other) : _r(other.r()),
                                    _g(other.g()),
                                    _b(other.b()),
-                                   _a(other.a()) {}
+                                   _a(other.a()),
+                                   _name(other.name()) {}
 
 const Color Color::pink(1.0, 0.0, 1.0, 1.0, "Pink");
 const Color Color::red(1.0, 0.0, 0.0, 1.0, "Red");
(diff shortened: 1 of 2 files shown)


src/py-opentimelineio/opentimelineio-bindings/otio_imath.cpp
```diff
@@ -101,7 +101,7 @@ static void define_imath_2d(py::module m) {
             });
 
     py::class_<IMATH_NAMESPACE::Box2d>(m, "Box2d", py::module_local())
-        .def(py::init<>())
+        .def(py::init<>([]() { return IMATH_NAMESPACE::Box2d(IMATH_NAMESPACE::V2d(0.0, 0.0)); }))
         .def(py::init<IMATH_NAMESPACE::V2d>())
         .def(py::init<IMATH_NAMESPACE::V2d, IMATH_NAMESPACE::V2d>())
```
reinecke: Looks like there may be a couple sneaky ones!

src/py-opentimelineio/opentimelineio-bindings/otio_imath.cpp
```diff
@@ -101,7 +101,7 @@ static void define_imath_2d(py::module m) {
             });
 
     py::class_<IMATH_NAMESPACE::Box2d>(m, "Box2d", py::module_local())
-        .def(py::init<>())
+        .def(py::init<>([]() { return IMATH_NAMESPACE::Box2d(IMATH_NAMESPACE::V2d(0.0, 0.0)); }))
         .def(py::init<IMATH_NAMESPACE::V2d>())
```
darbyjohnston: @reinecke In the meeting today we were wondering if this also needs to be initialized:
```
.def(py::init<IMATH_NAMESPACE::V2d>())
```
I believe this is OK as is, since this constructs a `Box2d` from a `V2d`. Since the `V2d` is now initialized to zero with this change, the `Box2d` will also be initialized to zero:
```
v = otio.schema.V2d() # initialized to zero
box = otio.schema.Box2d(v) # also initialized to zero
```

PR #1958, opened 2025-10-23
Reduce dependabot frequency

darbyjohnston:
As discussed in the TSC meeting today, it would be nice to reduce the frequency of dependabot PRs.


jhodges10 (approved):
Great idea, looks good to me!

ssteinbach (approved):

PR #1970, opened 2025-11-06
Fixed "Unknown Distribution Format" in PyPI publish action

reinecke:
Fixed an issue where the PyPI publish would throw this error on release:
```
Checking dist/opentimelineio-0.18.0.tar.gz: PASSED
Checking dist/wheel-macos-14-cp310: ERROR    InvalidDistribution: Unknown distribution format:                      
         'wheel-macos-14-cp310'      
```

It appears that multi-matched artifacts would each extract to their own dir rather than merge flatly - so the wheel was in a subdir with the artifact name instead of directly in the `dist` dir.


jhodges10:
Looks like it'll work to me! Let's give it a shot.

PR #1961, opened 2025-10-28
Update codecov action to v5

jminor:
OTIO's code coverage reports have been stuck in a stale state for a while now. Pull requests report coverage based on a commit from Apr 14, 2024 instead of the current main. For example:

> ✅ All modified and coverable lines are covered by tests.
✅ Project coverage is 85.10%. Comparing base ([c0e97b0](https://app.codecov.io/gh/AcademySoftwareFoundation/OpenTimelineIO/commit/c0e97b0c195881bec166c0a581dbae2999217265?dropdown=coverage&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=AcademySoftwareFoundation)) to head ([646eaec](https://app.codecov.io/gh/AcademySoftwareFoundation/OpenTimelineIO/commit/646eaec9c81d1e76ac0640788029bafe0a4e936d?dropdown=coverage&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=AcademySoftwareFoundation)).
⚠️ Report is 80 commits behind head on main.
> ❌ Your changes status has failed because you have indirect coverage changes. Learn more about [Unexpected Coverage Changes](https://docs.codecov.com/docs/unexpected-coverage-changes) and [reasons for indirect coverage changes](https://docs.codecov.com/docs/unexpected-coverage-changes#reasons-for-indirect-changes).
> The "base" commit that codecov compares against seems to be this one: https://github.com/AcademySoftwareFoundation/OpenTimelineIO/commit/c0e97b0c195881bec166c0a581dbae2999217265 for reasons that are unclear.

In an attempt to get this fixed I have just re-read the instructions https://github.com/codecov/codecov-action, re-entered the repo's CODECOV token in the "secrets" settings area, and in this PR, updated the codecov action to v5.

Let's see if that helps...

PR #1977, opened 2025-11-09
Run tests on wheels

### LINKED ISSUE #1027: Add Test pass for build artifacts
## Feature Request

With github actions producing wheels to deliver to pypi, it would be helpful to have a separate part of our build process to validate those exact artifacts before they can become candidates for pypi delivery.

## Description

At a minimum, smoke tests could be run that did something simple like:

1. Import otio
2. create an example timeline
3. write it to disk
4. read the timeline back into disk

The testing should be run on target platforms called out as supported.

In a perfect world, the full unittest suite would be run against the wheels that are being delivered.

## Context

Some things to consider:

- In testing, it is important to fail as fast as possible. I think we'd favor relying on the build time unittests a bit more in exchange for faster iteration time on build failures.
- At the time of writing, there are 17 wheels being built. Running tests to cover every one of these may not be worth the effort. It may be best to choose a representative set of platform/python version combos to maximize coverage while minimizing the variant combos of test machines.

reinecke:
This change re-shuffles the CI so that:

- A "smoke test" set of builds are run on one OS and generate coverage
- A separate "build and test only" set of platforms/python version run off the sdist (currently mingw)
- All wheels are built after smoke tests pass
- The test suite is run for each OS/python combo against the wheels

~Ideally this would take the suggestion from @JeanChristopheMorinPerso and run against `sdist` - that change hasn't been made.~
Update: after investigation, cibuildwheel isn't really meant to happen from sdist sources. I did a pass to make our build and test only phase run off it though - since those platforms would be running off sdist in the wild.

Additionally, if we could capture coverage from one of the cibuildwheel runs, we could probably eliminate  the smoke test phase.

- [x] ~Consider dropping the smoke test and just harvest coverage from testing the wheels~
    - Update: we decided against this because you need particular build options to generate needed coverage artifacts
- [x] Investigate having the wheel build to happen off the sdist
- [x] Currently the build and test only phase is hitting a test error - getting it to find the sdist tar.gz when the name is indeterminate (includes the version) is proving tricky
- [ ] Re-shuffle the order of repo checkout and wheel install in the wheel test action to match the test only step
- [ ] Reduce the matrix on the smoketest build so that it only does the one it needs for coverage - consider renaming explicitly as "generate_coverage" or similar

Fixes #1027 
This supersedes #1961 - the changes have been added to this PR

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -197,7 +197,7 @@ jobs:
           python -m pip install --break-system-packages --upgrade -r tests/requirements.txt
       - name: Build and Install OpenTimelineIO
         run: |
-          python -m pip install ./sdist/opentimelineio-*.tar.gz -v --break-system-packages
+          pkg=`ls -1 sdist/opentimelineio-*.tar.gz` python -m pip install $pkg -v --break-system-packages
       - name: Run tests w/ python coverage
         run: make ci-postbuild
 
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -121,9 +121,12 @@ jobs:
           python -m pip install --upgrade pip setuptools wheel "flake8>=3.5" check-manifest
       - name: Run check-manifest and lint check
         run: make ci-prebuild
-      - name: Build and Install
+      - name: Install test dependencies
         run: |
-          python -m pip install -r tests/requirements.txt -v --break-system-packages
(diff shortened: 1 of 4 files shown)


reinecke:
I'm ready for feedback on this - just the one open questions about whether or not we really need the smoketests or if I should move the coverage reports to the tests we run on the equivalent wheel.

PR #1989, opened 2026-01-08
VFX Reference Platform version matrix

jminor:
This PR adds VERSIONS.md with a table describing which versions of OTIO support which versions of the VFX Reference Platform. This question comes up often enough, and it it awkward to find historical information, that I figured it would be helpful to have this reference. Also, the resulting table makes it clear that OTIO supports the current year + 3 prior years.

Here is the table as an image for convenience in case the PR doesn't render it: 
image


JeanChristopheMorinPerso:
@jminor I think you will need to add an exclude on the new file in `MANIFEST.in` to fix https://github.com/AcademySoftwareFoundation/OpenTimelineIO/actions/runs/20802341301/job/60044736977#step:7:20

reinecke (approved):

PR #1991, opened 2026-01-16
Fix issues when building OTIO in debug on Windows (and using Debug Python)

### LINKED ISSUE #1173: Windows Python debug pyd suffix
### Build Problem

I note in the current build scripts that the `d` suffix is added for debug builds on Windows

https://github.com/PixarAnimationStudios/OpenTimelineIO/blob/main/CMakeLists.txt#L166-L171

However, in the Python build scripts, the .pyds have a `_d` suffix for a debug Python build on Windows, see

https://github.com/python/cpython/blob/3.7/PC/python_ver_rc.h#L13-L17

Thanks,
Mark

cedrik-fuoco-adsk:

```Fixes #1173```

**Summarize your change.**

- Updated the Windows Debug library postfix to _d to match Python’s expected debug naming convention.
- Fixed pybind11 GIL assertion risk by lazily initializing `_value_to_any` (avoid constructing py::none() during static initialization) in `otio_utils.cpp`

OpenRV builds OTIO from source to avoid compatibility issues with its custom Python distribution. 
When pairing Debug OTIO with Debug Python on Windows, I ran into:

- A Debug-postfix mismatch (Python expects `_d`), which breaks resolution/loading of `pyd`.
- pybind11 GIL assertions caused by creating Python objects during static initialization (before the interpreter/GIL is guaranteed to be ready).


rogernelson (approved):

JeanChristopheMorinPerso:
Interesting finding! We fixed a similar issue 4 years ago in https://github.com/AcademySoftwareFoundation/OpenTimelineIO/pull/1070.

PR #1992, opened 2026-01-18
Add missing export markers (OTIO_API) for constructors

jlskuz:
This is a follow up to #1953
I am working on MSVC support for Kdenlive (currently MinGW is used for Windows) and hit

```
kdenliveLib.lib(otioexport.cpp.obj) : error LNK2019: unresolved external symbol "public: __cdecl opentimelineio::v0_18_1::GeneratorReference::GeneratorReference(class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> > const &,class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> > const &,class std::optional<class opentime::v0_18_1::TimeRange> const &,class opentimelineio::v0_18_1::AnyDictionary const &,class opentimelineio::v0_18_1::AnyDictionary const &,class std::optional<class Imath_3_1::Box<class Imath_3_1::Vec2<double> > > const &)" (??0GeneratorReference@v0_18_1@opentimelineio@@QEAA@AEBV?$basic_string@DU?$char_traits@D@std@@V?$allocator@D@2@@std@@0AEBV?$optional@VTimeRange@v0_18_1@opentime@@@4@AEBVAnyDictionary@12@2AEBV?$optional@V?$Box@V?$Vec2@N@Imath_3_1@@@Imath_3_1@@@4@@Z) referenced in function "private: void __cdecl OtioExport::exportClip(class std::shared_ptr<class TimelineItemModel> const &,struct OtioExport::ClipData const &,struct opentimelineio::v0_18_1::SerializableObject::Retainer<class opentimelineio::v0_18_1::Track> &)" (?exportClip@OtioExport@@AEAAXAEBV?$shared_ptr@VTimelineItemModel@@@std@@AEBUClipData@1@AEAU?$Retainer@VTrack@v0_18_1@opentimelineio@@@SerializableObject@v0_18_1@opentimelineio@@@Z)
bin\cachetest.exe : fatal error LNK1120: 1 unresolved externals 
```

this brought to my attention that only some classes have `OTIO_API` on their constructors, while others, such as `GeneratorReference`, have not causing the above failure. As far as I understand all of these classes and ctors are public API however (please check and confirm) and hence should have `OTIO_API`.

@darbyjohnston Was #1953 just in complete or did you only add it to certain classes on purpose? This does probably not only affect the ctors, but also other class members. Should I add `OTIO_API` to all of them (either here or in a seperate PR)? Is it safe to assume that everything in the `public:` section of `src/opentimelineio/*.h` files is public api and hence should have `OTIO_API`?

### DIFF
--- a/src/opentimelineio/effect.h
+++ b/src/opentimelineio/effect.h
@@ -27,7 +27,7 @@ public:
     /// @param name The name of the effect.
     /// @param metadata The metadata for the clip.
     /// @param enabled Whether the effect is enabled.
-    Effect(
+    OTIO_API Effect(
         std::string const&   name        = std::string(),
         std::string const&   effect_name = std::string(),
         AnyDictionary const& metadata    = AnyDictionary(),
--- a/src/opentimelineio/freezeFrame.h
+++ b/src/opentimelineio/freezeFrame.h
@@ -25,7 +25,7 @@ public:
     ///
     /// @param name The name of the time effect.
     /// @param metadata The metadata for the time effect.
-    FreezeFrame(
+    OTIO_API FreezeFrame(
         std::string const&   name     = std::string(),
         AnyDictionary const& metadata = AnyDictionary());
 
--- a/src/opentimelineio/gap.h
+++ b/src/opentimelineio/gap.h
@@ -47,6 +47,7 @@ public:
     /// @param markers The list of markers for the gap. Note that the
     /// the gap keeps a retainer to each marker.
     /// @param metadata The metadata for the gap.
+    OTIO_API
     Gap(RationalTime                duration,
         std::string const&          name     = std::string(),
         std::vector<Effect*> const& effects  = std::vector<Effect*>(),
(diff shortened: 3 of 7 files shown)


darbyjohnston:
Hi @jlskuz, thanks for catching this, there definitely may have been places I missed adding the exports. I think the rule of thumb was all member functions that are not inline, but all of the details are here:
https://github.com/PixarAnimationStudios/OpenUSD/blob/dev/pxr/base/arch/export.h

Were you seeing errors about the `ErrorStatus` class?

jlskuz:
No, I did not see an error about the `ErrorStatus` class. The only real world issue I saw with Kdenlive was related to `GeneratorReference`

darbyjohnston:
I was wondering because it looks like the `ErrorStatus` members are inline, which I didn't think needed the export decorations (inline makes them visible by default).

jlskuz:
Oh, yes I confused at least one operator with a conatructor

darbyjohnston:
Maybe remove the exports from the `ErrorStatus` constructors and operators, otherwise this looks good to me.

darbyjohnston:
Hi @jlskuz, just wanted to check in and see if you agreed with my comment about removing the changes for the inline constructors, otherwise this looks good to me. If you don't have the time I can also make the changes.

jlskuz:
Thanks for the ping and sorry, I forgot about it... I just made the fix and removed the api markers from `ErrorStatus`

darbyjohnston:
No problem, thanks for making the updates.

PR #1993, opened 2026-02-20
Replace CMake generated version files with checked-in versions

darbyjohnston:
In PR #1907 we added version headers so applications could identify which OpenTimelineIO they were using. However this also introduced a dependency on CMake to generate the headers, which can cause issues for other projects that use OTIO without CMake (for example the OTIO Swift bindings).

This PR replaces the CMake file generation with checked-in files, and instead has CMake parse the version from the checked-in header files. This is similar to what OpenEXR does for their version handling.


src/opentime/version.h
```diff
@@ -0,0 +1,17 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Contributors to the OpenTimelineIO project
+
+#pragma once
+
+#define OPENTIME_VERSION_MAJOR 0
+#define OPENTIME_VERSION_MINOR 19
+#define OPENTIME_VERSION_PATCH 0
+#define OPENTIME_VERSION v0_19_0
+#define OPENTIME_VERSION_NS v0_19
```
meshula: Here's a variant that leaves no room for human forgetfulness

#define OPENTIME_VERSION_MAJOR 0
#define OPENTIME_VERSION_MINOR 19
#define OPENTIME_VERSION_PATCH 0
#define OPENTIME_VERSION
v##OPENTIME_VERSION_MAJOR####OPENTIME_VERSION_MINOR####OPENTIME_VERSION_PATCH
#define OPENTIME_VERSION_NS
v##OPENTIME_VERSION_MAJOR##_##OPENTIME_VERSION_MINOR

src/opentimelineio/version.h
```diff
@@ -3,11 +3,11 @@
...
 
-#define OPENTIMELINEIO_VERSION_MAJOR @PROJECT_VERSION_MAJOR@
-#define OPENTIMELINEIO_VERSION_MINOR @PROJECT_VERSION_MINOR@
-#define OPENTIMELINEIO_VERSION_PATCH @PROJECT_VERSION_PATCH@
-#define OPENTIMELINEIO_VERSION v@PROJECT_VERSION_MAJOR@_@PROJECT_VERSION_MINOR@_@PROJECT_VERSION_PATCH@
-#define OPENTIMELINEIO_VERSION_NS v@PROJECT_VERSION_MAJOR@_@PROJECT_VERSION_MINOR@
+#define OPENTIMELINEIO_VERSION_MAJOR 0
+#define OPENTIMELINEIO_VERSION_MINOR 19
+#define OPENTIMELINEIO_VERSION_PATCH 0
+#define OPENTIMELINEIO_VERSION v0_19_0
```
meshula: same DRY note as above

meshula (changes requested):
looking good, minor request

src/opentime/version.h
darbyjohnston: Maybe I'm missing something, but that doesn't seem to be working for me using VS2022. I get: `vOPENTIME_VERSION_MAJOR_OPENTIME_VERSION_MINOR_OPENTIME_VERSION_PATCH`.

meshula:
Sorry, I don't know how I typed 4 #'s in a row. I'm still in favor of concatenation even with a helpy macro!

darbyjohnston:
> Sorry, I don't know how I typed 4 #'s in a row. I'm still in favor of concatenation even with a helpy macro!

That requires updating every source file with the new macro. That seems like a lot of work to not duplicate a number, plus it also adds complexity which I think undermines the benefits of DRY.

jminor:
Trying this PR out on the Swift bindings here: https://github.com/OpenTimelineIO/OpenTimelineIO-Swift-Bindings/pull/78

jminor:
This seems to be working well for the Swift bindings.

Related: I noticed one typo in the README.md which says
```
namespace otio = opentimelineio::OPENTIMELINEIO_VERSION;
```
but I think it should be
```
namespace otio = opentimelineio::OPENTIMELINEIO_VERSION_NS;
```

darbyjohnston:
Good catch, I updated the example code to use our namespace macro `OTIO_NS`.

jminor:
Oh cool. I'll update the Swift bindings to use that macro too.

@meshula are you okay if we skip your suggestion about the ## macro concatenation?

meshula (approved):
fancy macros don't block merging, thanks for the PR!

PR #1996, opened 2026-03-03
Disable pr-tools.yml due to security vulnerability

jminor:
It has been brought to our attention that use of `pull_request_target` creates an opportunity for bad actors to do damage by posting pull requests that run undesirable code within this project's base repo. This PR disables the one workflow that uses that feature.

I opted to leave that code in place, just commented out, in the hopes that some alternate safer implementation becomes available.

Read more here:
- https://www.stepsecurity.io/blog/hackerbot-claw-github-actions-exploitation
- https://paulserban.eu/blog/post/why-is-pullrequesttarget-so-dangerous-a-security-explainer/


jminor:
FYI, we have some limited protection from this attack due to this setting on the repo which requires a maintainer to approve actions from new contributors, but a misleading PR could trick someone into allowing it to run. 
image

PR #1998, opened 2026-03-05
Replace GitHub action version tags with commit SHAs

darbyjohnston:
This PR replaces the version tags used in the GitHub actions with their corresponding commit SHAs. This is intended to improve security since the SHAs are immutable while the tags can potentially change.

Note that it looked like `pypa/gh-action-pypi-publish` was pointing to a branch name, so I used the latest version of v1 (v1.13.0).

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -10,11 +10,11 @@ jobs:
   check-links:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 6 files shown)


jminor:
After we land this, we can consider enabling this toggle in the repo settings that says "Require actions to be pinned to a full-length commit SHA"

image

jminor (approved):

PR #2002, opened 2026-03-24
Fix various warnings

### LINKED ISSUE #1408: Address build-time warnings
We aim to meet the OpenSSF Best Practices passing or higher badge level. One of the requirements is to address compiler warnings in the project's source code.

I don't see any warnings in our GitHub Action build logs - is that because there are none, or because they are suppressed/hidden?

Details from OpenSSF Best Practices:
> The project MUST address warnings. [warnings_fixed]  Hide details
These are the warnings identified by the implementation of the [warnings](https://bestpractices.coreinfrastructure.org/en/projects/2288#warnings) criterion. The project should fix warnings or mark them in the source code as false positives. Ideally there would be no warnings, but a project MAY accept some warnings (typically less than 1 warning per 100 lines or less than 10 warnings).

darbyjohnston:
Fixes #1408

This PR fixes various warnings in the OTIO build:
* C++ warnings
* CMake warnings
* ~~setuptools warnings~~ (these were actually necessary for `check-manifest`)

We still have a couple of C++ warnings about functions that have been deprecated, these should probably stay on to remind us to remove them at some point.

The setuptools warnings:
```
warning: no previously-included files found matching 'CHANGELOG.md'
```
Seem to be coming from excluding files in the manifest that were not included in the first place, but it would be good for someone with more Python knowledge to double check this. There are still a couple of warnings about temporary files (*.pyc, .DS_Store, etc.) that seem OK to leave in.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,8 @@
 # FindPython's Development.Module component was added in 3.18.
 # Require 3.18.2+ because pybind11 recommends it.
 cmake_minimum_required(VERSION 3.18.2)
+# This policy CMP0177 can be removed when we upgrade to CMake >= 3.31.
+cmake_policy(SET CMP0177 NEW)
 
 #------------------------------------------------------------------------------
 # Project Metadata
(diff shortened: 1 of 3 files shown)


jhodges10 (approved):
Consensus with the TSC call is that this looks good to merge, given all the tests are passing and it's a very minimal change.

PR #2015, opened 2026-04-04
Stream and Channel Mapping Support

### LINKED ISSUE #145: Media Reference Stream Specification support
Currently media references are focussed on locating specific assets on and identifying timing metadata for them.

When OTIO is used to drive a system that will concretely pull together media data, it will likely be necessary to identify a specific stream within a file, or even a specific channel within an audio stream.

reinecke:

Fixes #145

**Summarize your change.**

To for an explanation of the implementation, check out [the docs](https://github.com/reinecke/OpenTimelineIO/blob/ea0588fb19e887bce318ec14df7b1451e08b556a/docs/tutorials/channel-mapping.md).

This PR implements [the proposal](https://gist.github.com/ssteinbach/d98a25774c3b12c8863daaf0b2b1b6c7?permalink_comment_id=4714232) that was created to address #145 and add the ability to do stream/channel addressing and mapping from sources into the composition.

The C++ implementation of the mappings was built following the pattern of Clip.media_references.

Unittests were added for the new features.

Assisted-by: Claude Sonnet 4.6


reinecke:
@mrlimbic here is my first pass at stream/channel mapping. Please feel free to provide feedback here.

docs/tutorials/channel-mapping.md
```diff
@@ -0,0 +1,373 @@
+# Stream and Channel Mapping
+
+## Overview
+
+OTIO's stream and channel mapping support allows media references to describe the individual streams they contain (video eyes, audio channels, etc.) and for timeline items to select or remix those streams via effects. The system has three layers: **addressing** a stream within a container, **describing** a stream's semantic role, and **selecting or mixing** streams on a clip.
```
darbyjohnston, line 5: I haven't seen "video eyes" before, is that a convention?

docs/tutorials/channel-mapping.md
```diff
@@ -0,0 +1,373 @@
...
+           address=otio.schema.IndexStreamAddress(index=1),
+           kind=otio.schema.TrackKind.Audio,
+       ),
+   })
+   ```
+
+2. **Additional streams** SHOULD prefix an `Identifier` value to form a unique key within the media reference.
+
+   ```python
+   # A music stem alongside the audio mix
```
darbyjohnston, line 155: I thought this was a typo ('stem' -> 'stream'), but I guess `stem` is an audio term?

docs/tutorials/channel-mapping.md
```diff
@@ -0,0 +1,373 @@
...
+---
+
+## Layer 1: Addressing a Stream Within a Container
+
+A `StreamAddress` identifies where a specific stream lives inside a media file. The appropriate `StreamAddress` subclass depends on how the container format identifies its streams. The `StreamAddress` subclass should be chosen based on how closely it matches the container's native identification scheme.
+
+| Class | Use when |
+|---|---|
+| `IndexStreamAddress` | The container uses a single integer index per stream (e.g. WAV channel index, ffmpeg stream index) |
+| `StreamChannelIndexStreamAddress` | The container organises media into tracks, each with one or more channels (e.g. MP4/MOV, MXF) |
```
darbyjohnston, line 16: Could we simplify this class name a bit, maybe `ChannelStreamAddress`?

src/opentimelineio/audioMixMatrix.h
```diff
@@ -0,0 +1,81 @@
...
+///     }
+///   }
+/// @endcode
+class OTIO_API_TYPE AudioMixMatrix : public Effect
+{
+public:
+    /// @brief This struct provides the AudioMixMatrix schema.
+    struct Schema
+    {
+        static char constexpr name[]  = "AudioMixMatrix";
```
darbyjohnston, line 44: Thanks, I like the explicit type here vs. the `static auto constexpr name` we have in the rest of the headers. :)

src/opentimelineio/indexStreamAddress.h
```diff
@@ -0,0 +1,42 @@
...
+        static char constexpr name[]  = "IndexStreamAddress";
+        static int constexpr version = 1;
+    };
+
+    using Parent = StreamAddress;
+
+    OTIO_API explicit IndexStreamAddress(int64_t index = 0);
+
+    /// @brief Return the stream index.
+    OTIO_API int64_t index() const noexcept { return _index; }
```
darbyjohnston, line 27: I'm not sure if the export macros are needed on inline functions, we should double check that.

src/opentimelineio/mediaReference.h
```diff
@@ -66,15 +68,37 @@ class OTIO_API_TYPE MediaReference : public SerializableObjectWithMetadata
...
+    /// Keys should use values from `StreamInfo::Identifier` to signal "primary"
+    /// streams (e.g. `StreamInfo::Identifier::monocular` for a "traditional"
+    /// video).
+    /// Additional streams should prefix a `StreamInfo::Identifier` value to
+    /// form a key unique within that media reference (e.g. something like
+    /// "music_stereo_right" could be used for a music stem provided in addition
+	/// to the "primary" audio).
+    /// However, keys may be any unique string —
+    /// for example, spatial audio objects whose identity is only meaningful
+    /// alongside positional renderomg metadata, or production audio tracks
```
darbyjohnston, line 84: `renderomg` -> `rendering`?

src/opentimelineio/streamAddress.cpp
```diff
@@ -0,0 +1,27 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Contributors to the OpenTimelineIO project
+
+#include "opentimelineio/streamAddress.h"
+
+namespace opentimelineio { namespace OPENTIMELINEIO_VERSION_NS {
+
+StreamAddress::StreamAddress()
+    : Parent()
```
darbyjohnston, line 9: I think this is redundant, by default C++ will call the base class constructor with no arguments.

src/opentimelineio/streamAddress.h
```diff
@@ -0,0 +1,39 @@
...
+    /// @brief This struct provides the StreamAddress schema.
+    struct Schema
+    {
+        static char constexpr name[]  = "StreamAddress";
+        static int constexpr version = 1;
+    };
+
+    using Parent = SerializableObject;
+
+    OTIO_API StreamAddress();
```
darbyjohnston, line 30: We could use ` = default` here to save a bit of code since the constructor is empty. Just a suggestion though, I'm not sure we use `default` anywhere else yet. At some point we could also do a separate pass over the code replacing empty ctors/dtors with `default`.

src/opentimelineio/streamSelector.h
```diff
@@ -0,0 +1,63 @@
...
+#include "opentimelineio/version.h"
+
+#include <vector>
+
+namespace opentimelineio { namespace OPENTIMELINEIO_VERSION_NS {
+
+/// @brief An effect that selects specific named output streams from an item.
+///
+/// Use this to select a stereo view, specific audio channels, etc.
+/// The clip will expose these streams out with the same naming.
```
darbyjohnston, line 16: Minor, but I thought the wording on this was a little confusing:
```
The clip will expose these streams out with the same naming.
```
Maybe:
```
The stream names match the names within the clip.
```

src/opentimelineio/stringStreamAddress.h
```diff
@@ -0,0 +1,42 @@
...
+    /// @brief This struct provides the StringStreamAddress schema.
+    struct Schema
+    {
+        static char constexpr name[]  = "StringStreamAddress";
+        static int constexpr version = 1;
+    };
+
+    using Parent = StreamAddress;
+
+    OTIO_API explicit StringStreamAddress(std::string const& address = std::string());
```
darbyjohnston, line 24: Minor, but we could add a Doxygen comment here for completeness:
```
/// @brief Create a new StringStreamAddress.
```

src/py-opentimelineio/opentimelineio/algorithms/timeline_algo.py
```diff
@@ -5,9 +5,7 @@
 
 import copy
 
-from . import (
-    track_algo
-)
+from . import track_algo
```
darbyjohnston, line 8: It looks like the changes in timeline_algo.py and track_algo.py are just formatting?

docs/tutorials/channel-mapping.md
```diff
@@ -0,0 +1,373 @@
...
+
+When generating these keys, try to avoid names that might collide with future `Identifier` constants unless you think your use case is in line with the precedent in that constant. In these cases, consider submitting a new constant to OTIO to include.
+
+---
+
+## Layer 3: Selecting or Mixing Streams on a Clip
+
+Stream effects are `Effect` subclasses that attach to a clip's `effects` list and transform which streams are visible downstream and under what names. They operate on the stream identifiers declared in `available_streams`.
+
+### StreamSelector
```
darbyjohnston, line 173: Just curious if we need the `StreamSelector` class, it seems like you could do the same thing with `StreamMapper`?

darbyjohnston (changes requested):
Overall the C++ code looks fine, it seems to follow all of our conventions, I just left a few minor comments. Thanks for including the Doxygen markup and export macros.

I only skimmed the Python code, it would be good to get someone with more Python experience to take a look.

At some point it might be nice to add C++ tests, but for now the Python tests exercise the functionality.

darbyjohnston:
Just to continue the discussion from today's OTIO meeting:

If multiple media references expose different stream information, how is that handled by the stream effects?

For example, say you have a clip with two media references, one with 5.1 audio and the other with mono:
* Clip
    * Media Reference: "5.1"
        * Stream Info: "left"
        * Stream Info: "surround_center_front"
        * ...
    * Media Reference: "mono"
        * Stream Info: "monaural"

How would you use an `AudioMixMatrix` effect to make the clip stereo? Either use one effect with all of the streams:
* AudioMixMatrix
    * "left"
        * "left", 1.0
        * "surround_center_front", 0.707
        * ...
        * "monaural", 1.0
    * "right"
        * "right", 1.0
        * "surround_center_front", 0.707
        * ...
        * "monaural", 1.0

Or use multiple `AudioMixMatrix` effects?
* AudioMixMatrix
    * "left"
        * "left", 1.0
        * "surround_center_front", 0.707
        * ...
    * "right"
        * "right", 1.0
        * "surround_center_front", 0.707
        * ...
* AudioMixMatrix
    * "left"
        * "monaural", 1.0
    * "right"
        * "monaural", 1.0

docs/tutorials/channel-mapping.md
reinecke: Hah - that should probably be written as "Video Views" if I follow the multi-view nomenclature.

docs/tutorials/channel-mapping.md
reinecke: Yeah - [audio "stems"](https://en.wikipedia.org/wiki/Stem_(audio)) are a term used to refer to component audio tracks. It's common, for instance, to deliver a version of a film with dialog "stems" separate from music and effects "stems" to allow more flexibility when cutting trailers or marketing material.

PR #2017, opened 2026-04-27
Fix null pointer dereferences after dynamic_cast

jminor:
Certain invalid OTIO object graphs with cycles may cause crashes due to code that assumes a dynamic_cast always returns a valid object. This PR introduces tests which demonstrate the issue and a fix to prevent the crash in those cases.

Note that the discovery of this issue and portions of the test code and fix were assisted by Claude Code Opus 4.6 and Opus 4.7. Since OTIO's policy on LLM use is still in discussion, this PR is marked as a draft.

### DIFF
--- a/src/opentimelineio/algo/editAlgorithm.cpp
+++ b/src/opentimelineio/algo/editAlgorithm.cpp
@@ -182,8 +182,18 @@ overwrite(
             if (!isEqual(second_duration.value(), 0.0))
             {
                 auto second_item = dynamic_cast<Item*>(items.front()->clone());
-                trimmed_range    = second_item->trimmed_range();
-                source_range     = TimeRange(
+                if (!second_item)
+                {
+                    // clone() failed or returned an object that is not an Item.
+                    // Bail out before dereferencing nullptr (CWE-476).
+                    if (error_status)
+                        *error_status = ErrorStatus(
+                            ErrorStatus::INTERNAL_ERROR,
+                            "failed to clone item for overwrite split");
+                    return;
+                }
+                trimmed_range = second_item->trimmed_range();
+                source_range  = TimeRange(
                     trimmed_range.start_time() + first_duration
                         + range.duration(),
                     second_duration);
@@ -363,6 +373,16 @@ insert(
         if (!isEqual(second_source_range.duration().value(), 0.0))
         {
             auto second_item = dynamic_cast<Item*>(item->clone());
+            if (!second_item)
+            {
+                // clone() failed or returned an object that is not an Item.
+                // Bail out before dereferencing nullptr (CWE-476).
+                if (error_status)
+                    *error_status = ErrorStatus(
+                        ErrorStatus::INTERNAL_ERROR,
+                        "failed to clone item for insert split");
+                return;
+            }
             second_item->set_source_range(second_source_range);
             composition->insert_child(insert_index + 1, second_item);
         }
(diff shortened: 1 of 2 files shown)


darbyjohnston:
There are a lot of unrelated formatting changes; could you ask Claude to make another pass just fixing the bugs?

meshula:
LOL @darbyjohnston It was a long time ago, but I remember when humans were able to amend whitespace issues! What times those were! ;)

meshula (commented):
Fix LGTM. I can't see any reason why returning an error status instead of dereferencing null is a regression under any circumstance.

meshula (approved):

jminor:
Clarification: The removal of trailing whitespace came from my IDE, not the LLM.

jminor:
I figured out why those files didn't already adhere to our clang-format rules. Here's a fix for that: https://github.com/AcademySoftwareFoundation/OpenTimelineIO/pull/2018

Once that lands, I can rebase this PR so it won't have that distraction.

jminor:
Will rebase again after #2019 lands.

jminor:
Okay, this is rebased on to main now that the whitespace & formatting from #2018 and #2019 are resolved. That makes this PR's changes clean & tidy.

jminor:
Hmm... the coverage report claims that the newly updated lines are not being run by the tests. Let me verify locally.

jminor:
Oh, I see the coverage report comes from the Python tests (which run some of the C++ code) but not from the run of the C++ tests, so this area of code *is* tested, but doesn't show up in the coverage report.

Let's handle that separately from the PR.

src/opentimelineio/algo/editAlgorithm.cpp
```diff
@@ -182,8 +182,18 @@ overwrite(
             if (!isEqual(second_duration.value(), 0.0))
             {
                 auto second_item = dynamic_cast<Item*>(items.front()->clone());
-                trimmed_range    = second_item->trimmed_range();
-                source_range     = TimeRange(
+                if (!second_item)
+                {
+                    // clone() failed or returned an object that is not an Item.
```
darbyjohnston, line 187: Totally fine to leave this in, but checking a nullptr maybe doesn't need an explanation. I find sometimes the generated code is a bit verbose with explaining changes.

src/opentimelineio/algo/editAlgorithm.cpp
```diff
@@ -182,8 +182,18 @@ overwrite(
...
+                    // Bail out before dereferencing nullptr (CWE-476).
+                    if (error_status)
+                        *error_status = ErrorStatus(
+                            ErrorStatus::INTERNAL_ERROR,
+                            "failed to clone item for overwrite split");
```
darbyjohnston, line 192: I was curious where the `overwrite split` came from, it doesn't look like it's mentioned in the `overwrite` function. Maybe it came from the unit test comment which is kind of cool:
```
// overwrite() splits a single clip whose middle is overwritten.
```

tests/test_editAlgorithm.cpp
```diff
@@ -2131,6 +2131,126 @@ main(int argc, char** argv)
               TimeRange(RationalTime(0.0, 24.0), RationalTime(10.0, 24.0)) });
     });
 
+    // Regression Test: Null Pointer Dereference from Unchecked dynamic_cast
```
darbyjohnston, line 2134: This seems like a lot of code to add for some nullptr checks? But just my opinion, curious what you think?

Also if metadata cycles are an issue it seems like we should add specific tests for that, instead of testing it through the edit algorithms.

jminor: This metadata cycle was the only way I found to actually cause the nullptr scenario. There might be others that I don't know about.

darbyjohnston: Ah, in that case do we need the tests? It seems like a roundabout way to exercise what is a relatively simple fix. If we want to test when `clone` fails for code coverage it seems like some sort of function mocking might be more appropriate.

jminor:
Closing in favor of #2020

PR #2018, opened 2026-04-28
Clang format fix

jminor:
The Makefile `format` rule had some mistakes that prevented `make format` from applying the clang-format rules to all of the source files. This PR fixes the Makefile and applies the current clang-format rules to all the `.cpp` and `.h` source files that were previously missed.

PR #2019, opened 2026-04-28
More clang format

jminor:
This is a follow up to #2018 which applies clang-format to even more cpp files, specifically the code in `tests/` and `examples/` folders.

I split this into separate commits for easier review.

Sorry I missed these in the previous PR.


darbyjohnston (approved):
The Makefile changes look fine.

PR #2020, opened 2026-05-01
Fix null pointer dereferences after dynamic_cast (rewritten)

jminor:
Certain invalid OTIO object graphs may cause crashes due to code that assumes a dynamic_cast always returns a valid object. This PR introduces tests which demonstrate the issue and a fix to prevent the crash in those cases.

I made sure to write the tests first, ensure that they caused the crash (SEGFAULT) and then made the fix to each of the affected functions. I introduced a new ErrorStatus::CANNOT_CLONE_ITEM for this specific case. The code may also return ErrorStatus::TYPE_MISMATCH in a related edge case, but I was not able to create a scenario which triggers that case.

Note that the *discovery* of this issue was assisted by Claude Code Opus 4.6 and Opus 4.7, but all of the code in this PR was written by a human (me). This PR replaces https://github.com/AcademySoftwareFoundation/OpenTimelineIO/pull/2017 which had LLM-generated code in it.

### DIFF
--- a/src/opentimelineio/algo/editAlgorithm.cpp
+++ b/src/opentimelineio/algo/editAlgorithm.cpp
@@ -362,7 +362,24 @@ insert(
         // Clone the item for the second partially overwritten item.
         if (!isEqual(second_source_range.duration().value(), 0.0))
         {
-            auto second_item = dynamic_cast<Item*>(item->clone());
+            auto cloned_item = item->clone();
+            if (!cloned_item)
+            {
+                if (error_status)
+                    *error_status = ErrorStatus(
+                        ErrorStatus::INTERNAL_ERROR,
+                        "clone failed");
+                return;
+            }
+            auto second_item = dynamic_cast<Item*>(cloned_item);
+            if (!second_item)
+            {
+                if (error_status)
+                    *error_status = ErrorStatus(
+                        ErrorStatus::INTERNAL_ERROR,
+                        "clone was unexpected type");
+                return;
+            }
             second_item->set_source_range(second_source_range);
             composition->insert_child(insert_index + 1, second_item);
         }
--- a/src/opentimelineio/algo/editAlgorithm.cpp
+++ b/src/opentimelineio/algo/editAlgorithm.cpp
@@ -181,9 +181,26 @@ overwrite(
             composition->insert_child(insert_index, item);
             if (!isEqual(second_duration.value(), 0.0))
             {
-                auto second_item = dynamic_cast<Item*>(items.front()->clone());
-                trimmed_range    = second_item->trimmed_range();
-                source_range     = TimeRange(
+                auto cloned_item = items.front()->clone();
+                if (!cloned_item)
+                {
+                    if (error_status)
+                        *error_status = ErrorStatus(
+                            ErrorStatus::INTERNAL_ERROR,
+                            "clone failed");
+                    return;
+                }
+                auto second_item = dynamic_cast<Item*>(cloned_item);
+                if (!second_item)
+                {
+                    if (error_status)
+                        *error_status = ErrorStatus(
+                            ErrorStatus::INTERNAL_ERROR,
+                            "clone was unexpected type");
+                    return;
+                }
+                trimmed_range = second_item->trimmed_range();
+                source_range  = TimeRange(
                     trimmed_range.start_time() + first_duration
                         + range.duration(),
                     second_duration);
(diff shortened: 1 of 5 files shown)


meshula (approved):
This looks more rigorous than the initial fix.

src/opentimelineio/algo/editAlgorithm.cpp
```diff
@@ -181,9 +181,22 @@ overwrite(
...
             if (!isEqual(second_duration.value(), 0.0))
             {
-                auto second_item = dynamic_cast<Item*>(items.front()->clone());
-                trimmed_range    = second_item->trimmed_range();
-                source_range     = TimeRange(
+                auto cloned_item = items.front()->clone();
+                if (!cloned_item)
+                {
+                    if (error_status)
+                        *error_status = ErrorStatus::CANNOT_CLONE_ITEM;
```
darbyjohnston: I think this error status should be set in `clone()`, so it is set only once and not at each call site. Something like:
```
auto cloned_item = items.front()->clone(error_status);
if (!cloned_item || (error_status && is_error(error_status))
    return;
```
```
SerializableObject*
SerializableObject::clone(ErrorStatus* error_status) const
{
    ...
    SerializableObject*r = e._root.type() == typeid(SerializableObject::Retainer<>)
               ? std::any_cast<SerializableObject::Retainer<>&>(e._root)
                     .take_value()
               : nullptr;
    if (!r && error_status)
        *error_status = ErrorStatus::CANNOT_CLONE_ITEM;
    return r;
}
```

darbyjohnston (changes requested):
Thanks, I think this is clearer about what is being tested than the previous PR. When we are happy with this one, try asking Claude to compare the two and see what it thinks about the changes.

src/opentimelineio/algo/editAlgorithm.cpp
jminor: Good call. I overlooked the fact that clone() accepts an error_status. This is fixed now. As a bonus, I didn't need to invent a new CANNOT_CLONE_ITEM error status, so I took that out too.

tests/test_editAlgorithm.cpp
```diff
@@ -2131,6 +2131,125 @@ main(int argc, char** argv)
...
     });
 
+    tests.add_test("regression: crash in slice", [] {
+        SerializableObject::Retainer<Clip> big_clip = new Clip(
+            "big clip",
+            nullptr,
+            TimeRange(RationalTime(0.0, 24.0), RationalTime(24.0, 24.0)));
+        big_clip->metadata()["cycle"] = big_clip;
+
+        SerializableObject::Retainer<Clip> small_clip = new Clip(
```
darbyjohnston: It doesn't look like `small_clip` is being used in this test?

tests/test_editAlgorithm.cpp
```diff
@@ -2131,6 +2131,125 @@ main(int argc, char** argv)
...
+        small_clip->metadata()["cycle"] = small_clip;
+
+        SerializableObject::Retainer<Gap> gap = new Gap(
+            TimeRange(RationalTime(0.0, 24.0), RationalTime(20.0, 24.0)),
+            "gap");
+
+        SerializableObject::Retainer<Track> track = new Track();
+        track->append_child(small_clip);
+        track->append_child(gap);
+        track->append_child(small_clip);
```
darbyjohnston: Is adding the `small_clip` twice intentional?

tests/test_editAlgorithm.cpp
```diff
@@ -2131,6 +2131,125 @@ main(int argc, char** argv)
               TimeRange(RationalTime(0.0, 24.0), RationalTime(10.0, 24.0)) });
     });
 
+    tests.add_test("regression: crash in slice", [] {
```
darbyjohnston: Not a big deal, but since the crash is fixed maybe the test name should change?

jminor: Rephrased to say "fails gracefully" instead.

tests/test_editAlgorithm.cpp
jminor: Removed.

tests/test_editAlgorithm.cpp
jminor: Clarified.

darbyjohnston (approved):

PR #2021, opened 2026-05-08
Add core C++ support for otioz and otiod, take 2

### LINKED ISSUE #1869: Add core C++ support for otioz and otiod

darbyjohnston:
Fixes #1869

This PR adds support for otioz and otiod bundles to the C++ API. This PR is a second attempt at adding the functionality, superseding PR #1901.

The python otioz and otioz adapters still exist and work the same, but now they call into the C++ code for the bundle functionality.

Support for image sequences and multiple media references has been added.

Other changes:

* Added the library "minizip-ng" for working with ZIP files. This also relies on either "zlib-ng" or "zlib" for compression. This seemed like a reasonable choice since OCIO also uses it.
* Added a C++ example.
* Replaced most of the Python bundle tests with C++ tests. Minimal Python tests remain to verify the adapters still work.

Questions:
* Should the otioz/otiod file versions be bumped?
* Should we use a third party library for URL handling? We currently only need a single function, to return a filename from a URL.

Follow up work for other PRs:
* Add a flag to otiotool for specifying the relative path to media
* Consider using  ISO/IEC 21320-1 for the otioz bundle format (https://en.wikipedia.org/wiki/ZIP_(file_format))

Assisted-by: Claude:claude-opus-4-7 [code-review] [debugging]


src/opentimelineio/bundle.h
```diff
@@ -0,0 +1,106 @@
...
+    /// @brief Get a file from a URL.
+    std::optional<std::string> file_from_url(std::string const& url);
+
+    /// @brief Options for writing bundles.
+    struct OTIO_API_TYPE WriteOptions
+    {
+        /// @brief Base directory for resolving relative media reference paths.
+        /// If a media reference URL resolves to a relative path, it is resolved
+        /// against this directory before being added to the bundle.
+        std::optional<std::string> relative_media_path;
```
ssteinbach: Maybe `relative_media_path_base_dir` would be a better name (based on the comment)?

src/opentimelineio/bundle.h
```diff
@@ -0,0 +1,106 @@
...
+        
+        /// @brief Convert the media reference paths to absolute paths.
+        ///
+        /// If this is set to true for otioz files, an extract_path must also be set.
+        bool absolute_media_reference_paths = false;
+    };
+
+    /// @brief Calculate the total uncompressed size of the files that would be
+    /// written to a bundle, without actually writing it. This is useful for
+    /// estimating the disk space required.
```
ssteinbach: I'm assuming this also does a validity check per the error policy?

```suggestion
    /// @brief Check the timeline against the error policy to see if a bundle 
    /// can be made correctly.  If so, return the total uncompressed size of 
    /// the files that would be written to a bundle, without actually writing it. 
    /// This is useful for estimating the disk space required.
```

src/opentimelineio/bundle.h
```diff
@@ -0,0 +1,106 @@
...
+        /// @brief Number of spaces for JSON indentation.
+        int indent = 4;
+    };
+
+    /// @brief Options for reading bundles.
+    struct OTIO_API_TYPE ReadOptions
+    {
+        /// @brief Extract the contents of the otioz bundle to this directory,
+        /// which must not already exist.
+        std::optional<std::string> extract_path;
```
ssteinbach, line 63: I see, if this is false, then the only thing read from the bundle is the timeline itself, correct?  Maybe we should reflect that behavior in the comment there too.

src/opentimelineio/bundle.h
```diff
@@ -0,0 +1,106 @@
...
+        ErrorStatus*       error_status = nullptr);
+
+    /// @brief Write a timeline and it's referenced media to an otiod bundle.
+    OTIO_API bool write_otiod(
+        Timeline const*     timeline,
+        std::string const&  path,
+        WriteOptions const& options      = WriteOptions(),
+        ErrorStatus*        error_status = nullptr);
+
+    /// @brief Read a timeline from an otiod bundle.
```
ssteinbach, line 103: What do you think about also noting here that whether or not the media is extracted depends on the `ReadOptions`?

ssteinbach (commented):
Thanks for doing this work!  Some questions and notes about the interface.   As far as I can tell, this is 100% compatible with the way bundles work right now, right?  Or is it the new library that would warrant the version bump?

src/opentimelineio/bundle.h
darbyjohnston: Sure, that sounds good.

darbyjohnston:
> As far as I can tell, this is 100% compatible with the way bundles work right now, right? Or is it the new library that would warrant the version bump?

The file format is the same, I was just thinking about the library change. Like if we wanted to identify if a particular bundle was written with the original code vs. the new code. On second thought that probably shouldn't change the spec version, if it was useful we could add it as metadata to the timeline.

src/opentimelineio/bundle.h
darbyjohnston: Oh interesting, I thought the dry run was just for computing the size, I hadn't thought about using it to check for errors. The code is mostly the same so I think it will work, should I add some tests for it?

darbyjohnston: I added dry runs for the C++ tests that check the media policy.

src/opentimelineio/bundle.h
darbyjohnston: I renamed it `relative_media_base_dir` to keep it a bit shorter.

src/opentimelineio/bundle.cpp
```diff
@@ -0,0 +1,890 @@
...
+                resolved = relative_media_base_dir / resolved;
+            auto const bundle_path = (std::filesystem::u8path(media_dir) /
+                source_path.filename()).u8string();
+            out.insert({ resolved.u8string(), bundle_path });
+            return true;
+        }
+
+        // Process all media references according to the policy. The list of
+        // files to be added to the bundle is returned.
+        bool process_media_references(
```
meshula, line 142: I think it would be nice if this routine returned something more substantial for error reporting; if you had the stringstream facility you use elsewhere, you could explain which files violated which policies why. That wouldn't contribute to correct functionality but might be very valuable to users.

tests/utils.cpp
```diff
@@ -18,6 +32,46 @@ assertFalse(bool value)
...
+
+TempDir::TempDir()
+{
+    // Create a unique name from a GUID.
+    GUID guid;
+    CoCreateGuid(&guid);
+    const uint8_t* guidP = reinterpret_cast<const uint8_t*>(&guid);
+    std::stringstream ss;
+    ss << std::hex << std::setfill('0');
+    for (int i = 0; i < 16; ++i)
```
meshula: I don't think speed matters in this routine, but this is measurably the slowest possible way to make a string, hex converted or not, using std ;)

tests/utils.cpp
```diff
@@ -18,6 +32,46 @@ assertFalse(bool value)
...
+    if (!std::filesystem::create_directory(path, ec))
+        throw std::runtime_error("cannot create temp directory");
+    _path = path;
+}
+
+#else // _WIN32
+
+TempDir::TempDir()
+{
+    auto const path = (std::filesystem::temp_directory_path() / "XXXXXX").u8string();
```
meshula: I'm not sure why windows gets a UUID and other platforms are X-Rated. Maybe just make a common unique identifier routine, perhaps building from pid. I don't think you need to be as safe as a UUID nor as dangerous as tripleX times 2.

meshula (commented):
overall looks really good. general notes

- maybe clean up TempDir as noted?
- I think you could wire through much more descriptive error reporting for the CLI (or please correct me if I'm not spotting a verbose forensics path)
- please beware of string construction stringstream, as I noted it's not especially problematic (aside from performance) where you are using it, but especially for string construction it is loaded with footguns due to locale, and you haven't enforced the C locale which at least makes its behavior reproducible from computer to computer.

Please consider using zstd; to my knowledge it's bullet proof vs common attacks, offers excellent compression at I believe best speeds, and is also a very lightweight library that is easy to vendor. https://github.com/facebook/zstd

These are minor compared to the overall solidity of the work.

src/opentimelineio/bundle.cpp
darbyjohnston: In addition to the bool return value an `error_status` is also passed in for details. For example this is one of the possible errors (bundle.cpp:104):
```
if (error_status) {
    std::stringstream ss;
    ss << "media file " << source_path << " would overwrite " << i->second;
    *error_status = ErrorStatus(ErrorStatus::FILE_WRITE_FAILED, ss.str());
}
```
Are there other errors you were thinking of?

tests/utils.cpp
darbyjohnston: It's probably pretty fast compared to creating the 4GB test files. :)

tests/utils.cpp
darbyjohnston: AFAIK there is no `mkdtemp` on Windows. After some searching it looked like the `guid` method was one of the suggested alternatives or `GetTempFileNameA`.  It would be nice if there was some sort of utility library all the ASWF projects could use for this stuff.

darbyjohnston:
> https://github.com/facebook/zstd

That's a different compression scheme from deflate? We could possibly add that as a separate feature at some point.

Somewhat related, what do you think about having otioz files conform to ISO/IEC 21320-1?  (https://en.wikipedia.org/wiki/ZIP_(file_format)) As I understand it, it's a minimal ZIP format  for the same use cases as otioz.

What do you suggest instead of `stringstream`? Aren't we constrained by our C++ version from using `std::format`?

src/opentimelineio/bundle.cpp
meshula: well, specifically I thought all the return false conditional might deserve similar. e.g.

                                   if (!register_bundle_file(
                                        path,
                                        relative_media_base_dir,
                                        paths,
                                        out,
                                        error_status))
                                        return false;
                                }

unless that detailed reporting into error_status occurs inside register_bundle_file and I simply didn't piece it together in my head that there is detailed reporting handed off via the called function?

But I thought all of them could use a look. e.g.

                                if (error_status)
                                    *error_status = ErrorStatus(
                                        ErrorStatus::FILE_WRITE_FAILED,
                                        "media reference is not a file");
                                return false;

If you reported the text of the media reference it would be helpful to someone studying a log trying to understand what went wrong.

meshula: > the text of the media reference
The text of the media reference, or where it was encountered, basically "which breadcrumbs would help someone understand how to address the issue"

tests/utils.cpp
meshula: haha OMG TIL that mkdtemp says "put at least six X characters here".  Insert "crazy face emoji". Could you add a little comment please? I didn't know this and thought it was simply peculiar and possibly left over temp code.

meshula: also, it's still true that I am allergic to constructing strings with stringstream for many reasons, so although I don't specifically object to the pattern above, it is vaguely disconcerting for the reasons of performance and std::locale both being latent footguns. I don't disagree with your notion of a canonical UUID routine, but again point out that "Universally" unique IDs mean no collisions in the universe and UUIDs are therefore a wee bit expensive especially for a temp file name construction!

constructive alternatives

1) use GetTempFileNameW() on windows mkdtemp on posix.

2) 

```cpp
// Get the system's legitimate temp directory
    auto temp_dir = fs::temp_directory_path();
//  Use a fast, thread-safe pseudo-random generator for a local ID
// A high-entropy 64-bit hex string is plenty. mt19937 is terrible for games and most random needs tbh, but fine here.
    static thread_local std::mt19937_64 generator(std::random_device{}());
    uint64_t rand_num = generator();
// fast locale-independent string construction with hex formatting
    std::string filename = std::format("tmp_{:x}.tmp", rand_num);
```

src/opentimelineio/bundle.cpp
darbyjohnston: I added the name of the media reference to the error message.

I also added some code I missed for copying metadata to the missing references, so you can set the "media policy" to "missing_if_not_file", run the "dry_run" function, and see which media references were converted to missing.

darbyjohnston: I made the existing error messages more verbose, including the relevant filenames in the text. I also replaced the use of std::stringstream with string concatenation.

tests/utils.cpp
darbyjohnston: #2 seems like a better alternative since it's OS independent. I updated the PR with Claude's take on it, check it out and see what you think.

meshula:
afaik USD itself uses ISO/IEC 21320-1? so that makes sense to me. 

What do you suggest instead of stringstream? Aren't we constrained by our C++ version from using std::format?

Allocate space and fill it in trivially C style. C still works :)

meshula (approved):
Thanks for addressing the feedback!

jminor (approved):

PR #2023, opened 2026-05-18
Marker RGB Colors

### LINKED ISSUE #1873: Marker RGB colors
OTIO Markers have a "color" attribute which is a string taken from a list of color names ("RED", "GREEN", "YELLOW", etc.) However, some systems specify named colors that are not in this list, or allow arbitrary RGB values. In order to support a wider set of marker colors, we could switch OTIO to use an RGB triple like `(255, 127, 0)` or `(1.0, 0.5, 0.0)` and include helper functions to convert to and from lists of color names.

See prior work & discussion here: https://github.com/AcademySoftwareFoundation/OpenTimelineIO/issues/27

To limit the scope of this a bit, let us assert that markers never have a transparency/alpha, and that marker colors are not managed through colorspaces (sRGB, etc.). If anyone can show a counter example in an existing application, format, or workflow, please post it here.

semagnum:
Fixes #1873

**Summarize your change.**

Part of ASWF Dev Days 2026. Removes redundant `Marker::Color` enum and replaces it with the standard `Color` class recently added for metadata.

- replace `Marker::Color` and its `std::string _color` attribute with an optional `Color` equivalent (defaults kept the same)
- updates `Marker` to schema 3
    - adds upgrade function: if the color name matches any of the previous `Marker::Color` enum names, it converts it to the `Color` equivalent. Otherwise, it preserves the original name
    - adds a downgrade function: if the `Color` name matches any of the `Marker::Color` enum names, it renames it to the all-caps name. Otherwise, it preserves the original name
- maps the `MarkerColor` Python name to use `Color`. We can mark the name for deprecation in the future, but I figured this would be helpful for backwards compatibility for now
- related issue, found when updating tests: adds `!=` operator for `Color` class, binds the equality functionality into Python

- Marker-related baseline files updated
- updated `test_marker.py` tests, added a test for upgrading schema to 3
- updated `test_clip.cpp`, `test_clip.py`, `test_console.py` that references marker colors

### DIFF
--- a/docs/tutorials/otio-serialized-schema-only-fields.md
+++ b/docs/tutorials/otio-serialized-schema-only-fields.md
@@ -31,15 +31,6 @@ parameters:
 
 ## Module: opentimelineio.core
 
-### Color.1
-
-parameters:
-- *a*
-- *b*
-- *g*
-- *name*
-- *r*
-
 ### Composable.1
 
 parameters:
@@ -152,6 +143,15 @@ parameters:
 - *name*
 - *source_range*
 
+### Color.1
+
+parameters:
+- *a*
+- *b*
+- *g*
+- *name*
+- *r*
+
 ### Effect.1
 
 parameters:
@@ -224,7 +224,7 @@ parameters:
 - *name*
 - *time_scalar*
 
-### Marker.2
+### Marker.3
 
 parameters:
 - *color*
--- a/docs/tutorials/otio-serialized-schema.md
+++ b/docs/tutorials/otio-serialized-schema.md
@@ -55,27 +55,6 @@ parameters:
 
 ## Module: opentimelineio.core
 
-### Color.1
-
-*full module path*: `opentimelineio.core.Color`
-
-*documentation*:
-
-```
-:class:`Color` is a definition of red, green, blue, and alpha double floating point values, allowing
- conversion between different formats. To be considered interoperable, the sRGB transfer function 
-encoded values, ranging between zero and one, are expected to be accurate to within 1/255 of the 
-intended value. Round-trip conversions may not be guaranteed outside that. This Color class is meant
- for use in user interface elements, like marker or clip coloring, NOT for image pixel content.
-```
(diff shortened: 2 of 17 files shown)


timlehr:
Hey @semagnum, thank you, this is a great addition. The code looks good, but I am wondering how to best approach the release of this change, since some adapters are likely using the marker color and expecting it to be a string value. We might have to do pass over the common adapters, check where we need to do some changes.

@apetrynet can you assign the devdays26 badge on this?

reinecke:
Suggestion from @timlehr:

> When we update the adapters, they can look a the schema version on `Marker` and use that to determine whether they should use old style or new style color.

reinecke:
@semagnum Sorry to do this - would you mind doing a rebase on the current main for this? When that's done I'll be able to merge this in - sorry about that!

semagnum:
I had done the rebase, is this good to go?

timlehr:
@apetrynet @reinecke This should be good to merge.

apetrynet:
@timlehr Uhm.. Seems like the branch is out of date again(?) The merge button is grayed out.

apetrynet:
Hi @semagnum 
I've submitted a [PR](https://github.com/AcademySoftwareFoundation/OpenTimelineIO/pull/2027) to get the Read the Docs build to run.
You might need to do a final rebase after it lands before we land this one. Sorry about that.

apetrynet:
The RTD PR has landed.

apetrynet:
Hi, @semagnum 
I'm so sorry about this, but GitHub pulled the rug under us and upgraded the macos-latest runner ahead of time which broke the build. We've now pinned the runner to macos-15 which fixed it, but unfortunately that means yet another rebase for you.
There used to be a way for us to do a rebase on the GitHub page, but it seems to be gone or I don't have  the permissions.

Apologies again, for the bumpy landing.

apetrynet:
Thank you for your contribution @semagnum !

PR #2025, opened 2026-06-07
Fix edit algorithms deleting the wrong child via remove_child(<pointer>)

TroyHernandez:
Composition only declares `remove_child(int index, ErrorStatus*)`, but `overwrite()` and `insert()` in `editAlgorithm.cpp` call it with a `Composable*` / `Retainer`:

```cpp
composition->remove_child(transition);     // overwrite() and insert()
composition->remove_child(items.back());   // overwrite()
```

The pointer argument decays to `bool` (always `true`) → `int(1)`, so these calls remove the child at **index 1** (or, via `adjusted_vector_index`, the last child when `size() == 1`), **not** the item that was passed.

### Impact
- `overwrite()` over a multi-item span deletes the wrong clips: the partially-overwritten clip is removed while a fully-overwritten clip survives in place.
- Transition removal in `overwrite()`/`insert()` removes index 1 instead of the intended transition.

### Repro
Track `A|M|B` (each 5/8/5 frames), `overwrite(X, track, [8, 18))` — a span that partially covers `M` and fully covers `B` to the track end:

```
before (buggy):  A[0+5] B[20+5] X[2+10]   # M removed, B kept — wrong
after  (fixed):  A[0+5] M[10+3] X[2+10]   # M trimmed, B removed — correct
```

### Fix
Pass the already-computed (and bounds-checked) `index` for transitions, and `index_of_child(items.back())` in the removal loop.

### Tests
- Adds `test_edit_overwrite_partial_first_full_last`, which fails before this change and passes after.
- The existing `test_editAlgorithm` suite continues to pass.

I verified the fix and the regression test by building the library locally (the test fails on the unfixed tree, passes on the fixed tree).

### DIFF
--- a/src/opentimelineio/algo/editAlgorithm.cpp
+++ b/src/opentimelineio/algo/editAlgorithm.cpp
@@ -119,7 +119,7 @@ overwrite(
                         || static_cast<size_t>(index)
                                >= composition->children().size())
                         continue;
-                    composition->remove_child(transition);
+                    composition->remove_child(index);
                 }
             }
         }
@@ -254,7 +254,8 @@ overwrite(
             // Remove the completely overwritten items.
             while (!items.empty())
             {
-                composition->remove_child(items.back());
+                composition->remove_child(
+                    composition->index_of_child(items.back()));
                 items.pop_back();
             }
 
@@ -291,7 +292,7 @@ insert(
                     || static_cast<size_t>(index)
                            >= composition->children().size())
                     continue;
-                composition->remove_child(transition);
+                composition->remove_child(index);
             }
         }
     }
--- a/src/opentimelineio/serializableObject.h
+++ b/src/opentimelineio/serializableObject.h
@@ -626,7 +626,7 @@ public:
 
         T* operator->() const noexcept { return value; }
 
-        operator bool() const noexcept { return value != nullptr; }
+        explicit operator bool() const noexcept { return value != nullptr; }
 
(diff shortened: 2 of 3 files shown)


jminor:
Oh, good catch!

Would you mind adding `explicit` to the `operator bool()` on Retainer? I think that would have caused this error to be caught by the compiler & will help prevent folks using OTIO from falling into this same trap.

(cc @meshula to check this)

TroyHernandez:
Done in 6670f49 (also merged main to pick up #2020, which touches the same functions).

- The library, all C++ tests, and the Python bindings compile with zero call-site changes; full CI matrix is green.
- I verified the guard works: compiling the original bug pattern, `composition->remove_child(some_retainer)`, now fails with `invalid user-defined conversion from 'Retainer<Composable>' to 'int'`.
- For @meshula re: compatibility, the break surface is narrower than "explicit operator bool" usually implies. Since `operator T*()` is still implicit, boolean contexts (`if (r)`, `&&`, `!`) and plain `bool b = r;` all still compile via the pointer conversion. The only thing `explicit` forbids is using `bool` as a bridge to another type (`Retainer` -> `bool` -> `int`), which is exactly the trap this PR fixes. I'd expect zero impact on correct downstream code.

meshula (commented):
Thanks!

meshula (approved):
Thank yo for this!!

jminor (approved):

PR #2028, opened 2026-06-23
Pin CI to macos-15, since macos-latest is now using macOS 26

timlehr:
Github changed `macos-latest` runners to use macOS 26 now, which seems to not be incompatible with our current core library build matrix.

Reference: https://github.com/actions/runner-images/issues/14167


apetrynet (approved):
I agree on removing macos-14 in the same PR as it's EOL in a weeks time. Then we won't run into this same issue at that time
LGTM

PR #2030, opened 2026-07-02
Add cut_diff example: structural editorial diff between two cuts

chaoz23:
This adds an `examples/cut_diff.py` script that answers *"what changed between cut A and cut B"* — which clips were **added, removed, retimed, moved, or shifted** — in the spirit of existing examples like `conform.py` and `shot_detect.py`. The OTIO docs list "Shots Added or Removed From The Cut" as a first-class use case; this example implements it.

Because both inputs go through adapters, the cuts don't need to share a format — an EDL can be diffed against an FCP XML export of the revision:

```
$ python cut_diff.py editors_cut.edl finishing.fcpxml
1 retimed, 1 shifted (7 unchanged)
  ~ ZZ100_503A shortened by 12f (28f -> 16f)
  . ZZ100_504C shifted 12f earlier
```

Design notes:
- Clips are **matched by identity** (media `target_url` + source in-point), not timeline position — one insertion shifts every downstream timecode, so positional diffing reports everything as changed.
- "Moved" is detected by **relative order** (longest increasing subsequence over matched clips), so a gap inserted by a lift-trim doesn't read as every downstream clip moving.
- Ripple slides report as "shifted", separate from "retimed", so one trim doesn't read as N retimes.
- Scope is structural editorial only (clips/timing/order); effects and transitions are intentionally out, per their tool-specific serialization.
- `--json` output and `diff(1)` exit codes for scripting.

Validated against sample files from the cmx3600 and fcpx-xml adapter test suites, including hand-modified known-change cases. A packaged version with a test suite and an MCP server lives at [chaoz23/otio-diff](https://github.com/chaoz23/otio-diff); this example is self-contained.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/examples/cut_diff.py
+++ b/examples/cut_diff.py
@@ -1,0 +1,330 @@
+#!/usr/bin/env python
+#
+# SPDX-License-Identifier: Apache-2.0
+# Copyright Contributors to the OpenTimelineIO project
+
+"""Example OTIO script that diffs two cuts of a timeline structurally:
+which clips were added, removed, retimed, moved, or shifted between
+cut A and cut B.
+
+Because both inputs are parsed through OTIO adapters, the two cuts do not
+need to be the same format — e.g. an EDL can be diffed against an FCP XML
+export of a revised cut.
+
+Approach: a cut is not positionally diffable (inserting one clip shifts
+every downstream timecode), so clips are MATCHED BY IDENTITY first —
+``(media target_url, source in-point)`` — then classified:
+
+  added     in B, not in A
+  removed   in A, not in B
+  retimed   same clip, trimmed duration changed
+  moved     relative order among matched clips changed
+  shifted   only slid on the timeline (the ripple of an upstream edit)
+
+Duration is compared as an attribute after matching, so an out-point trim
+reads as "retimed" rather than removed+added. "moved" uses relative order
+(longest increasing subsequence over the matched clips) rather than absolute
+index, so a gap inserted by a lift-trim or a deleted event does not read as
+every downstream clip having moved. Duplicate identities (same source used
+twice) are paired as a multiset.
+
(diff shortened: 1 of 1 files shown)


jminor:
See also: https://github.com/AcademySoftwareFoundation/OpenTimelineIO/pull/1922

chaoz23:
Thanks — hadn't seen #1922. That's the ambitious integrated-tool path: an `otiotool` feature with a visual annotated-OTIO output, aimed at #26. From the outside it looks stalled, though — still a draft, no commits since Aug 2025, and currently conflicting with `main`.

This PR is deliberately narrower and at a different layer: a self-contained `examples/cut_diff.py` alongside `conform.py` and `shot_detect.py`. It touches no core code, CI is green, and it can land independently today. If #1922 revives later as the full integrated tool, this example points readers straight to it — they complement rather than compete. I've added a header note in the example cross-referencing `otiotool`/#1922 and #26.

Proposal: **merge this as an example**. If you'd rather hold it, I'll close it — no debate needed.

chaoz23:
Hi maintainers — following up on this example PR (cut_diff: structural editorial diff between two cuts). All checks are passing (build, docs, DCO, EasyCLA). Happy to make any adjustments if you have feedback. Thanks for taking a look!

jminor:
Hi @chaoz23 - the OpenTimelineIO Technical Steering Committee is actively debating this project's policy for LLM-assisted code contributions. You'll have to wait until we finalize that policy before moving forward with this PR. Thanks for your patience.

chaoz23:
Understood, and thanks for the heads-up, I'll wait for the TSC's policy and won't push in the meantime. Whenever it lands, happy to adapt this PR to whatever disclosure/provenance requirements you settle on (the AI-assistance note in the description was intentional on my part, in that spirit). Appreciate the transparency about the process.

Issue #1985, opened 2025-12-11
Clarity requested on Media Reference's target_url format


hww-clementpoulain:
Hi there,

I was wondering if any of those 2 pages were slightly out of date regarding how the Media Refence target_url should be formatted by write adapters.
Of if read adapters should be handling both cases?

The read the docs page [here](https://opentimelineio.readthedocs.io/en/latest/tutorials/write-an-adapter.html#media-references) mentions it should have a `file://` prefix.
Where the unit tests [here](https://github.com/AcademySoftwareFoundation/OpenTimelineIO/blob/2429dd5cd35a16949b7b675c24aad61e123f064d/tests/test_media_reference.py#L47) mention it should be raw.

The confusion emerged from Adobe Premiere Beta 26.1 exporting OTIO files with media references with said `file://` prefix.
And DaVinci Resolve 20.2 failing to import OTIO files formatted that way. (manually removing those prefixes let DaVinci Resolve import them just fine)
Which has been an issue reported a year ago in the Adobe community [here](https://community.adobe.com/t5/premiere-pro-beta-discussions/now-in-beta-otio-import-and-export/m-p/14938190#M10400) , which can still be replicated today with above - recent - versions.

Best,
C

reinecke:
Hi @hww-clementpoulain -

I'm going to preface this with one thing: My memory of how this all applies to Windows-style paths is weak, so most of what I say here is for POSIX style paths. There are some people in the community that may be able to provide more detail on the Windows path front if you're interested.

In OTIO we specify that URLs should be used - but these can be absolute or relative and technically be of various schemes (`file://`, `https://`, `s3://`, etc.).

URLs like `/Volumes/My_Working_HD/media/sc1_take1.mov` (forgive the bad faux filename) are still URLs, but relative anchored at the root. One way to think about this is with HTML links - many don't point to `https://` full URLs - this allows them to behave the same when loading from a local filesystem or a remote URL. If you have an HTML file at `https://mywebsite.com/about/index.html` with a link like `<a href="/index.html">Home</a>` - this is a relative URL because the system infers the `https://mywebsite.com` part of it.

[...]

PR #2033, opened 2026-07-13
Use url_utils.url_from_filepath() for file:// URLs in docs/examples

anAirdrop:
## Summary
Fixes the `target_url` examples in `docs/tutorials/write-an-adapter.md` and two example scripts (`examples/conform.py`, `examples/shot_detect.py`) that built `file://` URLs by string-concatenating `"file://" + path`. This only produces a valid URI for POSIX-absolute paths; for relative or Windows drive-letter paths, the first path segment is silently parsed as the URL's host (RFC 3986), producing a corrupted URL.

`opentimelineio.url_utils.url_from_filepath()` already handles this correctly (see `tests/test_url_conversions.py`), this PR just points the docs and examples at it instead of hand-rolling the URL.

Related to #1985. This is the narrow docs/examples fix mentioned in that thread, scoped separately from the larger Application Integrator's Guide rewrite being discussed there.

## Test plan
- `url_utils.py` itself is unchanged and already covered by `tests/test_url_conversions.py`.
- `examples/conform.py` and `examples/shot_detect.py` have no existing dedicated test coverage (they are standalone CLI examples, not part of the installed package), did not want to expand the scope of this PR to add new test infrastructure for the examples directory.
- Verified `python -m py_compile` on both edited example scripts.

### DIFF
--- a/docs/tutorials/write-an-adapter.md
+++ b/docs/tutorials/write-an-adapter.md
@@ -165,9 +165,18 @@ Note that all metadata should be nested inside a sub-dictionary (in this example
 Clip media (if known) should be linked like this:
 ```python
 clip.media_reference = otio.schema.ExternalReference(
-    target_url="file://example/movie.mov"
+    target_url=otio.url_utils.url_from_filepath("/show/seq/shot/movie.mov")
 )
 ```
+Prefer `otio.url_utils.url_from_filepath()` over building the `file://` URL by
+hand (e.g. `"file://" + path`). A hand-built URL is easy to get subtly wrong —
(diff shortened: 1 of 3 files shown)


reinecke:
@anAirdrop Thanks for the fixes, these look great! Would you mind just addressing the [DCO signoff issue](https://github.com/AcademySoftwareFoundation/OpenTimelineIO/pull/2033/checks?check_run_id=87535361294#check_run_87535361294) and then I think we should be able to merge!

anAirdrop:
done!  let me know if I can do anything else.

PR #2034, opened 2026-07-16
Restore Python url_utils module

darbyjohnston:
PR #2021 removed the `url_utils` package as part of transitioning .otioz support to C++. However `url_utils` was part of the public Python API and may be used by other code (e.g., #1664). This PR restores the module and also replaces Python code with C++ functionality where applicable.

Note that the `file_bundle_utils.py` was also removed as part of the C++ support. This seems unlikely it was ever used outside of the otioz/otiod adapters, so I did not restore it.

Assisted-by: Claude:claude-opus-4-7 [code-review] [debugging]


timlehr:
@reinecke @jminor can we approve / merge this? This is a blocker for AAF adapter tests too.

reinecke (approved):

PR #2039, opened 2026-09-11
Fixes for API exports

darbyjohnston:
This PR has various fixes for the API exports:
* Change OTIO_EXPORTS and OPENTIME_EXPORTS to PRIVATE
* Add missing exports
* Remove the exports from templates which are inlined
* Add missing includes

Assisted-by: Claude:claude-opus-5 [debugging]

PR #2041, opened 2026-09-11
Fix macOS Python 3.14 std::bad_any_cast by setting hidden visibility preset

### LINKED ISSUE #2035: Python 3.14 Throws a Runtime exception
As reported in [this thread](https://academysoftwarefdn.slack.com/archives/CMQ9J4BQC/p1787048598732269), OTIO in python 3.14 on macos arm64 throws a runtime error loading `.otio` files:

```python
File "/Users/al/Library/Python/3.14/lib/python/site-packages/opentimelineio/adapters/__init__.py", line 94, in from_filepath
   return plugins.ActiveManifest().from_filepath(outext)
       ~~~~~~~~~~~~~~~~~~~~~~^^
  File "/Users/al/Library/Python/3.14/lib/python/site-packages/opentimelineio/plugins/manifest.py", line 332, in ActiveManifest
   _MANIFEST = load_manifest()
  File "/Users/al/Library/Python/3.14/lib/python/site-packages/opentimelineio/plugins/manifest.py", line 317, in load_manifest
   plugin_manifest = manifest_from_file(builtin_manifest_path)
  File "/Users/al/Library/Python/3.14/lib/python/site-packages/opentimelineio/plugins/manifest.py", line 49, in manifest_from_file
   result = core.deserialize_json_from_file(filepath)
RuntimeError: bad any cast
```

It is reported the same thing works fine on linux.

This should be addressed and we should enable CI for python 3.14.

mcoliver:
Fixes #2035.

When building the Python bindings (where the core libraries are linked statically into `_otio.so`), pybind11 aggressively forces `-fvisibility=hidden`. Because the core library targets lacked visibility configuration, they were compiled with `-fvisibility=default`.

On macOS, this mismatch generates duplicate `type_info` structures for template types like `Retainer<>`. Since macOS `libc++` evaluates `type_info::operator==` via strict pointer equality, the dispatch table lookup inside `otio_utils.cpp` throws a `std::bad_any_cast` when crossing the boundary from `pybind11` to the core library.  

**_Note the reason this probably doesn't show up on Linux is because Linux's libstdc++ compares by name as a fallback_**

By setting `CMAKE_CXX_VISIBILITY_PRESET hidden` and `CMAKE_VISIBILITY_INLINES_HIDDEN ON` globally, the core library compiles with the exact same visibility rules as `pybind11`. The static linker then correctly collapses all `type_info` weak symbols into a single memory address, resolving the runtime exception on Python 3.14 macOS.


jminor (approved):

darbyjohnston:
@jminor This change breaks shared builds due to some missing exports. The symbols with missing exports get hidden by this which then cause linker errors. We should merge #2039 to fix the missing exports which might also make this change unnecessary. I also need to add a CI job that builds shared libraries to catch things like this.

mcoliver:
@darbyjohnston is correct.  Apologies for not thinking deeper on the core issue.  I was trusting the CI too much.  I do think keeping the hidden visibility preset is good to have as a hardening measure in the event things are missed down the road but hopefully shared build CI should flag that.

darbyjohnston:
@mcoliver No problem, the exports are tricky to get right. I merged #2039 that should hopefully fix the missing ones. I also created a PR to add shared library builds to the CI to help catch these issues (#2057).

PR #2046, opened 2026-09-12
Add AI Policy

reinecke:
**Summarize your change.**

This PR adds an AI coding policy to provide clarity for contributors using AI coding assistants.

This policy is forked from the [thoughtful policy](https://github.com/AcademySoftwareFoundation/OpenImageIO/blob/main/docs/dev/AI_Policy.md) provided by OpenImageIO.

PR #2048, opened 2026-09-13
Fix trimmed ranges of deeply nested children

### LINKED ISSUE #2043: trimmed_range_of_child broken for ≥3-level nesting
An LLM-based code review claims that there is a bug here:
> composition.cpp:397-400: missing current = parent; at the end of the loop (compare range_of_child at line 349). Stale current makes index_of_child fail with NOT_A_CHILD_OF for any timeline→track→stack→clip hierarchy. Poisons trimmed_range_in_parent, children_in_range, algo::fill.

Could we get a volunteer to verify if this is a real bug? It sounds plausible, and could be easily verified by writing a test case and/or a test OTIO file that creates the scenario described.

jhuanxx44:

Fixes #2043

**Summarize your change.**

`Composition::trimmed_range_of_child` fails with `NOT_A_CHILD_OF` (exposed as `NotAChildError` in Python) when it traverses three or more parent compositions. The first iteration advances `current`, but subsequent iterations leave it pointing to an earlier descendant, so the next parent cannot find it among its direct children.

Advance `current` after accumulating each parent range, matching the traversal in `range_of_child`.

- C++ regression: a `Stack → Track → Stack → Clip` hierarchy with a nonzero track offset; checks both the error status and the returned range.
- Python regression: one through four parent compositions, accumulated track offsets, and an outer source range that either clips the result or excludes the clip entirely.
- Both regressions fail before the production change and pass after it.

Validated locally on macOS arm64 with AppleClang 21 and Python 3.14.6:

- C++ static Debug build: all 10 CTest targets pass.
- `make test`: 482 tests run, 2 existing skips.
- `make lint`, `make manifest`, `clang-format --dry-run --Werror` on the changed C++ files, and `git diff --check` pass.

Assisted-by: Codex / GPT-6


jminor:
Hi @jhuanxx44 - thanks, this PR is looking pretty good. We'll arrange for an OTIO committer to review it. Out of curiosity, how did you come across this issue, and what lead you to contribute to OTIO?

jhuanxx44:
Hi @jminor, thanks for taking a look!

I’m exploring a video editing project of my own and will need a timeline model, which is how I came across OTIO. Around the same time, I saw #2043, where you were looking for someone to verify a bug identified by an LLM-based review, and it seemed like something I could help with.

To be upfront, I used Codex to help investigate and implement the change, hence the `Assisted-by:` trailer in the PR. I reviewed the diff myself and ran the full local test and lint suite; the new tests failed before the fix and passed afterward.

I’m happy to answer any questions from the committer and follow whatever conventions OTIO prefers for AI-assisted contributions.

jminor (approved):

jminor:
We're actually right in the midst of establishing our AI policy, and this PR matches very closely to what we're aiming for. Your timing is perfect. See details here: https://github.com/AcademySoftwareFoundation/OpenTimelineIO/pull/2046

Issue #2042, opened 2026-09-11
Some documentation is missing


jminor:
The OTIO documentation here https://opentimelineio.readthedocs.io/ is incomplete. There are some classes, functions, and concepts that are missing documentation entirely, or need additional information. We would love for a volunteer to read/skim through the documentation as-is and make a checklist of areas that are missing.

Examples:
- Item.1 is missing a doc string: https://opentimelineio.readthedocs.io/en/stable/tutorials/otio-serialized-schema.html#item-1
- MediaReference.1 is missing a doc string: https://opentimelineio.readthedocs.io/en/stable/tutorials/otio-serialized-schema.html#mediareference-1
- RationalTime.1 has docs, but they are too short and there's a math formatting problem: https://opentimelineio.readthedocs.io/en/stable/tutorials/otio-serialized-schema.html#rationaltime-1
- etc.

If we had a detailed checklist, then we could more easily ask people to fill in those gaps & track our progress on improving the documentation.

Also: fixes/additions to the documentation content itself are quite welcome.

ruthraguru248:
Hi! I'd like to work on this issue as contribution to OpenTimelineIO. I can review the documentation, identify the missing/incomplete areas, and create the requested checklist. Could you please assign this issue to me? Thank you!

PR #2052, opened 2026-09-16
Add missing docstrings for Item, MediaReference, RationalTime

hikmetba-bit:
## Summary
Addresses the three concrete examples named in #2042:
- `Item` had no class docstring at all.
- `MediaReference` had no class docstring at all.
- `RationalTime`'s docstring used a Sphinx `:math:` role (`:math:\`rt.value/rt.rate\``) that renders as raw, confusing text outside of the actual Sphinx build — e.g. in the autogenerated `docs/tutorials/otio-serialized-schema.md`, or in plain `help()` output.

While in there, I also documented `Item.source_range`/`color`/`markers`/`effects`, `MediaReference.available_range`/`available_image_bounds`/`is_missing_reference`, and `SerializableObjectWithMetadata.name`/`metadata`. That last one is inherited by essentially every schema class, so it fixes the empty `name`/`metadata` parameter descriptions across the whole generated schema doc, not just on `Item`/`MediaReference`.

This is #2042's three named examples plus one high-leverage base class — not the full documentation checklist the issue asks for, which is a much bigger, open-ended audit better done incrementally.

## Important note for the merger
`docs/tutorials/otio-serialized-schema.md` is generated by `make doc-model-update`, which requires building the native `_otio` extension. I don't have a C++ toolchain in this sandbox, so I could not run it, and deliberately left that generated file untouched rather than hand-edit it (the name/metadata change alone would touch dozens of unrelated lines across the whole file, which is exactly the kind of thing that file's generator, not a human, should produce). The project's own test suite should flag that it's stale — running `make doc-model-update` before merge will refresh it correctly.

## Test plan
- Could not build/run the test suite locally (no C++ toolchain available in this environment).
- Verified by manual review: cross-checked every new `:class:`~X`` reference against the actual bound class names in the same file, confirmed the `def_property`/`def_property_readonly` docstring-argument pattern against existing working examples in this same file (e.g. `Item.enabled`), and traced through `autogen_serialized_datamodel.py` to confirm exactly which docstrings feed the generated schema doc (via `inspect.getdoc`) before deciding what to change.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


hikmetba-bit:
Closing as a duplicate of #2051, which I opened concurrently in a separate session and just consolidated with the extra property-level docstrings from this PR (Item.source_range/color/markers/effects/available_image_bounds, MediaReference.available_range/available_image_bounds/is_missing_reference, RationalTime.value/rate). Apologies for the noise — #2051 now has everything from both.

PR #2053, opened 2026-09-17
Add the markers argument to the Track constructor.

peter-targett:
Now any markers assigned at the track level item are initialised.

Moved the markers argument ahead of color to try and maintain some consistency with the parent class.

Simple test added.

Changed required because otherwise there is no way to assign markers to a track in the C++ API - motivated by recent changes to the AAF adapter to add write-marker support.


jminor:
It looks like this API change causes some tests to fail. This change would require users of OTIO to update their code. For most API changes we try to deprecate the old way, but still support it for a release. Unlike Python where keyword arguments can be in any order, this is tricky to do with a C++ constructor. As much as I'd like to keep the argument order consistent with the parent class, it might not be worth the hassle in this case.

For your use case, are you able to construct a Track and then assign markers to it afterwards?

peter-targett:
> As much as I'd like to keep the argument order consistent with the parent class, it might not be worth the hassle in this case.

I thought this would be a possible issue, I will change it.

> For your use case, are you able to construct a Track and then assign markers to it afterwards?

As mentioned, it's motivated by the AAF adapter updates, it only looks for markers in the track. In the C++ API there is no way to assign markers after constructor use, you can only query the markers after the fact.

jminor:
Track inherits from Item, so you can modify the list of markers like this:

```
SerializableObject::Retainer<Track> track = new Track();
SerializableObject::Retainer<Marker> marker(new Marker("my_marker", TimeRange()));
track->markers().push_back(marker);
```

Here's an example in the Raven source code that works to add markers to a Clip or Track or Stack (though Raven seems to have a bug in _displaying_ track markers...) https://github.com/OpenTimelineIO/raven/blob/624951e89bb531a376327e0233c0e6ed6dffccb7/editing.cpp#L190

PR #2055, opened 2026-09-19
Fix null pointer dereference in Composable::available_image_bounds

### LINKED ISSUE #2045: Crash in available_image_bounds error code path
An LLM-based code review claims that there is a bug here:

> composable.cpp:79 — available_image_bounds writes *error_status without a null check; the parameter defaults to nullptr. gap->available_image_bounds() crashes.

Could we get a volunteer to verify if this is a real bug? It sounds plausible, and could be easily verified by writing a test case and/or a test OTIO file that creates the scenario described.

La1itchauhan:
Fixes #2045

`Composable::available_image_bounds(ErrorStatus* error_status)` was writing directly to `*error_status` without checking if the pointer was null. Because `error_status` defaults to `nullptr`, calling `available_image_bounds()` on `Composable` or derived classes that do not override it (such as `Gap` and `Transition`) caused a segmentation fault.

- Added a null check before writing `ErrorStatus::NOT_IMPLEMENTED` to `*error_status`, consistent with `Composable::duration`.
- Resolves the crash for all non-overriding subclasses (including `Gap` and `Transition`) inheriting the base implementation.

- Added `test_composable_available_image_bounds` in `tests/test_composition.cpp` verifying that calls with default `nullptr` and explicit `nullptr` do not crash, and that passing an `ErrorStatus` pointer correctly reports `NOT_IMPLEMENTED`.
- Added `test_available_image_bounds` in `tests/test_item.py` to cover the Python bindings.


jminor:
Hi @La1itchauhan - thanks, this PR is looking pretty good. We'll arrange for an OTIO committer to review it.

Out of curiosity, how did you come across this issue, and what lead you to contribute to OTIO?

